PR #40, opened 2012-08-17
Support for Maxon Datamax GPS reporting protocol

arekinath:
Hi!

I added support for the protocol spoken by a Maxon Datamax MA100-1010 modem-router (which has GPS support). It's possible some other things speak the same -- it's just NMEA with "$GPFID" to give the IMEI/identity of the unit, and no return from the server necessary.

Enjoy!

### DIFF
--- a/default.cfg
+++ b/default.cfg
@@ -83,7 +83,7 @@
     <entry key="http.enable">true</entry>
     <!--<entry key="http.address"></entry>-->
     <entry key="http.port">8082</entry>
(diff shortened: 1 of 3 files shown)


tananaev:
It's been a long time and I'm not sure if you are still using Traccar, but just want to let you know that I modified T55 protocol decoder to support Maxon devices, and removed separate Maxon decoder as it's redundant now.

PR #612, opened 2014-03-17
Atrack Multiple Positions Data Support

kero99:
Atrack, Add multiple positions protocol support. When Atrack devices lost GSM signal, send multiple positions in same data packets when signal is available again.


tananaev:
Support for Atrack protocol has been extended to include decoding of long date format and all additional information.

PR #737, opened 2014-06-01
new method getDistanceInMetersFromPosition(Position secondPosition) to class Position

31neKl31n3N4chtmu51k:
Implementation of method that uses the ‘haversine’ formula to calculate
the great-circle distance between two points, the Position itself and
the Position passed as argument. The return value is in meters.


tananaev:
There is a `DistanceCalculator` utility class to calculate distances.

PR #935, opened 2014-10-22
Added eSkyWireless 610 Protocol

jon-stumpf:
Here is my work that added the eSkyWireless 610 protocol.  This protocol may apply to other models of eSkyWireless devices but that has not been tested.  I will send the protocol documentation for your website.
- jss


tananaev:
Can you please remind me what happened with this pull request. Why we didn't finish it?

PR #963, opened 2014-11-05
Reverse Geocoding parsing to shorter format

vzoltan:
Reverse Geocoding parsing to shorter format:
City, Street, Number


tananaev:
I have implemented proper parsing of address components and in future the address format would be configurable.

PR #1018, opened 2014-12-28
Huabao808

sandroacoelho:
Huabao VT808 protocol
--------------------------------------------------------------------------------------------------------------------------------
This is a work in progress HUABAO implementation. Once I have some vendor details, it will be finished.

[Huabao 808 protocol] (https://dl.dropboxusercontent.com/u/22845186/huabao/Huabao-808-protocol-en.pdf)


sandroacoelho:
Hi @tananaev. Could you please heck if the code is fine for merging?

About your questions: 

>> 2. What is the purpose of Huabao808 class? I think all the logic from that class should be directly in the decoder class.

I thought that might be better if all the rules from VT808 protocol were encapsulated. Beside this, VT808 protocol is "conversation based" and requires an answer in order to send the location.

>> 1. Frame decoder is incorrect. Seems like you have some frame decoding functionality in protocol decoder, but it should be moved to frame decoder class.
>> 3. I think it is unnecessary to save data in member variables. Instead you should put data directly into Position instance. Having so many variables makes the decoder more complicated and harder to understand and maintain.

I have made a big refactoring in order to keep it simple.  Please, let me know if there is some issues on it.

Best,

tananaev:
> I thought that might be better if all the rules from VT808 protocol were encapsulated. Beside this, VT808 protocol is "conversation based" and requires an answer in order to send the location.

Decoder class is for the encapsulation. Why do you need another one?

Frame decoder is still incorrect.

Overall I don't see any major changes from the original version.

tananaev:
I have finally implemented Huabao protocol in Traccar.

PR #1028, opened 2015-01-13
Saving Teltonika device power state.

pallavt:
Works only when DIN 1 is connected to ignition and sending of DIN 1 state is enabled


tananaev:
Thanks for the pull request.

PR #1091, opened 2015-02-23
Implemented AutoFon 4.5 protocol (also works for StarLine M10/M11/M16/M17 devices)

vitalidze:
I was working on it to make Starline M17 device able to send positions to traccar. Basically, protocol is something like previous version of `AutoFon` protocol, but with some significant differences.

I have a question about binary protocols (galileo, autofon, etc.) - their decoder are stateful in fact (stores device identifier after authentication packet received). How traccar supposed to handle multiple devices with same protocol? I guess for now the only way is to start another instance of traccar on a different port.

### DIFF
--- a/default.cfg
+++ b/default.cfg
@@ -387,4 +387,8 @@
     <entry key='gosafe.enable'>true</entry>
     <entry key='gosafe.port'>5078</entry>
 
+    <!-- AutoFon 4.5 / Starline M10/M11/M16/M17 server configuration -->
+    <entry key='autofon45.enable'>true</entry>
+    <entry key='autofon45.port'>5079</entry>
+
 </properties>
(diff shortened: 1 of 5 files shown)


tananaev:
New decoder instance is created for each connection, so there should be no problem with having state in decoder.

Can you provide documentation for the protocol?

vitalidze:
I have Starline M17 device. After spending some time googling I have found these [docs for M11+](http://www.ultrastar.ru/8636/10346) and then I have found that M10/M11/M16/M17 are actually similar to autofon 4.5. [This forum thread](http://forum.autofon.ru/viewtopic.php?f=8&t=37) was useful too.

Such protocol implementation works fine with my Starline M17.

tananaev:
Thanks.

tananaev:
I have decided to merge AutoFon and AutoFon45 protocol. Is it possible for you to check if it still works for your devices?

vitalidze:
I will check later. Thanks for the notification.

PR #1116, opened 2015-03-12
Update Gps103ProtocolDecoder.java

prsolucoes:
Added speed as optional, because server is canceling the request because the speed that is empty.


tananaev:
Thanks for the pull request.

PR #1223, opened 2015-05-27
Update ServerManager.java

ke5stl:
Please allow the Queclink GL200 to support UDP as well as TCP connections.  GL300 model devices (compatible with GL200 protocol, it seems) support UDP mode as well as TCP.  UDP uses less battery though, so may allow the device to run a little bit longer on a charge.


tananaev:
For UDP protocol there is no need for frame decoder.

ke5stl:
Thank you for saying so.  I tested it locally, and it indeed, works fine without the frame decoder!

tananaev:
Done. Thanks for the pull request.

PR #1229, opened 2015-05-31
Update DataManager.java

ke5stl:
allow DataManager to use a container provided datasource, as would be the case when being used with Glassfish, where the jdbc datasource is configured in the container, and available via jndi lookup from the InitialContext.

### DIFF
--- a/src/org/traccar/database/DataManager.java
+++ b/src/org/traccar/database/DataManager.java
@@ -73,7 +73,7 @@ public class DataManager {
 
         String jndiName = properties.getProperty("database.jndi");
         
-        if(jndi != null) {
+        if(jndiName != null) {
             try {
             DataSource ds = (DataSource) new InitialContext().lookup(jndiName);
             dataSource = ds;
--- a/src/org/traccar/database/DataManager.java
+++ b/src/org/traccar/database/DataManager.java
@@ -94,6 +94,7 @@ public class DataManager {
                 DriverManager.registerDriver(new DriverDelegate(d));
             } else {
                 Class.forName(driver);
+                
             }
         }
         
--- a/src/org/traccar/database/DataManager.java
+++ b/src/org/traccar/database/DataManager.java
@@ -27,6 +27,7 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.text.ParseException;
 import java.util.*;
+import javax.naming.InitialContext;
 import javax.sql.DataSource;
 
 import org.traccar.Context;
@@ -70,6 +71,17 @@ public class DataManager {
         
         useNewDatabase = Boolean.valueOf(properties.getProperty("http.new"));
 
+        String jndiName = properties.getProperty("database.jndi");
+        
+        if(jndi != null) {
+            try {
+            DataSource ds = (DataSource) new InitialContext().lookup(jndiName);
+            dataSource = ds;
+
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+        } else {
         // Load driver
         String driver = properties.getProperty("database.driver");
         if (driver != null) {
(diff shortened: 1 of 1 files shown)


ke5stl:
FYI: I am using Traccar server on Glassfish, and I find it beneficial to be able to configure/manage the jdbc pool in Glassfish, and have the Traccar app use the jdbc datasource.

tananaev:
How do you run Traccar in Glassfish?

ke5stl:
Glassfish is a container like Tomcat is, but, it provides the option of offering EE capabilities too.  Glassfish is able to run .war projects, much like Tomcat does.  Glassfish is the reference implementation of a Java EE application server.  Mostly, I use the traccar server layer as a collector along side other collectors I have.  The collected data is viewable via my own custom web map, which also leverages OpenLayers.  But, I customize your code a little bit to suit my particular needs.  eg: I allow traccar to 'auto register' devices rather than requiring manual registration, since that's more convenient than having to manually register every device, for my usage.  While I have my own web front end, having the ability to have smart phones run the tracker app is great!  You've done a great job with traccar.  I'd like to roll some of my minor changes into your codebase, and so far, you've been very accommodating.

tananaev:
I know what Glassfish is, but Traccar is not a web application. It is a Java SE service.

ke5stl:
ha ha.  I just downloaded your latest server zip file, and yes, I see that there's just a .jar there.

ok.  So, I mentioned that I am using traccar-server as a collector of data.  What I do is put your source code into my web application source project, and have a context listener that starts up your server functionality.  Also, at a point within the TrackerStreams, I have a modification that causes all received data to be added to the other collected data (recall, I said that traccar is one of the collectors I have).  This data is then distributed to clients via a WebSocket connection, after it is potentially masked or modified depending on ownership and labelling requirements that are in effect at the time.

I have been thinking about how to better provide separation and scale options, and I'm thinking that perhaps something like having multiple collectors forwarding data to a JMS Topic might be a good way to allow scaleability to be improved.

then, the topic listeners would be able to also be scaled and separated from the collector side of things.

ke5stl:
perhaps using something like Hazelcast could also be rolled in to provide a shared current location database for trackers that's in-memory but replicated across collection nodes.  note: I am not trying to provide historical data, just the current / last known location.  so you see, my needs are a little bit different than what you're aiming to provide with your web front end.  but, the ability to collect data from so many different types of tracker in a well defined way is great.    

One other major difference, is that I'm streaming live updates to the web clients.  I don't make the client poll for data updates.  but, I also don't try to show historical data.  that might be something of interest at some point, but it's not the focus for me right now.

ke5stl:
the idea re scale are not as result of any issues that I've had so far, but, I am just looking to be able to have high availability should one server have a problem for whatever reason - might not be to do with the code at all.  could be the disk getting full, for example, or, perhaps the host service has to do a maintenance window due to a failure somewhere.

tananaev:
Anyway, this code change doesn't make much sense at this point. We can discuss scalability problems in another thread if you want, but it seems like the way you use Traccar is very different from the default setup.

As for current data, if you look at the latest code, there is already in-memory cache and an API to get current data live (it uses long polling instead of web sockets).

ke5stl:
is the in memory cache something that's able to be replicated across multiple jvm instances?

tananaev:
No, there is no replication support. I want to keep it simple (at least at this point) because from my experience when you try to implement redundancy it actually results in more problems because of increased complexity.

ke5stl:
I can understand your concern about complexity.  as far as the code change not making sense, why do you think that?  I've proposed that the change be done in a way that's entirely compatible with how you already do your configuration.  the change would be impactless to your existing method of operation, but would add support for being used within a web container environment, where the jdbc connection is managed in the container.

tananaev:
OK, I can merge the change, but can you please fix code formatting first.

ke5stl:
Thank you!  Please take a look now.  Let me know if any further tidyup is needed.

tananaev:
It didn't allow for automatic merge, so I copied the changes manually. Check if everything is correct.

ke5stl:
looks good to me.  thank you!

PR #1240, opened 2015-06-09
Moved all the SQL queries to the configuration file

demianalonso:


demianalonso:
Some issues about adding compatibility for MySQL that you might want to check:

* MySQL doesn't allow unique columns to have a size larger than 768 bytes. So I changed:
   - user.email to be varchar(256), which I think is the max for an email either way
   - device.uniqueId to be varchar(256) , which I am not sure if I am truncating some GPS Tracker id

* Since MySQL and h2 have different ways of calculating a SHA256, I moved the logic to Java, and so the queries are simpler and compatible.

* I had to add some backticks ( `  ) to some columns definitions that are keywords for MySQL. I haven't found any docs about using backticks in h2, but it seems to work nonetheless.

tananaev:
Can you please make sure there are no tabs in the code. All tabs should be replaced with 4 spaces.

Everything else looks good. I also want to move hash function to some other class instead of having it directly in User, but it's a minor thing.

Another thing I'm not sure about is the idea of storing password as bytes. String seems to be more portable solution. Say if we want to do hashing on the client side or something like that it would be easier to use string (e.g. hex representation of hash.). Can you think of pros and cons of both solutions?

demianalonso:
Hi,
  here goes the replie:
1) tabs => Gone :)
2) Moved the sha256 to a helper class (Hashing)
3) I stored as a byte array since thats what's the API returned, so that's just the rawest format. As for portability I think it would be the same, but I am don't have a strong opinion of either way. I will change it to be encoded in base64.


Also, I noticed that there is a salt column that it is not being used. Maybe I can also add the proper password functionality as recommended here (https://crackstation.net/hashing-security.htm). Do you think I should try that too?

tananaev:
Thanks.

As for salt, in initial version I didn't want to implement it just to save some time and release new interface earlier as for most users regular hashing should be enough (huge improvement over plain text like in previous version). But if you have time and can implement it without major changes to json and database serialization code, it would be awesome.

tananaev:
There is a problem with the change: `java.nio.charset.StandardCharsets` doesn't exist in Java 6.

tananaev:
Fixed the problem in ae56bc9e1f212bf96c3573b43fb8b11983bbb3e1.

PR #1242, opened 2015-06-10
Changing selectDevices query

al3x1s:
I think using "JOIN" instead of "IN" can get better performance


tananaev:
Thanks for the pull request.

PR #1243, opened 2015-06-10
removing double quotes

al3x1s:
got this warning if it has double quotes

======
2015-06-10 17:41:00  WARN: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '"SELECT userId, deviceId FROM user_device;"' at line 1 - com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException (NativeConstructorAccessorImpl.java:-2)

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -162,7 +162,7 @@
     </entry>
 
     <entry key='database.getPermissions'>
-        "SELECT userId, deviceId FROM user_device;"
+        SELECT userId, deviceId FROM user_device;
     </entry>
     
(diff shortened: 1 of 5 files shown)


tananaev:
I have reverted the merge because it has some other changes as well.

al3x1s:
wow, sorry Anton, I just make some change in my fork after that commit, I did not know it would affect the pull request

tananaev:
Please create another pull request.

al3x1s:
ok, I already did it

PR #1245, opened 2015-06-11
Removing double quotes 

al3x1s:
got this warning if it has double quotes

=====
2015-06-10 17:41:00 WARN: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '"SELECT userId, deviceId FROM user_device;"' at line 1 - com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException (NativeConstructorAccessorImpl.java:-2)


tananaev:
Thanks for the fix.

PR #1252, opened 2015-06-15
Implemented password hashing using a salt, following this code&guidel…

demianalonso:
I implemented password hashing following the guidelines on this site: https://crackstation.net/hashing-security.htm

I tested it using MySQL and H2 and it works on both databases.

There are some things that might be worth reviewing (plus any other you can think of :) )

1) It is not the same idea the password that comes from a web request that the password (hash) that is stored in the database. Yet, given the current JSON conversions, I was a bit stucked in reading the input password directly from the User object (since that one is loaded from the web request). 

That seems a bit inconsistent, but the request body reader cannot be reused to recreate another json to read the password. Maybe we can extend the current request to create the associated json object, and store it through the life of the request. In this way we could parse different objects from the same request if needed, or retrieve a particular property.

2) The Hashing class is no longer necessary but it does have a handy method to hash a String. Should be keep it just in case or erase it since it is no longer used?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ nbactions.xml
 .settings
 .idea
 web/.idea
+.idea
+traccar.iml
--- a/debug.xml
+++ b/debug.xml
@@ -133,7 +133,7 @@
     <entry key='database.loginUser'>
         SELECT * 
         FROM user 
-        WHERE email = :email AND password = :password;
+        WHERE email = :email;
     </entry>
 
     <entry key='database.selectUsersAll'>
@@ -141,8 +141,8 @@
     </entry>
 
     <entry key='database.insertUser'>
-        INSERT INTO user (name, email, password, admin) 
-        VALUES (:name, :email, :password, :admin);
+        INSERT INTO user (name, email, password, salt, admin) 
+        VALUES (:name, :email, :password, :salt, :admin);
     </entry>
         
     <entry key='database.updateUser'>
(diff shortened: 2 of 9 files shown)


tananaev:
1) Can we have another field `hashedPassword` or something like that to avoid passing password separately to data manager methods? Then `setPassword` method can set both plain text `password` and `hashedPassword` fields. In the database we would save only `hashedPassword`. Do you see any problems with this approach?

2) I don't like the `PasswordHash` class because it uses different license. It doesn't really have much logic. I think it would be good to re-implement hashing with salt in original `Hashing` helper class.

Everything else looks good.

demianalonso:
1) I changed to work like that (I pushed the change so you can see it).
There two things to notice about it:
  a) I changed the serialization so a property can be marked to be ignored in the serialization (for the hashed password and salt). There is the obvious problem that we don't want to transmit the hashed password + salt through the network (since all privacy and secrecy would be lost :) ), and because the frontend seem to post back those values too. And when that happens, it sometimes overrides the logic in the setPassword, since the setHashedPassword is executed after setPassword (messing up everything). 

  b) To answer your question, I would say that the disadvantage is that not always all actions matches directly to an object in the model. And perhaps you sometimes need to load different objects that might not be "persistent models". 
In this case it seems to me that we are adding a password field to the model just to solve the a serialization issue. But since the getter has some logic I think is a corner case (that is way I implemented it :) ). So maybe we can wait to a more problematic case to make a bigger change.


2) I thought the licences were compatible so I took the code since it seemed to be tested and debugged already. But if that is a problem we can try to rewrite those parts. It might take a bit of time since opening the blackbox will require me to learn a bit more about hashing :)

tananaev:
OK, let's leave the class for now. Thanks for the pull request.

PR #1302, opened 2015-07-16
Fix: Excluded slf4j dependency from com.ning.async-http-client as it …

demianalonso:
…conflicts with the log4j dependency


tananaev:
Thanks

PR #1317, opened 2015-07-28
Implementing Totem commands support

al3x1s:


tananaev:
What does this command do?

al3x1s:
Just get the device imei.
if you want, give me a second, I'll add more commands to this protocol

tananaev:
Do we actually need this command? What's the practical application for it? I think it would be better to have location request command or something like that.

tananaev:
Also, I noticed there is some problem with frame decoder. We need to fix it properly instead of catching the exception. Can you please provide a log file with HEX messages to and from device.

al3x1s:
just for to test it and check if everything works, I'll remove then

I will remove the try-catch too

tananaev:
OK, please update the pull request when you implement some useful commands so I can integrate it into the Traccar project. Thanks.

al3x1s:
For this protocol, needs to set a device password to send a command, so I added this key **KEY_GPS_PASSWORD** to  **Command.java**

this is an example how it works:
``` javascript
jQuery.ajax({
        type: "POST",
        url: "http://URL:8082/api/command/send",
        dataType: "text",
        data: JSON.stringify({
            deviceId: 2,
            type: "engineStop",
            other: {"gpsPassword": "000000"}
        })
    });
```
but I don't know what changes should I make on your website for this to work

tananaev:
That's a good point. Many devices require password. I need to think of some generic solution.

By the way, your unit test fails.

al3x1s:
sorry, I forgot update it
I already fix it, try again

tananaev:
Following code is incorrect. Need to implement it properly.
<pre>buf.readBytes(buf.readableBytes());</pre>

Everything else looks good. What is the default password? For now I think we can hardcode the default one and I will fix it properly in the next version. I want to release a new version next week, so I don't have time to add it into this version.

al3x1s:
Sorry sir, I don't know how implement a properly solution
the gps response is like these:

**for stop engine**
```
Receive:'025'OK
*000000,025,C,1#
```

**for resume engine**
```
Receive:'025'OK
*000000,025,C,0#
```


**Default password is 000000**

tananaev:
Can you please provide original HEX messages of the tracker's response.

al3x1s:
**Stop engine response**
```
526563656976653a27303235274f4b0d0a2a3030303030302c3032352c432c3123
```

**resume engine response**
```
526563656976653a27303235274f4b0d0a2a3030303030302c3032352c432c3023
```

tananaev:
So, there is no line break at the end?

al3x1s:
No, there isn't

tananaev:
Anyway, thanks for the pull request. I will think how to fix it properly and merge it to the main branch as soon as I find some free time to do it.

al3x1s:
Great! 
Thanks for support

tananaev:
I have made some minor modifications and added default password. You can pull the changes from master if you want to be in sync.

al3x1s:
already did it, thanks

PR #1328, opened 2015-08-01
Added speed and course information to the traccar log. This would hel…

asatsi:
Added speed and course information as well to the traccar log. This would help render the track on a different UI. As well this information could be pushed into elasticsearch via logstash for better analytics.


tananaev:
Why is there an extra comma at the end?

tananaev:
Also, why did you create a new pull request? I think you need to read how to properly do pull request. You should be able to update an old one.

asatsi:
Not very familiar with 'git' yet. Thanks for bearing with me. Please check now and let me know.

tananaev:
Looks good now. Thanks.

PR #1339, opened 2015-08-07
Correction of error in refresh page

rafaelnoob:


tananaev:
I have already fixed the problem.

PR #1356, opened 2015-08-17
Patch 1

mtrntx:


tananaev:
First of all, please give this pull request some meaningful name.

Secondly, it doesn't compile because you haven't implemented interface method.

PR #1379, opened 2015-08-27
Added smoke test for JNDI context initialisation, in pom.xml jetty version is in property now

vitalidze:
As we discussed yesterday in #1377 I have added a simple smoke test, which will fail if it won't be able to initialise the JNDI context.

Also I have added `jetty.version` property which is re-used in all jetty dependencies definition since now there are four such records in `pom.xml` file.


tananaev:
Thanks.

Issue #1395, opened 2015-09-04
Expose new UI API for the 'old' web UI


vitalidze:
Right now there are two options for the web UI:

1) New UI (default), which will expose several servlets and also serves contents under folder from the `web.path` configuration property. ExtJS UI located in that folder by default is communicating with the backend through the API calls.

2) Old UI, which will deploy a `traccar-web.war` file (or actually any other .war file) and then leaves everything to that web application.

I want to have a possibility to mix this. In particular I would like to have an API from the new UI and to not serve the contents of folder under `web.path`. I believe this is possible and quite simple to do. In theory I will need only these:

````
        ServletContextHandler servletHandler = new ServletContextHandler(ServletContextHandler.SESSIONS);
        servletHandler.setContextPath("/api");
        servletHandler.addServlet(new ServletHolder(new AsyncServlet()), "/async/*");
        servletHandler.addServlet(new ServletHolder(new ServerServlet()), "/server/*");
        servletHandler.addServlet(new ServletHolder(new UserServlet()), "/user/*");
        servletHandler.addServlet(new ServletHolder(new DeviceServlet()), "/device/*");
        servletHandler.addServlet(new ServletHolder(new PositionServlet()), "/position/*");
        servletHandler.addServlet(new ServletHolder(new CommandServlet()), "/command/*");
        servletHandler.addServlet(new ServletHolder(new MainServlet()), "/*");
````

To be more particular I need the Device, Position and Command servlets. But I guess from the maintenance point of view it will be more simple to move this servlet initialisation to a separate method. Then just call it from two constructors in `WebServer` class. Maybe I am missing something.

Please let me know what do you think? If you confirm then I will implement it and do a pull request.

tananaev:
I think it's a good idea to be able to expose APIs without any web interface or with old/custom version of the UI. If you expect a lot of changes, please describe the plan in more details, so I can review it before you start implementing.

vitalidze:
Actually I thought about putting `ServletContextHandler servletHandler` initialisation into a separate method in `WebServer` class and then just call it from both constructors. In case of constructor that runs `.war` file I would use `HandlerList` similar to the use in another constructor. In such case the API would always be available nevertheless of the web UI implementation.

If you want to have an option to expose API without any web UI then I guess it's better to have such configuration option and then I would stick with a single constructor of `WebServer`, which will do the right initialisation based on the configuration options. So in such case there could be just a single configuration option like `web.type` with values:

[...]

PR #1397, opened 2015-09-05
Initialise API for old interface, added possibility to expose only API

vitalidze:
Implemented three ways to initialise web server as discussed in #1395:

* with new web application ("web.type" = "NEW")
* with old web application ("web.type" = "OLD")
* only api ("web.type" = "API")

Please check and let me know if you have any questions.


tananaev:
Can you please change values to lower case. Everything is the config is lower case, so these values would look weird.

vitalidze:
Done. Please check.

PR #1413, opened 2015-09-19
Hungarian translation

zmitya:
Hi Anton,

A am not sure how to add a new lang to your project I just started to translate it. Let me know if I need to translate more files.


tananaev:
That looks correct. Thanks for pull request.

tananaev:
I have moved Traccar translations to the Transifex platform to be able to manage string changes and get translations for all the available languages quickly.

I would appreciate if you can become a translator for the Hungarian language on the website:

https://www.transifex.com/traccar/traccar/

There are only a couple of new string that require translation.

PR #1422, opened 2015-09-23
Add OpenCage ReverseGeodecoder

StefaanVanDooren:
Reverse geocoder for OpenCage

http://geocoder.opencagedata.com/


tananaev:
Thanks again.

PR #1433, opened 2015-09-28
Add support for BlackKite protocol.

bravegnu:
BlackKite is closely based on the Galileo protocol. We derive code
from the Galileo protocol and make changes as required to meet
BlackKite protocol specifications.


tananaev:
Thanks for the pull request. Can you provide some information about the brand or device? Is there an official website?

bravegnu:
@tananaev The device information is available from http://www.zilogic.com/blog/blackkite-fleet.html The user manual is should become available shortly.

PR #1435, opened 2015-09-28
Fix invalid (My)SQL syntax when quoting reserved keywords

TwizzyDizzy:
Hi folks,

I'm quite irritated by the SQL used to create the traccar database. Obviously there has gone some effort into quoting reserved keywords (though not all, for example, in MySQL, "status" in the device table would be another reserved keyword that is left unquoted) but the wrong delimiter for reserved keywords has been used - " instead of `.

Let me be clear on this: I gather the SQL in the file does currently work with the database driver used in traccar (that maybe fixes this on-the-fly, because my initial database has been created), but in itself, this is no valid MySQL (this is fixed by this commit). I don't know, whether there may be another delimiter for MSSQL or ORACLE, but as stated in 

> https://dev.mysql.com/doc/refman/5.5/en/keywords.html

the ` character is the one to turn to when using reserved keywords in MySQL.

This may be a bigger issue and other database folks (like MSSQL or ORACLE) might want to get involved.

Cheers
Thomas

### DIFF
--- a/setup/unix/traccar.xml
+++ b/setup/unix/traccar.xml
@@ -31,7 +31,7 @@
     </entry>
 
     <entry key='database.createSchema'>
-        CREATE TABLE "user" (
+        CREATE TABLE `user` (
         id INT PRIMARY KEY AUTO_INCREMENT,
         name VARCHAR(128) NOT NULL,
         email VARCHAR(128) NOT NULL UNIQUE,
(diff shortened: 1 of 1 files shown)


tananaev:
Double quotes (" ") is a standard (ANSI) SQL quoting mechanism. MySQL supports it as well, but you need to enable it explicitly by specifying `SQL_MODE=ANSI_QUOTES`.

It can be done for Traccar database connection:

<pre>&lt;entry key='database.url'&gt;jdbc:mysql://...&amp;amp;sessionVariables=sql_mode=ANSI_QUOTES&lt;/entry&gt;</pre>

TwizzyDizzy:
Ah! I've just seen, that the documentation has this sql mode specified in the documentation.

> https://www.traccar.org/mysql/

Then my issues must have been due to always using the DB connection string from my config file of the last version (my server is running since 2.x).

Thanks for the clarification, Anton!

Cheers
Thomas

PR #1470, opened 2015-10-22
UDP stops listening after failure

vitalidze:
We have a 'gator' device, which uses 'gator' protocol (port 5052). From my understanding in current traccar version (3.1) it is working through UDP protocol (and BTW we don't know how to configure it for TCP, but it's not the case). We were messing with device configuration and received following packets:

````
2015-10-22 00:11:31 DEBUG: [70E792A4: 5052 < 61.245.163.243] HEX: 242421000643e30282070d
2015-10-22 00:11:45 DEBUG: [70E792A4: 5052 < 61.245.163.243] HEX: 242400
2015-10-22 00:11:45  WARN: [70E792A4] error - Not enough readable bytes - Need 2, maximum is 0 - IndexOutOfBoundsException (... < GatorProtocolDecoder.java:59 < ExtendedObjectDecoder.java:39 < ...)
````

After that message traccar stops listening on UDP port 5052 and not displayed in `netstat -nlu` command. I know that device sends an invalid message, but why does it stops listening after that failure? Please advise.

### DIFF
--- a/src/org/traccar/MainEventHandler.java
+++ b/src/org/traccar/MainEventHandler.java
@@ -20,6 +20,7 @@ import org.jboss.netty.channel.ChannelHandlerContext;
 import org.jboss.netty.channel.ChannelStateEvent;
 import org.jboss.netty.channel.ExceptionEvent;
 import org.jboss.netty.channel.MessageEvent;
+import org.jboss.netty.channel.socket.DatagramChannel;
 import org.jboss.netty.handler.timeout.IdleStateAwareChannelHandler;
 import org.jboss.netty.handler.timeout.IdleStateEvent;
 import org.traccar.helper.Log;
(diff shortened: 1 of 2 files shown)


vitalidze:
Please see changes right in this issue, which was converted to pull request.

tananaev:
Not sure if it's a good idea to have a network server and client in a unit test, but let's leave it for now. Thanks for the fix.

vitalidze:
Thank you for merging. I agree that it may not be a good idea. I couldn't find a better way to test it. Probably it can be called 'integration test' since it brings up the full decoder. Anyway, if it will cause you problems - feel free to remove it.

tananaev:
I have removed the unit test for now. Can't think of any unit test for it.

Also, it could potentially fail if system already has some other server listening on the port. Of if user doesn't have permissions to open the port.

vitalidze:
Ok, no problem. The only way to create a good test without opening ports is to use mocks for the netty abstractions (like channels, bootstraps, etc.). But you don't use any mocking library for now and I don't know whether you are fine with something like [mockito](http://mockito.org/). Besides, you are already using some kind of mocks in your tests, so maybe it's worth to use mocking library for that. If you want I may try to implement similar test with mocks.

tananaev:
I don't think there is much value for unit testing this particular thing, so let's leave it as it is for now.

vitalidze:
Sure. What do you think about using traccar built from the current state of codebase? Is it stable enough for production?

tananaev:
I think it's stable. I didn't change the back-end much. Mostly working on the web interface features.

PR #1484, opened 2015-10-28
Another version of GlobalSat protocol

vitalidze:
I have a weird device [euspro sf-400](http://eus.com.ua/gps-monitoring/gps-tracker-sf400.html), which I want to connect with traccar. Unfortunately it does not have a protocol documentation, so I decided to break bad and just write to some port and read messages. Fortunately (I hope) the protocol has text format. Here are couple of message examples:

    $80050377796567,0,13,281015,173437,E08513.28616,N5232.85432,222.3,0.526,,07*37!

    $80050377796567,0,18,281015,191919,E08513.93290,N5232.42141,193.4,37.647,305.40,07*37!

(_I have replaced real IMEI and lon/lat numbers, but other fields were left untouched_)

The message format looks very close to the "alternative" version of `globalsat` protocol. However, there are couple of differences:

1) `fix` has something different (two digits number, `13` and `18` in these cases) and I don't think it should be used for identification whether received position or not.

2) `course` may be missing like in first message

3) I am not sure about last parameter `07*37`, but it looks like it is the number of satellites and the [HDOP](http://www.globalsat.ru/faq/gps_trekery_globalsat_tr-151_faq/tr_151_i_hdop) parameter separated with asterisk.

Here is my updated pattern of `globalsat` alternative for this protocol:

````java
            .txt("$")
            .num("(d+),")                        // imei
            .num("d+,")                          // mode
            .num("(d+),")                         // fix
            .num("(dd)(dd)(dd),")                // date (ddmmyy)
            .num("(dd)(dd)(dd),")                // time (hhmmss)
            .xpr("([EW])")
            .num("(ddd)(dd.d+),")                // longitude (dddmm.mmmm)
            .xpr("([NS])")
            .num("(dd)(dd.d+),")                 // latitude (ddmm.mmmm)
            .num("(d+.?d*),")                    // altitude
            .num("(d+.?d*),")                    // speed
            .num("(d*.?d*),")                    // course
            .num("(d+)[,\\*]")                   // satellites
            .num("(d+.?d*)")                     // hdop
            .compile();
````

So, what should I do? Update the pattern of the `globalsat` protocol or add the new one? Please advise.

### DIFF
--- a/src/org/traccar/protocol/GlobalSatProtocolDecoder.java
+++ b/src/org/traccar/protocol/GlobalSatProtocolDecoder.java
@@ -190,7 +190,7 @@ public class GlobalSatProtocolDecoder extends BaseProtocolDecoder {
             .number("(d+.?d*),")                 // altitude
             .number("(d+.?d*),")                 // speed
             .number("(d+.?d*)?,")                // course
-            .number("(d+)[,\\*]")                  // satellites
+            .number("(d+)[,*]")                  // satellites
             .number("(d+.?d*)")                  // hdop
             .compile();
 
--- a/src/org/traccar/protocol/GlobalSatProtocolDecoder.java
+++ b/src/org/traccar/protocol/GlobalSatProtocolDecoder.java
@@ -180,7 +180,7 @@ public class GlobalSatProtocolDecoder extends BaseProtocolDecoder {
             .text("$")
             .number("(d+),")                     // imei
             .number("d+,")                       // mode
-            .number("(d+),")                      // fix
+            .number("(d+),")                     // fix
             .number("(dd)(dd)(dd),")             // date (ddmmyy)
             .number("(dd)(dd)(dd),")             // time (hhmmss)
             .expression("([EW])")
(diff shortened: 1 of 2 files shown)


vitalidze:
I have updated it and added tests. Please see changes right in this ticket. Also the `valid` field was set up incorrectly since https://github.com/tananaev/traccar/commit/091edb35f76bf277ec0e22a738dd2825dfbe827f

tananaev:
Few comments about the change:

1. Can you please change course pattern from "(d*.?d*)," to "(d+.?d*)?," to make it clear that whole course field is optional. The you can use `if (parser.hasNext())` in the decoder method.

2. Why do you have a back slash in the satellites pattern field? I don't see any example where it can be a back slash there.

3. Are you sure that the last field is HDOP? I suspect it's some kind of checksum which usually comes after the asterisk.

vitalidze:
Thanks for reviewing.

1. Done

2. I was trying to escape the asterisk like for a regular expression pattern. Looks like it is done automatically in `PatternBuilder.number()` method. My bad, I have updated it.

3. No, I am not sure that it is HDOP. But it doesn't look like a check sum since it is the same in all messages. What should I do with it?

tananaev:
2. It's not done automatically, but as far as I know you don't need to escape asterisk in square brackets because it can't mean repetition in that particular context.

3. Let's leave it like this for now.

Thanks for contributing to the project.

vitalidze:
You are welcome. Thanks for the explanation :+1:

PR #1509, opened 2015-11-06
Add support to authorization basic http header in alternative to…

guterresrafael:
…cookie and java session.
Modification to allow traccar api stateless.


guterresrafael:
Class Base64 only java 1.8. I will refactor

tananaev:
You can use `DatatypeConverter` to decode base64.

PR #1534, opened 2015-11-17
add support for configurable handlers

ivanfmartinez:
make easier to test handlers without changing the traccar code

### DIFF
--- a/src/org/traccar/BasePipelineFactory.java
+++ b/src/org/traccar/BasePipelineFactory.java
@@ -19,6 +19,7 @@ import java.net.InetSocketAddress;
 import org.jboss.netty.buffer.ChannelBuffer;
 import org.jboss.netty.buffer.ChannelBuffers;
 import org.jboss.netty.channel.ChannelEvent;
+import org.jboss.netty.channel.ChannelHandler;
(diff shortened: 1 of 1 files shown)


tananaev:
Can you please explain the use case a bit more. Currently I don't see much value in this modification.

ivanfmartinez:
I'm making some tests to process the positions received in traccar.

Currently I'm testing mqtt publishing of data.

With this modification I can plug a handler without changing any other code in traccar base.
I have just to put the class name on the xml file.

This will make possible to people create own handlers and attach to traccar at runtime.

tananaev:
Did some cleanup (2776f3e5a7d837feff46c003ec32c162ac71b970), but no changes to the logic.

nwohaibi:
Dear @ivanfmartinez ,

Is it possible to have an example on how to use this feature?
Is it used for pluging extra logic to a current protocol OR pluging a new protocol instead of a existing one?

ivanfmartinez:
This feature is used to plug additional handlers to process data received by any protocol

For anyone interested in testing my handler for publishing data on mqtt : https://github.com/ivanfmartinez/ivanfm-traccar-mqtt

PR #1542, opened 2015-11-21
Add support to generic resources and others

guterresrafael:
Add base servlet for working with rest resources with generic type.
Refactor in core classes for support parameter Class<?> for base servlet.
Duplicate api urls to support plural to REST "Pattern"
Add support to url template for REST. Example: DELETE http://<host>/api/device/3
Change order of authentication for priority session
Implement Command Pattern to customize execution in flow of methods add, update, delete.


tananaev:
I've thought about this and it doesn't make sense to support two versions of the API. I guess REST API makes more sense, so we should go with it.

What do you think about using `javax.ws.rs` to implement the API?

Instead of long polling (/async) interface, I think it makes sense to use web sockets.

Any thoughts?

sosoxuc:
Web sockets are great but must consider Jetty capabilities. Whether it supports NIO and if each WS session will require separate Java Thread (This was a case in older Tomcat versions)

tananaev:
Good point. I need to investigate it. What about implementing proper REST API using JAX-RS?

sosoxuc:
That would be great

I can help implementing REST API

sosoxuc:
Have a meeting with my friend tomorrow and will review source code.

On Saturday we have whole day free so cad do actual work

sosoxuc:
We will test Jetty against WebSockets and tell you results

guterresrafael:
Great! I will refactor to use JAX-RS.

sosoxuc:
Hi @tananaev, @guterresrafael 

@zgnachvi has [tested Jetty against WebSockets](https://github.com/zgnachvi/cartrack-sockets/)

[here you can see results](https://github.com/zgnachvi/cartrack-sockets/blob/master/README.md)

And conclusion: Jetty WebSocket does not propagate threads per connection.

so you can **start using WebSockets and it would be great**

Also question. How will do REST API migration. **Author** or @guterresrafael or me and @zgnachvi

We have free time on Saturday and can help a lot i think

Thanks

sosoxuc:
Want to thank @zgnachvi one more time

tananaev:
I want to keep old API for the next release to give users some time to switch, so we need to implement REST alongside the old version. But at the same time it should be as independent as possible and allow for easy removal of the old API.

sosoxuc:
I think it's OK

So on Saturday we will implement REST

Would be great to ask @guterresrafael for help because he has already done most of the work

guterresrafael:
Hi,
I tested the jboss RESTEasy but could not support HATEOAS.
What JAX-RS implementation will be used? any preference?

tananaev:
I was thinking about Jersey.

sosoxuc:
The same for us :smile: 
We also consider Jersey

sosoxuc:
Hi @tananaev 

Here is the first REST method by @zgnachvi

https://github.com/sosoxuc/traccar/commit/f612e1c094e3e32d8d5b2dfa5abd8caf716b7c70

Please review code if style is good and code is valid

Thanks

sosoxuc:
P.S. We want to translate UI on Georgian
I have added i10n file manually but want to do this on https://www.transifex.com
Need your approval

tananaev:
I reviewed the commit and here are the comments:

1. Please use "rest" instead of "jaxrs" everywhere. It would make code more clear for people who don't know what JAX-RS is.
2. Please remove "jersey.version" from maven file. Only use variable if you need several dependencies with the same version (like Jetty).
3. RESP API should be a replacement of the old API and it should work with web interface as well. I think we need a config parameter to switch between old API and REST (not via "web.type").
4. I don't think login is a good example of REST. Also, I would rename it to "session" as REST should operate with resources.

First two things are very minor.

P.S. I added you on Transifex

zgnachvi:
Hi @tananaev , thx for review :)
i replaced jaxrs with rest and added config parameter for api -api.provider
I agree you about item 4, but this time, i think, it's not critical. We can change REST URL any time :)
Here is commit ttps://github.com/sosoxuc/traccar/commit/2b34fd60dfd9697d9e52b4587b0b6f8d248a8145

sosoxuc:
As I guess @zgnachvi did not changed methods at all, He just migrated on JAX RS, so if one changes config parameter, web must work with JAX RS as it did

guterresrafael:
@tananaev, in my last commit I removed http methods support, rollback all servlets and move all classes of api to package org.traccar.api.
I will finish implement support for roles (@RolesAllow) and refactor servlets to resources. What do you thing of creating a table to user roles? Please review the code and see if that was.

@zgnachvi @sosoxuc 
restfull api not use sessions.
we need to organize ourselves to not do the same things.

tananaev:
@guterresrafael, I believe we should allow both cookie sessions (because it makes web client code easier) and regular HTTP authentication. But if it's too hard to implement sessions we can drop support for it.

I don't think we need a separate table for roles, but again I don't know much about implementation.

I will review code as soon as I find some free time.

tananaev:
In general I like the implementation. I'll try to make it less "generic" because I think it makes the code more complicated than it should be. As soon as I find time I'll proceed with the merge.

tananaev:
I did a lot of refactoring before committing the changes. More changes are coming soon, but I think the base for REST API is there.

PR #1559, opened 2015-11-27
rest api -filters, error handlers, MainResource

sosoxuc:
As I guess @zgnachvi did not changed methods at all, He just migrated on JAX RS, so if one changes config parameter, web must work with JAX RS as it did


tananaev:
I can see there is some new commits that are not relates to REST API.

tananaev:
I think we'll go with implementation from #1542.

tananaev:
I have merged the REST API implementation.

PR #1601, opened 2015-12-17
create a PositionUtil class to process text templates

ivanfmartinez:
This class can be used as base to WebDataHandler, and also provide easy way to generate messages to be used on alerts.


tananaev:
It would nice to have something even more generic to be able to use it for database queries templates.

ivanfmartinez:
I can split the method to separate the replacing from the fields definition.
But the use in database queries should be limited, because of possible injections that can be made.
In my mind to use in database it should an different way using PreparedStatements not just string replace...

tananaev:
I know that it shouldn't be just string replacement. That's why I'm saying that the code should be more generic to support parsing needed for database queries.

tananaev:
Closing for now. Please reopen if you decide to update the code.

PR #1606, opened 2016-01-04
Commands sending integration

vitalidze:
I want to implement a possibility to send commands from my Traccar Web UI mod. Some time ago I have already sent a pull request that allows to run both `traccar-web.war` and original traccar API for the same purpose. However, during implementation I have faced following issues:

* HTTP sessions are not shared between API and web app deployed from `traccar-web.war`. I suggest to share them.
* permissions manager object is required to validate permissions on the API side. However, it is not being initialized in 'old' web app mode. So it should be initialized if any web app is going to be started.
* the key used to look up for a user identifier in HTTP session changed from 3.2 (from `user` to `userId`). For the compatibility purpose I think that, if possible, we can stick with the `userId` attribute name and don't change it in future and if the storage format will change then leave the attribute name, and introduce new one as needed.

What do you think? I can submit a pull request with the changes. Actually they are very small, say "cosmetic" changes.

### DIFF
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -168,10 +168,7 @@ public final class Context {
         }
 
         if (config.getBoolean("web.enable")) {
-            if (config.getString("web.type", "new").equals("new")
-                    || config.getString("web.type", "new").equals("api")) {
-                permissionsManager = new PermissionsManager(dataManager);
-            }
+            permissionsManager = new PermissionsManager(dataManager);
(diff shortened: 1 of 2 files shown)


vitalidze:
I have just attached changes to this ticket. Please observe and let me know if you have any questions.

tananaev:
Looks good. Thanks for pull request.

vitalidze:
Thank you! Hope I will integrate your new functions in next release :+1: 

Happy New Year!

PR #1723, opened 2016-02-12
Sending custom commands to Minifinder protocol

hanselsen:
This is my first Pull request so sorry for any errors.

Ciao. Hans


tananaev:
So, how do you send this command?

hanselsen:
These changes are the minimum needed for sending commands to the minifinders. I plan on sending it using the API.

tananaev:
I don't think it's going to work like this. You need to provide device id in the command and then you need to convert rest of the command model into a valid device format.

hanselsen:
I want to use the '/raw' endpoint. That way I can provide a deviceId and a
'raw' string.
The StringEncoder is needed anyhow, because it needs to be send in HEX.

Op za 13 feb. 2016 om 07:12 schreef Anton Tananaev <notifications@github.com

tananaev:
Raw commands are not supported in the new API.

hanselsen:
I use the old API because I use the UI mod from  http://traccar.litvak.su/.
Would that be an issue?

Op za 13 feb. 2016 11:45 schreef Anton Tananaev <notifications@github.com>:

tananaev:
I plan to remove old API eventually.

hanselsen:
Ok, that's a shame and that also means that I am going to use the new API.
There are two solutions I can think of.

1. Implement every command that I'm going to send. But then I'll still need
to have payload.

2. Add a RawCommand class or something. And somehow make it possible to
send this too.

For example, I want to setup my Minifinders using GPRS with a update
interval. This would look like: 123456M,[seconds].

What would you advise, because I want to help you get a very generic
application as well.

H

Op za 13 feb. 2016 13:08 schreef Anton Tananaev <notifications@github.com>:

PR #1755, opened 2016-02-26
Handle OPTIONS preflight for CORS, FIX issue 1751

rsmithsa:
* Change to CORS handling to handle pre-flight OPTIONS requests.
* Fix to return origin correctly in the "Access-Control-Allow-Origin" header.
* Returns the origin rather than "*" to allow basic auth to function. This behavior can be disabled by setting "web.origin" to any value which will never match a hostname sent by the browser. e.g. "<entry key='web.origin'>DISABLED</entry>"


4nonym0us:
:+1:

tananaev:
I have merged the changes with minor modifications. Thanks for pull request.

tananaev:
This change actually broke the REST API. Pre-flight OPTIONS fails with:

> Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.

tananaev:
After more investigation, it looks like it got broken somewhere else.

LucasBSC:
Is it broken still? Regards

PR #1794, opened 2016-03-17
Minifinder needs a stringEncoder as well as an ObjectEncoder

hanselsen:
MiniFinderProtocol should be complimented by the stringEncoder


tananaev:
Thanks for the pull request.

Issue #1859, opened 2016-04-09
MessageBodyWriter not found problem


ulion:
current master branch code, after build, the web ui call to server api failed with 500 status.

in the wrapper log:

```
INFO   | jvm 1    | 2016/04/09 07:33:05 | Apr 09, 2016 7:33:05 AM org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor aroundWriteTo
INFO   | jvm 1    | 2016/04/09 07:33:05 | SEVERE: MessageBodyWriter not found for media type=application/json, type=class org.traccar.model.Server, genericType=class org.traccar.model.Server.
INFO   | jvm 1    | 2016/04/09 07:33:05 | Apr 09, 2016 7:33:05 AM org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor aroundWriteTo
INFO   | jvm 1    | 2016/04/09 07:33:05 | SEVERE: MessageBodyWriter not found for media type=application/json, type=class java.util.HashMap, genericType=class java.util.HashMap.
INFO   | jvm 1    | 2016/04/09 07:33:05 | Apr 09, 2016 7:33:05 AM org.glassfish.jersey.server.ServerRuntime$Responder process
INFO   | jvm 1    | 2016/04/09 07:33:05 | SEVERE: Error occurred when processing a response created from an already mapped exception.
```

I used NetBeans built the jar on osx. it worked, but somehow it stop working any more, no matter which commit I checked out to build.

Any idea what's wrong with it?

tananaev:
Web UI is not an official Traccar project. Please contact author for support.

ulion:
All related code are in this repo, am I wrong?

2016-04-09 12:54 GMT+08:00 Anton Tananaev <notifications@github.com>:

> Web UI is not an official Traccar project. Please contact author for
> support.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/tananaev/traccar/issues/1859#issuecomment-207706356>
>



-- 
Ulion

tananaev:
Why are you using glassfish container? Please read this:

https://www.traccar.org/build/

ulion:
In the pom.xml file, jersey-media-json-jackson package is depended, and inside it (v2.22.2) it use glassfish container.

tananaev:
I guess you need to debug to find out why it's happening.

ulion:
[...]

PR #1860, opened 2016-04-09
Fix new api java<->json conversion problem #1859

ulion:
According to the jackson doc: https://jersey.java.net/documentation/latest/user-guide.html#json.jackson

I added one line to the code and resolved the problem.


tananaev:
Thanks for pull request. I still don't quite understand why it works fine for me, but not for you.

PR #1893, opened 2016-04-21
With itemclick listener it's possible to jump to the device even if t…

gaborgsomogyi:
…here is only one

### DIFF
--- a/web/app/view/Devices.js
+++ b/web/app/view/Devices.js
@@ -116,7 +116,7 @@ Ext.define('Traccar.view.Devices', {
     }],
 
     listeners: {
-        selectionchange: 'onSelectionChange'
+        itemclick: 'onItemClick'
     },
 
     columns: [{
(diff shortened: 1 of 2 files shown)


tananaev:
There is a button to center selected object. This seems like workaround other that a correct solution. Also, it's not consistent with the rest of the grids.

gaborgsomogyi:
Should this button should work immediately? As far as I've seen it jumps only if an update comes from the device.

tananaev:
Currently it doesn't, but it makes sense to fix that button to jump immediately.

gaborgsomogyi:
If that would jump immediately then this modification is useless :)

tananaev:
Are you interested in implementing the change?

gaborgsomogyi:
Sure, originally I'm java and C++ dev but I'm sure it's not brain surgery

PR #1894, opened 2016-04-21
Zoom to the streets when showing a device just like on the android ve…

gaborgsomogyi:
…rsion


tananaev:
Many people including myself get annoyed when map changes zoom automatically every time. So, if we implement it, there must be a way to opt-out.

Hardcoded value for zoom is also not a good idea.

gaborgsomogyi:
Hmmm, seems like I didn't fully understand the zoom in the Account till now. If I set this properly then the result will be perfect.

PR #1898, opened 2016-04-23
H02 protocol arm and disarm commands

gaborgsomogyi:


tananaev:
Thanks for pull request. Overall it looks good. Just a few comments:

1. Why do we need `getActualDateTime` method?
2. Why do we need to test that method?
3. I don't want to add mockito just to test one method.

gaborgsomogyi:
Thanks for the comments.

1. The code is not testing but mocking the time which is in the packet (give back hardcoded value in the test)
2. If the time is always the actual time how would you test the packet content? Checking the string without time part?
3. In general without mocking only integration tests can be done (2-3 instance behavior together)

tananaev:
If you want to make it testable, you can have `encodeCommand` method take time as an argument. It would be much cleaner solution and mockito is not required.

gaborgsomogyi:
Thrown away mockito.

tananaev:
Merged. Thanks.

gaborgsomogyi:
Thanks for the opportunity. I will implement the remaining commands for this protocol. But the next step would be to reach them from the web side. Can we fill somehow the command list based on BaseProtocol.supportedCommands or something similar?

PR #1899, opened 2016-04-23
H02 engine start/stop and periodic update commands

gaborgsomogyi:


tananaev:
What do those numbers mean?

tananaev:
I have done some minor refactoring to be more consistent with other code.

Please run checkstyle, findbugs and pmd checks on your code before submitting pull request. You can do it simply by running `mvn site` command.

gaborgsomogyi:
OK, next time I'll execute the analyzers.

gaborgsomogyi:
12. start cut off fuel and electric S20

* XX, YYYYYYYYYY, S20, HHMMSS, C, time1, time2,...Time30 #

C: the ultimate power

C = 0: dynamic cut off fuel and electric.detect the engine instantaneous operation, cut off the

circuit 5 seconds when the engine is running , forcing the engine stopped.

C = 1 or other digital: static cut off fuel and electric. Engine is not detected, the power relay a

lways suck close that Circuit is disconnected.

Dynamic cut off electric does not consume battery power, the engine must be connected to th

e signal (charge indicator) . If you don’t want to connect the signal of engine can use the way of static cut off fuel and ele

ctric.

When the early warning alert, ban to start engine function way is only use dynamic cut off fu

el and electric, so if you want to support a fortification function (handle or fortify the remote

control)

must be connected to the engine signals.

Time: the duration of the action, scope 1 ~ 30, units are seconds, outside the range as5

Example:*TH,0000000000,S20,130305,1,3,10,3,5,5,3,5,3,5,3,5# ;

Car receiver first check set_flag after receiving the command , if it not support off the oil, wi

ll return information directly

*TH,2020916012,V4,S20,ERROR,130305,050316,A,2212.8745,N,11346.6574,E,14.28,028,

220902,FFFFFBFF#

Restore the cutting of oil set the the time1 on S20 command as the “0” , It will reply :

*TH,2020916012,V4,S20,OK,130305,050316,A,2212.8745,N,11346.6574,E,14.28,028,220902,F

FFFFBFF#

gaborgsomogyi:
This is a snippet from the specification. I've sniffed the competitor webpages and they're sending also this pattern :)

PR #1920, opened 2016-05-02
Command framework

gaborgsomogyi:
Can we add something like this or shall we just add further commands statically to the UI? Actually it would be good to have Arm and Disarm commands.


strasharo:
Arm and disarm would be very helpful indeed.

tananaev:
I like the idea in general. I was actually thinking about implementing something like this myself.

Some comments:

1. I don't think we need `database.selectLatestPosition`query.
2. Do we actually need `SupportedCommand` model?
3. I don't think we need separate `name` parameter. We can just use command type. Back-end doesn't have to know about localization string ids.

gaborgsomogyi:
1. Yeah, right as I see the connection manager loads all the positions. I'll refactor it.

gaborgsomogyi:
2 and 3. How do you it think would be better? In the dialog we have 2 fields at the moment:
            store: 'SupportedCommands',
            displayField: 'name',
            valueField: 'key',
If we don't have SupportedCommands and separate fields how can we differentiate human readable and backend key things? Most probably that's my poor extjs knowledge :)

tananaev:
2. We still need a store, but do we need a separate model?

3. You can use same key (command type) for both. I don't see why we need two field that basically return the same information.

gaborgsomogyi:
1. No model is necessary. As I've seen in the docu the fields config option in the store creates internally a model
2. I've made the adaptation but it's not working (backend sends back the supported commands but the combo does not show it). Can you help me out?
3. With the actual implementation there is no i18n. It could come if the data is in the combo

tananaev:
I'll take a look when I have some time.

gaborgsomogyi:
Wow, seems like the data is there and can be selected but the characters are minimized. Maybe the displayField is not applied somehow.

gaborgsomogyi:
Now I see the problem. The backend side model is a must. I'm fixing it.

tananaev:
Can you rename `SupportedCommand` to `CommandType`? I think it makes more sense that way.

gaborgsomogyi:
Yeah, it's on the way. But I don't find any clean solution for i18n. Ideas?

tananaev:
You can load strings in the store.

gaborgsomogyi:
Yeah but I have only 1 field. Is it possible to add an extra field which comes from i18n?

tananaev:
I don't see why not. You can add as many new fields in JavaScript as you want.

gaborgsomogyi:
I'm just in a ramp-up in the extjs area. I just didn't know whether it's possible.

gaborgsomogyi:
Now it works like a charm :)

gaborgsomogyi:
One thing is missing. Not all the commands have i18n entry.

tananaev:
Translations can be added later. There is another problem. How would command that require additional parameters work?

gaborgsomogyi:
About the parameter I was thinking a lot. At the moment I see mainly 2 options:

1. Argument types like "frequency" can be sent through together with the type
2. There would be an additional service where the UI can request what kind of arguments are required by the command

I personally prefer the 1. solution. CommandType can be like a function signature in the java world, function name + parameters. The UI must read this data and the fields can be dynamically added.

tananaev:
I think the easiest option is to basically hard-code it in the web interface. There won't be too many commands and parameters for one type are always the same no matter what the protocol is.

Thank you very much for the pull request. I'll review it as soon as possible and if everything is fine, I'll merge it.

tananaev:
I've finished the review. Mostly looks good.

One thing is that I think you shouldn't change `ServerManager` class. Instead you can use `ActiveDevice` class to get protocol.

Another very minor comment - I think you should rename `key` to `type` or something in `CommandType` model. It made sense when you had `key` and `value`, but now just `key` looks a bit weird and confusing. 

Everything else looks good. Thanks again for contributing to the project.

gaborgsomogyi:
OK, I'll do the changes...

About the parameters basically you are right, the easiest way is to hardcode it. If parameter type is X then parameters Y are required.
On the other hand I've made a small proof of concept. Namely the commandtypes resource now sends back something like this:

[{"key":"engineStop","params":[]},{"key":"positionPeriodic","params":["frequency"]},{"key":"alarmDisarm","params":[]},{"key":"alarmArm","params":[]},{"key":"engineResume","params":[]}]

With this parameters can be turned on per command and protocol.
If all the protocols are using the exact same parameters for a specific command then this flexibility is not needed.

tananaev:
I think hard-coded parameters would do for now. Let me know when modifications are finished and I'll merge the request.

gaborgsomogyi:
Yeah, that's right the more I implementing this dynamic thing the more I have the feeling it's an overkill. The code is just compiling...

gaborgsomogyi:
The changes are there. Just came to my mind the command response topic. At the moment the commands are fire and forget things. Is it possible to receive packets which are not position related?

gaborgsomogyi:
In case of alarm on/off the user doesn't know whether the command was successfully executed or failed.

tananaev:
Merged. Thanks for the pull request.

As for command response, please create a separate thread to discuss it. It's a complicated issue.

PR #1941, opened 2016-05-12
Fix "Requests to this API must be over SSL" when use API Key

orcoliver:
When is set geocoder google and `geocoder.key` , the API response is:

```json
{
   "error_message" : "Requests to this API must be over SSL. Load the API with \"https://\" instead of \"http://\".",
   "results" : [],
   "status" : "REQUEST_DENIED"
}
```


tananaev:

orcoliver:
Thanks to you for this great work and support.

PR #1947, opened 2016-05-16
Implemented few commands for Wondex protocol.

Abyss777:
I've implemented few commands for Wondex protocol. It was inspired by implementation of "TotemProtocolEncoder". As I understood "KEY_DEVICE_PASSWORD" is not implemented neither in database nor in GUI, and I hardcoded it like in Totem.

### DIFF
--- a/src/org/traccar/protocol/WondexProtocol.java
+++ b/src/org/traccar/protocol/WondexProtocol.java
@@ -32,8 +32,7 @@ public class WondexProtocol extends BaseProtocol {
         super("wondex");
         setSupportedCommands(
                 Command.TYPE_REBOOT_DEVICE,
-                Command.TYPE_POSITION_SINGLE,
-                Command.TYPE_CUSTOM);
+                Command.TYPE_POSITION_SINGLE);
     }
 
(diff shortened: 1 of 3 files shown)


tananaev:
You don't need to implement `TYPE_CUSTOM` command. It's automatically supported by all protocols.

Abyss777:
OK. I've removed it.

tananaev:
Please run `mvn site` and fix any issues reported by checkstyle and other tools in your code. For example, I can see you are using tabs instead of spaces in a few places.

Abyss777:
Sorry for my code style issues. I believe next time I will be more careful. 

PS: Please add this parameter to checkstyle.xml to avoid false complains about "End line at the end of file" on Windows platform
```
-    <module name="NewlineAtEndOfFile"/>
+    <module name="NewlineAtEndOfFile">
+        <property name="lineSeparator" value="lf" />
+    </module>
```

tananaev:
Merged. Thanks for the pull request

Please send me the checkstyle modifications as a pull request as well.

PR #1951, opened 2016-05-17
Lock EOL to "lf" to avoid warnings on windows platform

Abyss777:
It is need to lock EOL to "lf" because there is false warning about absent EOL at the end of file on windows  platform.


tananaev:

PR #1953, opened 2016-05-17
Custom command support for web interface

Abyss777:
1. Added TYPE_CUSTOM to list of supported commands for all protocols
2. Extended Web UI

### DIFF
--- a/src/org/traccar/BaseProtocol.java
+++ b/src/org/traccar/BaseProtocol.java
@@ -42,6 +42,7 @@ public abstract class BaseProtocol implements Protocol {
 
     public void setSupportedCommands(String... commands) {
         supportedCommands.addAll(Arrays.asList(commands));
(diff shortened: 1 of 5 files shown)


tananaev:
Looks good. The only thing is that Traccar uses Transifex for localization, so you need to add English string first, and then when it's merged go to the Transifex platform and do a translation there:

https://www.transifex.com/traccar/traccar

Abyss777:
I've removed Russian string and added more English strings that autogenerated by https://github.com/tananaev/traccar/blob/master/web/app/store/CommandTypes.js#L38

I've tried to repeat your style e.g. all words capitalized. Added verb "set" for `sosNumber` and `silenceTime` and I doubt about verb for `voiceMessage`

tananaev:
Merged. Thanks for the pull request.

PR #1963, opened 2016-05-20
Store command result to position for wondex

Abyss777:
Firstly I've made a device identification on ping packet (because it is a binary I convert it in the framedecoder to string and pass to the protocoldecoder). It helps to avoid problems then command results came before any positions received and helps to better handle device online status.

Secondly I store command result to "result" field.

And Fixed wrong class name in WondexFrameDecoderTest

It is better to identify device in framedecoder (and don't answer on ping packets) but there is not `remoteAddress` to pass `identify` function.

### DIFF
--- a/src/org/traccar/MainEventHandler.java
+++ b/src/org/traccar/MainEventHandler.java
@@ -51,7 +51,7 @@ public class MainEventHandler extends IdleStateAwareChannelHandler {
             s.append("speed: ").append(String.format("%.1f", position.getSpeed())).append(", ");
             s.append("course: ").append(String.format("%.1f", position.getCourse()));
             if ((position.getAttributes().get(Event.KEY_RESULT) != null)) {
-                s.append(", result:").append(position.getAttributes().get(Event.KEY_RESULT));
+                s.append(", result: ").append(position.getAttributes().get(Event.KEY_RESULT));
             }
             Log.info(s.toString());
 
--- a/src/org/traccar/MainEventHandler.java
+++ b/src/org/traccar/MainEventHandler.java
@@ -50,8 +50,9 @@ public class MainEventHandler extends IdleStateAwareChannelHandler {
             s.append("lon: ").append(String.format("%.5f", position.getLongitude())).append(", ");
             s.append("speed: ").append(String.format("%.1f", position.getSpeed())).append(", ");
             s.append("course: ").append(String.format("%.1f", position.getCourse()));
-            if ((position.getAttributes().get(Event.KEY_RESULT) != null)) {
-                s.append(", result: ").append(position.getAttributes().get(Event.KEY_RESULT));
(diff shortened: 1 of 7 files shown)


src/org/traccar/protocol/WondexFrameDecoder.java
```diff
@@ -40,6 +42,9 @@ protected Object decode(
             if (channel != null) {
                 channel.write(frame);
             }
+            // Pass deviceId to protocol decoder
+            long deviceId = ((Long.reverseBytes((frame.getLong(0)))) >> 32) & 0xFFFFFFFFL;
+            return ChannelBuffers.copiedBuffer("$ID:" + String.valueOf(deviceId), StandardCharsets.US_ASCII);
```
tananaev: This is not a correct way of handling this situation. If you need this frame, you need to pass it to decode as is.

tananaev:
Commented in the code. If you need to handle both binary and strings, you should remove `StringDecoder` from the pipeline and handle frames in the protocol decoder.

Abyss777:
"remove StringDecoder" is a key.
Firstly I'd tried to do it right way, but StringDecoder confused me. 

PS: it was before I better understand arcitecture of traccar (pipelines and handlers)

Abyss777:
And I think it is good idea to write command result to log

src/org/traccar/protocol/WondexProtocolDecoder.java
```diff
@@ -59,23 +61,26 @@ public WondexProtocolDecoder(WondexProtocol protocol) {
     protected Object decode(
             Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
 
-        if (((String) msg).startsWith("$ID:")) {
-            identify(((String) msg).substring(4), channel, remoteAddress);
+        if (((ChannelBuffer) msg).getUnsignedByte(0) == 0xD0) {
```
tananaev: Create a variable of type ChannelBuffer as in any other binary decoder.

src/org/traccar/MainEventHandler.java
```diff
@@ -49,6 +50,9 @@ public void messageReceived(ChannelHandlerContext ctx, MessageEvent e) {
             s.append("lon: ").append(String.format("%.5f", position.getLongitude())).append(", ");
             s.append("speed: ").append(String.format("%.1f", position.getSpeed())).append(", ");
             s.append("course: ").append(String.format("%.1f", position.getCourse()));
+            if ((position.getAttributes().get(Event.KEY_RESULT) != null)) {
```
tananaev: Get result once for better performance.

tananaev:
Some more very minor comments.

Abyss777:
OK. Done.

tananaev:
Merged. Thanks.

PR #1978, opened 2016-05-26
Simplification of the hierarchy of models classes

Abyss777:
I've decided to do it with separate pull requests to better changes tracing. (Here a lot of files has been changed)


tananaev:
Merged. Thanks.

PR #1987, opened 2016-06-01
Disable command button if device is not online

Abyss777:
Here is some small fool-tolerance fix that disables the command button on the devices panel if a device is not online (or become not online)

### DIFF
--- a/web/app/view/DevicesController.js
+++ b/web/app/view/DevicesController.js
@@ -30,6 +30,11 @@ Ext.define('Traccar.view.DevicesController', {
                     selectDevice: 'selectDevice',
                     selectReport: 'selectReport'
                 }
+            },
+            store: {
+                '*': {
+                    update: 'onUpdateDevice'
+                }
             }
         }
     },
@@ -117,5 +122,9 @@ Ext.define('Traccar.view.DevicesController', {
         if (position !== undefined) {
             this.getView().getSelectionModel().deselectAll();
         }
+    },
+
+    onUpdateDevice: function () {
+        this.view.fireEvent('selectionchange',this.view.getSelectionModel());
     }
 });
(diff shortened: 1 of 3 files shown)


tananaev:
It doesn't seem like a right approach. You can register listener inside `DevicesController` class. You don't need to modify the store.

Abyss777:
I know that it is not very good to do it in `store` but I could not find any events in `Ext.grid.Panel` or `Ext.grid.column.Column` that fired on store update.

tananaev:
Check this example:

https://github.com/tananaev/traccar/blob/master/web/app/view/StateController.js#L32

Abyss777:
Shame on me, miss right solution.

web/app/view/DevicesController.js
```diff
@@ -30,6 +30,11 @@ Ext.define('Traccar.view.DevicesController', {
                     selectDevice: 'selectDevice',
                     selectReport: 'selectReport'
                 }
+            },
+            store: {
+                '*': {
```
tananaev: You need to call it only for device store, not for all the stores.

web/app/view/DevicesController.js
```diff
@@ -117,5 +122,9 @@ Ext.define('Traccar.view.DevicesController', {
         if (position !== undefined) {
             this.getView().getSelectionModel().deselectAll();
         }
+    },
+
+    onUpdateDevice: function () {
+        this.view.fireEvent('selectionchange',this.view.getSelectionModel());
```
tananaev: Can't we just call `onSelectionChange` method directly? Also, space is missing between arguments.

Abyss777:
I thought fire event would be in the spirit of asynchronous nature of javascript.

tananaev:
One last small thing. Please use `this.getView()` instead of `this.view`.

Abyss777:
Hope next time my pull requests need less correction.

tananaev:
Merged, thanks.

PR #1992, opened 2016-06-02
Show events in web-interface with help of toasts

Abyss777:
I think I've managed to find a way how send a complete data (an event and a position) to show in the  web-interface.


web/app/store/Events.js
```diff
@@ -0,0 +1,25 @@
...
+
+Ext.define('Traccar.store.Events', {
+    extend: 'Ext.data.Store',
+    model: 'Traccar.model.Event',
+
+    proxy: {
+        type: 'rest',
+        url: '/api/events'
+    }
+});
```
tananaev: New line is missing. By the way, there is style and issues check for JavaScript code as well. Use `jshint` and `jscs`.

web/app/controller/Root.js
```diff
@@ -126,6 +126,35 @@ Ext.define('Traccar.controller.Root', {
                     }
                 }
             }
+
+            if (data.events) {
+                array = data.events;
+                store = Ext.getStore('Events');
+                for (i = 0; i < array.length; i++) {
+                    store.add(array[i]);
+                    var text;
```
tananaev: All variables should be defined at the beginning of the method.

src/org/traccar/api/AsyncSocket.java
```diff
@@ -45,7 +47,9 @@ public AsyncSocket(long userId) {
     public void onWebSocketConnect(Session session) {
         super.onWebSocketConnect(session);
 
-        sendData(KEY_POSITIONS, Context.getConnectionManager().getInitialState(userId));
+        Map<String, Collection<?>> data = new HashMap<String, Collection<?>>();
```
tananaev: Traccar targets Java 7 and higher, so you can use diamond operator on the right hand side.

tananaev:
Added some comments.

Abyss777:
Fixed. Thanks for the tip about JSHint.

tananaev:
Merged. Thanks.

PR #2003, opened 2016-06-06
Fix connection check sql

drecchia:


tananaev:
Please use spaces instead of tabs.

tananaev:
Do we need to call `setConnectionInitSql`? Or is it optional?

drecchia:
It defines the behavior to test the retrieved connection, If you remove, you wont have the feature.

tananaev:
How about we define `SELECT 1;` as a default value (second parameter in `config.getString`), so we don't need to define it in every config. But it would still be possible to override it if required.

PR #2012, opened 2016-06-10
Geofences

Abyss777:
I reorganized a model hierarchy a bit.
Added two type of a geofences geometry.
Added an event handler for a geofences
The geofence handling happens in GeofenceManager, they cached there in different ways.
Added permissions to user for geofences.
Added ability to link the geofence to a group. 
Added ability to link the geofence to a device per user.
Added API to handle geofences, links and permissions (tested partially)

### DIFF
--- a/schema/changelog-3.6.xml
+++ b/schema/changelog-3.6.xml
@@ -19,6 +19,7 @@
       </column>
       <column name="deviceid" type="INT" />
       <column name="positionid" type="INT" />
+      <column name="geofenceid" type="INT" />
       <column name="attributes" type="VARCHAR(4096)">
         <constraints nullable="false" />
       </column>
@@ -29,6 +30,65 @@
     <addColumn tableName="devices">
       <column name="motion" type="VARCHAR(128)" />
     </addColumn>
+    <addColumn tableName="devices">
+      <column name="geofenceId" type="INT" />
+    </addColumn>
+
+    <createTable tableName="geofences">
+      <column name="id" type="INT" autoIncrement="true">
+        <constraints primaryKey="true" />
+      </column>
+      <column name="name" type="VARCHAR(128)">
+        <constraints nullable="false" />
+      </column>
+      <column name="description" type="VARCHAR(128)" />
+      <column name="area" type="VARCHAR(4096)">
+        <constraints nullable="false" />
+      </column>
+      <column name="attributes" type="VARCHAR(4096)">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <createTable tableName="user_geofence">
+      <column name="userid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="geofenceid" type="INT">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <addForeignKeyConstraint baseTableName="user_geofence" baseColumnNames="userid" constraintName="fk_user_geofence_userid" referencedTableName="users" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="user_geofence" baseColumnNames="geofenceid" constraintName="fk_user_geofence_geofenceid" referencedTableName="geofences" referencedColumnNames="id" onDelete="CASCADE" />
+
+    <createTable tableName="group_geofence">
+      <column name="groupid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="geofenceid" type="INT">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <addForeignKeyConstraint baseTableName="group_geofence" baseColumnNames="groupid" constraintName="fk_group_geofence_groupid" referencedTableName="groups" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="group_geofence" baseColumnNames="geofenceid" constraintName="fk_group_geofence_geofenceid" referencedTableName="geofences" referencedColumnNames="id" onDelete="CASCADE" />
+
+    <createTable tableName="user_device_geofence">
+      <column name="userid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="deviceid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="geofenceid" type="INT">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <addForeignKeyConstraint baseTableName="user_device_geofence" baseColumnNames="userid" constraintName="fk_user_device_geofence_userid" referencedTableName="users" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="user_device_geofence" baseColumnNames="deviceid" constraintName="fk_user_device_geofence_deviceid" referencedTableName="devices" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="user_device_geofence" baseColumnNames="geofenceid" constraintName="fk_user_device_geofence_geofenceid" referencedTableName="geofences" referencedColumnNames="id" onDelete="CASCADE" />
 
   </changeSet>
 </databaseChangeLog>
--- a/schema/changelog-3.6.xml
+++ b/schema/changelog-3.6.xml
@@ -30,9 +30,6 @@
     <addColumn tableName="devices">
       <column name="motion" type="VARCHAR(128)" />
     </addColumn>
-    <addColumn tableName="devices">
-      <column name="geofenceId" type="INT" />
-    </addColumn>
 
     <createTable tableName="geofences">
       <column name="id" type="INT" autoIncrement="true">
@@ -74,10 +71,7 @@
     <addForeignKeyConstraint baseTableName="group_geofence" baseColumnNames="groupid" constraintName="fk_group_geofence_groupid" referencedTableName="groups" referencedColumnNames="id" onDelete="CASCADE" />
     <addForeignKeyConstraint baseTableName="group_geofence" baseColumnNames="geofenceid" constraintName="fk_group_geofence_geofenceid" referencedTableName="geofences" referencedColumnNames="id" onDelete="CASCADE" />
 
-    <createTable tableName="user_device_geofence">
-      <column name="userid" type="INT">
-        <constraints nullable="false" />
-      </column>
+    <createTable tableName="device_geofence">
       <column name="deviceid" type="INT">
         <constraints nullable="false" />
       </column>
@@ -86,9 +80,8 @@
       </column>
     </createTable>
 
-    <addForeignKeyConstraint baseTableName="user_device_geofence" baseColumnNames="userid" constraintName="fk_user_device_geofence_userid" referencedTableName="users" referencedColumnNames="id" onDelete="CASCADE" />
-    <addForeignKeyConstraint baseTableName="user_device_geofence" baseColumnNames="deviceid" constraintName="fk_user_device_geofence_deviceid" referencedTableName="devices" referencedColumnNames="id" onDelete="CASCADE" />
-    <addForeignKeyConstraint baseTableName="user_device_geofence" baseColumnNames="geofenceid" constraintName="fk_user_device_geofence_geofenceid" referencedTableName="geofences" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="device_geofence" baseColumnNames="deviceid" constraintName="fk_user_device_geofence_deviceid" referencedTableName="devices" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="device_geofence" baseColumnNames="geofenceid" constraintName="fk_user_device_geofence_geofenceid" referencedTableName="geofences" referencedColumnNames="id" onDelete="CASCADE" />
 
   </changeSet>
 </databaseChangeLog>
--- a/src/org/traccar/BaseEventHandler.java
+++ b/src/org/traccar/BaseEventHandler.java
@@ -1,5 +1,7 @@
 package org.traccar;
 
+import java.util.Collection;
+
 import org.traccar.model.Device;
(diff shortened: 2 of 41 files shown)


ivanfmartinez:
I have used a very simplified implementation in my mqtt publishing, just with circles.

I have some cases where the areas (geofences in your case) overlaps. Your code just permit the user to be in a single geofence area each time.

Maybe you can adjust to make possible the overlay, and the in/out checking should be based in an state check based on previous position. 

This bring another problem that happens many times, with the Android devices. I have very bad reception signal in my area, inside my home the Android use cells for location instead of GPS. I'm in a hill and constantly it gets floating location from my current location to the cell tower area. Your code does not have option to work with this problem (like my code).

If this code become part of traccar I can rewrite my mqtt code that have my simple area locations and integrate directly in traccar.

I think if traccar starts using geospatial structures it should use an geospatial library for this like jts, geotools instead of creating own internal types and wkt processing. They will bring more dependencies but will be more expandable and future proof.

src/org/traccar/events/GeofenceEventHandler.java
```diff
@@ -0,0 +1,74 @@
...
+            return event;
+        }
+
+        Device device = dataManager.getDeviceById(position.getDeviceId());
+        if (device == null) {
+            return event;
+        }
+
+        Set<Long> geofences = geofenceManager.getAllDeviceGeofences(position.getDeviceId());
+        if (geofences == null) {
```
ivanfmartinez: I think it should control each geofence as separate in/out state, for doing this the BaseEventhandler should accept the return of an list of events instead of a single event. In this case will be possible to return more than one event like : 

EXIT : LOCATION 1
ENTER: LOCATION 2
ENTER: LOCATION 3

what you think about this ?

Abyss777:
Hello, @ivanfmartinez 
The problem of floating location partially handled with suppression of repeated event here and partially with filtering in other part of traccar I think.

I've tried geotools (fork of jts) to handle with geometry and it does not fit: it does not have circles. Too much calculations to just compare a distance between points with a radius.

About overlapping geofences... I don' think it is need to write too much code, but need changes in model and database. It is need to watch for ALL geofences of device, and it can be rather expensive if we have a lot of complicated areas.

Lets wait for @tananaev decision about this pull request, and if it is OK I'll try to implement it in the future.

Anyway, can you give an example then need overlapped areas?

schema/changelog-3.6.xml
```diff
@@ -29,6 +30,65 @@
...
+      <column name="geofenceid" type="INT">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <addForeignKeyConstraint baseTableName="group_geofence" baseColumnNames="groupid" constraintName="fk_group_geofence_groupid" referencedTableName="groups" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="group_geofence" baseColumnNames="geofenceid" constraintName="fk_group_geofence_geofenceid" referencedTableName="geofences" referencedColumnNames="id" onDelete="CASCADE" />
+
+    <createTable tableName="user_device_geofence">
+      <column name="userid" type="INT">
```
tananaev: Do we really need user in this table? You already have a connection between geofence and user in other table. Seems redundant and can lead to conflicts.

src/org/traccar/events/GeofenceEventHandler.java
tananaev: I agree that it should be possible to have multiple events. I'm guessing that some users would want to draw geofences next to each other and it's likely that object leaves one geofence and enters another one straight away.

Also, it should be possible to draw geofences on top of each other. It means that each object can be in multiple geofences at the same time.

src/org/traccar/geofence/GeofenceCircle.java
```diff
@@ -0,0 +1,82 @@
...
+    private double centerLongitude;
+    private double radius;
+
+    public GeofenceCircle() {
+        super();
+    }
+
+    public GeofenceCircle(String wkt) throws ParseException {
+        super();
+        fromWKT(wkt);
```
tananaev: Please use camel case. It should be `fromWkt`.

src/org/traccar/geofence/GeofencePolygon.java
```diff
@@ -0,0 +1,160 @@
...
+
+        int polyCorners = coordinates.size();
+        int i;
+        int j = polyCorners - 1;
+        double longitude360 = longitude + Coordinate.DEGREE360;
+        boolean oddNodes = false;
+
+        for (i = 0; i < polyCorners; j = i++) {
+            if (coordinates.get(i).getLon360() < longitude360
+                && coordinates.get(j).getLon360() >= longitude360
```
tananaev: Indentation should be 8 spaces instead of 4.

src/org/traccar/geofence/GeofencePolygon.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        if (!wkt.startsWith("POLYGON")) {
+            throw new ParseException("Mismatch geometry type", 0);
+        }
+        String content = wkt.substring(wkt.indexOf("(") + 1, wkt.indexOf(")"));
+        if (content == null || content.equals("")) {
+            throw new ParseException("No content", 0);
+        }
+        String[] commatokens = content.split(",");
```
tananaev: Please use camel case. Should be `commaTokens` or something.

src/org/traccar/model/Device.java
```diff
@@ -114,4 +114,13 @@ public void setMotion(String motion) {
         this.motion = motion;
     }
 
+    private long geofenceId;
+
+    public long getGeofenceId() {
```
tananaev: I'm guessing this should be a list.

tananaev:
Please update `debug.xml` config file as well. That's the one I use for development and it's always up to date.

In general code looks good. Left some comments there.

I don't think it's that important to handle floating location at this point. Can be an improvement idea for future.

If there are no good libraries, I think it's good to have own implementation. The code is really simple as far as I can see. We can always change it to use library if we can find a good one. Circle is not part of WKT standard I think, that's why it's not supported by libraries.

I think overlapping of geofences is required. I'm pretty sure people would want it. An example I can think of is that one set of geozones would be for example city suburbs and another set would be parking spots or something. So, parking would overlap with one of the suburbs.

schema/changelog-3.6.xml
Abyss777: I'll try to explain.
user_geofences - it is permissions - user can edit, delete and connect this geofences

user_device_geofence - it is link device-geofence, but per user. It solves the case then two users access to the same geofence but need to know enter/exit about two different devices.

For example: a sales company, a stockman need to be notified when comes semi-truck to warehouse to be prepared for unload it, sales manager need to be notified when other van leaved warehouse for delivery to call client. They both have access to "warehouse" geofence and both have access to semi-truck and van, but only one want to be notified about truck and other about van.

It just allow users to be flexible notified about only some of their devices entered/exited geofences.

src/org/traccar/model/Device.java
Abyss777: Should it be list for overlapping geofences?

ivanfmartinez:
"I've tried geotools (fork of jts) to handle with geometry and it does not fit: it does not have circles. Too much calculations to just compare a distance between points with a radius."

Geotools already have circle support : 
http://docs.geotools.org/stable/userguide/library/jts/geometry.html
Check "Custom Circles" .

The problem is if you instantiate a circle, you must define the number of points. But for this kind of work like checking if one point is inside a circle area the simple way is as you talked, check the distance of points:

don't instantiate a circle geometry, just calculate distance between circle center and the point you want to check, and check if distance is less than the range. 

As far I known officially the circle is not supported by wkt also.

The circle is an special case, that can be simplified or mapped, all others can be processed by an standard library directly.

schema/changelog-3.6.xml
tananaev: I think it's overkill. Plus it's not consistent with groups.

Regarding the example, stockman should have access to truck devices only, and salesman should have access to van devices only.

tananaev:
I agree that it would be nice to use a library, bet code seems really simple, so I think for now we can use our own implementation. It's easy to change anyway.

Abyss777:
I simplified device <-> geofence link.
Added support for overlapped geofences
Other style fixes

Made some base for notifications subsystem, because I think it is not good to blow `ConnectionManager` by inappropriate code. Logically send event notifications to web-client is still notifications, and should be made from something like `NotificationManager`

tananaev:
What about `user_device_geofence` table?

Abyss777:
Hm, I remember that added it.

tananaev:
Looks good now. Have you tested everything?

Abyss777:
Yes, but now I find out that https://github.com/tananaev/traccar/pull/2012/commits/6da14aded9775469a044d36e9a8ce5d8483f9003 did not help. Somewhere current device geofences is cleared and if device goes sleep in geofence it generate enter event again after wakeup. I need some time to fix it.

tananaev:
OK, let me know when everything is fixed.

Abyss777:
I've done two things:
getAllDevicesCached() should be used instead of getAllDevices() from everywhere. Only this objects will contain `geofencesIds`

Before now getDeviceById was called 10 times for every position, I've optimized to 8 times.

Now geofences work as should be.

PS: something need to be done with cached objects, I've faced to problems second time. Sometime it is inconsistent. May be it is need to write `DeviceManager`, `GroupManager`, move `lastPositions` cache from `ConnectionManager` somewhere and not to work with objects directly from DB.

Abyss777:
Sorry, missed Stack Overflow.

Abyss777:
I had been testing in two environment for a few hours: with test-generator.py and real devices.
Everything is OK without any errors and working as should.

src/org/traccar/BaseEventHandler.java
```diff
@@ -12,24 +14,32 @@ public boolean isLastPosition() {
         return isLastPosition;
     }
 
+    private Device device;
```
tananaev: Why are you storing in the event handler. I didn't actually notice it before, but `isLastPosition` shouldn't be here either. One handler can be used for different devices.

By the way, there is another typo that I just noticed. It should be `analyze` instead of `analize`.

src/org/traccar/database/DataManager.java
```diff
@@ -334,11 +346,34 @@ public void removeUser(long userId) throws SQLException {
                 .executeQuery(GroupPermission.class);
     }
 
-    public Collection<Device> getAllDevices() throws SQLException {
+    private Collection<Device> getAllDevices() throws SQLException {
         return QueryBuilder.create(dataSource, getQuery("database.selectDevicesAll"))
                 .executeQuery(Device.class);
     }
 
+    public Collection<Device> getAllDevicesCached() {
```
tananaev: Please explain the reason for this function. Why `getDeviceById` and `getDeviceByUniqueId` is not enough?

src/org/traccar/geofence/GeofenceCircle.java
```diff
@@ -0,0 +1,82 @@
...
+import org.traccar.helper.DistanceCalculator;
+
+public class GeofenceCircle extends GeofenceGeometry {
+
+    private double centerLatitude;
+    private double centerLongitude;
+    private double radius;
+
+    public GeofenceCircle() {
+        super();
```
tananaev: There is no need to call `super` explicitly. Same applies to other constructors.

tananaev:
Added some more comments. Also, can you please update/add license headers to all files except unit tests. Year should be 2016.

src/org/traccar/BaseEventHandler.java
Abyss777: `device` is always retrieved in `handlePosition` for every position just to determine if it is last
and retrieved again in 2 of 4 event handlers in `analyzePosition` why not to decrease `getDeviceById` calls

tananaev: Because what if it's called from different threads for different devices?

src/org/traccar/database/DataManager.java
Abyss777: `getAllDevices` were used in three places `GeofenceManager`, `api/resource/DeviceResource.java`
and `PermissionsManager` and return new collection of objects just created from db.
In first two places we do not need new clean objects without `geofencesIds` we need already filled objects with `geofenceIds`. They are already filled in `Map<> devicesById` and keep up to date by `GeofenceEventHandler`. Because of that we need function that return collection of cached devices `getAllDevicesCached()`, `getDeviceById` and `getDeviceByUniqueId` can't return whole collection to iterate or send with API.

src/org/traccar/BaseEventHandler.java
Abyss777: I do not argue, but may be I misunderstand something with OOP.
`anylizePosition` is called just after setting `device` in the same object instance in the same function. `device` can be different?

tananaev: Both can be different. That's exactly what I'm saying. Having `isLastPosition` there is also incorrect.

What can happen:

1. Thread 1 call method for device A, so handler assigns device A to the variable
2. Thread 2 call method for device B at the same time and re-assigns variable
3. Thread 1 continues execution for device A, but variable is already changed.

I know that's a very rare case, but it can definitely lead to problems.

Abyss777:
OK. Seems all fixed.

Cached objects is a separate topic for discussion.

Abyss777:
Two small additional fixes.

I don't understand how you can review such amount of code. If I as author miss a lot of issues.

warlockopy:
It will be a nice feature to link a geofence event with an schedule, example, geofence events are active between 10 pm, to 60 am, outside that schedule, events are ignored.

tananaev:
@warlockopy, it's definitely not going to be part of this pull request. I would recommend to create a separate feature request thread.

Abyss777:
I have some code for web-client to link/unlink geofences. Should I add it here or wait for approve and create new pull request?

tananaev:
@Abyss777, yes, please wait till we merge this request.

tananaev:
Merged, it looks fine. I'll test it and see if everything works fine for me.

Issue #1942, opened 2016-05-13
Secure Connection between GPS Client and Host


dschense:
Connection to the Webinterface could be secured over reverse Proxy with https. This is great and works like a charme.
But is it also possible to secure the Connection between the GPS Client (for me its a Android Mobile Device running the Traccar APP and connection over TCP Port 5055) and the Host running the Traccar instance?

The Connection looks like this:

```
2016-05-12 11:47:38  INFO: [36330907] connected
2016-05-12 11:47:39 DEBUG: [36330907: 5055 < #IPOfTheDevice] HEX: 
(This is not real. Just a example)
345634634gkdgekgegkegk34t4g34gß4kgß34kg3k3ßfk3ßfk3ßfk3ßfk3ßfk3ßtk3ßtk3ßk3ßtk3ßtk3dßkt3ßkdt3ßtk3ßtkd3ßtkd3ßtk3ßtk3ßtdk3ßtkg3ßtk3ßtk23tk34ß3ß6zj3ß6j3ß5j2ßd24k2´05k2´05kv2´5kv2´5kv2´5kv235kv34ßk6vb3ßzkvb3ßvb6k3ß6kv3ß6kvb3ß6kvb34ßkv3ßkct3ktß3kd3kdt3ükodäseoküghkwezwpjwzeüopzjwpjzw4jüjtzw3ßktd3ßk4dtwüßküktwpeojwrtwpejtzwjtfpüwjdpüwükdwoptfkweüzkweüzkowüozkwerjwwfeprjweijrzwüiejzüejzüwoezw+oekrzwerkfzwkozw+eokzw+zk+0zk+kzwk0ww304kzw30ktzw30ktzw3#tkw3tzk
2016-05-12 11:47:39 DEBUG: [36330907: 5055 > #IPOfTheDevice] HEX: 
(This is not real. Just a example)
345634634gkdgekgegkegk34t4g34gß4kgß34kg3k3ßfk3
2016-05-12 11:47:39  INFO: [36330907] disconnected
2016-05-12 11:47:22  INFO: [C9825705] id: 9, time: 2016-05-12 10:56:53, lat: 22.69108, lon: 27.67546, speed: 1.7, course: 22.1
```
It sends some Stuff to Host:5055

tananaev:
Traccar Client doesn't support secure connection at the moment, but it's certainly possible to add support for it. The problem is that it won't help with spoofing.

dschense:
[...]

PR #2016, opened 2016-06-11
 Secure Connection between GPS Client and Host #1942 

FX7:
This is putting the base in the backend for an extension of the OsmAndProtocol. Now the backend can use the key, stored for a device, to decode an request.

I havent made any extensions in the database scheme, nor any etensions in the server frontend, nor any extension of the client.

I hope with some help, this should become a working part ...

### DIFF
--- a/src/org/traccar/model/Device.java
+++ b/src/org/traccar/model/Device.java
@@ -15,7 +15,13 @@
  */
 package org.traccar.model;
 
+import java.security.MessageDigest;
+import java.util.Arrays;
 import java.util.Date;
+
+import javax.crypto.spec.SecretKeySpec;
+
+import org.traccar.helper.Log;
 
 public class Device {
 
@@ -99,6 +105,39 @@ public class Device {
 
     public void setGroupId(long groupId) {
         this.groupId = groupId;
-    }
+	}
 
+	private String cipherKey;
+
+	public String getCipherKey() {
+		return cipherKey;
+	}
+
+	public void setCipherKey(String cipherKey) {
+		this.cipherKey = cipherKey;
+	}
+
+	public SecretKeySpec getSecretKeySpec()
+	{
+		String keyString = getCipherKey();
+		if (keyString == null)
+		{
+			Log.warning("No cipherKey for device '" + getId() + "' set!");
+			return null;
+		}
+
+		try
+		{
+			byte[] key = (keyString).getBytes("UTF-8");
+			MessageDigest sha = MessageDigest.getInstance("SHA-256");
+			key = Arrays.copyOf(key, 16);
+			key = sha.digest(key);
+			return new SecretKeySpec(key, "AES");
+		} catch (Exception e)
+		{
+			Log.error("Exception '" + e.getMessage() + "' during generating SecretKeySpec for device '" + getId()
+					+ "'!");
+			return null;
+		}
+	}
 }
--- a/src/org/traccar/model/Device.java
+++ b/src/org/traccar/model/Device.java
@@ -19,6 +19,7 @@ import java.security.MessageDigest;
 import java.util.Arrays;
 import java.util.Date;
 
+import javax.crypto.SecretKey;
 import javax.crypto.spec.SecretKeySpec;
 
 import org.traccar.helper.Log;
@@ -117,10 +118,10 @@ public class Device {
 		this.cipherKey = cipherKey;
 	}
 
-	public SecretKeySpec getSecretKeySpec()
+	public SecretKey getSecretKey()
 	{
 		String keyString = getCipherKey();
-		if (keyString == null)
+		if (keyString == null || keyString.isEmpty())
 		{
 			Log.warning("No cipherKey for device '" + getId() + "' set!");
 			return null;
@@ -128,10 +129,11 @@ public class Device {
 
 		try
 		{
+			
 			byte[] key = (keyString).getBytes("UTF-8");
-			MessageDigest sha = MessageDigest.getInstance("SHA-256");
-			key = Arrays.copyOf(key, 16);
+			MessageDigest sha = MessageDigest.getInstance("SHA-1");
 			key = sha.digest(key);
+			key = Arrays.copyOf(key, 16);
 			return new SecretKeySpec(key, "AES");
 		} catch (Exception e)
 		{
--- a/src/org/traccar/protocol/OsmAndProtocolDecoder.java
+++ b/src/org/traccar/protocol/OsmAndProtocolDecoder.java
@@ -26,7 +26,8 @@ import java.util.Map.Entry;
 import java.util.Set;
 
 import javax.crypto.Cipher;
-import javax.crypto.spec.SecretKeySpec;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.IvParameterSpec;
 
 import org.jboss.netty.channel.Channel;
 import org.jboss.netty.channel.ChannelFutureListener;
@@ -51,6 +52,8 @@ public class OsmAndProtocolDecoder extends BaseProtocolDecoder {
 	public static final String PARAM_ID = "id";
 	public static final String PARAM_DEVICEID = "deviceid";
 	public static final String PARAM_CIPHER = "cipher";
+
+	public static final String CIPHER_TRANSFORMATION = "AES/CBC/PKCS5PADDING";
 
     public OsmAndProtocolDecoder(OsmAndProtocol protocol) {
         super(protocol);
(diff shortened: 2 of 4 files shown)


tananaev:
First of all, code does follow any of the code style requirements. Please run `mvn site` and make sure there are no new issues there.

Secondly, I don't really understand how it works. Does device send both key and encrypted message?

ivanfmartinez:
@tananaev  apparently the Android device send ID unencrypted, and all other data encrypted by the key. 

The server should be updated to have the key in the database. In my mind to use this the cipher type (hardcoded now as AES) should be also in database. Adding two fields to each device can make possible do define encryption type and key to be used and make possible to implement encryption in other protocols too.

The code just works for tests without adding key and creating UI to update the fields in the server side, but I think is a good idea and easy to implement in the server and in Android client at least.

FX7:
Hello and sorry that I may not be so quick responding. I hope to keep my response time in under one week ;-)

> First of all, code does follow any of the code style requirements. Please run mvn site and make sure there are no new issues there.
Im sorry. Maven is really new to me so I have to get into this tool. I will run this and hope I can fix all code issues.

As @ivanfmartinez mentioned: This is just the basic code for the issue. Main thing would be to get the enryption into the client. I think I can try this, but I dont have to much time :/
The idea of making the cipher type optional by device is a really good one. But I think for a first run, this is not essential.

tananaev:
Why not use standard HTTPS connection instead of re-inventing our own encryption?

ivanfmartinez:
@tanaev The https will exchange a lot of data just to stablish the secure connection. Considering that we need just few bytes with location information. I think that is a big overhead. For web browsing, and standard web usage https  the overhead is low, but in this specific topic is very high.

The code does not re-invent it just uses an existing algorithm, which make data secure (if no one get access to the encryption key). And according to Android documentation the cipher is available since level 1 API. The implementation probably will be easy in the Android side.

@FX7 making the cipher configurable make possible to use in different devices, as the java  on server probably have more options than the smaller Android devices.

FX7:
Also https needs lots of stuff to handle with (CAs, self-signed certs, expirering certs ...).
And as @ivanfmartinez said: I dont reinvent stuff, just using std encryption algorithm.

Sorry, that I dindt find time so far for cleaning up the code or even handle the client side. I hope I can do some of the stuff in this week.

tananaev:
I still think HTTPS is a better option. All you need is a certificate (can possibly be self signed, but ideally real one). Connection can be maintained to avoid overhead.

ivanfmartinez:
Lets encrypt make easy to create and maintain valid certificates, its a good solution for future implementations. My traccar server web interface is only accessible using https, using letsencrypt and nginx as reverse proxy.

Using self signed are not necessary. But the client should recognize the certificate authority. New devices accepts lets encrypt, older devices not, and some does not accept to install new root certificates.

How a connection can be maintained in a device that is moving(using different cells, the IP will be changing) and have limited battery ?

I never seen an implementation of SSL/TLS with disconnection and reconnection without negotiation . Do you have any reference to read about this ?

tananaev:
There is no reconnection. That's not how mobile network works.

ivanfmartinez:
In some cases I have seen the device using same IP in different locations, but sometimes the IP has changed (valid one not CGNAT or something like internal ip) in the same place. 

On server side, if the client does not send data for some time the server will disconnect by timeout. 

If is possible to have a single SSL negotiation and use the same connection for long time, without making the device awake to keep connection alive It will be really the best option. I did not know how to do this.

FX7:
Besides that there is some work on the client side to be done for a https connection, whats on the server side and where to "hook in"? How to get a protocol using https and still have access to the webinterface? Or is there already some way to do so?

ivanfmartinez:
On the server side a simple way to do the https is to use an reverse proxy that supports https, if clients support SNI you can share the IP with many apps just using different names. If clients does not support SNI you should have one IP for each SSL certificate.

On client  side if using the standard routines and a trusted certificate, there are no big changes to replace http to https.

Apparently all Android after 4.2 supports SNI for all java applications.

FX7:
I did this for the WebInterface. But can I also do this for the device protocol? But that would mean, I need one domain name for the webinterface like webinterface.tracking.domain and one for the device protocol like protocolX.tracking.domain. And last but not least, I would need a unique name for each protocol, right?

That configuration wouldnt need any server code or configuration changes. With server I mean the traccar server not the configuration of the webserver that is providing the proxy.

On the client side some code extensions have to be made. So that the client can handle https.

Are all my presumptions correct?

tananaev:
You can use ports with HTTPS as well. For example:

https://example.com:5055

Not sure if you can get a real SSL certificate for it though.

FX7:
SSL certificates dont no anything about ports. They are just for a domain.
Only the browser / client coult react to this and say "somethings maybe wrong". But I didnt know about such a "feature".

Ok, back to the point: This way I dont have to make any code changes on the traccar server side, correct? I have to config a proxy webserver to listen to the port and then redirect all traffic to traccar.

But the client has to be extended, to know about https, so there is some coding to do on this side, correct?

tananaev:
Yes, it looks like only Traccar Client has to be modified.

ivanfmartinez:
If the certificate is valid and recognized by the Android device, probably the change will be just make possible to select https intead of hardcoded http.

https://github.com/tananaev/traccar-client-android/blob/master/app/src/main/java/org/traccar/client/ProtocolFormatter.java

dschense:
u think its enough to change:

```
Uri.Builder builder = new Uri.Builder();
        builder.scheme("http").encodedAuthority(address + ':' + port)
```

to

```
Uri.Builder builder = new Uri.Builder();
        builder.scheme("https").encodedAuthority(address + ':' + port)
```

tananaev:
Should be enough.

dschense:
fair enough to try

dschense:
Ive tried this.
I changed the builder sheme to https and build the app.
then set up reverse proxy with nginx from incoming port with ssl to traccar serverport 5055 http.
but this does not work.

ivanfmartinez:
What was exactly the problem ? 

The nginx received the request ? Do you have used a port different of 5055 in nginx  for the SSL service ? 

If nginx did not received the request the problem is on the client. If nginx received the request maybe is a problem with your nginx configuration.

There are any error in Android logcat  ?

dschense:
I dont know yet.

nginx gets nothing. i used port 5550 for incoming and port 5055 for the reverse.

and this is what I get in logcat:

`06-21 16:49:22.490 2511-2511/org.traccar.client I/PositionProvider: provider location
06-21 16:49:22.490 2511-2511/org.traccar.client I/PositionProvider: backup provider stop
06-21 16:49:22.490 2511-2511/org.traccar.client I/PositionProvider: location new
06-21 16:49:22.491 2511-2511/org.traccar.client D/TrackingController: write (id:0 time:1466527762 lat:70.0 lon:6.0)
06-21 16:49:28.269 2511-2511/org.traccar.client D/TrackingController: retry
06-21 16:49:45.290 2511-2511/org.traccar.client I/PositionProvider: provider location
06-21 16:49:45.290 2511-2511/org.traccar.client I/PositionProvider: backup provider stop
06-21 16:49:45.290 2511-2511/org.traccar.client I/PositionProvider: location new
06-21 16:49:45.293 2511-2511/org.traccar.client D/TrackingController: write (id:0 time:1466527785 lat:70.0 lon:6.0)
06-21 16:49:58.274 2511-2511/org.traccar.client D/TrackingController: read
06-21 16:49:58.279 2511-2511/org.traccar.client D/TrackingController: send (id:1 time:1466527027 lat:70.0 lon:6.0)
06-21 16:50:13.354 2511-2511/org.traccar.client D/TrackingController: retry
06-21 16:50:43.364 2511-2511/org.traccar.client D/TrackingController: read
06-21 16:50:43.377 2511-2511/org.traccar.client D/TrackingController: send (id:1 time:1466527027 lat:70.0 lon:6.0)
06-21 16:50:58.460 2511-2511/org.traccar.client D/TrackingController: retry
`

ivanfmartinez:
https://github.com/tananaev/traccar-client-android/blob/master/app/src/main/java/org/traccar/client/RequestManager.java

replace the 
"        } catch (IOException error) {
            return false;"

with 
"        } catch (IOException error) {
            Log.e(RequestManager.class.getSimpleName(), "Request error : "+ request, error);
            return false;"

with this change the error will be at Logcat.

@tananaev this logging can help to detect communication errors in the Android client.

FX7:
Perfect! But if I get it right, you changed the hard coded http to hard coded https, right? I think this should be part of the address thats inputet by the user.
Or maybe try https and if this is not working, use http as fallback?

ivanfmartinez:
@FX7 I think the protocol should be configured and not keep trying the two options without knowing which is correct. That hardcoded change is for testing the concept.

FX7:
You'r completely right. I just mentioned the "try both" method as fallback ;-)

dschense:
yep it is hardcoded by now.
I think its the best to do something like this:

In Client enter adress with full path like http://example.com or https://example.com

and then:

```
if (!url.StartsWith("http://", StringComparison.OrdinalIgnoreCase))
    builder.scheme("http")
    if (!url.StartsWith("https://", StringComparison.OrdinalIgnoreCase))
    builder.scheme("https")
```

or something like this

ivanfmartinez:
@dschense  something like this can be implemented without changing the current UI. 

But for compatibility it should consider the if address starts with "protocol://" use protocol as scheme, else use the standard http scheme.

have you made the test with the Log.e in RequestManager ?
Do you have used an valid certificate on your nginx or an self signed ?

dschense:
@ivanfmartinez 

yes. Ive added the Log.e:

`06-21 18:04:35.257 10790-10790/org.traccar.client I/PositionProvider: location new
06-21 18:04:35.259 10790-10790/org.traccar.client D/TrackingController: write (id:0 time:1466532274 lat:48.0 lon:9.0)
06-21 18:04:35.317 10790-10790/org.traccar.client D/TrackingController: read
06-21 18:04:35.318 10790-10790/org.traccar.client D/TrackingController: send (id:1 time:1466532274 lat:48.0 lon:9.0)
06-21 18:04:35.367 10790-10790/org.traccar.client D/TrackingController: delete (id:1 time:1466532274 lat:48.0 lon:9.0)
06-21 18:04:35.387 10790-10790/org.traccar.client D/TrackingController: read

And I used a valid cert by LetsEncrypt.`

ivanfmartinez:
previous log showed 
send
retry

this log :
send
delete

apparently its working, no error from RequestManager.

dschense:
yep. no error. The fault was i did not open the port in my iptables.

FX7:
I have contacted the developer of the alternative traccar client https://play.google.com/store/apps/details?id=de.wialonconsulting.wiatrack.traccar
He has implemented a https acces and i was lucky to get early access to this version: Its working!

So thank you all, for helping!

I would like to help implementing this in the "official" client, but I dont have to much android coding skills. :/

tananaev:
I will consider implementing HTTPS in official client for one of the next releases.

tananaev:
Closing this pull request.

PR #2044, opened 2016-06-23
Possibility to disable event and geofence subsystems

Abyss777:
I added  two parameters in config to disable events and geofences.
Added some code to web-client to disable geofence's buttons if they are disabled in server.

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -52,8 +52,6 @@
 
     <entry key='event.geofenceHandler'>true</entry>
 
-    <entry key='geofence.enable'>true</entry>
-
     <!-- DATABASE CONFIG -->
 
     <!--<entry key='database.driverFile'>hsqldb.jar</entry>-->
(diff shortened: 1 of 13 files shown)


debug.xml
```diff
@@ -41,16 +41,19 @@
...
 
     <entry key='event.overspeedHandler'>true</entry>
     <entry key='event.globalSpeedLimit'>90</entry>
 
     <entry key='event.motionHandler'>true</entry>
 
-
     <entry key='event.geofenceHandler'>true</entry>
 
+    <entry key='geofence.enable'>true</entry>
```
tananaev: We already have a flag for geofencing.

tananaev:
I don't really think we need to disable API or handle it in web interface.

Abyss777:
You are right, second flag is unnecessary.
Removed changes in web-client. Is it OK if there will be "NullPointerException" in API response if geofences are disabled?

PR #2052, opened 2016-06-25
Notifications via email

Abyss777:
I added sending notifications via email. Settings from Config have priority before users settings.

Added notifications settings and dialog to web-client as discussed. It is not ordinary object like device or user, it is like link, but with attributes. Id is used only for database actions. Notification with empty attributes is removed from database (like unlink).

Extended user model to get/store mail settings.

I did not add attributes to Server model because they can be requested via API without login. The mail settings should be somehow hidden from everybody.

Email Templates is hardcoded for now.

I used javamail 1.4.7 because something wrong with maven repositories and 1.5.5 couldn't be used for runtime.
Inline conditionals are very convenient, I removed AvoidInlineConditionals from checkstyle tho.

### DIFF
--- a/setup/unix/traccar.xml
+++ b/setup/unix/traccar.xml
@@ -28,20 +28,6 @@
 
     <entry key='event.geofenceHandler'>true</entry>
 
-    <!--<entry key='mail.smtp.host'>smtp.example.com</entry>
-    for STARTTLS
-    <entry key='mail.smtp.port'>587</entry>
-    <entry key='mail.smtp.starttls.enable'>true</entry>
-    for SSL
-    <entry key='mail.smtp.port'>465</entry>
-    <entry key='mail.smtp.ssl.enable'>true</entry>
-
-    <entry key='mail.smtp.from'>traccar@example.com</entry>
-
-    <entry key='mail.smtp.auth'>true</entry>
-    <entry key='mail.smtp.username'>traccar@example.com</entry>
-    <entry key='mail.smtp.password'>password</entry>-->
-
     <!-- DATABASE CONFIG -->
 
     <entry key='database.driver'>org.h2.Driver</entry>
--- a/setup/unix/traccar.xml
+++ b/setup/unix/traccar.xml
@@ -28,6 +28,20 @@
 
     <entry key='event.geofenceHandler'>true</entry>
 
+    <!--<entry key='mail.smtp.host'>smtp.example.com</entry>
+    for STARTTLS
+    <entry key='mail.smtp.port'>587</entry>
+    <entry key='mail.smtp.starttls.enable'>true</entry>
+    for SSL
+    <entry key='mail.smtp.port'>465</entry>
+    <entry key='mail.smtp.ssl.enable'>true</entry>
+
+    <entry key='mail.smtp.from'>traccar@example.com</entry>
+
+    <entry key='mail.smtp.auth'>true</entry>
+    <entry key='mail.smtp.username'>traccar@example.com</entry>
+    <entry key='mail.smtp.password'>password</entry>-->
+
     <!-- DATABASE CONFIG -->
 
     <entry key='database.driver'>org.h2.Driver</entry>
@@ -250,6 +264,27 @@
 
     <entry key='database.unlinkDeviceGeofence'>
         DELETE FROM device_geofence WHERE deviceId = :deviceId AND geofenceId = :geofenceId;
+    </entry>
+
+    <entry key='database.selectNotifications'>
+        SELECT * FROM notifications;
+    </entry>
+
+    <entry key='database.insertNotification'>
+        INSERT INTO notifications (userId, type, attributes)
+        VALUES (:userId, :type, :attributes);
+    </entry>
+
+    <entry key='database.updateNotification'>
+        UPDATE notifications SET
+        userId = :userId,
+        type = :type,
+        attributes = :attributes
+        WHERE id = :id;
+    </entry>
+
+    <entry key='database.deleteNotification'>
+        DELETE FROM notifications WHERE id = :id;
     </entry>
 
     <!-- PROTOCOL CONFIG -->
--- a/setup/windows/traccar.xml
+++ b/setup/windows/traccar.xml
@@ -28,20 +28,6 @@
 
     <entry key='event.geofenceHandler'>true</entry>
 
-    <!--<entry key='mail.smtp.host'>smtp.example.com</entry>
-    for STARTTLS
-    <entry key='mail.smtp.port'>587</entry>
-    <entry key='mail.smtp.starttls.enable'>true</entry>
-    for SSL
-    <entry key='mail.smtp.port'>465</entry>
-    <entry key='mail.smtp.ssl.enable'>true</entry>
-
-    <entry key='mail.smtp.from'>traccar@example.com</entry>
-
-    <entry key='mail.smtp.auth'>true</entry>
-    <entry key='mail.smtp.username'>traccar@example.com</entry>
-    <entry key='mail.smtp.password'>password</entry>-->
-
     <!-- DATABASE CONFIG -->
 
     <entry key='database.driver'>org.h2.Driver</entry>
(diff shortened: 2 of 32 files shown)


setup/unix/traccar.xml
```diff
@@ -28,6 +28,20 @@
 
     <entry key='event.geofenceHandler'>true</entry>
 
+    <!--<entry key='mail.smtp.host'>smtp.example.com</entry>
```
tananaev: We don't need comments in release config. Better to put it in the documentation later.

setup/windows/traccar.xml
```diff
@@ -28,6 +28,20 @@
 
     <entry key='event.geofenceHandler'>true</entry>
 
+    <!--<entry key='mail.smtp.host'>smtp.example.com</entry>
```
tananaev: Same here.

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -0,0 +1,164 @@
...
+            + "Point: http://www.openstreetmap.org/?mlat=%3$f&mlon=%4$f#map=16/%3$f/%4$f%n"
+            + "Time: %2$tc%n";
+
+    public static String formatTitle(long userId, Event event, Position position) {
+        Device device = Context.getIdentityManager().getDeviceById(event.getDeviceId());
+        StringBuilder stringBuilder = new StringBuilder();
+        Formatter formatter = new Formatter(stringBuilder, Locale.getDefault());
+
+        switch (event.getType()) {
+            case Event.TYPE_COMMAND_RESULT  : formatter.format(TITLE_TEMPLATE_TYPE_COMMAND_RESULT, device.getName());
```
tananaev: Please format all switches in following way:
```
case SOMETHING:
    something();
    something(
            "some long string");
    break;
```

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -0,0 +1,164 @@
...
+                    position.getLatitude(),
+                    position.getLongitude(),
+                    position.getSpeed());
+                                              break;
+            case Event.TYPE_GEOFENCE_ENTER  : formatter.format(MESSAGE_TEMPLATE_TYPE_GEOFENCE_ENTER,
+                    device.getName(),
+                    position.getFixTime(),
+                    position.getLatitude(),
+                    position.getLongitude(),
+                    (Context.getGeofenceManager() != null)
```
tananaev: Parentheses is not needed here afaik.

src/org/traccar/notification/NotificationMail.java
```diff
@@ -0,0 +1,133 @@
...
+        Properties mailServerProperties;
+        Session mailSession;
+        MimeMessage mailMessage;
+
+        String from = null;
+        String username = null;
+        String password = null;
+
+        try {
+        User user = dataManager.getUser(userId);
```
tananaev: Wrong indentation.

web/app/view/NotificationsController.js
```diff
@@ -0,0 +1,83 @@
...
+                    },
+                    scope: this,
+                    callback: function (records, operation, success) {
+                        var i, index, attributes, storeRecord;
+                        if (success) {
+                            for (i = 0; i < records.length; i++) {
+                                index = this.getView().getStore().find('type', records[i].getData().type);
+                                attributes = records[i].getData().attributes;
+                                storeRecord = this.getView().getStore().getAt(index);
+                                if (typeof attributes.web != 'undefined') {
```
tananaev: Why do you need to check if it's undefined?

tananaev: Also, do we really need to parse it and put into the record?

tananaev:
I haven't gone too deep into the logic. I assume everything is tested and works, right?

web/app/view/NotificationsController.js
Abyss777: I did not plan to do that, but unfortunately `Columns` does not support nested objects in `dataIndex`

tananaev: Maybe we can add them into the database as columns? I don't think we would have too many options. Is it better/cleaner?

Abyss777: It will ruin our idea about easy extendable notification object.
I can try to do it via `renderers`. it will look more cleaner, but still we will have to construct request from plane object.

tananaev: I was thinking attributes for all extra data, but whatever you think is cleaner solution.

Abyss777: OK, I'll show you code, hope today. And will decide if it fits.

Abyss777: I did changes. Never used plane record field, only it's name.
`onBeforeCheckChange` and `onCheckChange` can be combined in one, but I thought it looks clearer.
What do you think?

Abyss777:
Fixed other issues

> I haven't gone too deep into the logic. I assume everything is tested and works, right?

Yes, I think so.
I tested it in two environments linux/windows, with real and generated data (but only with one device)
H2 on windows and MySQL on linux.
It is my usual environments.

Tested with 3 mail servers:

- `smtp.gmail.com` in 2 auth configs: SSL and STARTTLS
- `smtp.mail.ru` in one auth config SSL
- internal company server without any authorization

We must allow less secure apps for Gmail https://support.google.com/accounts/answer/6010255?hl=en
Hope we will add OAuth2 in future.

Abyss777:
Just because of perfectionism made code more universal, to not lock on 'attributes'

Abyss777:
A bit more perfectionism. Separated some code from huge `sendMailSync`

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -0,0 +1,158 @@
...
+                    position.getLatitude(), position.getLongitude(), position.getSpeed());
+                break;
+            case Event.TYPE_GEOFENCE_ENTER:
+                formatter.format(MESSAGE_TEMPLATE_TYPE_GEOFENCE_ENTER, device.getName(), position.getFixTime(),
+                    position.getLatitude(), position.getLongitude(), Context.getGeofenceManager() != null
+                    ? Context.getGeofenceManager().getGeofence(event.getGeofenceId()).getName() : "");
+                break;
+            case Event.TYPE_GEOFENCE_EXIT:
+                formatter.format(MESSAGE_TEMPLATE_TYPE_GEOFENCE_EXIT, device.getName(), position.getFixTime(),
+                    position.getLatitude(), position.getLongitude(), Context.getGeofenceManager() != null
```
tananaev: This is really hard to read. That's why I didn't want to use ternary operator. Do we actually need to check for null? Is it possible to receive geofence event without having a manager?

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -0,0 +1,158 @@
...
+                break;
+            case Event.TYPE_DEVICE_ONLINE:
+                formatter.format(MESSAGE_TEMPLATE_TYPE_DEVICE_ONLINE, device.getName(), event.getServerTime());
+                break;
+            case Event.TYPE_DEVICE_OFFLINE:
+                formatter.format(MESSAGE_TEMPLATE_TYPE_DEVICE_OFFLINE, device.getName(), event.getServerTime());
+                break;
+            case Event.TYPE_DEVICE_MOVING:
+                formatter.format(MESSAGE_TEMPLATE_TYPE_DEVICE_MOVING, device.getName(), position.getFixTime(),
+                    position.getLatitude(), position.getLongitude());
```
tananaev: Indentation is incorrect. It should be double (8 spaces) for function parameters.

tananaev:
Just few more minor comments. Otherwise looks fine.

tananaev:

Abyss777:
Yes, it is almost impossible, but we should not forget about this check if we will format events from history.
Fixed indentation.

tananaev:
@Abyss777, can you please take a look:

https://www.traccar.org/forums/topic/smtp-config-for-notifications/

PR #2057, opened 2016-06-28
JsonConverter.objectToJson now can return Lists of objects or primitives

Abyss777:
I noticed that `device.deofenceIds` is not sended to websocket, but by API it is accessible.
I've extended `objectToJson` to convert `List`s to JSON and `arrayToJson` to convert primitives to JSON. I've added most used primitives just to complete.


tananaev:

PR #2077, opened 2016-07-04
Load area for editing geofence.

Abyss777:
There is nice feature, polygon can be edited without drawing new but circle should be redrawn.


Abyss777:
Added center to editing geofence.
Unfortunately [getInteriorPoint()](http://openlayers.org/en/v3.15.1/apidoc/ol.geom.Polygon.html#getInteriorPoint) does not work.
I can write math function to calculate real polygon center, but I think first point is enough.

tananaev:

PR #2081, opened 2016-07-06
Granit protocol implementation

Abyss777:
Here is granit v3 protocol implementation.

It works OK with real device, but there can be some pitfalls. I need a day to drive more and test while you are reviewing.

I've added `Command.TYPE_IDENTIFICATION` as well. I'm sure most of devices has command to get hardware and firmware versions. I'll implement it for wondex soon.

By the way it is 113 protocol supported by traccar. Really impressive number.


test/org/traccar/protocol/GranitProtocolDecoderTest.java
```diff
@@ -0,0 +1,33 @@
...
+    @Test
+    public void testDecode() throws Exception {
+
+        GranitProtocolDecoder decoder = new GranitProtocolDecoder(new GranitProtocol());
+
+        verifyPosition(decoder, binary(ByteOrder.LITTLE_ENDIAN,
+                "2b525243427e1a003e2934757c57b8b03c38d279b4e61e9bd7006b000000001c00002a45330d0a"));
+        
+        verifyPositions(decoder, binary(ByteOrder.LITTLE_ENDIAN,
+        		"2b444441547e84003e290401d01690737c57b8903c383c7fa0e5081b64006b000000001c0000b8803c388e7fe7e5102197006c000000001c0000b8813c38ad7f02e6042035006c000000001d0000b8813c38bf7f13e6001d1e006c000000001d0000b8813c38bf7f13e6001d00006c000000001d0000b8903c38977f34e6091065006c000000001e000014002a39320d0a"));
```
tananaev: Please remove tabs from the code.

src/org/traccar/protocol/GranitFrameDecoder.java
```diff
@@ -0,0 +1,40 @@
...
+import org.jboss.netty.handler.codec.frame.FrameDecoder;
+import org.traccar.helper.StringFinder;
+
+public class GranitFrameDecoder extends FrameDecoder {
+
+    @Override
+    protected Object decode(
+            ChannelHandlerContext ctx, Channel channel, ChannelBuffer buf) throws Exception {
+
+        int index = buf.indexOf(buf.readerIndex(), buf.writerIndex(), new StringFinder("\r\n"));
```
tananaev: Can we use `LineBasedFrameDecoder` instead of custom implementation?

src/org/traccar/protocol/GranitProtocolDecoder.java
```diff
@@ -0,0 +1,215 @@
...
+public class GranitProtocolDecoder extends BaseProtocolDecoder {
+
+    private static final int HEADER_LENGTH = 6;
+
+    public GranitProtocolDecoder(GranitProtocol protocol) {
+        super(protocol);
+    }
+
+    public static void appendChecksum(ChannelBuffer buffer, int length) {
+        buffer.writeByte(0x2a); // asterisk
```
tananaev: You can replace codes with actual symbols. Like `'*'` for asterisk. Same for end line symbols.

src/org/traccar/protocol/GranitProtocolDecoder.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+        buf.skipBytes(1); //StatMess
+    }
+
+    @Override
+    protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
+
+        ChannelBuffer buf = (ChannelBuffer) msg;
+
+        if (hasDeviceId() &&  buf.toString(StandardCharsets.US_ASCII).contains("OK")
```
tananaev: I think it would be more efficient and cleaner to call `buf.toString(StandardCharsets.US_ASCII)` once.

tananaev:
Added some comments. I'll take a look at the documentation a bit later. Possibly I'll have more.

src/org/traccar/protocol/GranitProtocolEncoder.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class GranitProtocolEncoder extends BaseProtocolEncoder {
+    @Override
+    protected Object encodeCommand(Command command) {
+
+        ChannelBuffer commandString;
+
+        switch (command.getType()) {
+        case Command.TYPE_IDENTIFICATION:
+            commandString = ChannelBuffers.directBuffer(ByteOrder.LITTLE_ENDIAN, 12);
```
tananaev: I think it's better to use `dynamicBuffer` other than hardcoding length.

Abyss777: What is the difference? There is no `dynamicBuffer` constructor with `ByteOrder` and without `estimatedLength`. We have to set length anyway.

tananaev: Less hardcoded values is better. For dynamic buffer you can use length 0 (EMPTY_BUFFER) and it will automatically resize to needed size when you add data.

Abyss777: Hm, but we do not need little-endian. I'll replace to `dynamicBuffer`

tananaev: You need little endian in `GranitProtocolDecoder`. Similar issue there where you format responses.

Abyss777:
There is no information about max packet length in documentation, but I have never seen more than one `block`, it is 142 bytes. I set `LineBasedFrameDecoder` `maxLength` to 1024 just in case.

Other issues fixed.

src/org/traccar/protocol/GranitProtocol.java
```diff
@@ -0,0 +1,52 @@
...
+                Command.TYPE_REBOOT_DEVICE,
+                Command.TYPE_POSITION_SINGLE);
+    }
+
+    @Override
+    public void initTrackerServers(List<TrackerServer> serverList) {
+        TrackerServer server = new TrackerServer(new ServerBootstrap(), this.getName()) {
+            @Override
+            protected void addSpecificHandlers(ChannelPipeline pipeline) {
+                pipeline.addLast("frameDecoder", new LineBasedFrameDecoder(1024, true, false));
```
tananaev: Why do you use 3-argument constructor here?

src/org/traccar/protocol/GranitProtocolDecoder.java
```diff
@@ -0,0 +1,212 @@
...
+        position.set(Position.KEY_SATELLITES, BitUtil.from(satDel, 4));
+
+        int pdop = BitUtil.to(satDel, 4);
+        position.set("pdop", pdop);
+
+        int lonDegrees = buf.readUnsignedByte();
+        int latDegrees = buf.readUnsignedByte();
+        int lonMinutes = buf.readUnsignedShort();
+        int latMinutes = buf.readUnsignedShort();
+        double latitude = latDegrees + (double) latMinutes / 60000;
```
tananaev: `(double) latMinutes / 60000` can be replaced with `latMinutes / 60000.0` for readability

src/org/traccar/protocol/GranitProtocolDecoder.java
```diff
@@ -0,0 +1,212 @@
...
+        position.set(Position.KEY_DISTANCE, buf.readShort());
+
+        int analogIn1 = buf.readUnsignedByte();
+        int analogIn2 = buf.readUnsignedByte();
+        int analogIn3 = buf.readUnsignedByte();
+        int analogIn4 = buf.readUnsignedByte();
+
+        int analogInHi = buf.readUnsignedByte();
+
+        analogIn1 = analogInHi << 8 & 0x300 | analogIn1;
```
tananaev: Can you please explain a little bit about what's happening here.

src/org/traccar/protocol/GranitProtocolDecoder.java
```diff
@@ -0,0 +1,212 @@
...
+        position.set(Position.PREFIX_ADC + 3, analogIn3);
+        position.set(Position.PREFIX_ADC + 4, analogIn4);
+
+        position.setAltitude(buf.readUnsignedByte() * 10);
+
+        short diOut = buf.readUnsignedByte();
+        for (int i = 0; i < 8; i++) {
+            position.set(Position.PREFIX_IO + (i + 1), BitUtil.check(diOut, i));
+        }
+        buf.skipBytes(1); //StatMess
```
tananaev: what's this?

tananaev:
Some more very minor comments and questions. Sorry for being fussy.

src/org/traccar/protocol/GranitProtocol.java
Abyss777: Two reasons :
- To remove line ending at the end
- To decode as much as possible if we get too big packet.
If it is unnecessary I'll remove.

tananaev: Here is the 1-argument constructor implementation:
```java
public LineBasedFrameDecoder(final int maxLength) {
    this(maxLength, true, false);
}
```

So it basically does the same thing.

src/org/traccar/protocol/GranitProtocolDecoder.java
Abyss777: Very odd structure.
As I understand analog DAC is 10 bit, and analogInX is 8 low bits, and two other bits in analogInHi in according place.

analogInHi[0-1] is high bits of analogIn1
analogInHi[2-3] is high bits of analogIn2
analogInHi[4-5] is high bits of analogIn3
analogInHi[6-7] is high bits of analogIn3

I just move them on right place (to 8-9) clear with `& 0x300` and `or` with analogInX

src/org/traccar/protocol/GranitProtocolDecoder.java
Abyss777: Are you about `StatMess` ? I don't know what it is, the explanation is

> unsigned char	StatMess;	// Состояние буфера статусных 
> 					// сообщений

Abyss777:
Done fixes.

tananaev:
Merged, thanks.

PR #2084, opened 2016-07-06
Revert GranitFrameDecoder

Abyss777:
`LineBasedFrameDecoder` did not fit at all, It handles both \n and \r\n. 0x0a appears rather frequently in binary data.
`DelimiterBasedFrameDecoder` is better but we can't exclude appearing 0x0d0a in binary data, for example I can set deviceId to 2537 and never handle any position. 

Because of that I've added reading frame based on `length` field of packet. Not binary packets handled as usual with \r\n delimiter.

### DIFF
--- a/src/org/traccar/protocol/GranitFrameDecoder.java
+++ b/src/org/traccar/protocol/GranitFrameDecoder.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2016 Anton Tananaev (anton.tananaev@gmail.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


tananaev:
This all looks very messy. I read the documentation again and I suggest following algorithm for frame decoder:

1. Find end of line index
2. Try to find '~' index
3. If end of line comes first, return result
3. If symbol '~' comes before end of line, read length of binary data (length field comest right after '~'), go to point 1 and search from the end of binary data

Also, please add frame decoder unit tests.

tananaev:
I also took another look at the protocol decoder, and it seems incorrect that you convert the whole message into a string even if it has binary data it in. Can't you just convert the header?

This check is definitely incorrect:
```java
bufString.contains("OK")
```
Binary data can contain anything including these two symbols.

Abyss777:
I hurried, I'm sorry.
I've got your point about tilde, It is sign of binary data. I've done almost as you said.

To ProtocolDecoder...
with `bufString.contains("OK")` I wanted to cover cases `ACCESS OK` and `ERASE OK`
Header length can be less 6, `OK` is 2, `ERROR` is 5. But I agree something should be done with it.

tananaev:
Merged, thanks. Please send me another pull request if you manage to fix issues in decoder.

PR #2085, opened 2016-07-07
Simplify GranitProtocolDecoder command result detection

Abyss777:
I've used the same idea with tilde and assumed that tilde has only packets with points.
If there is no tilde it is command result.
And now there is no binary to string conversion.

### DIFF
--- a/src/org/traccar/protocol/GranitProtocolDecoder.java
+++ b/src/org/traccar/protocol/GranitProtocolDecoder.java
@@ -136,7 +136,7 @@ public class GranitProtocolDecoder extends BaseProtocolDecoder {
 
         int indexTilde = buf.indexOf(buf.readerIndex(), buf.writerIndex(), new StringFinder("~"));
 
-        if (hasDeviceId() &&  indexTilde == -1) {
+        if (hasDeviceId() && indexTilde == -1) {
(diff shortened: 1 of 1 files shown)


src/org/traccar/protocol/GranitProtocolDecoder.java
```diff
@@ -132,10 +133,11 @@ private static void decodeStructure(ChannelBuffer buf, Position position) {
...
 
         ChannelBuffer buf = (ChannelBuffer) msg;
-        String bufString = buf.toString(StandardCharsets.US_ASCII);
-        if (hasDeviceId() &&  bufString.contains("OK") || bufString.startsWith("ERROR")
-                || bufString.startsWith("+PR") || bufString.startsWith("+IDNT")
-                || bufString.startsWith("+BBMD")) {
+
+        int indexTilde = buf.indexOf(buf.readerIndex(), buf.writerIndex(), new StringFinder("~"));
+
+        if (hasDeviceId() &&  indexTilde == -1) {
```
tananaev: Please remove double space.

tananaev:
One tiny comment. Otherwise looks good.

Abyss777:
Checkstyle missed it, and me too.
Done.

tananaev:
Merged. Thanks again.

PR #2091, opened 2016-07-08
Pass whole record to attribute editor not to lose "dirtyness"

Abyss777:
Most of time, after editing only user attributes, the its record is not synced to server.
I've decided to pass whole record to editor, and set `dirty` if something changed.


tananaev:
Merged, thanks.

PR #2096, opened 2016-07-11
Carcell protocol implementation

drecchia:
This is the tcp cr2000 protocol implementation ( www.carcell.com.br )

mvn site - ok


src/org/traccar/protocol/CarcellProtocolDecoder.java
```diff
@@ -0,0 +1,162 @@
...
+            .compile();
+
+    @Override
+    protected Object decode(
+            Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
+
+        Pattern pattern = null;
+        String sentence = (String) msg;
+        
+        if ( sentence.indexOf("CEL,") < 0 ) {
```
tananaev: Please follow same code style as everywhere else. There should be no spaces after and before parentheses.

src/org/traccar/protocol/CarcellProtocolDecoder.java
```diff
@@ -0,0 +1,162 @@
...
+        position.set(Position.KEY_ARCHIVE, parser.next().equals("%"));
+        position.setValid(true);
+        
+        if (!identify(parser.next(), channel, remoteAddress)) {
+            return null;
+        }
+        
+        position.setDeviceId(getDeviceId());
+        
+        if ( PATTERN_GPS == pattern ) {
```
tananaev: Same here.

src/org/traccar/protocol/CarcellProtocolDecoder.java
```diff
@@ -0,0 +1,162 @@
...
+            .number("(d),")                      // block
+            .number("(d),")                      // ignition
+            .number("(d),")                      // cloned
+            .expression("([AF])")                // panic
+            .number("(d),")                      // painel
+            .number("(d+),")                     // battery voltage
+            .any()                               // full format
+            .compile();
+    
+    private static final Pattern PATTERN_CEL = new PatternBuilder()
```
tananaev: Please combine two patterns. They are almost identical.

tananaev: See `H02ProtocolDecoder` for an example of what you can do.

tananaev:
Please fix comments.

src/org/traccar/protocol/CarcellProtocolDecoder.java
```diff
@@ -0,0 +1,147 @@
...
+        
+        if (!identify(parser.next(), channel, remoteAddress)) {
+            return null;
+        }
+        
+        position.setDeviceId(getDeviceId());
+        
+        if (sentence.indexOf("CEL,") < 0) {
+            position.setLatitude(parser.nextCoordinate(CoordinateFormat.HEM_DEG_MIN_MIN));
+            parser.next();
```
drecchia: Please have a look here, when I changed to groupBegin/or/groupEnd I have to "jump" groups that didnt match. Whats the proper way to deal with this situation using your Parser ?

src/org/traccar/protocol/CarcellProtocolDecoder.java
```diff
@@ -0,0 +1,147 @@
...
+
+        position.setDeviceId(getDeviceId());
+
+        if (sentence.indexOf("CEL,") < 0) {
+            position.setLatitude(parser.nextCoordinate(CoordinateFormat.HEM_DEG_MIN_MIN));
+            parser.next();
+            parser.next();
+            position.setLongitude(parser.nextCoordinate(CoordinateFormat.HEM_DEG_MIN_MIN));
+            parser.next();
+            parser.next();
```
drecchia: Please have a look if groupBegin/End usage is correct, since it generate this extra groups.

tananaev: I would combine both coordinates in a single group. Then instead of a wall of `next()` calls and a weird check for `CEL` at the beginning, you can use `hasNext(N)`.

tananaev:
That's not exactly what I was thinking. Can you please do similar to `TelicProtocolDecoder`.

drecchia:
mvn site - pass

src/org/traccar/protocol/CarcellProtocol.java
```diff
@@ -0,0 +1,55 @@
...
+                Command.TYPE_ENGINE_STOP,
+                Command.TYPE_ENGINE_RESUME);
+    }
+
+    @Override
+    public void initTrackerServers(List<TrackerServer> serverList) {
+        serverList.add(new TrackerServer(new ServerBootstrap(), this.getName()) {
+            @Override
+            protected void addSpecificHandlers(ChannelPipeline pipeline) {
+                // inbound
```
tananaev: Can you please remove unnecessary comments. Also, there is a reason why decoders and encoders are mixed. For example, string encoder should be before object decoder, because if object decoder want to send something back (e.g. acknowledgement), there is a good chance that the message should be converted from string to buffer.

In general please follow existing code styles and patterns.

tananaev:
Just one last comment. Everything else looks good.

drecchia:
Damm I messed this pull request again! I will create a clean one.

PR #2097, opened 2016-07-11
Fix "from" in mail notifications

Abyss777:
According to https://javamail.java.net/nonav/docs/api/com/sun/mail/smtp/package-summary.html

> mail.smtp.from	
> Email address to use for SMTP MAIL command. This sets the envelope return address. Defaults to msg.getFrom() or InternetAddress.getLocalAddress(). NOTE: mail.smtp.user was previously used for this.

Should work and my local server on Zimbra and Gmail understand this, but mail.ru even did not accept mails until I set it via `mailMessage.setFrom`

Hope should fix this https://www.traccar.org/forums/topic/smtp-config-for-notifications/#post-6870

Other changes:
- better handle empty mail user attributes

- fix deleting attribute in attribute editor
- fix duplicating attribute if name changed

PS: sorry for duplicated commits, there was conflict and git solved it like this.

### DIFF
--- a/web/app/view/AttributesController.js
+++ b/web/app/view/AttributesController.js
@@ -40,17 +40,22 @@ Ext.define('Traccar.view.AttributesController', {
         store.addListener('add', function (store, records, index, eOpts) {
             for (i = 0; i < records.length; i++) {
                 this.getView().record.get('attributes')[records[i].get('name')] = records[i].get('value');
-                this.getView().record.dirty = true;
             }
+            this.getView().record.dirty = true;
         }, this);
         store.addListener('update', function (store, record, operation, modifiedFieldNames, details, eOpts) {
             if (operation === Ext.data.Model.EDIT) {
+                if (record.modified.name !== record.get('name')) {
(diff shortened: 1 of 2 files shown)


web/app/view/AttributesController.js
```diff
@@ -40,17 +40,22 @@ Ext.define('Traccar.view.AttributesController', {
...
+                if (record.modified.name !== record.get('name')) {
+                    delete this.getView().record.get('attributes')[record.modified.name];
+                }
                 this.getView().record.get('attributes')[record.get('name')] = record.get('value');
                 this.getView().record.dirty = true;
             }
         }, this);
         store.addListener('remove', function (store, records, index, isMove, eOpts) {
-            delete this.getView().record.get('attributes')[records[index].get('name')];
+            for (var i = 0; i < records.length; i++) {
```
tananaev: Variable should be declared at the beginning of the function.

tananaev:
Merged. There is a minor comment, but it can be fixed later. For future, please run `jshint` and `jscs` for JavaScript code.

PR #2099, opened 2016-07-11
Revert back groups cache and split GeofenceManager.refresh() on 3 functions

Abyss777:
- Split geofence refresh on 3 functions
- Enable back groupsCache

For now groupsCache have unwanted update only if device(group) linked to deleted group.

I'm researching foreign keys, but have some issues (null and not null incompatibility I think)

For demo servers you can null dead links via sql, it can help for a while.


tananaev:
Merged, thanks.

PR #2103, opened 2016-07-12
Add Constraints for groupIds

Abyss777:
Here are changelogs to add constraints to `groupId`s of devices and groups.

Unfortunately, `preConditions` can be only at the beginning (before or after changeSet) and I have to split changelogs.

Tested on H2, MySQL, PosgresSQL and MS SQL.

For MS SQL users we can comment in documentation that can be created trigger do the same as constraint, but not via liquibase.


schema/changelog-3.7.xml
```diff
@@ -0,0 +1,24 @@
...
+    <update tableName="devices">
+        <column name="groupid"/>
+        <where>groupid NOT IN (SELECT id FROM groups)</where>
+    </update>
+
+    <addForeignKeyConstraint baseColumnNames="groupid" baseTableName="devices" constraintName="fk_device_group_groupid" onDelete="SET NULL" onUpdate="RESTRICT" referencedColumnNames="id" referencedTableName="groups"/>
+
+    <update tableName="groups">
+        <column name="groupid"/>
+        <where>groupid NOT IN (SELECT id FROM (SELECT id FROM groups) AS groups_1)</where>
```
tananaev: Can't the nested statement be replaced with simple `SELECT id FROM groups`?

tananaev:
Also, please use 2-space indentation for changelogs. In some places you have 4 spaces.

schema/changelog-3.7.xml
Abyss777: Unfortunately not. It's a trick to update data in the same table. At least MySQL do not allow that and we have to use intermediate table.

Abyss777:
Done

tananaev:
I'm sure I had added another comment, but I don't see it now. Would it be possible to combine changesets into a single file?

schema/changelog-3.7.xml
tananaev: It's annoying how each database engine has its own constraints.

Can you please rename `groups_1` to something better like `group_ids` or something.

Abyss777:
Sorry, for misunderstanding.
Combined in one file.

schema/changelog-3.7.xml
```diff
@@ -0,0 +1,35 @@
...
+    <update tableName="groups">
+      <column name="groupid"/>
+      <where>groupid NOT IN (SELECT id FROM (SELECT id FROM groups) AS groups_1)</where>
+    </update>
+
+  </changeSet>
+
+  <changeSet author="author" id="changelog-3.7-notmssql">
+
+    <preConditions onFail="CONTINUE">
```
tananaev: Should we change it to `MARK_RAN`? Otherwise it will try to execute it every time:

> CONTINUE: Skip over change set. Execution of change set will be attempted again on the next update. Continue with change log.

tananaev:
One comment. Also, can you please rename `groups_1` to something more meaningful?

Abyss777:
Sorry, I'm hurrying again...
Fixed

tananaev:
Merged, thanks. Hopefully it solves those invalid links problems, but we still need to handle it in code for MS SQL.

Abyss777:
Very strange, DB do not allow add devices with groupId = 0, but! allow add groups with groupId = 0

Abyss777:
Can we add here https://github.com/tananaev/traccar/blob/master/src/org/traccar/database/QueryBuilder.java#L178

if value == 0 `statement.setNull` 

like here https://github.com/tananaev/traccar/blob/master/src/org/traccar/database/QueryBuilder.java#L204

?

PR #2107, opened 2016-07-12
Added new suntech commands

drecchia:


src/org/traccar/protocol/SuntechProtocolEncoder.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Anton Tananaev (anton.tananaev@gmail.com)
+ * Copyright 2016 Anton Tananaev (anton.tananaev@gmail.com)
```
tananaev: Don't change the year. Add range like `2015 - 2016`.

tananaev:
Added a comment. Also, have you run `mvn site` and checked reports to see if there are any issues there? The line seems too long in your code.

drecchia:
mvn site - ok

src/org/traccar/protocol/SuntechProtocol.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Anton Tananaev (anton.tananaev@gmail.com)
+ * Copyright 2015-2016 Anton Tananaev (anton.tananaev@gmail.com)
```
tananaev: Can you please follow existing style. There should be spaces around dash.

src/org/traccar/protocol/SuntechProtocolEncoder.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Anton Tananaev (anton.tananaev@gmail.com)
+ * Copyright 2015-2016 Anton Tananaev (anton.tananaev@gmail.com)
```
tananaev: same

src/org/traccar/protocol/SuntechProtocolEncoder.java
```diff
@@ -25,6 +25,13 @@
     protected Object encodeCommand(Command command) {
 
         switch (command.getType()) {
+            case Command.TYPE_REBOOT_DEVICE:
+                return formatCommand(command, "SA200CMD;{%s};02;Reboot\r", Command.KEY_UNIQUE_ID);
+            case Command.TYPE_POSITION_SINGLE:
+                return formatCommand(command, "SA200GTR;{%s};02;\r", Command.KEY_UNIQUE_ID);
+            case Command.TYPE_OUTPUT_CONTROL:
+                return formatCommand(command, "SA200CMD;{%s};0{%s};Enable{%s}\r",
+                        Command.KEY_UNIQUE_ID, Command.KEY_DATA, Command.KEY_DATA);
```
tananaev: Wait, what is in `KEY_DATA`?

drecchia: I am not sure, I was expecting it is some params sent by the interface.
I sow a similar usage on Pt502ProtocolEncoder.

drecchia: Where the value should be 1 or 2 to determine which output gonna be controlled.

tananaev: I suspect it is, but one of the two parameters should output index (`Command.KEY_INDEX`), the other one should be output value (`Command.KEY_DATA`). In your case both are `Command.KEY_DATA`.

drecchia: No, both params has to be the same.
Example: 
=> SA200CMD;123456;01;Enable1  ( 01 and Enable1 )
=> SA200CMD;123456;02;Enable2 ( 02 and Enable 2 )

If i make it as two different params would be error prone.

tananaev: OK, your code is completely incorrect. First parameter is protocol version and should be `02` all the time as far as I can tell. Also, you should be able to both enable and disable outputs. And the only the last number if index, but it's not data like it your code.

drecchia: You are absolutely right, sorry, long night. New code on the way...

tananaev:

Issue #2101, opened 2016-07-11
Geofence notification - add script call in addition to mail/web


TwizzyDizzy:
Hi Anton,

first and foremost: thanks for the great new feature "geofencing". This is highly appreciated!

What I was thinking of: in order to keep the notification stuff generic, one could add the option to call an external resource. This could be:

* a URI which receives the relevant information via POST request
* a script call on the local server (this would be the most generic version and would enable nearly any kind of integration)

The crucial part would be to transmit the data in a generic fashion so the external script could utilize them.

Cheers
Thomas

tananaev:
It would be possible to have something similar to data forwarding that we have, but for now I guess email is pretty generic way.

TwizzyDizzy:
Unfortunately with mail it's not really possible to trigger something external with just this hook :|

tananaev:
@Abyss777, what do you think? How hard would it be to add another option to forward event to a URL (something similar to `forward.enable` and `forward.url`).

Abyss777:
It is in my nearest plans. Only question, what we will forward? event object (as json for example) or preformated message?
Forwarding to script is insecure, we have to sonsider it.

TwizzyDizzy:
I'd suggest sending a serialized object instead of a preformatted message. The former is much more easy to handle. Preformatted messages would lead to parsing this message which is simply not neccessary.

Cheers
Thomas

tananaev:
Parsing form parameters should be even easier than JSON.

Abyss777:
We can even POST only event id. Details can be retrieved via API.
Actually other side will request `position` after receiving event in most cases to get whole information.
Or we can POST two objects: event and position, as we do for web-client.

1. Just event id
2. Only event via parameters (not json)
3. Event and position (if need)

What is better?

tananaev:
I guess JSON would be easier to implement, so we can go with that option.

TwizzyDizzy:
My two cents: From a scaling perspective (assuming a huge install of Traccar) it would probably be better to send as much information as possible with the request so the server doesn't have to handle additional API requests because some information in the notification was missing.

Cheers
Thomas

tananaev:
I agree that it would be better to send all information straight away.

Abyss777:
OK, I see your points.

TwizzyDizzy:
[...]

PR #2110, opened 2016-07-13
Added possibility to forward all events via POST request.

Abyss777:
As requested in #2101 added forwarder for all events via HTTP POST requests.

Here is example
````
POST / HTTP/1.1
Content-Type: application/json; charset=utf-8
User-Agent: Traccar Server
Authorization: Basic QWxhZGRpbjpPcGVuU2VzYW1l
Content-Length: 801
Connection: keep-alive
Host: localhost
Accept: */*

{"event":{"positionId":15608,"serverTime":"2016-07-13T17:36:25.595+05:00","geofenceId":4,"type":"geofenceExit","deviceId
":4,"id":3452,"attributes":{}},"position":{"address":"125 Varick St, New York, NY, US","protocol":"osmand","latitude":40
.7250605714,"longitude":-74.0058465714,"altitude":0.0,"valid":true,"outdated":false,"fixTime":"2016-07-13T17:36:25.000+0
5:00","deviceTime":"2016-07-13T17:36:25.000+05:00","speed":0.0,"course":7.19870889662,"deviceId":4,"id":15608,"attribute
s":{"ip":"0:0:0:0:0:0:0:1"}},"device":{"name":"Test","id":4,"status":"online","groupId":0,"uniqueId":"123456789012345","
geofenceIds":[],"positionId":15608,"lastUpdate":"2016-07-13T17:36:25.469+05:00"},"geofence":{"name":"The Beginning","des
cription":"","area":"CIRCLE (40.722412 -74.006288, 250)","id":4,"attributes":{}}}
````

I have some doubts about sending `device` and `geofence` but  because of https://github.com/tananaev/traccar/issues/2101#issuecomment-232027790 I've added it.

### CODE REVIEW

File: src/org/traccar/notification/NotificationForward.java [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private static final String USER_AGENT = "Traccar Server";
+
+    private static final String KEY_POSITION = "position";
+    private static final String KEY_EVENT = "event";
+    private static final String KEY_GEOFENCE = "geofence";
+    private static final String KEY_DEVICE = "device";
+
+    public static void forwardEvent(Event event, Position position) {
```
tananaev: I think it's better to add this class to `Context` instead of having static method. It would be more consistent with other modules.

Abyss777:
`updateDeviceStatus` and `updateLatestPosition` happen very frequently, I've replaced `getDeviceById` to direct `get` from Map. Should decrease locks dramatically.

Forwarder moved to `Context`

Abyss777:
Another thing from `jstack`, actually we do not need update geofences after `add` and `edit` user. Only after deleting we need update users geofences.

tananaev:
Looks good, thanks.

jlms77:
Hi, i'm testing the forward functionality, i manage to capture the position data using the forward_enable and forward_url pointing to another server where we are already capturing the position data.

On the other hand we have configured event.forward.enable and event.forward.url but i can't see any traffic on the server coming from traccar. This is meant to function similar to forward_enable?

We are trying to send SMS notifications using an existent SMS gateway so we want to have all the events forwarded for processing on an external server.

Thank you in advance.

tananaev:
Event forwarding works similar way to position forwarding. Are you using latest official release of Traccar?

jlms77:
Yes i've installed the version 3.8 (traccar-linux-3.8.zip) downloaded from here on an Ubuntu server. And im using traccar ui mod.  I have added this two lines to the config:

    <entry key='event.forward.enable'>true</entry>
    <entry key='event.forward.url'>http://externalserverip/</entry>

Now i dont have nothing to receive in the external server but im expecting at least some logs in the apache access.log, right now i cant see nothing.

tananaev:
What UI mod are you talking about? Have you removed anything from the config to install the mod?

PR #2113, opened 2016-07-14
Fixes for new foreign key constraints 

Abyss777:
Solution to store `long == 0` as `NULL` in DB is not very nice, but simple. The other one is create "Zero-Group" but I do not like it, to many unpredictable side effects and changes.

 I tested a bit, but did not see any regressions in other places we store `long`s.

Also there was bug in SQL request and warning in browser when editing groups

PS: also I can replace `if (value == 0)` to `if (value == 0 && name.toLowerCase().equals("groupid"))`. It will minimize side effects (if they exist)

### DIFF
--- a/src/org/traccar/database/QueryBuilder.java
+++ b/src/org/traccar/database/QueryBuilder.java
@@ -175,7 +175,11 @@ public final class QueryBuilder {
     public QueryBuilder setLong(String name, long value) throws SQLException {
         for (int i : indexes(name)) {
             try {
-                statement.setLong(i, value);
+                if (value == 0) {
+                    statement.setNull(i, Types.INTEGER);
+                } else {
+                	statement.setLong(i, value);
+                }
             } catch (SQLException error) {
                 statement.close();
                 connection.close();
--- a/src/org/traccar/database/QueryBuilder.java
+++ b/src/org/traccar/database/QueryBuilder.java
@@ -178,7 +178,7 @@ public final class QueryBuilder {
                 if (value == 0) {
                     statement.setNull(i, Types.INTEGER);
                 } else {
-                	statement.setLong(i, value);
+                    statement.setLong(i, value);
                 }
             } catch (SQLException error) {
                 statement.close();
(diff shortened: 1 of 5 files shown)


src/org/traccar/database/QueryBuilder.java
```diff
@@ -175,7 +175,11 @@ public QueryBuilder setInteger(String name, int value) throws SQLException {
     public QueryBuilder setLong(String name, long value) throws SQLException {
         for (int i : indexes(name)) {
             try {
-                statement.setLong(i, value);
+                if (value == 0) {
+                    statement.setNull(i, Types.INTEGER);
+                } else {
+                	statement.setLong(i, value);
```
tananaev: Replace tab with spaces.

tananaev:
My only concern is that if we really want to save 0 for something, it won't work. I guess if we don't have anything else of type `long`, we should be fine.

Abyss777:
We have. `server.zoom` and `user.zoom` other `long` are `id`s that can't be `0`.
Maybe it is really better to use `if (value == 0 && name.toLowerCase().equals("groupid"))`, rather ugly but without side effects.

tananaev:
I think we should add another method `setLong` with third parameter. Something like `nullIfZero`. Then in `setObject` check if method name ends with `Id` (case sensitive), then call method with `true`, otherwise call old version.

Abyss777:
Done

src/org/traccar/database/QueryBuilder.java
```diff
@@ -185,6 +185,23 @@ public QueryBuilder setLong(String name, long value) throws SQLException {
         return this;
     }
 
+    public QueryBuilder setLong(String name, long value, boolean nullIfZero) throws SQLException {
```
tananaev: Please make `setLong` function with 2 parameters call this one to avoid duplication.

Abyss777:
OK

tananaev:
Merged, thanks.

PR #2115, opened 2016-07-15
Fix geofences not showed in old browsers and IE

Abyss777:
`startsWith()` is rather new and supported only in new browsers https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

Investigating this post https://www.traccar.org/forums/topic/problem-editing-geofence/


tananaev:
Merged, thanks.

tananaev:
Can you please comment of forum that the issue is fixed.

PR #2116, opened 2016-07-15
Move code related to devices to DeviceManager

Abyss777:
Key changes:
1. Move unusual for `DataManager` code to `DeviceManager`
2. Move `lastPositions` from `ConnectionManager`
3. Remove `ConnectionManager` from tests and extend `IdentityManager`

That allowed:
1. Decrease locks for device manipulation (removed 2 `updateDeviceCache`)
2. Decrease locks in ConnectionManager (some `synchronised` replaced to `ConcurentHashMap`, some removed)
3. LastPosition set is in one place now (DB, cached devices objects, cached last positions) that prevent  out of sync.
4. Place for future extending device objects (attributes, sensors states and others)

Note: please note that device status via websocket is updating at the level of `DefaultDataHandler` not at `MainEventHandler`. It is before `EventHandlers`. Because sometimes toast appeared earlier than device moved on map and for not split the logic.

### DIFF
--- a/src/org/traccar/BaseProtocolDecoder.java
+++ b/src/org/traccar/BaseProtocolDecoder.java
@@ -47,7 +47,9 @@ public abstract class BaseProtocolDecoder extends ExtendedObjectDecoder {
             Device device = Context.getIdentityManager().getDeviceByUniqueId(uniqueId);
             if (device != null) {
                 deviceId = device.getId();
-                Context.getConnectionManager().addActiveDevice(deviceId, protocol, channel, remoteAddress);
+                if (Context.getConnectionManager() != null) {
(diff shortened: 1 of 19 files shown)


tananaev:
Looks good. Have you tested everything? Do you want me to try to install it on the demo server?

Abyss777:
Yes, I have.
Creating, Editing, Removing device work well. Last positions and InitialState work too, webclient get all information as usual.

It would be greate if you try it on one demo server, but only if you have time.
Also if you know how to measure thread locks before and after upgrade it would be perfect.
I've tried JMX, it good but i can't find summs of tread locks and waits, only per thread.

tananaev:
Mostly I want to see if it's stable. I have installed new version on the demo server. Let's see if it works without any issues for a few days.

Abyss777:
OK, Let's watch.

PR #2118, opened 2016-07-15
TEST - Travis improvment - DO NOT ACCEPT

drecchia:
This build should be marked as failed when validated by travis-ci.


drecchia:
YEAHHH!! Broken build!
Anton, what you think about my changes on pom.xml ?

tananaev:
I can't see any actual changes because you changed spaces to tabs, so it highlights everything. I suggest to revert back to spaces.

PR #2119, opened 2016-07-15
Travis improvment

drecchia:
This changes will make travis-ci fail the build if there is checkstyle violations.


pom.xml
```diff
@@ -134,6 +134,31 @@
             </resource>
         </resources>
 
+        <pluginManagement>
```
tananaev: Do we need separate `pluginManagement`? Why can't we just add this to plugins?

drecchia: Unfortunately if you wanna make the project compatible with eclipse, m2e plugin requires that this custom maven call needs to be inside pluginManagement.

tananaev: Why do we need custom call? Can't this just be part of default build process? Like unit tests.

drecchia: Because mvn site is cpu intensive, slow and do stuff that we dont need for a a pull request integration.
The question should be, why not have a custom call ?

tananaev: Why are you talking about `mvn site` if the question is about `checkstyle`? Style check is pretty quick. It is faster than compilation of the code. It makes sense to have it as a part of build process. Similar to unit tests.

drecchia: Ok, Maybe I started with the wrong foot here, lets start over.

1. Today in trunk if you wanna run just checkstyle, you do "mvn checktyle:check", but since maven-checkstyle-plugin is under "reporting" group, the previous check wont read checkstyle.xml and will return 8421 violations.

2. Previous travis-ci config was running "mvn test" as the default ( https://docs.travis-ci.com/user/languages/java ) so my intention is to keep integration test as light as possible and avoid "trash" generated by "mvn site" since its not useful in pull request integration tests.

3. The way I found to achieve this is to configure the goal on build->plugins->maven-checkstyle-plugin and then run "mvn checktyle:check" with the parameter "failsOnError" activated ( so the build will fail ).

4. This works on console, but eclipse m2e plugin starting complaining about project and stops solving artifacts. The workaround is to move config to build->pluginManagement->plugins->maven-checkstyle-plugin.

So looks to me that we have two options:
1. Move checkstyle config to "reporting" group and do a mvn site on travis-ci.
2. Go with isolated goal as the pull request submitted.

That's your call.
Regards

tananaev: You keep avoiding my questions. Why can't we put checkstyle task as a part of standard build process?

drecchia: The point here is to validate the pull request.
What do you mean by "standard build process" ?

tananaev:
I've added checkstyle to the validate build step.

drecchia:
And about the travis-ci config ?

tananaev:
It's a part of standard validation phase. Travis-ci will call it automatically.

drecchia:
As I said, custom goals out of pluginManagement are "incompatible" with eclipse m2e plugin.
( https://www.eclipse.org/m2e/documentation/m2e-execution-not-covered.html )

Now almost every eclipse developer will got:

> Description	Resource	Path	Location	Type
> Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (execution: checkstyle, phase: validate)	pom.xml	/traccar	line 146	Maven Project Build Lifecycle Mapping Problem
>

tananaev:
Works for me. I just tested on the latest version of Eclipse. You just need to install required plug-ins. I think `m2e-code-quality` is the right one. Eclipse installed it for me automatically.

drecchia:
yes, you are right again. Thanks

tananaev:
Thanks for idea anyway. Style check is very fast, so it makes sense to do it all the time to pick up issues earlier.

PR #2123, opened 2016-07-18
Attributes for devices

Abyss777:
Hello, lets continue.

Added attributes to Device model, actually without payload, but I'm planning to use it at least to store device pin-code to send commands.

Also:
- Fixed license header for `DeviceManager.java`
- Fixed editing attributes for not saved (new) object (`AttributesController.js`)

PS: I've noticed some repeated code in controllers, it is need to be optimized with inheritance.

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -159,11 +159,11 @@
     </entry>
 
     <entry key='database.insertDevice'>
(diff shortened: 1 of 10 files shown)


tananaev:
Looks good, thanks.

Pin code for commands is one place where we need to use this. Another one is attributes like timezone. Currently you can set timezone (e.g. GT06 protocol) only for all devices. In some cases it has to be done per device.

PR #2126, opened 2016-07-18
Use password from attributes to send command

Abyss777:
If you don't mind I used `Command.KEY_DEVICE_PASSWORD` as attribute name or may be we should use some prefix? like `"Command."+Command.KEY_DEVICE_PASSWORD`
Tested with wondex device.

### DIFF
--- a/src/org/traccar/protocol/TotemProtocolEncoder.java
+++ b/src/org/traccar/protocol/TotemProtocolEncoder.java
@@ -28,12 +28,10 @@ public class TotemProtocolEncoder extends StringProtocolEncoder {
     protected Object encodeCommand(Command command) {
 
         command.set(Command.KEY_DEVICE_PASSWORD, "000000");
-        if (Context.getDeviceManager() != null) {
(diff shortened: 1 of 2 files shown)


src/org/traccar/protocol/TotemProtocolEncoder.java
```diff
@@ -16,17 +16,25 @@
...
+import org.traccar.model.Device;
 
 public class TotemProtocolEncoder extends StringProtocolEncoder {
 
     @Override
     protected Object encodeCommand(Command command) {
 
-        // Temporary put default password
         command.set(Command.KEY_DEVICE_PASSWORD, "000000");
+        if (Context.getDeviceManager() != null) {
```
tananaev: Do we really need all these null checks? I think if we get to the encoder stage, it means that device definitely exists.

Also, you shouldn't call `toString()` method. Just cast result to string.

Abyss777:
You are right.
Done

tananaev:
Merged, thanks.

PR #2127, opened 2016-07-19
Move device id to a session

tananaev:
Changed `BaseProtocolDecoder` class to store device id in a session.

This solves potential problem with UDP connections.

@Abyss777, can you please take a quick look. There are a lot of changes and I don't trust myself to not make any silly mistakes there. Also, I think now we can store position sequence id in the session. That was the main purpose of this refactoring.


Abyss777:
I've found a problem.  Devices never become offline.
You removed `Context.getConnectionManager().addActiveDevice` because of that https://github.com/tananaev/traccar/blob/master/src/org/traccar/database/ConnectionManager.java#L56 will never set `DEVICE_OFFLINE`

tananaev:
Fixed the problem. Thanks.

Have you finished the review? Is there anything else?

Abyss777:
Other seems OK, but...
I'm thinking of another thing. What happen if there is short tcp connection lost, and device will try to reconnect while previous session still active. Will it be replaced? I do not see that, looks like old session will be returned. Or am I wrong?

Abyss777:
I'm wrong. There will be a new instance of Decoder and new instance of session.
Everything is OK.

tananaev:
Merged. Thanks for reviewing the code.

PR #2129, opened 2016-07-19
Split locks in GeofenceManager.refresh() and try to avoid crosslocks with devices.

Abyss777:
This should fix some deadlocks


tananaev:
Merged, thanks.

PR #2132, opened 2016-07-20
Move groups related actions to DeviceManager

Abyss777:
Also removed cache updates from `getXxxxById`, but leaved in getAllXxxx, and added to constructor.

PS: suddenly I've noticed that Eclipse do checkstyle on the fly. I configured it to use project `checkstyle.xml` but it added `.checkstyle` config file. If you don't mind I'll add it to github, or If not I can put it to `.gitignore`.


src/org/traccar/database/DeviceManager.java
```diff
@@ -43,14 +48,20 @@
...
 
     public DeviceManager(DataManager dataManager) {
         this.dataManager = dataManager;
         this.config = Context.getConfig();
-        dataRefreshDelay = config.getLong("database.refreshDelay", DataManager.DEFAULT_REFRESH_DELAY) * 1000;
+        dataRefreshDelay = config.getLong("database.refreshDelay", DEFAULT_REFRESH_DELAY) * 1000;
         if (dataManager != null) {
             try {
+                updateDeviceCache(true);
+                updateGroupCache(true);
```
tananaev: Should we init group cache first? Because devices reference it...

tananaev:
Please add `.checkstyle` to git ignore. It can be automatically generated, so no need to check it in.

Abyss777:
You are right.
Done.

tananaev:
Merged, thanks.

PR #2136, opened 2016-07-21
Migrate from locks to ConcurrentHashMaps in DeviceManager

Abyss777:
Key changes:

- Incrementally update cache,  `updateDeviceCache` looks monstrous but works OK (please check it more carefully, I could miss something)
- Migrate from locks to maps, construct maps happen on first update to initialize it with right size to avoid resizing delays on every `put` (default initial capacity is 16)
- Device initialized as offline and stopped (status and motion can be easily removed from DB now)

Tested only with few devices, It would be good if you test it on one of demo servers.

### DIFF
--- a/src/org/traccar/database/DeviceManager.java
+++ b/src/org/traccar/database/DeviceManager.java
@@ -70,7 +70,7 @@ public class DeviceManager implements IdentityManager {
     private void updateDeviceCache(boolean force) throws SQLException {
 
         long lastUpdate = devicesLastUpdate.get();
-        if (force || System.currentTimeMillis() - lastUpdate > dataRefreshDelay
+        if ((force || System.currentTimeMillis() - lastUpdate > dataRefreshDelay)
                 && devicesLastUpdate.compareAndSet(lastUpdate, System.currentTimeMillis())) {
             GeofenceManager geofenceManager = Context.getGeofenceManager();
             Collection<Device> databaseDevices = dataManager.getAllDevices();
@@ -233,7 +233,7 @@ public class DeviceManager implements IdentityManager {
     private void updateGroupCache(boolean force) throws SQLException {
 
         long lastUpdate = groupsLastUpdate.get();
-        if (force || System.currentTimeMillis() - lastUpdate > dataRefreshDelay
+        if ((force || System.currentTimeMillis() - lastUpdate > dataRefreshDelay)
                 && groupsLastUpdate.compareAndSet(lastUpdate, System.currentTimeMillis())) {
             Collection<Group> databaseGroups = dataManager.getAllGroups();
             if (groupsById == null) {
--- a/src/org/traccar/database/DeviceManager.java
+++ b/src/org/traccar/database/DeviceManager.java
@@ -18,15 +18,13 @@ package org.traccar.database;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.traccar.Config;
 import org.traccar.Context;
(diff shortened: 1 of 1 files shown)


src/org/traccar/database/DeviceManager.java
```diff
@@ -43,14 +41,12 @@
     private final DataManager dataManager;
     private final long dataRefreshDelay;
 
-    private final ReadWriteLock devicesLock = new ReentrantReadWriteLock();
-    private final Map<Long, Device> devicesById = new HashMap<>();
-    private final Map<String, Device> devicesByUniqueId = new HashMap<>();
-    private long devicesLastUpdate;
+    private Map<Long, Device> devicesById;
+    private Map<String, Device> devicesByUniqueId;
+    private AtomicLong devicesLastUpdate = new AtomicLong(0);
```
tananaev: You don't need to explicitly specify 0. Default constructor initializes value to 0.

src/org/traccar/database/DeviceManager.java
```diff
@@ -72,158 +68,125 @@ public DeviceManager(DataManager dataManager) {
...
+            GeofenceManager geofenceManager = Context.getGeofenceManager();
+            Collection<Device> databaseDevices = dataManager.getAllDevices();
+            if (devicesById == null) {
+                devicesById = new ConcurrentHashMap<>(databaseDevices.size());
+            }
+            if (devicesByUniqueId == null) {
+                devicesByUniqueId = new ConcurrentHashMap<>(databaseDevices.size());
+            }
+            Set<Long> databaseDevicesIds = new HashSet<>();
+            Set<String> databaseDevicesUniqIds = new HashSet<>();
```
tananaev: Please use full word `Unique`.

src/org/traccar/database/DeviceManager.java
```diff
@@ -72,158 +68,125 @@ public DeviceManager(DataManager dataManager) {
...
     private void updateDeviceCache(boolean force) throws SQLException {
-        boolean needWrite;
-        devicesLock.readLock().lock();
-        try {
-            needWrite = force || System.currentTimeMillis() - devicesLastUpdate > dataRefreshDelay;
-        } finally {
-            devicesLock.readLock().unlock();
-        }
-
+        boolean needWrite = force || System.currentTimeMillis() - devicesLastUpdate.get() > dataRefreshDelay;
```
tananaev: I don't think it's the right way to use atomics. I should it's better to use `compareAndSet`. What do you think?

tananaev:
Done with review. A couple of small comments. Otherwise looks good.

src/org/traccar/database/DeviceManager.java
Abyss777: Unfortunately `compareAndSet` check if value is equal not less or more.

Abyss777: I see that `AtomicLong` is not fit. May be `volatile` will be enough?

tananaev: No, I was thinking about more complicated check. Something like this:
```java
long lastUpdate = devicesLastUpdate.get();
if (force || System.currentTimeMillis() - lastUpdate > dataRefreshDelay) {
    if (devicesLastUpdate.compareAndSet(lastUpdate, System.currentTimeMillis())) {
```

src/org/traccar/database/DeviceManager.java
```diff
@@ -68,8 +68,10 @@ public DeviceManager(DataManager dataManager) {
     }
 
     private void updateDeviceCache(boolean force) throws SQLException {
-        boolean needWrite = force || System.currentTimeMillis() - devicesLastUpdate.get() > dataRefreshDelay;
-        if (needWrite) {
+
+        long lastUpdate = devicesLastUpdate.get();
+        if (force || System.currentTimeMillis() - lastUpdate > dataRefreshDelay
```
tananaev: I believe you need to add extra parentheses for `||`.

src/org/traccar/database/DeviceManager.java
Abyss777: I get your point. My understanding atomic is not good enough.

Abyss777:
Done.

tananaev:
Just added a small comment. Sorry :)

Abyss777:
You are right. Done.

tananaev:
Looks good. I hope it works as well.

Abyss777:
I've tested as could do. But without high concurrent.

tananaev:
I guess we can try with the next release. There were some database changes, so I don't want to update database schema on demo server before next official release.

PR #2140, opened 2016-07-22
Update swagger.json

Abyss777:
- actualized recent changes
- replaced `number` to `integer` types where needed (according to https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types)
- added `attributes`
- added missed fields in `Position` model

PS: I can generate whatever I want from `swagger.json`, but generating `swagger.json` is a real quest that I've failed. I've done it manually in http://editor.swagger.io

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -159,6 +159,59 @@
                 }
             }
         },
+        "/devices/geofences": {
+            "post": {
+                "consumes": [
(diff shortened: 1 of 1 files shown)


tananaev:
Thanks. Merged.

PR #2149, opened 2016-07-25
Move lastPosition update to the end of pipeline

Abyss777:
- Optimize MotionEventHandler to do not use Device.motion
- Remove Device.motion from model and database
- some optimizations

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -167,7 +167,7 @@
     </entry>
 
     <entry key='database.updateDeviceStatus'>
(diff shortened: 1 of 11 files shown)


tananaev:
One thing unrelated to this pull request, but I just noticed it now. There is an extra space in `SPEED_THRESHOLD` variable definition. Please fix.

tananaev:
Pull request looks fine. Can we remove status as well?

Abyss777:
Space fixed.
Yes, we can. Should I do it here, or in other pull request?

tananaev:
Let's do another pull request for it. Thanks

PR #2154, opened 2016-07-25
Group and server attributes

Abyss777:
Added attributes to group and server

Controllers with attributes combined to one. Any suggestions to cut the name?

PS: we should keep in mind, that sever object can be requested without authentication, and do not store sensitive information in server attributes.


web/app/view/GroupDialog.js
```diff
@@ -40,5 +40,18 @@ Ext.define('Traccar.view.GroupDialog', {
             displayField: 'name',
             valueField: 'id'
         }]
-    }
+    },
+
+    buttons: [{
```
tananaev: Should we move this to the base class. Most (or all?) models have attributes now.

web/app/view/ServerDialog.js
```diff
@@ -85,5 +85,18 @@ Ext.define('Traccar.view.ServerDialog', {
             fieldLabel: Strings.settingsTwelveHourFormat,
             allowBlank: false
         }]
-    }
+    },
+
+    buttons: [{
```
tananaev: Same for all dialogs.

web/app/view/UserDialogController.js
```diff
@@ -48,19 +44,5 @@ Ext.define('Traccar.view.UserDialogController', {
             });
         }
         button.up('window').close();
-    },
-
-    showAttributesView: function (button) {
```
tananaev: What happened to this method? Is it defined in the base class?

tananaev:
Done with review.

web/app/view/UserDialogController.js
Abyss777: It was the same as in `DeviceDialogController`.
I moved it to `BaseEditDialogWithAttributesController` actually git recognized it like renaming from `DeviceDialogController`.

tananaev: OK, please fix other comments and I'll merge it.

web/app/view/GroupDialog.js
Abyss777: It was my first decision, but was something wrong and page stopped loading.
Geofences also have attributes, but we do not use it for now. I'll try how it will look with 4 buttons ( + Area).

tananaev: We can have custom bottom bar for geofences, but share same implementation for other dialogs. If it doesn't work, I guess you missed something. I don't see why it wouldn't work.

Abyss777:
Hm, I've broken git resolving conflict. I'd better do a new pull request.

PR #2155, opened 2016-07-25
Added group and server attributes

Abyss777:
Sorry for double. Git still confuses me sometime.

I've managed to move everything to `BaseEditDialog`. The problem was that I tried to add
````
requires: [
        'Traccar.view.Attributes'
    ],
````
but it  should not be there.

Geofence buttons I leaved as was.


tananaev:
Merged, thanks.

If you want to learn more about git, I would recommend to read "Pro Git". It's the best book about git. You don't really need to read all of it. Probably first half should be enough to be very confident with git in most use cases.

Abyss777:
I'm really sorry. Chrome cache screwed me.
`requires` is necessary but `AttributeDialog` uses `BaseEditDialog` that uses `attributesView` that uses `AttributeDialog`  ... there is a cycle. 

I'll fix it as soon as possible. Today or tomorrow morning.

tananaev:
Send me a pull request when it's ready.

PR #2158, opened 2016-07-25
Fix attribute editing

Abyss777:
We can't avoid code duplicating. Requires are reverted. Attribute dialog extends `BaseDialog` directly.
This fixes cycle in requires.

Tested more carefully, there is no issues now.
Sorry again.


tananaev:
Merged. Thanks

PR #2159, opened 2016-07-25
Fixed commands for compatibility with some chinese devices

orcoliver:
Some chinese devices as BW08, not receive the "Relay,1#" command and not cut the engine.
From communication protocol, I change the command for "DYD,000000#".
Regards.


tananaev:
Two things:

1. You need to fix checkstyle errors
2. It should be configurable what format to use

orcoliver:
Hi Anton,
Thankyou for response.
Corrected the first point after merge with upstream.
For the second point, I don't understand if you are referencing to the command or the style.
Regards.

PD: Sorry for my bad english.

tananaev:
I am talking about command itself. One format works for some devices, other format works for other devices, so it should be configurable (at least through the config file).

tananaev:
I have implemented support for both options (see 989956915085944d6d5ac550020a7a854c68b541).

PR #2161, opened 2016-07-26
Add email configuration options for geofencing notification.

jenkstom:
Add email settings to config files for geofence notifications.

### DIFF
--- a/setup/unix/traccar.xml
+++ b/setup/unix/traccar.xml
@@ -24,6 +24,14 @@
     <entry key='event.motionHandler'>true</entry>
     <entry key='event.geofenceHandler'>true</entry>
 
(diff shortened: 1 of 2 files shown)


tananaev:
I don't think it makes much sense to add those lines because there are no meaningful default values. User still has to configure it manually.

jenkstom:
Could I add them as XML comments? It took me a while to figure out why the notifications weren't working. I had to dig through the source code and I don't know much about Java. It would have been easier if it were in the config file.

tananaev:
You could search on the forum. I also plan to write some official documentation, but I haven't got time to do it yet.

tananaev:
Configuration has been updated:

https://www.traccar.org/configuration-file/

PR #2163, opened 2016-07-27
OverspeedEventHandler enhancement to use attributes

Abyss777:
I've changed OverspeedEventHandler to use speed limit from attributes. Attribute name is `speedLimit`
Attribute lookup starts from Device -> Groups -> Server and Device has higher priority

Because lookup groups is expensive, I've added `event.overspeed.groupsEnabled` config to enable it.
There is two idea: rise event on every "overspeeding" point (by default) or only once during period.
`event.overspeed.riseOnce` enables it.

Also I've removed Device.status from DB. It is really small commit to separate it.

Tested for a two days. Works OK.

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -45,6 +45,7 @@
     <entry key='event.suppressRepeated'>60</entry>
     <entry key='event.overspeedHandler'>true</entry>
     <entry key='event.overspeed.groupsEnabled'>true</entry>
+    <entry key='event.overspeed.riseOnce'>true</entry>
     <entry key='event.motionHandler'>true</entry>
     <entry key='event.geofenceHandler'>true</entry>
     <entry key='event.alertHandler'>true</entry>
--- a/debug.xml
+++ b/debug.xml
@@ -45,7 +45,7 @@
 
     <entry key='event.enable'>true</entry>
     <entry key='event.overspeedHandler'>true</entry>
-    <entry key='event.overspeed.riseOnce'>true</entry>
+    <entry key='event.overspeed.notRepeat'>true</entry>
     <entry key='event.motionHandler'>true</entry>
     <entry key='event.geofenceHandler'>true</entry>
     <entry key='event.alertHandler'>true</entry>
--- a/debug.xml
+++ b/debug.xml
@@ -42,7 +42,6 @@
     <entry key='logger.file'>target/tracker-server.log</entry>
 
     <entry key='event.enable'>true</entry>
-    <entry key='event.suppressRepeated'>60</entry>
     <entry key='event.overspeedHandler'>true</entry>
     <entry key='event.overspeed.groupsEnabled'>true</entry>
     <entry key='event.overspeed.riseOnce'>true</entry>
--- a/debug.xml
+++ b/debug.xml
@@ -41,9 +41,10 @@
     <entry key='logger.level'>all</entry>
     <entry key='logger.file'>target/tracker-server.log</entry>
 
+    <entry key='deviceManager.lookupGroupsAttribute'>true</entry>
+
     <entry key='event.enable'>true</entry>
     <entry key='event.overspeedHandler'>true</entry>
-    <entry key='event.overspeed.groupsEnabled'>true</entry>
     <entry key='event.overspeed.riseOnce'>true</entry>
     <entry key='event.motionHandler'>true</entry>
     <entry key='event.geofenceHandler'>true</entry>
--- a/debug.xml
+++ b/debug.xml
@@ -171,7 +171,7 @@
     </entry>
 
     <entry key='database.updateDeviceStatus'>
-        UPDATE devices SET status = :status, lastUpdate = :lastUpdate WHERE id = :id;
+        UPDATE devices SET lastUpdate = :lastUpdate WHERE id = :id;
     </entry>
 
     <entry key='database.deleteDevice'>
--- a/debug.xml
+++ b/debug.xml
@@ -44,7 +44,7 @@
     <entry key='event.enable'>true</entry>
     <entry key='event.suppressRepeated'>60</entry>
     <entry key='event.overspeedHandler'>true</entry>
-    <entry key='event.globalSpeedLimit'>90</entry>
+    <entry key='event.overspeed.groupsEnabled'>true</entry>
     <entry key='event.motionHandler'>true</entry>
     <entry key='event.geofenceHandler'>true</entry>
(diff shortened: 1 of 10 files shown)


src/org/traccar/events/OverspeedEventHandler.java
```diff
@@ -50,8 +55,48 @@ public OverspeedEventHandler() {
...
+        }
+        double oldSpeed = 0;
+        if (riseOnce) {
+            Position lastPosition = Context.getDeviceManager().getLastPosition(position.getDeviceId());
+            if (lastPosition != null) {
+                oldSpeed = lastPosition.getSpeed();
+            }
+        }
+        speedLimit = UnitsConverter.knotsFromKph(speedLimit);
+        if (speedLimit != 0 && speed > speedLimit && oldSpeed <= speedLimit) {
```
tananaev: Have you tested the logic? It seems like if `riseOnce` (by the way, you should probably rename it to something like `repeat`) is set to true, `oldSpeed` will be always 0 and you will always get an event. Am I missing something?

src/org/traccar/events/OverspeedEventHandler.java
```diff
@@ -50,8 +55,48 @@ public OverspeedEventHandler() {
 
         Collection<Event> events = new ArrayList<>();
         double speed = position.getSpeed();
+        double speedLimit = 0;
 
-        if (globalSpeedLimit != 0 && speed > globalSpeedLimit) {
+        if (device.getAttributes().containsKey(ATTRIBUTE_SPEED_LIMIT)) {
+            speedLimit = Double.parseDouble((String) device.getAttributes().get(ATTRIBUTE_SPEED_LIMIT));
+        }
+        if (speedLimit == 0 && checkGroups) {
```
tananaev: Should we move attribute lookup logic somewhere else. It will be reused for other attributes as well, right?

src/org/traccar/events/OverspeedEventHandler.java
Abyss777: Yes, I have.
As you suggested in email discussion, by default it should repeat events on every "overspeeding" point. `riseOnce` disables this behavior, "Once" is right word. 
Actually if `riseOnce` is false `oldSpeed` will be always zero, and third part of compare will be always true. (`oldSpeed <= speedLimit`)
Because of that whole compare will be true only if `speed > speedLimit` and it is default behavior.

src/org/traccar/events/OverspeedEventHandler.java
Abyss777: Yes, It can be used elsewhere . What is right place to it? `DeviceManager`?

tananaev: `DeviceManager` might be a good place, or just another utility class.

src/org/traccar/events/OverspeedEventHandler.java
tananaev: Wait, so we request events from the database on every position report? When do we remove events from the database?

Abyss777: Actually, yes.  We do it to avoid too frequent events. We never remove events from DB.

tananaev: I guess I missed it before. We should get rid of this behaviour as soon as possible. What if we have millions of events in the database? We should never request anything from the database for every position.

Abyss777: You are right. I'll do that. Events can be filtered later in reports ore somewhere else.

tananaev: Thanks. It applies for all events. I guess we can have another event filtering handler (which can cache old events or just type and timestamp or something).

Abyss777: But... we still requesting server object from DB...

tananaev: Good point. Maybe we should cache it as well, but server table always has only one record, so it shouldn't be as big of a problem as requesting events.

Abyss777: Actually `Server` is cached in `PermissionManager` but used only there and never updated.
For example updating `readOnly` via API won't affect until sever restart.
I'll organize it too.

tananaev: Great, thanks again.

Abyss777:
My mistake, Server object was actually being updated with `refresh()` but I remember case that it did not affect on `readOnly` .  Anyway  I added proxy functions to retrieve and update cached object. And made it `volatile` to work correctly from different threads.

Detached `lookupAttribute` to function and moved to `DeviceManager`. There is side effect. Because functions does not know type of looking attribute it stops looking just when find it. And if "speedLimit" is zero handler will always skip it. That means users can made exclusions for devices or groups just setting "speedLimit" to zero.

Removed all DB requests during analyzing positions.

Tested on only one real trip. Looks working OK.

debug.xml
```diff
@@ -41,10 +41,11 @@
     <entry key='logger.level'>all</entry>
     <entry key='logger.file'>target/tracker-server.log</entry>
 
+    <entry key='deviceManager.lookupGroupsAttribute'>true</entry>
+
     <entry key='event.enable'>true</entry>
-    <entry key='event.suppressRepeated'>60</entry>
     <entry key='event.overspeedHandler'>true</entry>
-    <entry key='event.globalSpeedLimit'>90</entry>
+    <entry key='event.overspeed.riseOnce'>true</entry>
```
tananaev: Can you please rename parameter to `repeat`. Rise once sounds a bit confusing.

tananaev:
Just one small comment. Otherwise looks good. Assuming that it works :)

debug.xml
Abyss777: The meaning is opposite. `notRepeat` is fit?

tananaev: We can just inverse the logic and set the name to "repeat". I guess
"notRepeat" is also OK.

Abyss777: Inversing the logic will inverse default behavior. If we omit parameter it is assumed as "false". 
Should I inverse default behavior?

tananaev: I guess we can. Default doesn't really matter that much because you can
always explicitly specify what you want.

Abyss777:
I've renamed. If need inversing, I'll do it a bit later.

tananaev:
Merged, thanks.

PR #2165, opened 2016-07-29
Highlight only active alarm

Abyss777:
If you don't mind I would highlight only active alarm.

### DIFF
--- a/web/app/view/State.js
+++ b/web/app/view/State.js
@@ -36,7 +36,7 @@ Ext.define('Traccar.view.State', {
         dataIndex: 'value',
         flex: 1,
         renderer: function (value, metaData, record) {
-            if (record.get('name') === 'Alarm') {
+            if (record.get('name') === 'Alarm' && value === Ext.Msg.buttonText.yes) {
                 metaData.tdCls = 'view-color-red';
             }
             return value;


tananaev:
I think we need to add `alarm` attribute only if there is an alarm. Also, we need to standardize alarm types. Also, there should be no boolean values, it should be type of the alarm.

Abyss777:
OK then I'll change Granit protocol decoder to add attribute only if alarm is active

tananaev:
Yes please. Use types from here:

https://github.com/tananaev/traccar/blob/master/src/org/traccar/model/Position.java#L69

Add new if some type is missing.

Also, it would be nice if you can change other protocols as well... if you have time.

Abyss777:
I'll try but I should have understanding type of alarm. If I'll find it in documentation. 
Actually, there is no type of alarm in Granit protocol, just `флаг тревоги`.
Will we add `ALARM_COMMON = "common";` ? or something like that?

Abyss777:
or `general` ?

tananaev:
Yes, I think we can add `general` type.

Abyss777:
What will we do with multiple alarms? For example, Ulbotech has 8 meaningful bits of alarm value, that can be fired at same time.

tananaev:
I guess we can add all of them into the same value (separated by comma or something). Another option is to show only most important one. I guess that it's unlikely that you get multiple alarms at the same time in the real life.

Abyss777:
I have some difficulties with understanding some alarms.
In the TK103 documentation:

> X The type of alarm for BO01X up alarm message.1character,16
advance system, ASCII character
0 Cut off vehicle oil
1  Happen accident
2 Vehicle rob (SOS help)
3 Vehicle anti-theft alarm
4 Vehicle low speed alarm
5 Vehicle over speed alarm
6 Alarm out of Geo-fence

What do you think "0 Cut off vehicle oil" can mean?

Also there is a problem on https://www.traccar.org/protocols/ I'm trying to download `tzone` documentation, but get `tlt2h` instead

tananaev:
I think cut off oil means stop/block the engine.

As for tzone, I've just updated the link.

PR #2166, opened 2016-07-29
Organized alarms in the first part of protocols

Abyss777:
I know it is not good to do not finished pull request, but it is a huge work to analyze documentation and decode alarms. Frankly, I want to avoid redoing it if I made a mistake in the beginning.
Please take a look if I do it in a right way of course if you have a time. If there is a principle problem let me know.

I decided to not combine different alarms and did like you in `MiniFinder`

It is not finished yet. I'll add other protocols gradually in this pull request.

### DIFF
--- a/src/org/traccar/model/Position.java
+++ b/src/org/traccar/model/Position.java
@@ -68,20 +68,12 @@ public class Position extends Message {
     public static final String PREFIX_COUNT = "count";
     public static final String PREFIX_BUTTON = "button";
 
-    public static final String ENDING_PRESSED = "Pressed";
-    public static final String ENDING_RELEASED = "Released";
-    public static final String ENDING_OPEN = "Open";
-    public static final String ENDING_CLOSE = "Close";
-    public static final String ENDING_ACTIVE = "Active";
-    public static final String ENDING_INACTIVE = "Inactive";
-
-
     public static final String ALARM_GENERAL = "general";
     public static final String ALARM_SOS = "sos";
     public static final String ALARM_VIBRATION = "vibration";
     public static final String ALARM_MOVEMENT = "movement";
+    public static final String ALARM_LOW_SPEED = "lowspeed";
     public static final String ALARM_OVERSPEED = "overspeed";
-    public static final String ALARM_OVERSPEED_END = "overspeedEnd";
     public static final String ALARM_FALL_DOWN = "fallDown";
     public static final String ALARM_LOW_BATTERY = "lowBattery";
     public static final String ALARM_MOTION = "motion";
@@ -91,19 +83,9 @@ public class Position extends Message {
     public static final String ALARM_GEOFENCE = "geofence";
     public static final String ALARM_GEOFENCE_ENTER = "geofenceEnter";
     public static final String ALARM_GEOFENCE_EXIT = "geofenceExit";
-    public static final String ALARM_ODB = "odb";
-    public static final String ALARM_POWER_LOW = "powerLow";
-    public static final String ALARM_PARKING = "parking";
-    public static final String ALARM_CHARGE_BEGIN = "chargeBegin";
-    public static final String ALARM_CHARGE_END = "chargeEnd";
-    public static final String ALARM_RFID_NEW = "rfidNew";
-    public static final String ALARM_SLEEP_INTO = "intoSleep";
-    public static final String ALARM_SLEEP_FROM = "fromSleep";
-    public static final String ALARM_TRACKER_ON = "trackerOn";
-    public static final String ALARM_SHUTDOWN = "shutdown";
-    public static final String ALARM_GPS_BLIND_ENTER = "gpsBlindEnter";
-    public static final String ALARM_GPS_BLIND_EXIT = "gpsBlindExit";
     public static final String ALARM_GPS_ANTENNA_CUT = "gpsAntennaCut";
+    public static final String ALARM_OIL_CUT = "oilCut";
+    public static final String ALARM_ACCIDENT = "accident";
 
 
     private String protocol;
--- a/src/org/traccar/model/Position.java
+++ b/src/org/traccar/model/Position.java
@@ -79,7 +79,6 @@ public class Position extends Message {
     public static final String ALARM_MOTION = "motion";
     public static final String ALARM_FAULT = "fault";
     public static final String ALARM_POWER_OFF = "powerOff";
-    public static final String ALARM_JAMMING = "jamming";
     public static final String ALARM_GEOFENCE = "geofence";
     public static final String ALARM_GEOFENCE_ENTER = "geofenceEnter";
     public static final String ALARM_GEOFENCE_EXIT = "geofenceExit";
--- a/src/org/traccar/model/Position.java
+++ b/src/org/traccar/model/Position.java
@@ -54,6 +54,7 @@ public class Position extends Message {
     public static final String KEY_APPROXIMATE = "approximate";
     public static final String KEY_THROTTLE = "throttle";
     public static final String KEY_MOTION = "motion";
+    public static final String KEY_GUARD = "guard";
 
     public static final String KEY_OBD_SPEED = "obd-speed";
     public static final String KEY_OBD_ODOMETER = "obd-odometer";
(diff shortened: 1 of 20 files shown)


src/org/traccar/protocol/NoranProtocolDecoder.java
```diff
@@ -90,7 +90,18 @@ protected Object decode(
 
             position.setValid(BitUtil.check(buf.readUnsignedByte(), 0));
 
-            position.set(Position.KEY_ALARM, buf.readUnsignedByte());
+            short alarm = buf.readUnsignedByte();
+            if (alarm != 0x80) {
```
tananaev: This check seems redundant.

tananaev:
I'll take a closer look tomorrow. It seems like some of those events are not really alarms. Need to think about how to handle it properly.

Abyss777:
I agree, some of them just `info`

src/org/traccar/model/Position.java
```diff
@@ -65,15 +66,45 @@
     public static final String PREFIX_ADC = "adc";
     public static final String PREFIX_IO = "io";
     public static final String PREFIX_COUNT = "count";
+    public static final String PREFIX_BUTTON = "button";
 
+    public static final String ENDING_PRESSED = "Pressed";
```
tananaev: I think all this (`ENDING_*`) should be just info. Possibly we can ignore it for now because most likely IO status is already saved in some other attribute.

src/org/traccar/model/Position.java
```diff
@@ -65,15 +66,45 @@
...
+    public static final String ENDING_ACTIVE = "Active";
+    public static final String ENDING_INACTIVE = "Inactive";
+
+
+    public static final String ALARM_GENERAL = "general";
     public static final String ALARM_SOS = "sos";
     public static final String ALARM_VIBRATION = "vibration";
     public static final String ALARM_MOVEMENT = "movement";
     public static final String ALARM_OVERSPEED = "overspeed";
+    public static final String ALARM_OVERSPEED_END = "overspeedEnd";
```
tananaev: I think we don't need this one. Just ignore it.

src/org/traccar/model/Position.java
```diff
@@ -65,15 +66,45 @@
...
     public static final String ALARM_FALL_DOWN = "fallDown";
     public static final String ALARM_LOW_BATTERY = "lowBattery";
     public static final String ALARM_MOTION = "motion";
     public static final String ALARM_FAULT = "fault";
+    public static final String ALARM_POWER_OFF = "powerOff";
+    public static final String ALARM_JAMMING = "jamming";
+    public static final String ALARM_GEOFENCE = "geofence";
+    public static final String ALARM_GEOFENCE_ENTER = "geofenceEnter";
+    public static final String ALARM_GEOFENCE_EXIT = "geofenceExit";
+    public static final String ALARM_ODB = "odb";
```
tananaev: I guess it should be OBD, not ODB. What does it mean?

src/org/traccar/model/Position.java
```diff
@@ -65,15 +66,45 @@
...
+    public static final String ALARM_POWER_LOW = "powerLow";
+    public static final String ALARM_PARKING = "parking";
+    public static final String ALARM_CHARGE_BEGIN = "chargeBegin";
```
tananaev: Most of the alarm below are not really alarms. I guess we can ignore them.

src/org/traccar/protocol/TotemProtocolDecoder.java
```diff
@@ -159,6 +159,34 @@ public TotemProtocolDecoder(TotemProtocol protocol) {
             .any()
             .compile();
 
+    private String decodeAlarm(String value) {
+        switch (value) {
```
tananaev: Can we do switch on a number value? It should be faster...

src/org/traccar/protocol/TotemProtocolDecoder.java
```diff
@@ -193,9 +221,10 @@ protected Object decode(
         position.setDeviceId(deviceSession.getDeviceId());
 
         if (pattern == PATTERN1 || pattern == PATTERN2) {
-
-            position.set(Position.KEY_ALARM, parser.next());
-
+            String alarm = decodeAlarm(parser.next());
```
tananaev: You can do something like this instead of adding a variable:
```java
if (parser.hasNext()) {
    position.set(Position.KEY_ALARM, decodeAlarm(parser.next()));
}
```
That's the code style I usually use for optional fields.

tananaev:
Added some comments. I think we can start with small number of decoders to avoid huge code reviews.

src/org/traccar/model/Position.java
Abyss777: Yes, sure it s typo. OBD is right.
As i understood from documentation device can read "CheckEngine" errors via OBD and signalize about them.

tananaev: I think it shouldn't be alarm. We already handle OBD DTC codes for some other protocols.

Abyss777:
Seems fix all comments
In defense of `ALARM_GPS_ANTENNA_CUT` - it can be useful to determine if driver want to hide his location and can be retrieved via alarm only.

Not sured in `ALARM_OIL_CUT`

Also added other protocols with simple `ALARM_GENERAL`

H02 decreased to 3 common types

src/org/traccar/model/Position.java
```diff
@@ -54,6 +54,7 @@
     public static final String KEY_APPROXIMATE = "approximate";
     public static final String KEY_THROTTLE = "throttle";
     public static final String KEY_MOTION = "motion";
+    public static final String KEY_GUARD = "guard";
```
tananaev: What does guard mean? Should it be renamed to "armed"?

src/org/traccar/model/Position.java
```diff
@@ -65,15 +66,26 @@
...
     public static final String ALARM_FALL_DOWN = "fallDown";
     public static final String ALARM_LOW_BATTERY = "lowBattery";
     public static final String ALARM_MOTION = "motion";
     public static final String ALARM_FAULT = "fault";
+    public static final String ALARM_POWER_OFF = "powerOff";
+    public static final String ALARM_GEOFENCE = "geofence";
+    public static final String ALARM_GEOFENCE_ENTER = "geofenceEnter";
+    public static final String ALARM_GEOFENCE_EXIT = "geofenceExit";
+    public static final String ALARM_GPS_ANTENNA_CUT = "gpsAntennaCut";
+    public static final String ALARM_OIL_CUT = "oilCut";
```
tananaev: I don't think we need oil cut alarm. I'm guessing that it means operator sent the command to stop the engine.

src/org/traccar/protocol/CarTrackProtocolDecoder.java
```diff
@@ -99,7 +99,7 @@ protected Object decode(
         odometer = odometer.replace("?", "F");
         position.set(Position.KEY_ODOMETER, Integer.parseInt(odometer, 16));
 
-        position.set(Position.KEY_ALARM, parser.next());
+        parser.next(); //there is no meaningful alarms
```
tananaev: Add a space after slashes.

src/org/traccar/protocol/Gps103ProtocolDecoder.java
```diff
@@ -102,6 +102,23 @@ public Gps103ProtocolDecoder(Gps103Protocol protocol) {
             .any()
             .compile();
 
+    private String decodeAlarm(String value) {
+        if (value == "tracker") {
```
tananaev: Do switch instead of ifs. Also, you can't really compare string like that in Java.

src/org/traccar/protocol/MeiligaoProtocolDecoder.java
```diff
@@ -151,6 +151,19 @@ private String getMeiligaoServer(Channel channel) {
         return server;
     }
 
+    private String decodeAlarm(short value) {
+        switch (value) {
+        case 0x01: return Position.ALARM_SOS;
```
tananaev: Return statement should be on the next line. For all cases here.

src/org/traccar/protocol/NavisProtocolDecoder.java
```diff
@@ -95,7 +95,11 @@ private ParseResult parsePosition(DeviceSession deviceSession, ChannelBuffer buf
 
         buf.skipBytes(6); // event time
 
-        position.set(Position.KEY_ALARM, buf.readUnsignedByte());
+        short guardStatus = buf.readUnsignedByte();
```
tananaev: Should this be renamed to "armedStatus" or something?

src/org/traccar/protocol/NoranProtocolDecoder.java
```diff
@@ -90,7 +90,16 @@ protected Object decode(
 
             position.setValid(BitUtil.check(buf.readUnsignedByte(), 0));
 
-            position.set(Position.KEY_ALARM, buf.readUnsignedByte());
+            short alarm = buf.readUnsignedByte();
+            if (alarm == 1) {
```
tananaev: I guess switch is better here as well.

src/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -75,6 +75,19 @@ public Tk103ProtocolDecoder(Tk103Protocol protocol) {
             .any()
             .compile();
 
+    private String decodeAlarm(int value) {
+        switch (value) {
+        case 0: return Position.ALARM_OIL_CUT;
```
tananaev: Return on the next line.

src/org/traccar/protocol/TotemProtocolDecoder.java
```diff
@@ -159,6 +159,17 @@ public TotemProtocolDecoder(TotemProtocol protocol) {
             .any()
             .compile();
 
+    private String decodeAlarm(Short value) {
+        switch (value) {
+        case 0x01: return Position.ALARM_SOS;
```
tananaev: Return on new line.

tananaev:
Sorry for delay. Just got time now to review everything. Added some minor comments, but in general looks good. Thanks

src/org/traccar/model/Position.java
Abyss777: it is from Navis protocol
````
 Статус охраны / режим
тревожного оповещения
U8 Битовое поле:
0…2 Биты
0 – наблюдение (нет охраны);
1 - профиль охраны №1 (полная охрана);
2 - профиль охраны №2;
3…5 Биты Резерв
6 Бит
0 – рабочий режим
1 – тестовый режим
7 Бит
0 – штатная работа
1 – режим тревожного оповещения
````
I translated it to "Guard status" If "armed" is better, then OK.

src/org/traccar/protocol/MeiligaoProtocolDecoder.java
Abyss777: I have a question, what is the right indentation for `case` ?
Eclipse wants the same level as `switch` but for example in 'OsmAndProtocolDecoder` it is on 4 spaces right.

src/org/traccar/model/Position.java
tananaev: Yes, I think "armed" is a better English word for it. Also, we use it for commands already, so it's also better for consistency.

Abyss777:
All done.

tananaev:
Merged, thanks.

PR #2180, opened 2016-08-02
Master2

ninioe:
Usage of getString(key, defaultVal) function instead of String.key.
It return the value in English if not exist in the current language, else the default value is returned (optional param)


tananaev:
I don't think it's needed. When I generate localization files, they include all strings, so it shouldn't be a problem.

ninioe:
But until all the localization files are ready it's better have the English translation than an empty string

ninioe:
I needed it yesterday

tananaev:
I believe you can get localization as well... using `translate.py` script.

ninioe:
This is great, I will check this tomorrow.
Can we also use the method I proposed so I could send updates to my server even if the translation is not complete yet?

‫ב-3 באוג׳ 2016, בשעה 18:55, ‏‏Anton Tananaev ‏<notifications@github.com> כתב/ה:‬

tananaev:
You don't need complete localization. It will automatically add English version if translation is missing. Just try the script.

ninioe:
it works, thanks

ninioe:
Hi Anton,
can you exclude the en.json from the translate.py script?
we can pull this file from github that is more updated anyway, and we don't want to override it.

tananaev:
There should be no difference between original and generated.

ninioe:
Ok, but the string for "remember me" is missing

tananaev:
It takes some time to automatically synchronize the source strings with transifex.

Issue #2182, opened 2016-08-03
No authentication problem


tananaev:
@Abyss777, can you please take a look at this issue:

https://www.traccar.org/forums/topic/traccar-mail-notifications

It seems like something fails if no authentication is provided.

Abyss777:
I think it can be closed 
https://www.traccar.org/forums/topic/traccar-mail-notifications/#post-7497

tananaev:
Thanks.

PR #2185, opened 2016-08-03
Mail optimization

Abyss777:
Sorry, I've made two of four common mistakes http://www.oracle.com/technetwork/java/javamail/faq/index.html#commonmistakes

Here is fix. Actually, It worked fine, excepting some parameters from user attributes do not applied until server restart.

Also I've added parsing `mail.smtp.ssl.trust` that allow connect to server with self-signed certificate.

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -63,6 +63,8 @@
     for SSL
     <entry key='mail.smtp.port'>465</entry>
     <entry key='mail.smtp.ssl.enable'>true</entry>
+    for self-signed server sertificates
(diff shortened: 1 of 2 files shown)


tananaev:
Merged. Thanks. Can you please comment of the forum thread if you haven't done it already.

Abyss777:
Actually, this changes is not connected to #2182 directly (I founded them while investigating)
I have some thoughts about problem, but need feedback from user.

tananaev:
Sorry, I thought it's a fix for the problem.

PR #2187, opened 2016-08-04
fix #2182

### LINKED ISSUE #2182: No authentication problem

Abyss777:
Fix NullPointer exception without `mail.smtp.username` in config.

I do not have excuses for me. This checks was in some commits of pull request, but why I decided to remove them... and why I decided that `put` allow `null` value...


tananaev:
Merged, thanks.

PR #2188, opened 2016-08-04
Add "Remember" to login panel

ninioe:
Add "Remember" to login panel


tananaev:
What is this? Why are you removing everything?

tananaev:
Also, please make some meaningful names for pull requests if you want to send something.

ninioe:
Ok

tananaev:
Please create another pull request instead of adding to this one.

ninioe:
why?
I will have to add a new branch to do this.
can I close this one and open it again instead?

tananaev:
I guess you can. Make sure it has a meaningful name.

ninioe:
Is the current name OK?

tananaev:
Name is OK, but I don't want to see all the changes that I've already merged. Also, why are removing so many files?

ninioe:
I Removed all the language files from my branch except for en

ninioe:
at the time I didn't know about the translate.py script

ninioe:
I can add those language files back to my branch to reduce the number of files changed

ninioe:
sorry for this, it just increase the number of files changed.
this because I translated more he strings on transifex.com and run the translate.py script

ninioe:
what do you suggest?
I can remove those language files from my branch again and put them back on git exclude list

tananaev:
I think you need to create a new branch and do the changes there. Creating branches with git is very easy. You can cherry pick your commits there as well.

ninioe:
why do I need another branch? I'm not having two versions.
Just tell me what you want me to change on this branch and I will.

tananaev:
Because that's the right way of working with git.

I just want to see the new changes.

ninioe:
All the changes there are new and related to my pull request except the changes to the languages files from the translate.py script.
If you want me to fix something just tell me where?

‫ב-7 באוג׳ 2016, בשעה 17:09, ‏‏Anton Tananaev ‏<notifications@github.com> כתב/ה:‬

tananaev:
Please remove all the changes to the language files.

As for the actual changes, I think it's not a good idea to store username and password in the cookie. Can't we just use session id or something? There must be a better way.

ninioe:
I can store the userId in the cookie

‫ב-7 באוג׳ 2016, בשעה 21:00, ‏‏Anton Tananaev ‏<notifications@github.com> כתב/ה:‬

ninioe:
We can't use session id because session expire usually after 20 minutes, also userId is not good because if a user can fake a userId in a cookie it is bad because he could login as other user or worst as admin.
I think saving user and password on non public computer or on mobile phone is not a bad option

ninioe:
About the language files in my branch.
If you run the translate.py script and update the language files on your master branch there will be no changes.
Or I can remove the files from my branch and download your files.
What ever you prefer?

tananaev:
I have updated all language files.

I don't think storing user id would be very secure. Is there a way to extend the session beyond restarting web browser?

ninioe:
I don't think keeping opened sessions on the server is wise, it may lead to memory leak.
You need to save a unique key on the client as a cookie to restore the session securely.

I can think of 2 ways to do this:
1. you can add a new field to the users table like: "secret_key" to match the cookie and produce this unique key each time a user login with remember checked.
2. you can use the combination of userId and the encrypted password on users table to do the match (no need to change the db)

tananaev:
I guess username and password would do for now. That seems like the easiest option. Can you please remove language files.

ninioe:
those language files are new, I can remove them from my branch or you could add them to yours?
about the cookies we can set them on the server side with encryption / decryption

tananaev:
Those languages have 0% translated strings, so I'm not going to add them yet. Please remove those.

I think implementing own encryption is not a good idea. If people want more security, they can just set up proper SSL connection.

ninioe:
Do you want me to add a class for encryp/decrypt like this for the cookies:
http://stackoverflow.com/questions/5307132/is-this-java-encryption-code-thread-safe

tananaev:
How does it add any security? Where would you store the keys?

ninioe:
it could be on the configuration file

ninioe:
here is the class

[DesEncrypter.java.txt](https://github.com/tananaev/traccar/files/406305/DesEncrypter.java.txt)

tananaev:
It should be on the client side as well for encryption, right? It means that everybody would get it anyway.

tananaev:
Merged the changes.

ninioe:
no, we could set the cookie on the server side as well

tananaev:
Server side cookie is a session. That's what I suggested before. You said it can lead to memory leaks.

tananaev:
I see what you mean. You want to encrypt on the server and send to the client to store it. It doesn't add much security though because anybody can copy and re-use encrypted cookie.

ninioe:
no, session use cookie to keep track the session id.
session is the information on the server, and cookie is saved on the client and sent to the server in the headers with the request

ninioe:
we could also use the user's salt for the key

tananaev:
How would it help and how would it work?

ninioe:
if someone steel the user's cookie he could generate a new password and the salt will change

tananaev:
Can you please fully explain how it adds any security.

ninioe:
when the user login with the remember checked the user and password data are encrypted on the server side (with the user's salt as the key) and set as a cookie for the next login session.

ninioe:
every time a user set a new password a new random salt is created:

    public static HashingResult createHash(String password) {
        byte[] salt = new byte[SALT_SIZE];
        RANDOM.nextBytes(salt);
        byte[] hash = function(password.toCharArray(), salt);
        return new HashingResult(
                DatatypeConverter.printHexBinary(hash),
                DatatypeConverter.printHexBinary(salt));
    }

tananaev:
I still don't understand how it would prevent someone from stealing user's cookie and logging in. Can you please explain that part.

ninioe:
no one can prevent from someone from stealing user's cookie if he has access to his computer.
but if the user change his password the cookie is useless.

the other way is to save a key to users table that is valid until he logout, but you will have to add a new field for this and add this to the db change log

ninioe:
in the other way we will save in the cookie the userId and the key that match in field "secret_key" in the users table.
on every login with remember a new key will be produced.

tananaev:
It all sounds overcomplicated. Let's just leave it as it is. SSL should prevent credentials from being stolen.

ninioe:
ok

PR #2190, opened 2016-08-05
Reports API

Abyss777:
Here is initial implementation of Reports API

All resources can return csv
- Routes can be requested for multiple devices and groups
- Events can be requested for multiple devices and groups and multiple types
- General report return distance, maximum and average speed

### DIFF
--- a/src/org/traccar/api/BaseResource.java
+++ b/src/org/traccar/api/BaseResource.java
@@ -15,6 +15,7 @@
  */
 package org.traccar.api;
 
+import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.SecurityContext;
 
 public class BaseResource {
@@ -22,12 +23,19 @@ public class BaseResource {
     @javax.ws.rs.core.Context
     private SecurityContext securityContext;
 
+    @javax.ws.rs.core.Context
+    private HttpHeaders headers;
+
     protected long getUserId() {
         UserPrincipal principal = (UserPrincipal) securityContext.getUserPrincipal();
         if (principal != null) {
             return principal.getUserId();
         }
         return 0;
+    }
+
+    protected String getAcceptHeader() {
+        return headers.getRequestHeaders().getFirst("Accept");
     }
 
 }
--- a/src/org/traccar/api/BaseResource.java
+++ b/src/org/traccar/api/BaseResource.java
@@ -15,16 +15,12 @@
  */
 package org.traccar.api;
 
-import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.SecurityContext;
 
 public class BaseResource {
 
     @javax.ws.rs.core.Context
     private SecurityContext securityContext;
-
-    @javax.ws.rs.core.Context
-    private HttpHeaders headers;
 
     protected long getUserId() {
         UserPrincipal principal = (UserPrincipal) securityContext.getUserPrincipal();
@@ -33,9 +29,4 @@ public class BaseResource {
         }
         return 0;
     }
-
-    protected String getAcceptHeader() {
-        return headers.getRequestHeaders().getFirst("Accept");
-    }
-
 }
--- a/src/org/traccar/reports/ReportUtils.java
+++ b/src/org/traccar/reports/ReportUtils.java
@@ -1,12 +1,7 @@
 package org.traccar.reports;
 
-import java.io.IOException;
-import java.io.OutputStream;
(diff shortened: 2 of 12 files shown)


src/org/traccar/api/BaseResource.java
```diff
@@ -30,4 +34,8 @@ protected long getUserId() {
         return 0;
     }
 
+    protected String getAcceptHeader() {
+        return headers.getRequestHeaders().getFirst("Accept");
```
tananaev: You should be using `Consumes` annotation instead of manually accessing headers.

src/org/traccar/web/CsvBuilder.java
```diff
@@ -0,0 +1,167 @@
...
+
+    private StringBuilder builder = new StringBuilder();
+
+    private void addLineEnding() {
+        builder.append(LINE_ENDING);
+    }
+    private void addSeparator() {
+        builder.append(SEPARATOR);
+    }
+    public void addLine(String value) {
```
tananaev: Why do we need all these `addLine` and `addField` for primitive types. They don't make much sense. Are they actually used?

src/org/traccar/web/CsvBuilder.java
```diff
@@ -0,0 +1,167 @@
...
+                    addLine((boolean) object);
+                    break;
+                default:
+                    addLine(object);
+                    break;
+            }
+        }
+    }
+
+    public String get() {
```
tananaev: Name `build` or something like that would be better.

src/org/traccar/reports/ReportUtils.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.Collection;
+
+import org.traccar.Context;
+
+public final class ReportUtils {
+
+    private ReportUtils() {
+    }
+
+    public static Collection<Long> getReportedDevices(Collection<Long> deviceIds, Collection<Long> groupIds) {
```
tananaev: I don't like the name of the function. It's a bit confusing. Can we rename it to something like `getDeviceList` or maybe `getSelectedDevices`.

src/org/traccar/reports/model/GeneralReport.java
```diff
@@ -0,0 +1,42 @@
+package org.traccar.reports.model;
+
+public class GeneralReport {
```
tananaev: Maybe rename it to `SummaryReport`? Because general sounds like some base class for all reports.

src/org/traccar/reports/model/GeneralReport.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private String deviceName;
+    public String getDeviceName() {
+        return deviceName;
+    }
+    public void setDeviceName(String deviceName) {
+        this.deviceName = deviceName;
+    }
+
+    private double distance = 0;
```
tananaev: No need to explicitly set to zero. Same for other variables.

tananaev:
Added some comments.

src/org/traccar/web/CsvBuilder.java
Abyss777: `addField` 100% used in `addLine(Object object)`
`addLine` for primitives actually never used now, I've made it for universality of `addArray`.
I can move code from that functions to `addLine(Object object)` and `addArray(Collection<?> array)` but I do not see any difference and we will win nothing I guess.

tananaev: It's never supposed to be used outside of the class, so I think we should remove it. There is also not much value even in private methods like this because they are only 2 lines of code each. Can you please remove them.

Abyss777:
All done except `addLine` and `addFiled`

tananaev:
Can you please remove `addLine` and `addField`.

src/org/traccar/web/CsvBuilder.java
Abyss777: OK, sure.

Abyss777:
Done

src/org/traccar/api/BaseResource.java
```diff
@@ -29,5 +29,4 @@ protected long getUserId() {
         }
         return 0;
     }
-
```
tananaev: Please revert this file.

tananaev:
Never mind. Everything else looks good, so I've merged it.

Abyss777:
Not got a chance :)

PR #2197, opened 2016-08-09
Initial implementation Report API for web interface

Abyss777:
- Changed JSON format to bare objects
- Fixed csv MIME type and added header to force browser download
- Combined some strings
- Fixed some `jscs` issues in other code
- Added `deviceId` to SummaryReport

Report still uses only one device and get all events without filtering. I wanted to get working code with not big pull request.
Route report works as before. Events and summary just fill the table.

It was not easy to find solution how download file via request with custom header `Accept`.

### DIFF
--- a/src/org/traccar/api/resource/ReportResource.java
+++ b/src/org/traccar/api/resource/ReportResource.java
@@ -17,12 +17,15 @@ import org.traccar.reports.Route;
 import org.traccar.web.JsonConverter;
 
 @Path("reports")
-@Consumes("application/json")
+@Consumes(javax.ws.rs.core.MediaType.APPLICATION_JSON)
 public class ReportResource extends BaseResource {
+
+    public static final String TEXT_CSV = "text/csv";
+    public static final String CONTENT_DISPOSITION_VALUE = "attachment; filename=report.csv";
 
     @Path("route")
     @GET
-    @Produces("application/json")
+    @Produces(javax.ws.rs.core.MediaType.APPLICATION_JSON)
     public Response getRouteJson(
             @QueryParam("deviceId") final List<Long> deviceIds, @QueryParam("groupId") final List<Long> groupIds,
             @QueryParam("from") String from, @QueryParam("to") String to) throws SQLException {
@@ -32,19 +35,19 @@ public class ReportResource extends BaseResource {
 
     @Path("route")
     @GET
-    @Produces("text/csv")
+    @Produces(TEXT_CSV)
     public Response getRouteCsv(
             @QueryParam("deviceId") final List<Long> deviceIds, @QueryParam("groupId") final List<Long> groupIds,
             @QueryParam("from") String from, @QueryParam("to") String to) throws SQLException {
         return Response.ok(Route.getCsv(getUserId(), deviceIds, groupIds,
                 JsonConverter.parseDate(from), JsonConverter.parseDate(to)))
-                .header("Content-Disposition", "attachment; filename=route.csv")
+                .header(javax.ws.rs.core.HttpHeaders.CONTENT_DISPOSITION, CONTENT_DISPOSITION_VALUE)
                 .build();
     }
 
     @Path("events")
     @GET
-    @Produces("application/json")
+    @Produces(javax.ws.rs.core.MediaType.APPLICATION_JSON)
     public Response getEventsJson(
             @QueryParam("deviceId") final List<Long> deviceIds, @QueryParam("groupId") final List<Long> groupIds,
             @QueryParam("type") final List<String> types,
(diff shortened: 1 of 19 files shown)


src/org/traccar/api/resource/ReportResource.java
```diff
@@ -32,12 +32,14 @@ public Response getRouteJson(
...
     @GET
-    @Produces("application/ms-excel")
+    @Produces("text/csv")
     public Response getRouteCsv(
             @QueryParam("deviceId") final List<Long> deviceIds, @QueryParam("groupId") final List<Long> groupIds,
             @QueryParam("from") String from, @QueryParam("to") String to) throws SQLException {
         return Response.ok(Route.getCsv(getUserId(), deviceIds, groupIds,
-                JsonConverter.parseDate(from), JsonConverter.parseDate(to))).build();
+                JsonConverter.parseDate(from), JsonConverter.parseDate(to)))
+                .header("Content-Disposition", "attachment; filename=route.csv")
```
tananaev: It would be nice to use some constants here. Header name must be available as constant somewhere already. Value can be a constant here (we can just use same filename like `report.csv`).

src/org/traccar/reports/Events.java
```diff
@@ -19,12 +19,13 @@ private Events() {
 
     public static String getJson(long userId, Collection<Long> deviceIds, Collection<Long> groupIds,
             Collection<String> types, Date from, Date to) throws SQLException {
-        JsonObjectBuilder json = Json.createObjectBuilder();
+        JsonArrayBuilder json = Json.createArrayBuilder();
         for (long deviceId: ReportUtils.getDeviceList(deviceIds, groupIds)) {
             Context.getPermissionsManager().checkDevice(userId, deviceId);
             for (String type : types) {
-                json.add(String.valueOf(deviceId), JsonConverter.arrayToJson(Context.getDataManager()
```
tananaev: It didn't work?

web/app/model/ReportSummary.js
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright 2015 Anton Tananaev (anton.tananaev@gmail.com)
```
tananaev: Please update year.

web/app/view/ReportController.js
```diff
@@ -77,6 +90,152 @@ Ext.define('Traccar.view.ReportController', {
...
+                    downloadUrl = URL.createObjectURL(blob);
+                    if (filename) {
+                        a = document.createElement("a");
+                        a.href = downloadUrl;
+                        a.download = filename;
+                        document.body.appendChild(a);
+                        a.click();
+                    }
+                    setTimeout(function () {
+                            url.revokeObjectURL(downloadUrl);
```
tananaev: Indentation is incorrect here.

web/app/view/ReportController.js
```diff
@@ -77,6 +90,152 @@ Ext.define('Traccar.view.ReportController', {
...
+                type = response.getResponseHeader('Content-Type');
+                blob = new Blob([response.responseText], { type: type });
+                if (typeof window.navigator.msSaveBlob !== 'undefined') {
+                    // IE workaround
+                    window.navigator.msSaveBlob(blob, filename);
+                } else {
+                    url = window.URL || window.webkitURL;
+                    downloadUrl = URL.createObjectURL(blob);
+                    if (filename) {
+                        a = document.createElement("a");
```
tananaev: Can we have a better name for `a`?

web/l10n/en.json
```diff
@@ -132,5 +132,12 @@
     "alarmFault": "Fault Alarm",
     "notificationType": "Type of Notification",
     "notificationWeb": "Send via Web",
-    "notificationMail": "Send via Mail"
+    "notificationMail": "Send via Mail",
+    "reportRoute": "Route",
+    "reportEvents": "Events",
+    "reportSummary": "Summary",
+    "reportCsv": "Csv",
```
tananaev: I guess it's better to make value upper-case `CSV`.

web/app/view/ReportController.js
```diff
@@ -77,6 +90,152 @@ Ext.define('Traccar.view.ReportController', {
     },
 
     selectReport: function (position, center) {
-        this.getView().getSelectionModel().select([position], false, true);
+        if (position instanceof Traccar.model.Position) {
+            this.getView().getSelectionModel().select([position], false, true);
+        }
+    },
+
+    doDownloadCsv: function (requestUrl, requestParams) {
```
tananaev: Rename to `downloadCsv`?

tananaev:
Added some comments.

src/org/traccar/reports/Events.java
Abyss777: It worked but did over-complicating in web part. Invent new model with `deviceId` and array and other things. Bare array of objects is easier and still can be filtered by `deviceId` on `Ext.store` side.

`ArrayBuilder` with `arrayToJson` for every device also did not fit because it produced array of arrays

web/app/view/ReportController.js
Abyss777: `elementA` is OK?

tananaev: Sure, it would make it much clearer.

src/org/traccar/api/resource/ReportResource.java
```diff
@@ -17,12 +17,15 @@
 import org.traccar.web.JsonConverter;
 
 @Path("reports")
-@Consumes("application/json")
+@Consumes(javax.ws.rs.core.MediaType.APPLICATION_JSON)
```
tananaev: Can we import instead of using full name?

Abyss777:
All done.
Also renamed `summary*` strings, I hope we will use it in "TripsReport"

web/app/view/ReportController.js
```diff
@@ -77,6 +90,152 @@ Ext.define('Traccar.view.ReportController', {
...
+
+    downloadCsv: function (requestUrl, requestParams) {
+        Ext.Ajax.request({
+            url: requestUrl,
+            method: 'GET',
+            params: requestParams,
+            headers: {
+                Accept: 'text/csv'
+            },
+            success: function (response) {
```
tananaev: This function looks a bit messy. Is there any cleaner way to do it?

Abyss777: It was the hardest part of code to pass header and force browser to download result.
`Ext.form.action.StandardSubmit` replaces header to his own, "nonstandart submit" do not force browser show "save as" dialog.

I can replace DOM manipulation to `Ext.DomHelper` but idea should stay the same

tananaev: So, we can't just download the file regular way?

Abyss777: Unfortunately, there is no regular way. Every way is a trick. (hidden iframes, self clicked links and so on)
Simpliest way is hiden `Ext.form.Panel`. Everything does ExtJS for us, but it replaces header to his own.

We can change server side to response with csv by default.

tananaev:
OK, let's leave it as it is for now. Everything else looks good.

PR #2202, opened 2016-08-10
Adjusted swagger.json

stevenbouma:
query parameters for GET /positions are not marked required. This allows generated clients to call /positions with or without additional parameters. Also updated Position model to include deviceId.


tananaev:

PR #2206, opened 2016-08-11
Ignition and Motor Hours

Abyss777:
I just wanted to do small pull request, but...

- Added `IgnitionEventHandler`
- Added Motor Hours to Summary report

Also
- Added Unit tests for some event handlers
- Added message template for alarms
- If you don't mind, I've added `ignition` to `test-generator.py`

### DIFF
--- a/src/org/traccar/events/IgnitionEventHandler.java
+++ b/src/org/traccar/events/IgnitionEventHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Anton Tananaev (anton.tananaev@gmail.com)
+ * Copyright 2016 Anton Tananaev (anton.tananaev@gmail.com), Andrey Kunitsyn (abyss@fox5.ru)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -40,8 +40,8 @@ public class IgnitionEventHandler extends BaseEventHandler {
 
         boolean ignition = false;
         Object ignitionObject = position.getAttributes().get(Position.KEY_IGNITION);
-        if (ignitionObject != null && Boolean.parseBoolean(ignitionObject.toString())) {
-            ignition = true;
+        if (ignitionObject != null && ignitionObject instanceof Boolean) {
+            ignition = (Boolean) ignitionObject;
         }
 
         boolean oldIgnition = false;
@@ -50,8 +50,8 @@ public class IgnitionEventHandler extends BaseEventHandler {
         if (lastPosition != null) {
             oldIgnitionObject = lastPosition.getAttributes().get(Position.KEY_IGNITION);
         }
-        if (oldIgnitionObject != null && Boolean.parseBoolean(oldIgnitionObject.toString())) {
-            oldIgnition = true;
+        if (oldIgnitionObject != null && oldIgnitionObject instanceof Boolean) {
+            oldIgnition = (Boolean) oldIgnitionObject;
         }
 
         if (ignition && !oldIgnition) {
--- a/src/org/traccar/events/IgnitionEventHandler.java
+++ b/src/org/traccar/events/IgnitionEventHandler.java
@@ -1,5 +1,6 @@
 /*
- * Copyright 2016 Anton Tananaev (anton.tananaev@gmail.com), Andrey Kunitsyn (abyss@fox5.ru)
+ * Copyright 2016 Anton Tananaev (anton.tananaev@gmail.com)
+ * Copyright 2016 Andrey Kunitsyn (abyss@fox5.ru)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/org/traccar/events/IgnitionEventHandler.java
+++ b/src/org/traccar/events/IgnitionEventHandler.java
@@ -17,22 +17,14 @@ package org.traccar.events;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.List;
 
 import org.traccar.BaseEventHandler;
 import org.traccar.Context;
-import org.traccar.database.GeofenceManager;
 import org.traccar.model.Device;
 import org.traccar.model.Event;
 import org.traccar.model.Position;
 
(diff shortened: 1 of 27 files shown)


src/org/traccar/events/IgnitionEventHandler.java
```diff
@@ -0,0 +1,67 @@
...
+            ignition = true;
+        }
+
+        boolean oldIgnition = false;
+        Object oldIgnitionObject = null;
+        Position lastPosition = Context.getIdentityManager().getLastPosition(position.getDeviceId());
+        if (lastPosition != null) {
+            oldIgnitionObject = lastPosition.getAttributes().get(Position.KEY_IGNITION);
+        }
+        if (oldIgnitionObject != null && Boolean.parseBoolean(oldIgnitionObject.toString())) {
```
tananaev: Ignition should be boolean value. If it's not, we can just ignore it.

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -70,6 +70,23 @@ private NotificationFormatter() {
             + "Point: http://www.openstreetmap.org/?mlat=%3$f&mlon=%4$f#map=16/%3$f/%4$f%n"
             + "Time: %2$tc%n";
 
+    public static final String TITLE_TEMPLATE_TYPE_ALARM = "%1$s: alarm!";
+    public static final String MESSAGE_TEMPLATE_TYPE_ALARM = "Device: %1$s%n"
+            + "Alarm: %5$s%n"
+            + "Point: http://www.openstreetmap.org/?mlat=%3$f&mlon=%4$f#map=16/%3$f/%4$f%n"
```
tananaev: Maybe we should have a special page in our system to show map instead of using OSM link? It definitely shouldn't be part of this pull request. Just an idea for future improvement.

src/org/traccar/reports/model/SummaryReport.java
```diff
@@ -47,4 +47,15 @@ public void setMaxSpeed(double maxSpeed) {
             this.maxSpeed = maxSpeed;
         }
     }
+
+    private long motorHours;
```
tananaev: Better to rename from `motor` to `engine`.

web/app/AttributeFormatter.js
```diff
@@ -34,6 +34,19 @@ Ext.define('Traccar.AttributeFormatter', {
         return Ext.getStore('DistanceUnits').formatValue(value, Traccar.app.getPreference('distanceUnit'));
     },
 
+    intervalFormatter: function (value) {
```
tananaev: There must be an easier way. Possibly ExtJS has some formatting functions.

web/l10n/en.json
```diff
@@ -26,6 +26,10 @@
     "sharedMute": "Mute",
     "sharedType": "Type",
     "sharedDistance": "Distance",
+    "sharedDayAbbreviation": "d",
```
tananaev: ExtJS probably has translation for it already. If we use their formatted, we won't need it here anymore.

tananaev:
Added some comments. By the way, if you want, you can add your name to copyright notices. Also, we can add you to readme file as a contributor or something because you've added a lot of functionality.

src/org/traccar/events/IgnitionEventHandler.java
Abyss777: I'm just afraid of ClassCastException and NullPointerException
````
if (oldIgnitionObject != null && oldIgnitionObject instanceof Boolean) {
    oldIgnition = (Boolean) oldIgnitionObject;
}
````
will be better?

File: web/app/AttributeFormatter.js (hunk 4 above)
Comment (Abyss777) [inline web/app/AttributeFormatter.js]: I could not find anything http://docs.sencha.com/extjs/6.0.1-classic/Ext.Date.html
Implementing own `formatFunction` won't be easier.
Microsoft suggesting the same algorithm  https://msdn.microsoft.com/en-us/library/ee532932(v=vs.94).aspx

File: src/org/traccar/events/IgnitionEventHandler.java (hunk 1 above)
Comment (tananaev) [inline src/org/traccar/events/IgnitionEventHandler.java]: Yes, it's better.

File: web/app/AttributeFormatter.js (hunk 4 above)
Comment (tananaev) [inline web/app/AttributeFormatter.js]: OK, I guess we have to keep this logic here. Do we need days?

Comment (Abyss777) [inline web/app/AttributeFormatter.js]: I was thinking...
It is engine hours, why not it be formatted like just hours, floating-point hours. I do not think some body wants to know how much seconds his engine worked.
It will look like:
`5.04 h` or `189.26 h`

What do you think?

Comment (tananaev) [inline web/app/AttributeFormatter.js]: I think we don't even need decimal part. Just hours should be enough.

File: src/org/traccar/events/IgnitionEventHandler.java [hunk 6]
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2016 Anton Tananaev (anton.tananaev@gmail.com), Andrey Kunitsyn (abyss@fox5.ru)
```
Abyss777: Does it looks OK?

tananaev: Just put it on the next line like this:
```
 * Copyright 2016 Anton Tananaev (anton.tananaev@gmail.com)
 * Copyright 2016 Andrey Kunitsyn (abyss@fox5.ru)
```

src/org/traccar/notification/NotificationFormatter.java
Abyss777: I like it. May be it can be special API request that load report with one position or something like that.

tananaev: Yes, just a separate HTML page or something.

Abyss777:
Seems done now.
Added myself to copyright, thank you for acknowledgment.
I'm not a programmer, but may be some day it will be advantage for hiring :)

tananaev:
Merged, thanks.

By the way, do we have fuel info in summary report? If not, can we add it?

Abyss777:
No, we don't. Sure, but we need discus it.

PR #2211, opened 2016-08-12
Changed way of summary distance calculation

Abyss777:
It is small this time

- `DistanceHandler` is always enabled now
- Summary distance is calculated by "odometer" if exist or by "distance" if not

### DIFF
--- a/src/org/traccar/BasePipelineFactory.java
+++ b/src/org/traccar/BasePipelineFactory.java
@@ -179,9 +179,6 @@ public abstract class BasePipelineFactory implements ChannelPipelineFactory {
         if (hemisphereHandler != null) {
             pipeline.addLast("hemisphere", hemisphereHandler);
         }
-        if (distanceHandler != null) {
-            pipeline.addLast("distance", distanceHandler);
-        }
         if (reverseGeocoderHandler != null) {
             pipeline.addLast("geocoder", reverseGeocoderHandler);
         }
@@ -198,6 +195,10 @@ public abstract class BasePipelineFactory implements ChannelPipelineFactory {
 
         if (coordinatesHandler != null) {
             pipeline.addLast("coordinatesHandler", coordinatesHandler);
+        }
+
+        if (distanceHandler != null) {
+            pipeline.addLast("distance", distanceHandler);
         }
 
         if (Context.getDataManager() != null) {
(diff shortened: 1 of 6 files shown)


Abyss777:
Oh, I've forgotten GranitProtocolDecoder sets `Position.KEY_DISTANCE`. What we should do? Remove it from ProtocolDecoder or store in other attribute?

tananaev:
What's distance in Granit protocol? Is it odometer value?

Abyss777:
No, it is distance between current and previous position.

tananaev:
I guess we can ignore it for now. I understand now why you wanted to make distance value to be only distance from last position. Maybe we can save it under some other key, or maybe store total distance under some other key.

Abyss777:
I like an idea to store total distance in other key.
Like `Position.KEY_TOTAL_DISTANCE = "totalDistance"`

`DistanceHandler` will calculate distance between current and last position and store it to `KEY_DISTANCE`  if it is not there already. I mean it won't replace device calculated distance.

Also `DistanceHandler` will sum `KEY_TOTAL_DISTANCE` and `KEY_DISTANCE`

`KEY_DISTANCE` is not very useful in this case but it will be uniform. May be there will be some other protocols that calculates distance by themselves. 

Is this logic fits?

tananaev:
Yes, I think it's a good idea.

Abyss777:
Hm, I've noticed another problem.  Some protocols set 0,0 coordinates to not valid positions (gps not fixed and others)

`KEY_TOTAL_DISTANCE` will rise dramatically for such positions. I suggest calculate distance only between valid positions for not valid it should be zero. 
It will cause loosing some small distances in case: valid-notValid-notValid-valid
but will exclude huge jumps of total distance.

tananaev:
I guess we can put filtering handler before distance calculation to solve this problem.

Abyss777:
Also one reason to not allow user reset total distance, it will cause mistakes in reports.

tananaev:
Is this pull request ready?

Abyss777:
It is look like, but I want test tomorrow on real trips.

tananaev:
OK, let me know how it goes.

Abyss777:
I've noticed the comment https://github.com/tananaev/traccar/blob/master/src/org/traccar/model/Position.java#L22
`"totalDistance"` seems wrong then. Should I use `"total-distance"` ?

tananaev:
Good question. Maybe it's better to change everything to lower-camel-case.

Abyss777:
May be, but it should be done in other pull request. There are a lot of changes in protocols and ObdDecoder and may be in other places.

Abyss777:
Seems done now.
Tested on real device total distance and emulate odometer for test device. Works OK.

tananaev:

PR #2215, opened 2016-08-15
Made speed threshold configurable

### LINKED ISSUE #2214: Speed threshold
@Abyss777, I think it's a good idea to make speed threshold in `MotionEventHandler` configurable. It should be really easy to do.

FIY:

https://www.traccar.org/forums/topic/move-and-stop-device-timeout/

Abyss777:
fix #2214 
- Made speed threshold configurable
- Added `Config.getDouble()`

### DIFF
--- a/src/org/traccar/Config.java
+++ b/src/org/traccar/Config.java
@@ -74,4 +74,16 @@ public class Config {
         }
     }
 
+    public double getDouble(String key) {
+        return getDouble(key, 0.0);
+    }
+
+    public double getDouble(String key, double defaultValue) {
+        if (properties.containsKey(key)) {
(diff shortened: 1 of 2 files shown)


nefertiti007:
Hello,

I have noticed a small situation with many GPS Devices, Please advice if that is covered here,.

The last position sent by the device would be speed 1 or 2 or 7 etc (a specific value which is not 0 or 0.01 etc.

In such case,. the device is stationary but the last position received has a speed value. Here would the movement stop be triggered, if for a certain time there is no movement (co-ordinate change) as I can see its based on speed value.

Abyss777:
@nefertiti007 
This changes can help, but like workaround.

tananaev:
Merged, thanks.

PR #2216, opened 2016-08-15
Convert speed units in mail notifications

Abyss777:
- Added speed converting in NotificationFormatter
- Used constants in UnitsConverter

Partially fix https://www.traccar.org/forums/topic/setting-the-speed-limit-and-speed-limit-notifications/


Abyss777:
Any comments for this pull request?

tananaev:
Weird, I'm pretty sure I've added a comment. Maybe there was some network issue and it wasn't published. Anyway, the only comment I has is that querying the users table every time might not be a very good idea. Possibly we need to cache it same as with devices.

Abyss777:
Totally agree.
I suggest cache it in `PermissionManager` because it is already cached there but do not have public interface.

Abyss777:
Sorry, I occasionally delete branch, I will reopen it after implementing cache.

Abyss777:
Oh, again I've broken git. I'll made a new pull request.

PR #2219, opened 2016-08-16
Alarm decoding for next protocols

Abyss777:
Added alarm decoding for 5 protocols that have documentation.

Not sure about Megastek, but according my intuition should work for all patterns.
I think we can change it in future if something wrong according to users feedback.


Abyss777:
Also added alarm decoder for GT06

tananaev:
Merged, thanks for pull request. I'll try to find some missing protocols.

PR #2222, opened 2016-08-17
Alarm decoding for few more protocols

Abyss777:
Fixed alarm decoding for more protocols 

There is no list of possible alarms in Xexun, in the tests only "help me!", I've added some from SMS answers that can be.


tananaev:
Merged, thanks.

PR #2223, opened 2016-08-17
Some fuel attributes organization

Abyss777:
- Added `KEY_FUEL_CONSUMPTION` and `KEY_FUEL_INSTANT_CONSUMPTION`
- Fixed wrong using of `KEY_FUEL` 

About Mta6: I'm not sure what means "Fuel consumption accumulator", but pretty sure it is not `KEY_FUEL`

### DIFF
--- a/src/org/traccar/model/Position.java
+++ b/src/org/traccar/model/Position.java
@@ -39,6 +39,8 @@ public class Position extends Message {
     public static final String KEY_LAC = "lac";
     public static final String KEY_CID = "cid";
     public static final String KEY_FUEL = "fuel";
+    public static final String KEY_FUEL_CONSUMPTION = "fuelConsumption";
+    public static final String KEY_FUEL_INSTANT_CONSUMPTION = "fuelInstantConsumption";
     public static final String KEY_RFID = "rfid";
     public static final String KEY_VERSION = "version";
     public static final String KEY_TYPE = "type";
--- a/src/org/traccar/model/Position.java
+++ b/src/org/traccar/model/Position.java
@@ -40,7 +40,6 @@ public class Position extends Message {
     public static final String KEY_CID = "cid";
     public static final String KEY_FUEL = "fuel";
     public static final String KEY_FUEL_CONSUMPTION = "fuelConsumption";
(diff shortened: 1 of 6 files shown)


src/org/traccar/model/Position.java
```diff
@@ -39,6 +39,8 @@
     public static final String KEY_LAC = "lac";
     public static final String KEY_CID = "cid";
     public static final String KEY_FUEL = "fuel";
+    public static final String KEY_FUEL_CONSUMPTION = "fuelConsumption";
+    public static final String KEY_FUEL_INSTANT_CONSUMPTION = "fuelInstantConsumption";
```
tananaev: What's instant consumption? Do we really need it?

Abyss777: It is consumption at the time calculated by ECU and accessible via OBD. Frankly, it is useless parameter even for engine diagnostic.

But why not to store it if it is exist. There are a lot of useless attributes we store in different protocols. Or we are trying to avoid it?

tananaev: Most stored attributes are there because someone asked to add it. I think we can remove this one because the one protocol that provides it, also provides regular consumption as well.

Abyss777:
Done.

src/org/traccar/protocol/Gps103ProtocolDecoder.java
```diff
@@ -189,7 +189,8 @@ protected Object decode(
             getLastLocation(position, dateBuilder.getDate());
 
             position.set(Position.KEY_ODOMETER, parser.nextInt());
-            position.set(Position.KEY_FUEL, parser.next());
+            parser.next();                                                          // instant fuel consumption
```
tananaev: Why there are so many spaces before comment?

Abyss777: Just to not to be in the middle of the code (up and down). How many would be OK?

tananaev: Just one space is the standard. Only exception is regular expressions because there is a comment for each line.

Abyss777:
Sorry, I'll remember.

tananaev:
Thanks.

Issue #2224, opened 2016-08-18
Trips reports


Abyss777:
I've almost finished trip reports.
Key features of trips detector:

1. configurable parameters:
````
double speedThreshold = Context.getConfig().getDouble("event.motion.speedThreshold", 0.01);
long minimalTripDuration = Context.getConfig().getLong("report.trip.minimalTripDuration", 300) * 1000;
double minimalTripDistance = Context.getConfig().getLong("report.trip.minimalTripDistance", 500);
long minimalParkingDuration = Context.getConfig().getLong("report.trip.minimalParkingDuration", 300) * 1000;
````
2. Trips less than minimal duration and minimal distance are parking time
3. Parking periods less than minimal duration do not  cut trips
4. Trip cut by end of period does not go in report

It looks like this:
![image](https://cloud.githubusercontent.com/assets/5688080/17772974/39195672-6565-11e6-8875-e8ca56f4518b.png)

Trips detector algorithm is not obvious. If there is something to change or add it is better to do it while I remember how it is work.

PR #2228, opened 2016-08-19
Trips reports

Abyss777:
- Added trips report, key features of trips detector are here #2224
- Refactored reports models and moved some code to `ReportUtils`
- Added tests for `ReportUtils`

I've added "Spent fuel" only to trips (not summary) because it is very unlikely that fuel increases during a trip. It is calculated only for 3 protocols for now. It is String and formatted in java, because web-client never knows which format is used for `KEY_FUEL`.

If you don't mind I've added `speed` to `test-generator.py`. It helps to test motion, overspeed, and trips detector.
Also I can add comments to `detectTrips` for how it is work explanation if need.

### DIFF
--- a/src/org/traccar/reports/ReportUtils.java
+++ b/src/org/traccar/reports/ReportUtils.java
@@ -49,7 +49,7 @@ public final class ReportUtils {
                 && lastPosition.getAttributes().containsKey(Position.KEY_ODOMETER)) {
             distance = new BigDecimal(lastPosition.getAttributes().get(Position.KEY_ODOMETER).toString())
                     .subtract(new BigDecimal(firstPosition.getAttributes().get(Position.KEY_ODOMETER).toString()))
-                    .multiply(new BigDecimal("1000"));
+                    .multiply(new BigDecimal(1000));
         } else if (firstPosition.getAttributes().containsKey(Position.KEY_TOTAL_DISTANCE)
                 && lastPosition.getAttributes().containsKey(Position.KEY_TOTAL_DISTANCE)) {
             distance = new BigDecimal(lastPosition.getAttributes().get(Position.KEY_TOTAL_DISTANCE).toString())
@@ -59,7 +59,7 @@ public final class ReportUtils {
         return distance.setScale(2, RoundingMode.HALF_EVEN).doubleValue();
     }
 
-    public static String calculateSpentFuel(Position firstPosition, Position lastPosition) {
+    public static String calculateFuel(Position firstPosition, Position lastPosition) {
 
         if (firstPosition.getAttributes().get(Position.KEY_FUEL) != null
                 && lastPosition.getAttributes().get(Position.KEY_FUEL) != null) {
--- a/src/org/traccar/reports/ReportUtils.java
+++ b/src/org/traccar/reports/ReportUtils.java
@@ -44,19 +44,18 @@ public final class ReportUtils {
     }
 
     public static double calculateDistance(Position firstPosition, Position lastPosition, boolean useOdometer) {
-        BigDecimal distance = new BigDecimal("0.0");
+        double distance = 0.0;
         if (useOdometer && firstPosition.getAttributes().containsKey(Position.KEY_ODOMETER)
                 && lastPosition.getAttributes().containsKey(Position.KEY_ODOMETER)) {
-            distance = new BigDecimal(lastPosition.getAttributes().get(Position.KEY_ODOMETER).toString())
-                    .subtract(new BigDecimal(firstPosition.getAttributes().get(Position.KEY_ODOMETER).toString()))
-                    .multiply(new BigDecimal(1000));
+            distance = (((Number) lastPosition.getAttributes().get(Position.KEY_ODOMETER)).doubleValue()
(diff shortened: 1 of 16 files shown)


src/org/traccar/reports/ReportUtils.java
```diff
@@ -35,4 +39,46 @@ private ReportUtils() {
         return result;
     }
 
+    public static double calculateDistance(Position firstPosition, Position lastPosition) {
+        return calculateDistance(firstPosition, lastPosition, true);
+    }
+
+    public static double calculateDistance(Position firstPosition, Position lastPosition, boolean useOdometer) {
+        BigDecimal distance = new BigDecimal("0.0");
```
tananaev: BigDecimal can take double value. String looks weird. Same for "1000" below. It can just be int.

src/org/traccar/reports/ReportUtils.java
```diff
@@ -35,4 +39,46 @@ private ReportUtils() {
...
+        } else if (firstPosition.getAttributes().containsKey(Position.KEY_TOTAL_DISTANCE)
+                && lastPosition.getAttributes().containsKey(Position.KEY_TOTAL_DISTANCE)) {
+            distance = new BigDecimal(lastPosition.getAttributes().get(Position.KEY_TOTAL_DISTANCE).toString())
+                    .subtract(new BigDecimal(firstPosition.getAttributes().get(Position.KEY_TOTAL_DISTANCE)
+                            .toString()));
+        }
+        return distance.setScale(2, RoundingMode.HALF_EVEN).doubleValue();
+    }
+
+    public static String calculateSpentFuel(Position firstPosition, Position lastPosition) {
```
tananaev: I think it can be just `calculateFuel`. More consistent with other methods.

src/org/traccar/reports/ReportUtils.java
tananaev: By the way, do we really need BigDecimal?

src/org/traccar/reports/model/BaseReport.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private double averageSpeed; // knots
+    public double getAverageSpeed() {
+        return averageSpeed;
+    }
+    public void setAverageSpeed(Double averageSpeed) {
+        this.averageSpeed = new BigDecimal(averageSpeed.toString()).setScale(3, RoundingMode.HALF_EVEN).doubleValue();
+    }
+
+    private double maxSpeed; //knots
```
tananaev: add space after //

tananaev:
I haven't had much time to look at the actual logic, but at first glance it looks OK. Have you tested it with real trips?

src/org/traccar/reports/ReportUtils.java
Abyss777: I get PMD complain "Avoid creating BigDecimal with a decimal (float/double) literal. Use a String literal". Int is OK.

As I understand using `BigDecimal.setScale` is best (precise) way to round floating point values.
And really sometimes I got *999999999 values subtracting two rounded distances.
I used double, and rounded it at the end of function, but got PMD warning, I converted double to string and ... saw that there is casting from odometer to double and from distance to double. I Thought why not to cast directly to BigDecimal and do math with it.

Abyss777:
Sure, in #2224 screenshot if my real trips and it correspond to facts 100%.
My device is rather precise and sleep on parking. On some other devices that always send positions may happen "false trips" but I think it can be handled with configuration and filters.

src/org/traccar/reports/ReportUtils.java
tananaev: We should not round anything on the server side. All the formatting should be on the client, so I guess we can use double here.

Abyss777:
OK. Used double.

I just was repairing my car, and tracker sent positions for whole period.
As I guessed logic produces some false trips. For such devices need additional filtering, but I think `CoordinatesHandler` will be enough in most cases

src/org/traccar/reports/ReportUtils.java
```diff
@@ -35,4 +39,45 @@ private ReportUtils() {
...
+        return calculateDistance(firstPosition, lastPosition, true);
+    }
+
+    public static double calculateDistance(Position firstPosition, Position lastPosition, boolean useOdometer) {
+        double distance = 0.0;
+        if (useOdometer && firstPosition.getAttributes().containsKey(Position.KEY_ODOMETER)
+                && lastPosition.getAttributes().containsKey(Position.KEY_ODOMETER)) {
+            distance = (((Number) lastPosition.getAttributes().get(Position.KEY_ODOMETER)).doubleValue()
+                    - ((Number) firstPosition.getAttributes().get(Position.KEY_ODOMETER)).doubleValue())
+                    * 1000;
```
tananaev: I think we should store odometer value in meters as well.

tananaev:
Merged, thanks.

src/org/traccar/reports/ReportUtils.java
Abyss777: Agree, It will be more uniform.

I was wondering and actually I'm not sure that odometer is always in kilometers. What do you think can it be in miles?

tananaev: There is no standard at the moment. Some values can be in miles, some and km and some in meters already.

PR #2231, opened 2016-08-19
Commands to Wialon protocol

Abyss777:
My device send answers for commands, for example USSD responce, in UTF8 charset.
I made it configurable, because I'm not sure that every device used UTF8. My be It should be UTF8 by default.

Also I added more optional fields to CommandDialog


tananaev:
Looks good. Thanks.

PR #2240, opened 2016-08-23
Odometer units for first part of protocols

Abyss777:
- Made changes to use meters in `KEY_ODOMETER`
- Other changes to use right keys in some protocols


tananaev:
Excellent, thanks for pull request.

Abyss777:
Oh, I've found the problem:
I've replaced in few places `parser.next()` to `parser.nextDouble()` or `parser.nextInt()` and they both return  0 or 0.0 even if match is optional. For example my wondex device do not send odometer, but it always 0.0 because of that calculated distance is always "0.0"

Add check `if (parser.hasNext())` for such optional cases will be correct?

tananaev:
Yes, it should be fine if you call `hasNext`.

PR #2242, opened 2016-08-23
Users cache

Abyss777:
- Implemented users cache

Also while testing I've noticed that some buttons and menu items is not hidden from readonly user, fixed it.

The only button for devices is "Send command". It is not forbidden in API. 
Is it OK that readonly user can send commands?


tananaev:
Looks good, thanks.

PR #2245, opened 2016-08-23
Added missing ports declaration.

drecchia:
Debug.xml has more ports than others XML files.


tananaev:
Thanks. I usually update production configs just before an official release.

PR #2247, opened 2016-08-24
Fixed optional odometer for wondex protocol and exclusion for 0.0 odometers

Abyss777:
Only wondex has optional odometer.

Anyway I've added check if both odometer values are zero they are not used to calculate distance.
There are a lot of "0.0" odometers in unit tests in different protocols. I think it is not rare case that it is zero.


tananaev:
Merged, thanks.

PR #2248, opened 2016-08-24
Convert speed units in mail notifications v2

Abyss777:
- Added speed converting in NotificationFormatter
- Used constants in UnitsConverter

Partially fix https://www.traccar.org/forums/topic/setting-the-speed-limit-and-speed-limit-notifications/

Sorry for duplicate.


tananaev:
Merged, thanks.

Issue #2251, opened 2016-08-25
Trip detector algorithm


Abyss777:
I tested a bit on other device and reach a deadlock, can't decide.

Here is exaggerated example of speed graph:
![image](https://cloud.githubusercontent.com/assets/5688080/17957330/47fb9d54-6aaa-11e6-8531-7e26b903134d.png)

What is this?
1. Parking period with two too small trips that should be filtered (4 minutes duration)
2. 3 minutes trip with small stop in the middle

1. Can cause cut real trip at the beginning and at the end
2. Can cause false trips.

My first version was 2 and now it is 1.  But I want come back to 2 again.

@tananaev what do you think?

Abyss777:
Or just made it configurable? "Greedy parking" vs. "greedy trips"

Abyss777:
One more example
![image](https://cloud.githubusercontent.com/assets/5688080/17960117/9094a5a2-6abd-11e6-8e19-961aa30a6405.png)
**Note:** C-D is more than minimal parking duration
A-B and E-F less than minimal trip duration

1. Variant says it is two trips **B-C** and **D-E**
2. Variant says it is one trip **A-F**

tananaev:
I think it's a good idea to make it configurable.

PR #2252, opened 2016-08-25
Made trips detector behavior configurable

Abyss777:
More explanation in #2251 

`report.trip.greedyParking = true` means algorithm tends to be in "parking state". Only continuous trips longer than minimal detected.
Default `report.trip.greedyParking = false` means algorithm tends to be in "trip state". Only continuous parking longer than minimal can divide trip.


tananaev:

PR #2254, opened 2016-08-25
Colour routes of different devices

Abyss777:
- Multiselect is deprecated for `combobox`, I used `tagfield`
- I decided to not use random for pick a color. Just used remainder on dividing `deviceId`. Color will stay the same for each device.
- Randomly generated set of colors (can be easily changed or added)
- `mapView.fit` did for first route only. Frankly, have no idea how to fit for all without edges calculating. 

Here is screenshot:

image

### DIFF
--- a/web/app/Style.js
+++ b/web/app/Style.js
@@ -38,7 +38,14 @@ Ext.define('Traccar.Style', {
     mapDefaultLon: -0.1275,
     mapDefaultZoom: 6,
 
-    mapRouteColor: 'rgba(21, 127, 204, 1.0)',
+    mapRouteColor: [
+        'rgba(21, 127, 204, 1.0)',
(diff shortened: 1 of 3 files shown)


web/app/Style.js
```diff
@@ -38,7 +38,14 @@ Ext.define('Traccar.Style', {
     mapDefaultLon: -0.1275,
     mapDefaultZoom: 6,
 
-    mapRouteColor: 'rgba(21, 127, 204, 1.0)',
+    mapRouteColor: [
+        'rgba(21, 127, 204, 1.0)',
```
tananaev: I think we shouldn't duplicate this array twice. Just use same for both.

Abyss777:
OK. Removed.

PR #2257, opened 2016-08-26
Fix telic protocol speed units

### LINKED ISSUE #2256: Incorrect speed with Telic devices
I noticed that speed is incorrect with Telic devices.

According to Telic protocol documentation the tracking device transmits speed in km/h, but protocol decoder doesn't convert it into knots. Therefore speed is stored incorrectly into the database.

Abyss777:
Fix #2256 
Here is also km/h https://github.com/tananaev/traccar/issues/691


tananaev:
Thanks.

PR #2258, opened 2016-08-26
Fix wialon protocol speed units

Abyss777:
Why 218 km/h top speed did not confuse me earlier...
Same problem, it is km/h according to documentation.


tananaev:
Thanks for pull request.

PR #2262, opened 2016-08-27
Prevent Device tagfield growing too much

Abyss777:
By default it grows indefinitely to right.
And it will grow down if limited only width.


tananaev:
Merged.

PR #2266, opened 2016-08-29
Actualize swagger.json with reports

Abyss777:
Added only json because of https://github.com/OAI/OpenAPI-Specification/issues/146


tananaev:
Merged, thanks.

PR #2269, opened 2016-08-29
Implemented "Configure" dialog for reports

Abyss777:
- Used intermediate store to list event types, because `tagfield` can't render names separately. 
- Added "All Events" just to optimize requests to DB.
- Store field values in parent panel object.
- Limited `tagfield` width, but not limited height. It grows down indefinitely.
- "Event Types" field showed only if "Events" report type are selected.

Empty dialog:
image

Fulfilled dialog:
image


tananaev:
I think we need to disable `Show` and `CSV` buttons if parameters are not configured for the report. Or maybe instead of disabling, if it's not configured and the button is clicked, it should show the configuration dialog.

web/app/view/ReportController.js
```diff
@@ -60,6 +60,15 @@ Ext.define('Traccar.view.ReportController', {
         dialog.show();
     },
 
+    updateButtons: function () {
+        var reportType, disabled, devices;
+        reportType = this.lookupReference('reportTypeField').getValue();
+        devices = this.deviceId && this.deviceId.length !== 0 || this.groupId && this.groupId.length !== 0;
+        disabled = !reportType || !devices;
```
tananaev: Don't we need time fields as well?

Abyss777:
I chose disable buttons. It looks more naturally.

web/app/view/ReportController.js
Abyss777: It configured by default from 30 minutes before to now. And can't be not configured.

Abyss777: Oh. Sorry. My mistake. It can be.

Abyss777:
Done. Added time check

web/app/store/ReportEventTypes.js
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+Ext.define('Traccar.store.ReportEventTypes', {
+    extend: 'Ext.data.Store',
+    fields: ['type', 'name'],
+
+    data: [{
+        type: '%',
```
tananaev: What does '%' mean? I can see it everywhere in the code. Can you please explain how it all works.

tananaev: Is it just a special string you picked to represent all events? Or does it have any meaning in extjs framework?

Abyss777: I used `LIKE` in sql request https://github.com/tananaev/traccar/blob/master/debug.xml#L244
`%` is wildcard for `LIKE`

tananaev: I think it's better to a more meaningful string. For example, 'allEvents'.

Abyss777: And convert it on server side?

tananaev: I see now. You use it directly. I guess then we should use it, but maybe we can have an alias/variable for it.

Abyss777: Alias will be good. Where is better place? `Traccar.Application` or `Traccar.store.ReportEventTypes` ? Or somewhere else?

tananaev: Good question. I guess in `ReportEventTypes` would make more sense.

Abyss777:
Something like this? Unfortunately it is impossible use `allEvents` at same store config.

tananaev:
No, not like that. You should use a static variable for it and not actual store instance.

Abyss777:
Sorry. I'm still not good in JS and extjs.
Also moved "All Events" creation to `ReportConfigController.init()`. It is just to use the same static variable.

web/app/view/ReportConfigController.js
```diff
@@ -0,0 +1,64 @@
...
+
+Ext.define('Traccar.view.ReportConfigController', {
+    extend: 'Ext.app.ViewController',
+    alias: 'controller.reportConfigDialog',
+
+    init: function () {
+        var store = this.lookupReference('eventTypeField').getStore();
+        if (store.getCount() === 0) {
+            store.add({
+                type: Traccar.store.ReportEventTypes.allEvents,
```
tananaev: Make sure that you always import all the types that you use. It might not give any errors at this point because it's probably imported somewhere already, but it might fail at some point in future or it might result in some issues after minification. Applies to everything, not just this line.

tananaev:
Just one small comment. Otherwise looks good.

web/app/view/ReportConfigController.js
Abyss777: All stores and models are imported in `Application.js`, I added `ReportEventTypes` there, I do not think we need to "requires" it again here.
Thanks, you are right, I missed `CustomTimeField`.
IDEs is relaxing in some aspects. I'll try to be more careful and do not do such mistakes.

tananaev: I know that it's going to work fine at this point, but you shouldn't assume that it would be the case all the time. You are basically making everything more tightly coupled than it should be. All dependencies should be explicitly added.

Abyss777: OK. But still I do not understand why we, for example, not import `Traccar.store.Devices` in `DevicesController.js` and should do it here.

tananaev: Because we never use `Traccar.store.Devices` there.

Abyss777: OK, bad example, but why we not requires `Traccar.Style` and `Traccar.AttributeFormatter` ?

tananaev: I think style is also a bad example. Technically you are right and it should be imported, but because it's used so much, I decided to make an exception and basically treat it as a global required for every class.

Abyss777: OK. I'll try to sum up:
1. We have to import stores and models in `Application.js` and should not import it again if access via `Ext.getStore()` or objects config (like grid or combobox).
We get error or warning if missed something.
2. We have to import all things that access directly except `Traccar.Style` and `Traccar.AttributeFormatter` because we usually just get 'undefined' with no error.
3. Chained import view is enough. I mean import view that imports controller.

Am I right?

One question:
If we create model or store via full name and `Ext.create()` should we import it again if it is already imported in `Application.js` ?

tananaev: I think the rule should be that if you use some class name, you should import it. When you use `Ext.getStore`, you don't use class name, you just use store id. For example, if you use store class name explicitly anywhere, you should import it. `Traccar.Style` is the only exception because it's used everywhere. `Traccar.AttributeFormatter` is not an exception. If there is no import for it, I guess we should add it. Chained import is not enough. I think it's better to always import any classes that are used explicitly.

Abyss777:
Looks added all missed now.

Abyss777:
History looks messy, I can squash it in new pull request.

tananaev:
I think it's fine. Merged. Thanks for pull request and fixing all the imports.

PR #2273, opened 2016-08-31
Fix geofences colors and separate geofences styles

Abyss777:
Sorry, I did not think that `mapColorReport` used to draw geofences. (it has "report" in name)

The quickfix was just replace `mapColorReport` to `mapRouteColor[0]` but it looks not very good and I decided to separate styles used for geofences.
It will allow some fine tuning and avoid such confuses in future.


tananaev:
Merged, thanks.

PR #2276, opened 2016-09-01
Rename phoneNumber constant and field in web

Abyss777:
In addition to https://github.com/tananaev/traccar/commit/ae5fe50416df190f5374d084526518a22096871e

### DIFF
--- a/web/l10n/en.json
+++ b/web/l10n/en.json
@@ -118,7 +118,7 @@
     "commandDeviceIdentification": "Device Identification",
     "commandIndex": "Index",
     "commandData": "Data",
-    "commandPhoneNumber": "Phone Number",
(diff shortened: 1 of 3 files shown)


web/l10n/en.json
```diff
@@ -118,7 +118,7 @@
     "commandDeviceIdentification": "Device Identification",
     "commandIndex": "Index",
     "commandData": "Data",
-    "commandPhoneNumber": "Phone Number",
+    "commandPhone": "Phone",
```
tananaev: I think it might be a good idea to keep text as "Phone Number". It would be clearer.

Abyss777:
I agree.

tananaev:
Merged.

PR #2279, opened 2016-09-01
Fix find index in BasePermissionsController.js

### LINKED ISSUE #2278: User Devices
The user devices display wrong devices
image

in BasePermissionsController.js line 39 the store find return the wrong index:
image

Also, a primary index is missing on "user_device" table (on fields: userid, deviceid)

Abyss777:
`find` by default uses RegExp
Should fix  #2278

### DIFF
--- a/web/app/view/BasePermissionsController.js
+++ b/web/app/view/BasePermissionsController.js
@@ -36,7 +36,7 @@ Ext.define('Traccar.view.BasePermissionsController', {
                         var i, index;
                         if (success) {
                             for (i = 0; i < records.length; i++) {
-                                index = this.getView().getStore().find('id', records[i].getData().id);
+                                index = this.getView().getStore().findExact('id', records[i].getData().id);
                                 this.getView().getSelectionModel().select(index, true, true);
                             }
                         }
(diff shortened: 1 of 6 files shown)


web/app/view/BasePermissionsController.js
```diff
@@ -36,7 +36,7 @@ Ext.define('Traccar.view.BasePermissionsController', {
                         var i, index;
                         if (success) {
                             for (i = 0; i < records.length; i++) {
-                                index = this.getView().getStore().find('id', records[i].getData().id);
+                                index = this.getView().getStore().findExact('id', records[i].getData().id);
```
tananaev: Why are we using find here instead of `getById`?

Abyss777: Frankly, do not now. I used old code:
https://github.com/tananaev/traccar/blob/v3.5/web/app/view/UserDevicesController.js#L37

But sure `getById` will be better

Abyss777:
Used `getById()`
Also replaced all `getData().*` to `getId()` and `get()`
Also used `findExact('type', ` in `NotificationsController.js`

tananaev:
Thanks for fixing the issue and cleaning up the code.

PR #2283, opened 2016-09-03
Adding location parameter to OsmAnd protocol

dough29:
As exposed on forum thread - https://www.traccar.org/forums/topic/tracking-through-http-url-request/ - add location parameter to OsmAnd protocol for coma separated lat & lon (“&location=60.0,30.0”).


src/org/traccar/protocol/OsmAndProtocolDecoder.java
```diff
@@ -104,6 +105,11 @@ protected Object decode(
                 case "lon":
                     position.setLongitude(Double.parseDouble(value));
                     break;
+                case "location":
+                    List<String> location = Arrays.asList(value.split(","));
```
tananaev: There is no need to convert it to list. Just use array.

tananaev:

PR #2287, opened 2016-09-05
Server/User dialog - retrieving map center from current map

Abyss777:
User and server dialog can retrieve map center parameters from current map
Grouped map center parameters, it is looks like this:
 
image

Used "crossview" events to request and receive map center
Had to fix button margin, because form do not set it by itself.

Have two doubts:
1.  First 5 `textfield`s end left a bit than `fieldset`. It is normal behavior, but can looks not perfect. (look at screenshot). Or may be we should group other parameters too.
2.  `UserDialogController` extends `ServerDialogController`, "server" may be should have some "base" name, but I do not know how to name it. "BaseEditWithMapCenterDialogController" looks long.

### DIFF
--- a/web/app/view/ServerDialog.js
+++ b/web/app/view/ServerDialog.js
@@ -66,9 +66,8 @@ Ext.define('Traccar.view.ServerDialog', {
             displayField: 'name',
             valueField: 'key'
         }, {
-            xtype: 'fieldset',
+            xtype: 'fieldcontainer',
             reference: 'mapCenter',
-            title: Strings.sharedMapCenter,
             defaultType: 'numberfield',
             items: [{
                 name: 'latitude',
@@ -81,12 +80,6 @@ Ext.define('Traccar.view.ServerDialog', {
             }, {
                 name: 'zoom',
                 fieldLabel: Strings.serverZoom
-            }, {
-                xtype: 'button',
-                margin: Traccar.Style.inFormButtonMargin,
-                enableToggle: false,
-                text: Strings.sharedGetFromMap,
-                handler: 'getFromMap'
             }]
         }, {
             xtype: 'checkboxfield',
@@ -94,5 +87,24 @@ Ext.define('Traccar.view.ServerDialog', {
             fieldLabel: Strings.settingsTwelveHourFormat,
             allowBlank: false
         }]
-    }
+    },
+
+    buttons: [{
+        text: Strings.sharedAttributes,
+        handler: 'showAttributesView'
+    }, {
+        glyph: 'xf276@FontAwesome',
+        minWidth: 0,
+        handler: 'getFromMap',
+        tooltip: Strings.sharedGetFromMap,
+        tooltipType: 'title'
+    }, {
+        xtype: 'tbfill'
+    }, {
+        text: Strings.sharedSave,
+        handler: 'onSaveClick'
+    }, {
+        text: Strings.sharedCancel,
+        handler: 'closeView'
+    }]
 });
--- a/web/app/view/ServerDialog.js
+++ b/web/app/view/ServerDialog.js
@@ -66,21 +66,22 @@ Ext.define('Traccar.view.ServerDialog', {
             displayField: 'name',
             valueField: 'key'
         }, {
-            xtype: 'fieldcontainer',
-            reference: 'mapCenter',
-            defaultType: 'numberfield',
-            items: [{
-                name: 'latitude',
-                fieldLabel: Strings.positionLatitude,
-                decimalPrecision: Traccar.Style.coordinatePrecision
-            }, {
-                name: 'longitude',
-                fieldLabel: Strings.positionLongitude,
-                decimalPrecision: Traccar.Style.coordinatePrecision
-            }, {
-                name: 'zoom',
-                fieldLabel: Strings.serverZoom
-            }]
+            xtype: 'numberfield',
+            reference: 'latitude',
+            name: 'latitude',
+            fieldLabel: Strings.positionLatitude,
+            decimalPrecision: Traccar.Style.coordinatePrecision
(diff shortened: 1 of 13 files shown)


tananaev:
I don't really like how it looks. Too complicated. Can we just leave it as it was and add a button (probably with just icon to save space) to the bottom bar to retrieve current map location?

Abyss777:
Moved button, now it looks like this:

image

Glyph is http://fontawesome.ru/icon/map-pin/

tananaev:
Why not this one: http://fontawesome.io/icon/map-marker/

I think it looks less confusing.

web/app/view/ServerDialog.js
```diff
@@ -66,24 +66,45 @@ Ext.define('Traccar.view.ServerDialog', {
...
-        }, {
-            xtype: 'numberfield',
-            name: 'longitude',
-            fieldLabel: Strings.positionLongitude,
-            decimalPrecision: Traccar.Style.coordinatePrecision
-        }, {
-            xtype: 'numberfield',
-            name: 'zoom',
-            fieldLabel: Strings.serverZoom
+            xtype: 'fieldcontainer',
```
tananaev: Why do we need a container here?

web/app/view/ServerDialogController.js
```diff
@@ -0,0 +1,42 @@
...
+            controller: {
+                '*': {
+                    setCenterFromMap: 'setCenterFromMap'
+                }
+            }
+        }
+    },
+
+    getFromMap: function (button) {
+        this.fireEvent('getMapCenter');
```
tananaev: I think instead of "center", we should use something like "state". Because zoom has nothing to do with center of the map.

web/app/view/UserDialog.js
```diff
@@ -70,24 +70,45 @@ Ext.define('Traccar.view.UserDialog', {
...
-        }, {
-            xtype: 'numberfield',
-            name: 'longitude',
-            fieldLabel: Strings.positionLongitude,
-            decimalPrecision: Traccar.Style.coordinatePrecision
-        }, {
-            xtype: 'numberfield',
-            name: 'zoom',
-            fieldLabel: Strings.serverZoom
+            xtype: 'fieldcontainer',
```
tananaev: Same here. I don't think we need a container.

web/app/view/ServerDialogController.js
```diff
@@ -0,0 +1,42 @@
...
+
+Ext.define('Traccar.view.ServerDialogController', {
+    extend: 'Traccar.view.BaseEditDialogController',
+    alias: 'controller.serverEditDialog',
+
+    config: {
+        listen: {
+            controller: {
+                '*': {
+                    setCenterFromMap: 'setCenterFromMap'
```
tananaev: I think this should be renamed as well. It's basically a response for callback, right? It shouldn't assume that it will always be used to "set" something. Maybe just "mapstate".

By the way, I looked at the extjs documentation and noticed that event names are all lowercase, so maybe we should follow same styles as well?

web/l10n/en.json
```diff
@@ -28,6 +28,7 @@
     "sharedDistance": "Distance",
     "sharedHourAbbreviation": "h",
     "sharedMinuteAbbreviation": "m",
+    "sharedGetFromMap": "Get from Map",
```
tananaev: Maybe change it to "sharedGetMapState": "Get Map State" to be more clear.

tananaev:
Added some more comments.

web/app/view/ServerDialog.js
Abyss777: Is it OK if I continue using `this.getView().down('numberfield[name="latitude"]')` to access values or it is better to add `reference` and use `lookupReference()` ?

tananaev: I think it's better to use reference for each field.

web/app/view/ServerDialogController.js
Abyss777: OK. here will be `mapstate` and `mapstaterequest` in `MapController`. (they should have different names)
Also I going to do additional commit to rename other events to lowercase.

tananaev: Yes, perfect, thanks.

Abyss777:
Looks like fixed all comments. Current look:

image

web/app/view/ServerDialogController.js
```diff
@@ -0,0 +1,41 @@
...
+            }
+        }
+    },
+
+    getMapState: function (button) {
+        this.fireEvent('mapstaterequest');
+    },
+
+    setMapState: function (lat, lon, zoom) {
+        this.getView().lookupReference('latitude').setValue(lat);
```
tananaev: You don't need to call `getView`, you can call `lookupReference` straight away from controller.

web/app/view/UserDialogController.js
```diff
@@ -15,7 +15,7 @@
  */
 
 Ext.define('Traccar.view.UserDialogController', {
-    extend: 'Traccar.view.BaseEditDialogController',
+    extend: 'Traccar.view.ServerDialogController',
```
tananaev: This is a bit confusing. Maybe it should be renamed from `ServerDialogController` to something like `MapPickerDialogController`.

tananaev:
A couple more small comments.

web/app/view/UserDialogController.js
Abyss777: Perfect, I just could not find right words.

web/app/view/ServerDialogController.js
Abyss777: Sorry, It is really unnecessary.

tananaev:
Merged, thanks.

PR #2307, opened 2016-09-12
Move column sets outside the method

### LINKED ISSUE #2301: In ReportController move columns outside of method
@Abyss777,

Not a critical issue, but I think it's better to move `routeColumns`, `eventsColumns`, `summaryColumns` and `tripsColumns` into the class instead of creating new arrays/objects every time `onTypeChange` method is called.

Abyss777:
fix #2301


tananaev:
Thanks.

PR #2317, opened 2016-09-14
Small csv export improvement

Abyss777:
- Make `attributes` always last column
- Format Map object as simple string instead of json

Here is an example
[report (4).zip](https://github.com/tananaev/traccar/files/471907/report.4.zip)

### DIFF
--- a/src/org/traccar/web/CsvBuilder.java
+++ b/src/org/traccar/web/CsvBuilder.java
@@ -80,7 +80,7 @@ public class CsvBuilder {
                         Map value = (Map) method.invoke(object);
                         if (value != null) {
                             String map = MiscFormatter.toJson(value).toString();
-                            map = map.replaceAll("[\\{|\\}|\"]", "");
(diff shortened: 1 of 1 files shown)


src/org/traccar/web/CsvBuilder.java
```diff
@@ -73,7 +79,10 @@ public void addLine(Object object) {
                     } else if (method.getReturnType().equals(Map.class)) {
                         Map value = (Map) method.invoke(object);
                         if (value != null) {
-                            builder.append(MiscFormatter.toJson(value).toString());
+                            String map = MiscFormatter.toJson(value).toString();
+                            map = map.replaceAll("[\\{|\\}|\"]", "");
```
tananaev: Why do you have `|` in the regex? I don't think you need those.

Abyss777:
You are right.

tananaev:
Merged, thanks.

PR #2339, opened 2016-09-18
Add jpkorjar protocol support

nyash:
Hi,

I have recently been tidying up shelves in my room and I found 2 GPS devices that I bought in 2008 or so. 

They are using their own text protocol. I added support for this protocol to traccar.

So far it works fine for me, so someone might find it useful if it becomes built-in.

There isn't really any documentation for this protocol but it's pretty self explainatory. 

The only problem I see is that at least one of my devices never reports the correct course (always returns the same number) but afaik it has been like that from the beginning (presumably some bug in the device's firmware)

The code could have been more compact if it used the regexp patterns, but I am not that well versed to do it right.

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -456,5 +456,6 @@
     <entry key='arknavx8.port'>5119</entry>
     <entry key='autograde.port'>5120</entry>
     <entry key='oigo.port'>5121</entry>
+    <entry key='jpkorjar.port'>5122</entry>
 
 </properties>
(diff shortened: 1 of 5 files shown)


tananaev:
Can you please provide some original HEX message samples from server log?

nyash:
Sure, here's some samples: http://pastebin.com/ZG7NfigL

Not many with little diversity since the device was mostly stationary. I am also not at home right now, so I can't collect more samples.

tananaev:
Merged, thanks for pull request. I will do some cleanup later.

By the way, are those real full messages? They just end with comma? Is that what device sends or is it a message after frame decoder already?

tananaev:
I have updated decoder to use regular expressions.

nyash:
Hi, thanks for the merge and cleanup. Won't need to keep my own version of traccar..

As for the messages yes, they are real that I pulled from the logs (and only randomized the latitude/longitude etc to not disclose my location). 

I was perplexed by the end comma as well, but it is what the device is actually sending.

PR #2348, opened 2016-09-20
Add parameter "report.ignoreOdometer"

Abyss777:
Added this parameter to flexibly configure distance calculation.
In continue https://www.traccar.org/forums/topic/%D0%BD%D0%B5-%D1%80%D0%B0%D1%81%D1%81%D1%87%D0%B8%D1%82%D1%8B%D0%B2%D0%B0%D0%B5%D1%82-%D0%BF%D1%80%D0%BE%D0%B9%D0%B4%D0%B5%D0%BD%D0%BE%D0%B5-%D1%80%D0%B0%D1%81%D1%81%D1%82%D0%BE%D1%8F%D0%BD%D0%B8-2/

### DIFF
--- a/src/org/traccar/database/DeviceManager.java
+++ b/src/org/traccar/database/DeviceManager.java
@@ -316,14 +316,85 @@ public class DeviceManager implements IdentityManager {
         groupsById.remove(groupId);
     }
 
-    public String lookupServerAttribute(long deviceId, String attributeName) {
-        return lookupAttribute(deviceId, attributeName, true);
+    public boolean lookupServerBoolean(long deviceId, String attributeName, boolean defaultValue) {
+        String result = lookupAttribute(deviceId, attributeName, true);
+        if (result != null) {
+            return Boolean.parseBoolean(result);
+        }
+        return defaultValue;
     }
 
-    public String lookupConfigAttribute(long deviceId, String attributeName) {
-        return lookupAttribute(deviceId, attributeName, false);
+    public String lookupServerString(long deviceId, String attributeName, String defaultValue) {
+        String result = lookupAttribute(deviceId, attributeName, true);
+        if (result != null) {
+            return result;
+        }
+        return defaultValue;
     }
 
+    public int lookupServerInteger(long deviceId, String attributeName, int defaultValue) {
+        String result = lookupAttribute(deviceId, attributeName, true);
+        if (result != null) {
+            return Integer.parseInt(result);
+        }
+        return defaultValue;
+    }
+
+    public long lookupServerLong(long deviceId, String attributeName, long defaultValue) {
+        String result = lookupAttribute(deviceId, attributeName, true);
+        if (result != null) {
+            return Long.parseLong(result);
+        }
+        return defaultValue;
+    }
+
+    public double lookupServerDouble(long deviceId, String attributeName, double defaultValue) {
+        String result = lookupAttribute(deviceId, attributeName, true);
+        if (result != null) {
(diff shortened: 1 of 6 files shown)


src/org/traccar/reports/Summary.java
```diff
@@ -60,7 +63,14 @@ private static SummaryReport calculateSummaryResult(long deviceId, Date from, Da
                 speedSum += position.getSpeed();
                 result.setMaxSpeed(position.getSpeed());
             }
-            result.setDistance(ReportUtils.calculateDistance(firstPosition, previousPosition));
+            boolean ignoreOdometer = false;
+            if (device.getAttributes().containsKey(ReportUtils.IGNORE_ODOMETER)) {
```
tananaev: It would be nice to move this functionality somewhere in common place because it would be useful in multiple places. For example, I'm already using something similar for GT06 protocol encoder.

src/org/traccar/reports/ReportUtils.java
```diff
@@ -30,6 +30,8 @@
     private ReportUtils() {
     }
 
+    public static final String IGNORE_ODOMETER = "report.ignoreOdometer";
```
tananaev: As soon as we have a separate function to get config parameter, I guess we won't need a constant for it because it will only be in one/two places.

src/org/traccar/reports/Summary.java
Abyss777: There is function lookupAttribute https://github.com/tananaev/traccar/blob/master/src/org/traccar/database/DeviceManager.java#L319
But it looks through device->group->server.
Two options:
1. Use it and may be do not use config because there is server attributes.
2. Add additional parameters to function to ignore groups/server

What is better? I like first one. It will allow to configure whole group (if enabled)

tananaev: Maybe we can have two functions:
1. `lookupServerAttribute` - device -> group -> server
2. `lookupConfigAttribute` - device -> group -> config

In both cases group lookup will be configurable.

Abyss777:
Added functions and removed constant

src/org/traccar/database/DeviceManager.java
```diff
@@ -316,7 +316,16 @@ public void removeGroup(long groupId) throws SQLException {
         groupsById.remove(groupId);
     }
 
-    public String lookupAttribute(long deviceId, String attributeName) {
+    public String lookupServerAttribute(long deviceId, String attributeName) {
```
tananaev: It's better, but it still looks a bit clumsy. What do you think about adding a few helper methods with default values. For example:
```Java
public boolean lookupServerBoolean(long deviceId, String attributeName, boolean defaultValue);
public boolean lookupServerString(long deviceId, String attributeName, String defaultValue);
```
Then we won't need parse string everywhere and check for nulls. Do you think it would be a good idea? I think we just need int, boolean and string as a start.

Abyss777: .. double as well.

Also I think  we use attributes with reserve. We should use them more in some protocols for timezones or chose subformat per device. Hope this helpers "help" in adoption.

There will be a lot of new functions (8 at least), leave them in DeviceManager or move attribute lookup to new class?

tananaev: I think if we keep them in the same area of the DeviceManager class, it should be fine. Don't need a separate class for it.

Abyss777:
I implemented helpers for 5 primitives

tananaev:
Perfect, thanks. Merged.

PR #2354, opened 2016-09-21
Add attributes aliases

Abyss777:
Added attributes aliases (server part)

- It is just cache, proxy functions and API resources
- Used `id` to simplify update and remove
- Added unique constraint to deviceId-attribute to exclude duplicate attributes per device

Sometimes I used `alias` instead of `attributeAlias` just not to do long names.

### DIFF
--- a/schema/changelog-3.8.xml
+++ b/schema/changelog-3.8.xml
@@ -8,7 +8,7 @@
 
   <changeSet author="author" id="changelog-3.8">
 
-    <createTable tableName="device_aliases">
+    <createTable tableName="attribute_aliases">
       <column name="id" type="INT" autoIncrement="true">
         <constraints primaryKey="true" />
       </column>
@@ -23,8 +23,8 @@
       </column>
     </createTable>
     
-    <addForeignKeyConstraint baseTableName="device_aliases" baseColumnNames="deviceid" constraintName="fk_device_aliases_deviceid" referencedTableName="devices" referencedColumnNames="id" onDelete="CASCADE" />
-    <addUniqueConstraint tableName="device_aliases" columnNames="deviceid, attribute" constraintName="un_deviceid_attribute" />
+    <addForeignKeyConstraint baseTableName="attribute_aliases" baseColumnNames="deviceid" constraintName="fk_attribute_aliases_deviceid" referencedTableName="devices" referencedColumnNames="id" onDelete="CASCADE" />
+    <addUniqueConstraint tableName="attribute_aliases" columnNames="deviceid, attribute" constraintName="uk_deviceid_attribute" />
 
     <update tableName="users">
       <column name="map" type="VARCHAR(128)"/>
--- a/schema/changelog-3.8.xml
+++ b/schema/changelog-3.8.xml
@@ -8,6 +8,24 @@
 
   <changeSet author="author" id="changelog-3.8">
 
+    <createTable tableName="device_aliases">
+      <column name="id" type="INT" autoIncrement="true">
+        <constraints primaryKey="true" />
+      </column>
+      <column name="deviceid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="attribute" type="VARCHAR(128)">
+        <constraints nullable="false" />
+      </column>
+      <column name="alias" type="VARCHAR(128)">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+    
+    <addForeignKeyConstraint baseTableName="device_aliases" baseColumnNames="deviceid" constraintName="fk_device_aliases_deviceid" referencedTableName="devices" referencedColumnNames="id" onDelete="CASCADE" />
+    <addUniqueConstraint tableName="device_aliases" columnNames="deviceid, attribute" constraintName="un_deviceid_attribute" />
+
     <update tableName="users">
       <column name="map" type="VARCHAR(128)"/>
       <where>map = 'osm'</where>
--- a/src/org/traccar/api/resource/AttributeAliasResource.java
+++ b/src/org/traccar/api/resource/AttributeAliasResource.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2016 Anton Tananaev (anton.tananaev@gmail.com)
+ * Copyright 2016 Andrey Kunitsyn (abyss@fox5.ru)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.api.resource;
+
+import java.sql.SQLException;
+import java.util.Collection;
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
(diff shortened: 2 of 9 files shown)


schema/changelog-3.8.xml
```diff
@@ -8,6 +8,24 @@
...
+      <column name="attribute" type="VARCHAR(128)">
+        <constraints nullable="false" />
+      </column>
+      <column name="alias" type="VARCHAR(128)">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+    
+    <addForeignKeyConstraint baseTableName="device_aliases" baseColumnNames="deviceid" constraintName="fk_device_aliases_deviceid" referencedTableName="devices" referencedColumnNames="id" onDelete="CASCADE" />
+    <addUniqueConstraint tableName="device_aliases" columnNames="deviceid, attribute" constraintName="un_deviceid_attribute" />
```
tananaev: Prefix should be "uk", not "un".

schema/changelog-3.8.xml
```diff
@@ -8,6 +8,24 @@
 
   <changeSet author="author" id="changelog-3.8">
 
+    <createTable tableName="device_aliases">
```
tananaev: Did we decide on this name? It sounds confusing. I think we should rename it to `attribute_aliases` or something like that.

src/org/traccar/api/resource/AttributeAliasResource.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Path("{id}")
+    @DELETE
+    public Response remove(@PathParam("id") long id) throws SQLException {
+        Context.getPermissionsManager().checkReadonly(getUserId());
+        if (!Context.getPermissionsManager().isAdmin(getUserId())) {
+            AttributeAlias attrbuteAlias = Context.getAliasesManager().getAttributeAliasById(id);
+            Context.getPermissionsManager().checkDevice(getUserId(),
+                    attrbuteAlias != null ? attrbuteAlias.getDeviceId() : 0);
```
tananaev: I think we don't need null check here. Just let it crash.

src/org/traccar/api/resource/AttributeAliasResource.java
```diff
@@ -0,0 +1,93 @@
...
+        return Response.ok(entity).build();
+    }
+
+    @Path("{id}")
+    @PUT
+    public Response update(@PathParam("id") long id, AttributeAlias entity) throws SQLException {
+        Context.getPermissionsManager().checkReadonly(getUserId());
+        if (!Context.getPermissionsManager().isAdmin(getUserId())) {
+            AttributeAlias oldAttrbuteAlias = Context.getAliasesManager().getAttributeAliasById(entity.getId());
+            if (oldAttrbuteAlias != null && oldAttrbuteAlias.getDeviceId() != entity.getDeviceId()) {
```
tananaev: Same here. I don't think we need to check. Just let it crash if old doesn't exist. I also think we can also not check if device id is the same, just check both. Less code.

src/org/traccar/api/resource/AttributeAliasResource.java
```diff
@@ -0,0 +1,93 @@
...
+        Context.getAliasesManager().addAttributeAlias(entity);
+        return Response.ok(entity).build();
+    }
+
+    @Path("{id}")
+    @PUT
+    public Response update(@PathParam("id") long id, AttributeAlias entity) throws SQLException {
+        Context.getPermissionsManager().checkReadonly(getUserId());
+        if (!Context.getPermissionsManager().isAdmin(getUserId())) {
+            AttributeAlias oldAttrbuteAlias = Context.getAliasesManager().getAttributeAliasById(entity.getId());
```
tananaev: There is a typo in variable name. Also, I think we should use `oldEntity` and `entity` names for consistency with other resources and places. Applies to other methods.

src/org/traccar/database/AliasesManager.java
```diff
@@ -0,0 +1,119 @@
...
+import org.traccar.model.AttributeAlias;
+
+public class AliasesManager {
+
+    private final DataManager dataManager;
+
+    private final Map<Long, Set<AttributeAlias>> deviceAliases = new ConcurrentHashMap<>();
+    private final Map<Long, AttributeAlias> aliasesById = new ConcurrentHashMap<>();
+
+
```
tananaev: Please remove extra line break here.

src/org/traccar/database/AliasesManager.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    public Set<AttributeAlias> getDeviceAttributeAliases(long deviceId) {
+        if (!deviceAliases.containsKey(deviceId)) {
+            deviceAliases.put(deviceId, new HashSet<AttributeAlias>());
+        }
+        return deviceAliases.get(deviceId);
+    }
+
+    public final void refresh() {
```
tananaev: We never call it outside? Should this just be moved into constructor.

src/org/traccar/database/AliasesManager.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    public void removeArrtibuteAlias(long attributeAliasId) throws SQLException {
+        dataManager.removeAttributeAlias(attributeAliasId);
+        AttributeAlias cachedAlias = aliasesById.get(attributeAliasId);
+        getDeviceAttributeAliases(cachedAlias.getDeviceId()).remove(cachedAlias);
+        aliasesById.remove(attributeAliasId);
+    }
+
+    public AttributeAlias getDeviceAliasByAttribute(long deviceId, String attribute) {
```
tananaev: I think this can be just `getAttributeAlias`.

src/org/traccar/database/AliasesManager.java
```diff
@@ -0,0 +1,119 @@
...
+    private final Map<Long, Set<AttributeAlias>> deviceAliases = new ConcurrentHashMap<>();
+    private final Map<Long, AttributeAlias> aliasesById = new ConcurrentHashMap<>();
+
+
+    public AliasesManager(DataManager dataManager) {
+        this.dataManager = dataManager;
+        refresh();
+    }
+
+    public Set<AttributeAlias> getDeviceAttributeAliases(long deviceId) {
```
tananaev: Word `Device` can be removed.

src/org/traccar/database/AliasesManager.java
```diff
@@ -0,0 +1,119 @@
...
+    public AttributeAlias getDeviceAliasByAttribute(long deviceId, String attribute) {
+        for (AttributeAlias alias : getDeviceAttributeAliases(deviceId)) {
+            if (alias.getAttribute().equals(attribute)) {
+                return alias;
+            }
+        }
+        return null;
+    }
+
+    public Collection<AttributeAlias> getUserDevicesAttributeAliases(long userId) {
```
tananaev: Can we rename this to `getAllAttributeAliases`?

src/org/traccar/database/AliasesManager.java
```diff
@@ -0,0 +1,119 @@
...
+        Collection<AttributeAlias> userDevicesAliases = new ArrayList<>();
+        for (long deviceId : Context.getPermissionsManager().getDevicePermissions(userId)) {
+            userDevicesAliases.addAll(getDeviceAttributeAliases(deviceId));
+        }
+        return userDevicesAliases;
+    }
+
+    public AttributeAlias getAttributeAliasById(long id) {
```
tananaev: Don't need `ById`. It's implied by arguments.

tananaev: Don't need `ById`. It's implied by arguments.

tananaev: Don't need `ById`. It's implied by arguments.

schema/changelog-3.8.xml
Abyss777: Yes we did https://github.com/tananaev/traccar/issues/2302#issuecomment-246158247
Anyway, `attribute_aliases` is much clearer

tananaev: Sorry about that. Yes, I think attribute aliases is better.

Abyss777:
Looks like fixed all comments.

tananaev:
Looks good, thanks.

PR #2355, opened 2016-09-21
Adds support for specifying a host.

dcbastos:
This change allows us to use this nice script with hosts other than localhost.

Shell variables unprotected by double-quotes can cause shell-commands to be executed by the script.  It's a security vulnerability.  The quotes make sure only quote-expansion is allowed.  For instance, variables will be substituted for their values, but no shell commands will be expanded.  I effectively consider quote-unprotected variables as syntax errors --- I never write them this way.

### DIFF
--- a/tools/hex.sh
+++ b/tools/hex.sh
@@ -10,26 +10,34 @@
 
 if [ $# -lt 2 ]
 then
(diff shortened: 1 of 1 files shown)


tananaev:
Great, thanks for pull request. Merged.

PR #2356, opened 2016-09-21
Avoids nullPointerException in case no config file was given.

dcbastos:


src/org/traccar/Context.java
```diff
@@ -137,10 +137,13 @@ public static EventForwarder getEventForwarder() {
     public static void init(String[] arguments) throws Exception {
 
         config = new Config();
-        if (arguments.length > 0) {
-            config.load(arguments[0]);
+        if (arguments.length <= 0) {
+            System.out.println("No configuration file given.");
```
tananaev: Use `Log.error` instead of `System.out.println`.

src/org/traccar/Context.java
```diff
@@ -137,10 +137,13 @@ public static EventForwarder getEventForwarder() {
...
+            System.exit(0);
```
tananaev: I don't like `System.exit` as well. Can we just throw an exception instead? That was we don't need log either.

dcbastos: We can throw an exception.  I think that's a good idea.  We also couldn't log because at that point we don't have the logger ready.

So, here's what I did.

```
    public static void init(String[] arguments) throws Exception {

        config = new Config();
        if (arguments.length <= 0) {
            throw new Exception("No configuration file given");
        }
     ... 

(I also removed the period in ``[n]o configuration file given.''  It's not a log, but I suppose the same rule will apply here as well.  Let me know if it doesn't.)

```
The result being 

```
--- exec-maven-plugin:1.2.1:exec (default-cli) @ traccar ---
Exception in thread "main" java.lang.Exception: No configuration file given
	at org.traccar.Context.init(Context.java:141)
	at org.traccar.Main.main(Main.java:35)

```

tananaev: I think it's better to use RuntimeException, but yes, I like they idea of just throwing it here. Also, I think it would be better to use "Configuration file is not provided" as a text.

dcbastos: Done.

tananaev:
Merged, thanks.

PR #2367, opened 2016-09-25
Implement server statistics collection

tananaev:
@Abyss777, can you please review the changes.

For now it creates a record every minute (mostly for testing), but I plan to change it to every day a bit later (before release).

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -353,6 +353,15 @@
         DELETE FROM attribute_aliases WHERE id = :id;
     </entry>
 
+    <entry key='database.selectStatistics'>
+        SELECT * FROM statistics;
+    </entry>
+
+    <entry key='database.insertStatistics'>
+        INSERT INTO statistics (captureTime, activeUsers, activeDevices, requests, messagesReceived, messagesStored, attributes)
+        VALUES (:captureTime, :activeUsers, :activeDevices, :requests, :messagesReceived, :messagesStored, :attributes);
+    </entry>
+
     <!-- PROTOCOL CONFIG -->
 
     <entry key='gps103.port'>5001</entry>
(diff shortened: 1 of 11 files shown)


Abyss777:
Useful feature.
On the first sight everything looks good.
If you wait for tomorrow I will compile and test more carefully.

tananaev:
Thanks. I think it would be useful to have some usage statistics for the server, so decided to implement something simple. In future we can extend it to return data per protocol or something, but general numbers are enough to start with.

Abyss777:
Code works OK, without issues.
But there is some problems in wording or idea.
For example, I started test-generator an look in db, first record is OK, but next has `activeusers` = 0 and `activedevices` = 0. It not looks like true, because device are active and sends positions and me watching on toasts in web-client.

Just an suggestion, may be better use `ConnectionManager`.`activeDevices` and `ConnectionManager`.`listeners`?

Abyss777:
But I just thought if we change the period to day, your variant will be more relevant.

tananaev:
You are right about zero values. Especially it's important for device, if one has been sending for more than a day, it won't be counted as active because there would be no new "identification".

I have changed the logic. It's more heavy because it will check devices list on every message, but it's more accurate. I also changed it to daily record. Can you please check quickly and then I'll merge it.

Abyss777:
It work OK without issues and more accurate now. But ...
I've replaced interval back to minute for test. And got strange fluctuation two times.
`mesagereceived` and `messagestored` are lost two times. I gues it can be more on many devices.
image



My suggestion, the reason is synchronization/multi-threading

tananaev:
Merging it for now.

I think what you are looking at is not a statistics issue. The problem is that script that generates data doesn't actually send in exact 1 second interval. There is some delay to send data, so it accumulates and sometimes it can get through only 59 messages instead of 60.

PR #2371, opened 2016-09-26
Change aliases API path

Abyss777:
As discussed here https://github.com/tananaev/traccar-web/pull/287#discussion_r80353811


tananaev:
Thanks.

PR #2387, opened 2016-10-01
For JT600c - added checksum verification and replying for alarms/positions

vitalidze:
Per protocol documentation and conversations with support of JointTech I have updated jt600 protocol implementation in following way:

1) For U01, U02 and U03 (location data) it will parse check sum, then do the calculation on traccar side and verify it with check sum from server. If check sum differs, then position is ignored. Otherwise traccar replies with the `(S39)` confirming receiving of position.

2) For U06 (alarm data) there are no checksum. However, such messages must be accepted by the server with `(S20)` reply. If such message is not accepted, then device sends it periodically, which floods the database.

Please check and let me know whether there are any issues with implementation. I am going to implement couple of commands for JT600 next.


src/org/traccar/protocol/Jt600ProtocolDecoder.java
```diff
@@ -201,24 +201,25 @@ private Position decodeW01(String sentence, Channel channel, SocketAddress remot
...
+            .number("(d+),")                                // course
+            .number("(d+),")                                // satellites
+            .number("(d+%),")                               // battery
+            .expression("([01]+),")                         // status
+            .number("(d+),")                                // cid
+            .number("(d+),")                                // lac
+            .number("(d+),")                                // gsm signal
+            .number("(d+),")                                // odometer
+            .number("(d+),")                                // serial number
+            .expression("([0-9A-F][0-9A-F])").optional()    // checksum
```
tananaev: I think we can use `.number("(xx)?")` instead. Easier to read and won't require to change comment indentation.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
```diff
@@ -256,11 +259,29 @@ private Position decodeU01(String sentence, Channel channel, SocketAddress remot
         position.set(Position.KEY_CID, parser.nextInt());
         position.set(Position.KEY_LAC, parser.nextInt());
         position.set(Position.KEY_GSM, parser.nextInt());
-        position.set(Position.KEY_ODOMETER, parser.nextLong() * 1000);
+        position.set(Position.KEY_ODOMETER, parser.nextLong() * 100);
```
tananaev: Are you sure about this? The documentation that I have says that it's in km.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
```diff
@@ -256,11 +259,29 @@ private Position decodeU01(String sentence, Channel channel, SocketAddress remot
...
+                    sendResponse(channel, "(S39)");
+                    return position;
+                } else {
+                    return null;
+                }
+            case "U06":
+                sendResponse(channel, "(S20)");
+                return position;
+            default:
+                return null;
```
tananaev: Why by default we ignore the message?

src/org/traccar/protocol/Jt600ProtocolDecoder.java
```diff
@@ -282,4 +303,18 @@ protected Object decode(
         return null;
     }
 
+    private byte checkSum(String sentence) {
```
tananaev: Use `Checksum` class instead.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
```diff
@@ -256,11 +259,29 @@ private Position decodeU01(String sentence, Channel channel, SocketAddress remot
...
+        position.set(Position.KEY_ODOMETER, parser.nextLong() * 100);
         position.set(Position.KEY_INDEX, parser.nextInt());
 
-        return position;
+        switch (messageType) {
+            case "U01":
+            case "U02":
+            case "U03":
+                String checkSum = parser.next();
+                int calculatedCheckSum = checkSum(sentence.substring(1, sentence.length() - checkSum.length() - 1));
```
tananaev: Do we really need to check a checksum? I don't do it for other protocols because it seems redundant as TCP and UDP protocols already have 16-bit checksum which is more reliable than 1-byte xor checksum here.

test/org/traccar/protocol/Jt600ProtocolDecoderTest.java
```diff
@@ -18,13 +18,13 @@ public void testDecode() throws Exception {
                 "2475605035891613002328091601152806086750106533350c00000000000a000000000000e1ff4f97007f1607"));
 
         verifyPosition(decoder, buffer(
-                "(3301210003,U01,040812,185302,T,22.564025,N,113.242329,E,5.21,152,9,32%,00000000000011,10133,5173,22,100,1)"));
+                "(3301210003,U01,040812,185302,T,22.564025,N,113.242329,E,5.21,152,9,32%,00000000000011,10133,5173,22,100,1,18)"));
```
tananaev: Please don't modify existing test cases because they are taken from real devices or documentation, so they are valid messages. Add new ones if required.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
vitalidze: I think I can. Will update and change the indentation back.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
vitalidze: Yes, I am pretty sure. Here is what documentation says:

> | NO. | Name | Value(Extended ASCII) | Description |
> | --- | --- | --- | --- | --- |
> | 19 |  Mileage | 100 | The current value of the mileage of 100 kilometers |

And it does matches with what is coming from device. Also I believe it's weird to count mileage by thousands of kilometers.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
vitalidze: I am just following protocol documentation. It clearly says that:

> When server center received the gps data (e.g. U01,U02,U03,U07),and check if the checksum
is correct or not. If it’s correct, so send (S39) command to tracker, and tracker send the next
GPS data; if incorrect, so tracker will resend the last GPS data again until the checksum is ok

Maybe there can be a case, when it does not match. I don't know.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
vitalidze: Because according to the protocol these are only messages types with location data. Otherwise, in theory, if message is handled incorrectly, it will start flooding like it was with `U06` type of messages. What's wrong with the approach of handling only known messages types?

src/org/traccar/protocol/Jt600ProtocolDecoder.java
vitalidze: This will require wrapping sub string into `ByteBuffer`. I think it's over-engineering. Maybe it's worth to add an `checkSum(String)` method to the utility class?

test/org/traccar/protocol/Jt600ProtocolDecoderTest.java
vitalidze: From my understanding current protocol implementation **always** sends check sum for U01, U02, U03 messages types. Are you absolutely sure that these are valid messages examples?

I have updated these tests because I was curious that the implementation supporting Uxx types of messages was only added recently (for example it does not exist in 3.7 for sure) and for the same guy, for whom I am working on these protocol updates. Thus, according to the documentation and to the week of testing, the correct message format is used in tests right now.

vitalidze:
I have made some updates. Other comments require discussion.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
tananaev: You haven't changed it back. They main reason I want it back so that git doesn't think that all lines have changed. I want to see only real changes.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
tananaev: So it's in km. It's not 100s of km. 100 is just and example value in their document. Traccar saves odometer in meters, so if original value is in km, it has to be multiplied by 1000.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
tananaev: OK, let's leave checksum.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
tananaev: Can you please quote the documentation where it says those are the only valid message types. What about U07 that you mentioned in the other comment?

src/org/traccar/protocol/Jt600ProtocolDecoder.java
tananaev: Yes, we can add another method there.

test/org/traccar/protocol/Jt600ProtocolDecoderTest.java
tananaev: Traccar should support both old and current versions of the protocol. As I said, test cases are always takes from real devices or official documentation, so they are valid messages.

tananaev:
Replied to some comments.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
vitalidze: Documentation describes only these types of messages. Since there is no description on other types, like U07, I consider them invalid. Again, other types of messages may require a special type of response. Otherwise, device repeats such messages and floods the server. Thus, I believe it's better to handle only known types of messages.

src/org/traccar/protocol/Jt600ProtocolDecoder.java
vitalidze: Well, okay. Will revert back.

test/org/traccar/protocol/Jt600ProtocolDecoderTest.java
vitalidze: Ok, I updated decoder and tests.

vitalidze:
I have made couple of updates.

tananaev:
Merged, thanks.

PR #2388, opened 2016-10-01
For JT600c - added engine start/stop, set timezone and reboot commands implementations

vitalidze:
I have added following commands implementations:

* Engine start
* Ending stop
* Set Timezone
* Reboot device

Also tests were added. All commands are sent via text, so I am not sure that it is acceptable for the binary version of protocol.

### DIFF
--- a/src/org/traccar/protocol/Jt600Protocol.java
+++ b/src/org/traccar/protocol/Jt600Protocol.java
@@ -17,6 +17,7 @@ package org.traccar.protocol;
 
 import org.jboss.netty.bootstrap.ServerBootstrap;
 import org.jboss.netty.channel.ChannelPipeline;
+import org.jboss.netty.handler.codec.string.StringEncoder;
 import org.traccar.BaseProtocol;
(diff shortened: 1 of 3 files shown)


src/org/traccar/protocol/Jt600Protocol.java
```diff
@@ -17,15 +17,21 @@
...
 import org.traccar.TrackerServer;
+import org.traccar.model.Command;
 
 import java.util.List;
 
 public class Jt600Protocol extends BaseProtocol {
 
     public Jt600Protocol() {
         super("jt600");
+        setSupportedCommands(Command.TYPE_ENGINE_RESUME,
```
tananaev: Please make sure formatting is consistent with other protocol classes.

vitalidze: Well, I think I have updated formatting to match other protocol classes. Please check.

tananaev:
Thanks.

PR #2392, opened 2016-10-03
Make status events switchable

Abyss777:
Made status events "disablable" in continue to https://www.traccar.org/forums/topic/event-onlineoffline/

### DIFF
--- a/src/org/traccar/database/ConnectionManager.java
+++ b/src/org/traccar/database/ConnectionManager.java
@@ -40,6 +40,7 @@ public class ConnectionManager {
     private static final long DEFAULT_TIMEOUT = 600;
 
     private final long deviceTimeout;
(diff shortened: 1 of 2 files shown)


src/org/traccar/database/ConnectionManager.java
```diff
@@ -40,13 +40,15 @@
     private static final long DEFAULT_TIMEOUT = 600;
 
     private final long deviceTimeout;
+    private final boolean statusHandler;
```
tananaev: I think it's better to rename variable to something like `enableStatusEvents`. We can keep config key the same for consistency with others, but it's better to change variable name to make source code more clear.

Abyss777:
Sure, done.

tananaev:
Merged, thanks.

PR #2397, opened 2016-10-03
Implement trakmate protocol

santushet:
Please advice any changes needs to be done. Thanks.
.Attached protocol document.
[TMD-200 Protocol.pdf](https://github.com/tananaev/traccar/files/506307/TMD-200.Protocol.pdf)

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -485,5 +485,7 @@
     <entry key='fifotrack.port'>5124</entry>
     <entry key='smokey.port'>5125</entry>
     <entry key='extremtrac.port'>5126</entry>
+    <entry key='trakmate.port'>5127</entry>
+    <entry key='trakmate.timezone'>19800</entry>
 
 </properties>
--- a/debug.xml
+++ b/debug.xml
@@ -486,6 +486,5 @@
     <entry key='smokey.port'>5125</entry>
     <entry key='extremtrac.port'>5126</entry>
     <entry key='trakmate.port'>5127</entry>
-    <entry key='trakmate.timezone'>19800</entry>
 
 </properties>
--- a/src/org/traccar/protocol/TkmateProtocolDecoder.java
+++ b/src/org/traccar/protocol/TkmateProtocolDecoder.java
@@ -20,10 +20,7 @@ public class TkmateProtocolDecoder extends BaseProtocolDecoder {
 
     public TkmateProtocolDecoder(TkmateProtocol protocol) {
         super(protocol);
-
-        if (Context.getConfig().hasKey(getProtocolName() + ".timezone")) {
-            timeZone.setRawOffset(Context.getConfig().getInteger(getProtocolName() + ".timezone") * 1000);
-        }
+        timeZone.setRawOffset(Context.getConfig().getInteger(getProtocolName() + ".timezone") * 1000);
     }
 
     private static final Pattern PATTERN_SRT = new PatternBuilder()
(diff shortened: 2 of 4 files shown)


src/org/traccar/protocol/TkmateProtocolDecoder.java
```diff
@@ -0,0 +1,239 @@
...
+            .any()
+            .compile();
+
+
+    private String decodeAlarm(int value) {
+        switch (value) {
+            case 1:
+                return Position.ALARM_SOS;
+            case 2:
+                return null;
```
tananaev: Remove all "null" cases. There is a default case for that.

src/org/traccar/protocol/TkmateProtocolDecoder.java
```diff
@@ -0,0 +1,239 @@
...
+        DeviceSession deviceSession = getDeviceSession(channel, remoteAddress, parser.next());
+        if (deviceSession == null) {
+            return null;
+        }
+        position.setDeviceId(deviceSession.getDeviceId());
+
+        position.setLatitude(parser.nextDouble());
+        position.setLongitude(parser.nextDouble());
+
+        DateBuilder dateBuilder = new DateBuilder(TimeZone.getTimeZone("Asia/Calcutta"))
```
tananaev: Why are you using this timezone?

santushet: While testing without timezone i have found that timezone(which is gmt+5.30) is being added to time field of packet. For example, ^TMSRT|354678456723764|12.59675|77.56789|**123456**|030414|1.03|1.01|#. After parsing this packet, time becomes 18:04:56.Hence I used particular timezone to avoid this problem. Please advice right approach to solve this problem.

src/org/traccar/protocol/TkmateProtocolDecoder.java
santushet: Removed.

src/org/traccar/protocol/TkmateProtocolDecoder.java
tananaev: Do you mean device sends local time?

santushet: Yes.

tananaev: Then timezone has be to configurable in device attributes and/or config file.

santushet: Timezone is configured in config file.

santushet:
This PR is tested and working fine.

debug.xml
```diff
@@ -485,5 +484,7 @@
     <entry key='fifotrack.port'>5124</entry>
     <entry key='smokey.port'>5125</entry>
     <entry key='extremtrac.port'>5126</entry>
+    <entry key='trakmate.port'>5127</entry>
+    <entry key='trakmate.timezone'>19800</entry>
```
santushet: No. This is offset only for GMT+5.30, Please suggest the alternative method.

tananaev: Even if the device is in another country with different timezone? Then make it a default value in `Context.getConfig().getInteger(...)`.

santushet: Added default timezone "UTC".

tananaev:
Merged, thanks.

PR #2407, opened 2016-10-05
GL200: Password for commands and heartbeat authentication

Abyss777:
- Remove hardcoded password from commands, there is no default one, it depends on device model.
- Authenticate a device on heartbeat stage and reply only to registered devices. It will make the device online and allow send commands before a position received


tananaev:
Thanks. Is it really password or just device model? We can use password either way, just curious.

Abyss777:
It is really password, It can be configured via `AT+GTCFG` command.
But out of the box it is same as device model.

tananaev:
I see. The it's perfect match.

PR #2413, opened 2016-10-06
GL200: Handle ACK packets as command result

### LINKED ISSUE #2408: GL200: handle command answers
If we send command...
According to protocol architecture device firstly answers with `+ACK:<COMMANDNAME>` packet then it send some `+RESP:...` packet with requested data like battery, coordinates and others or sends nothing if it was some configuration command.

Sometimes it is useful to understand if device accepted command. (It can ignore command if wrong password or wrong parameters)

Most `+ACK:` packets contains same (nothing) information as heartbeat (actually heartbeat packet is `+ACK:GTHBD`)

I want to handle such packets as command result and do something like `set(Position.KEY_RESULT,  "Accepted "+<COMMANDNAME>);`

And next answer `+RESP:``GTBAT|GTVER|GTTMZ|GTCSQ` and others plan to handle like `GTINF`

What do you think is it OK?
Or may be that `+REST:` packets also handle as command result? And requested information put in `Result` field?

Abyss777:
fix #2408

### DIFF
--- a/src/org/traccar/protocol/Gl200ProtocolDecoder.java
+++ b/src/org/traccar/protocol/Gl200ProtocolDecoder.java
@@ -37,12 +37,14 @@ public class Gl200ProtocolDecoder extends BaseProtocolDecoder {
         super(protocol);
     }
 
-    private static final Pattern PATTERN_HBD = new PatternBuilder()
-            .text("+ACK:GTHBD,")
+    private static final Pattern PATTERN_ACK = new PatternBuilder()
+            .text("+ACK:GT").expression("...,")
             .number("([0-9A-Z]{2}xxxx),")        // protocol version
             .number("(d{15}),")                  // imei
             .any().text(",")
-            .number("(xxxx)")
+            .number("(dddd)(dd)(dd)")            // date
+            .number("(dd)(dd)(dd),")             // time
+            .number("(xxxx)")                    // counter
             .text("$").optional()
             .compile();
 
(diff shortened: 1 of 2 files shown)


src/org/traccar/protocol/Gl200ProtocolDecoder.java
```diff
@@ -37,12 +37,14 @@ public Gl200ProtocolDecoder(Gl200Protocol protocol) {
         super(protocol);
     }
 
-    private static final Pattern PATTERN_HBD = new PatternBuilder()
-            .text("+ACK:GTHBD,")
+    private static final Pattern PATTERN_ACK = new PatternBuilder()
+            .text("+ACK:GT").expression("...,")
```
tananaev: Please move expression to the next line and add a comment (e.g. "type").

src/org/traccar/protocol/Gl200ProtocolDecoder.java
```diff
@@ -620,41 +639,42 @@ protected Object decode(
...
-                result = decodeIgn(channel, remoteAddress, sentence);
-                break;
-            case "IDA":
-                result = decodeIda(channel, remoteAddress, sentence);
-                break;
-            default:
-                result = decodeOther(channel, remoteAddress, sentence, type);
-                break;
-        }
+        if (typeIndex >= 4 && sentence.startsWith("+ACK", typeIndex - 4)) {
```
tananaev: Can't we just check `sentence.startsWith("+ACK")`?

Abyss777:
Done

tananaev:
Merged, thanks.

PR #2417, opened 2016-10-07
GL200: GTVER packets and a few small improvements

Abyss777:
- Changed a bit `GTINF` pattern, because it can come from buffer
- Added decoder for GTVER packets
- Added a few alarms

I think `gl200` improvement ended for now. Looks like traccar somehow handles all needed to me packets from GL300.


src/org/traccar/model/Position.java
```diff
@@ -93,6 +93,8 @@
     public static final String ALARM_BREAKING = "hardBreaking";
     public static final String ALARM_FATIGUE_DRIVING = "fatigueDriving";
     public static final String ALARM_POWER_CUT = "powerCut";
+    public static final String ALARM_JAMMING = "jamming";
+    public static final String ALARM_TEMPERATURE = "temperature";
```
tananaev: What does this alarm mean?

src/org/traccar/protocol/Gl200ProtocolDecoder.java
```diff
@@ -660,13 +717,16 @@ protected Object decode(
...
+                    result = decodeVer(channel, remoteAddress, sentence);
+                    break;
                 default:
                     result = decodeOther(channel, remoteAddress, sentence, type);
                     break;
             }
 
             if (result == null) {
-                result = decodeBasic(channel, remoteAddress, sentence);
+                result = decodeBasic(channel, remoteAddress, sentence, type);
```
tananaev: Ideally we would want to get rid of this method eventually. That's basically a backup for all un-handled types. It's just a comment, no action needed at this point.

src/org/traccar/model/Position.java
Abyss777: I think it is better to have `LOW_TEMPERATURE` and `HIGH_TEMPERATURE` alarms but it is impossible to determine from packet.

Here is quote from doc:
"The AT+GTTEM command is used to configure temperature alarm of the device. Based on the
working mode set, the device will report temperature alarm when its temperature is outside or
inside of a predefined range."

tananaev:
Merged, thanks.

PR #2436, opened 2016-10-14
Implement limiting list of users to be notified

### LINKED ISSUE #2061: Notification Feature Suggestion
Dear Anton & Team,.

I can see the email notification system being built, so was just thinking would the notification recipient email be set per device or per userid.

As per what I feel, per device would be a good idea, when we have groups/and possibility of a single device being under multiple accounts, thus users would be able to specify who to alert for a particular device.

-Sorry wasnt sure if I should post it here or in forum :) so did it here..

Abyss777:
fix #2061 
User can specify `notifyOnly` device attribute with list of emails separated by space.
Emails must be of existed users and that users must have access to the device.

### DIFF
--- a/src/org/traccar/database/NotificationManager.java
+++ b/src/org/traccar/database/NotificationManager.java
@@ -55,7 +55,7 @@ public class NotificationManager {
             Log.warning(error);
         }
 
-        Set<Long> users = Context.getPermissionsManager().getDeviceUsers(event.getDeviceId());
+        Set<Long> users = new HashSet<Long>(Context.getPermissionsManager().getDeviceUsers(event.getDeviceId()));
         String notifyOnly = Context.getDeviceManager().lookupConfigString(event.getDeviceId(),
                 "event.notifyOnly", null);
         if (notifyOnly != null) {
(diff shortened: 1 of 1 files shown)


src/org/traccar/database/NotificationManager.java
```diff
@@ -54,6 +57,15 @@ public void updateEvent(Event event, Position position) {
         }
 
         Set<Long> users = Context.getPermissionsManager().getDeviceUsers(event.getDeviceId());
+        Device device = Context.getIdentityManager().getDeviceById(event.getDeviceId());
+        if (device != null && device.getAttributes().containsKey("notifyOnly")) {
```
tananaev: Can't we use `lookupConfigString` instead? Ideally we want it to be configurable globally in the config instead of per-device.

Abyss777: For config it have to have other name. `event.notifyOnly` will be OK?

tananaev: I just read original thread and it doesn't make sense to me. What's the use case for this? I'll ask on the original issue.

Abyss777:
It is ready. If you decide it is worth to add.

Abyss777:
Found an issue and fixed now. Sorry.

tananaev:
Let's hold off on this while we discuss #2061. It's not clear for me what we are trying to solve.

Abyss777:
Close this as not actual.

PR #2442, opened 2016-10-16
Fix month decode in Navis protocol

EvgenyPopov72:
The numbering of the months in Navis protocol starts with 0


tananaev:
Thanks for pull request. Merged.

PR #2448, opened 2016-10-17
Implement export reports to excel

Abyss777:
Key features:
- Used jxls library with Apache POI engine
- Add helper class DeviceReport to pass structured data to template and do not do it manually
- Every device on own sheet
- Added start and end coordinates to TripReport
- Pass geofences names to events template

All formatting happen in templates 

Also
- Migrate csv export to /api/positions (with the same limitation of one device)


src/org/traccar/reports/Events.java
```diff
@@ -47,16 +67,69 @@ public static String getJson(long userId, Collection<Long> deviceIds, Collection
...
+            Collection<String> types, Date from, Date to) throws SQLException, IOException {
+        ArrayList<DeviceReport> devicesEvents = new ArrayList<>();
+        ArrayList<String> sheetNames = new ArrayList<>();
+        HashMap<Long, String> geofenceNames = new HashMap<>();
         for (long deviceId: ReportUtils.getDeviceList(deviceIds, groupIds)) {
             Context.getPermissionsManager().checkDevice(userId, deviceId);
+            SortedSet<Event> eventReports = new TreeSet<Event>(new Comparator<Event>() {
+                @Override
+                public int compare(Event e1, Event e2) {
+                    return e1.getServerTime().compareTo(e2.getServerTime());
```
tananaev: I think it's better to get ordered events from the database.

src/org/traccar/reports/Events.java
```diff
@@ -47,16 +67,69 @@ public static String getJson(long userId, Collection<Long> deviceIds, Collection
...
+                    Group group = Context.getDeviceManager().getGroupById(device.getGroupId());
+                    if (group != null) {
+                        deviceEvents.setGroupName(group.getName());
+                    }
+                }
+                deviceEvents.setObjects(eventReports);
+                devicesEvents.add(deviceEvents);
+            }
+        }
+        String templatePath = Context.getConfig().getString("report.events.template.excel",
```
tananaev: A couple of things here:

1. I don't think we need two subfolders for templates. Just put xlsx files into the "templates" folder.
2. I think it would be easier if in the config we just needed to configure path to the temples, not every file separately.

Abyss777: 1. I still hope we will implement templates for mails some day and they will be placed in `templates/mail/`
2. OK, I'll combine it.

tananaev: 1. I see. I guess it makes sense to have two levels then.

src/org/traccar/reports/Events.java
Abyss777: Can you advice how to make SQL request with unknown set of `type`s. `LIKE` do not have any "OR" constructions.
I have only idea to construct `type = ... OR type = ...OR type = ...` programmatically

Abyss777:
new("java.lang.String", "") was taken from documentation https://commons.apache.org/proper/commons-jexl/reference/syntax.html but agree "" is better.

Frankly, templates debugging was cut and try method, some of jexl constructions do not work, some of jxls constructions have syntax other than in documentation. I could not make work parameterized excel formulas, as they described in documentation and so on.

I'll try simplify templates as much as can.

Abyss777:
There is an issue. If address is empty there is a warning in logs. I've tried to use:
````
${util.hyperlink("".format("http://www.openstreetmap.org/?mlat=%1$f&mlon=%2$f#map=16/%1$f/%2$f", trip.startLat, trip.startLon), "".format("%1$f °, %2$f ° - %3$s", trip.startLat, trip.startLon, trip.startAddress))}
````
and get `40.723295 °, -74.006141 ° - null`

I think we have to use "if" in this place. or any other suggestions?

**Comment by Abyss777** [comment]
As about extra rows at the beginning, we can't avoid one first row:
- `js:erea` and `js:each` can't be combined to one cell.
- `js:each ... multisheet` copying as is, from A2:I8 to A2:I8

The second row I've removed.

File: src/org/traccar/reports/Events.java (hunk 1 above)
Comment (tananaev) [inline src/org/traccar/reports/Events.java]: You can use something like `type IN ('type1', 'type2', ...)`, but how is it related to this comment?

**Comment by tananaev** [comment]
For address, you can use ternary operator if it's supported:
```java
address != null ? address : ""
```

A small side note. I think there should be no space between coordinate value and degrees sign.

src/org/traccar/reports/Events.java
Abyss777: There is loop few lines down:
````
for (String type : types) {
    Collection<Event> events = Context.getDataManager().getEvents(deviceId, type, from, to);
````
It requests events for every type separately (except allTypes)
Every result is ordered by DB, but to leave it in natural sorting order for user, whole result have to be sorted too. I do it with `SortedSet`

I can replace this loop by changing SQL request and processing `Collection<String> types` to `('type1', 'type2' ...)`

Abyss777:
I've replaced it to (I think it is better to use just address, and replace it to coordinates if null) 
````
${util.hyperlink("".format("http://www.openstreetmap.org/?mlat=%1$f&mlon=%2$f#map=16/%1$f/%2$f", position.latitude, position.longitude), position.address == null ? "".format("%1$f°, %2$f°", position.latitude, position.longitude) : position.address)}
````

and still get
````
[qtp604151333-61] WARN org.apache.commons.jexl2.JexlEngine - org.jxls.expression.JexlExpressionEvaluator.evaluate@61![135,151]: 'util.hyperlink(''.format('http://www.openstreetmap.org/?mlat=%1$f&mlon=%2$f#map=16/%1$f/%2$f', position.latitude, position.longitude), position.address == null? ''.format('%1$f°, %2$f°', position.latitude, position.longitude) : position.address);' inaccessible or unknown property position
````

Abyss777:
````
${util.hyperlink("".format("http://www.openstreetmap.org/?mlat=%1$f&mlon=%2$f#map=16/%1$f/%2$f", position.latitude, position.longitude), position.getAddress() == null ? "".format("%1$f°, %2$f°", position.latitude, position.longitude) : position.address)}
````
did the trick

src/org/traccar/reports/Events.java
Abyss777: New question, how pass such combined string to QueryBuilder.setString (or java.sql.PreparedStatement.serString) to not escape apostrophe and  do not add new?

tananaev: I think we can just get all events and filter them on the server side. It would be cleaner and easier. What do you think?

Abyss777: Tend to agree. May be it will be best solution, at least better to have two SQL requests and hack `java.sql.PreparedStatement`

Abyss777:
Seems all done.

- Filter events on server side
- Use common path for templates
- Filter events about geofences that user do not have access (minor security fix)
- Removed unused function
- Visual templates improvements

Also added `allEvents` constant and reflex it in web part of pull-request. Added logic if types omitted that means all events.

tananaev:
Looks good, merged. Have you tested everything? Does it work with empty reports as well?

Abyss777:
Oh, sorry. The only thing I missed is empty reports. What behavior should be?

Abyss777:
Is it OK if will be empty sheets for every given device? (with header but without data)

tananaev:
I think that would be the best solution.

PR #2450, opened 2016-10-17
Log which filters have been used

renaudallard:
This is useful to know which filters have been used to filter out data if you enabled multiple filters.

### DIFF
--- a/src/org/traccar/FilterHandler.java
+++ b/src/org/traccar/FilterHandler.java
@@ -175,9 +175,13 @@ public class FilterHandler extends BaseDataHandler {
             filterType.append("Distance ");
         }
 
-        if (!"".equals(filterType.toString()) && !filterLimit(p)) {
+        if (filterType.length() > 0) {
             result = true;
         }
+
+       if (filterLimit(p)) {
+            result = false;
+       }
 
         if (result) {
             Log.info("Position filtered by " + filterType.toString() + "filters from " + p.getDeviceId());
--- a/src/org/traccar/FilterHandler.java
+++ b/src/org/traccar/FilterHandler.java
@@ -148,30 +148,30 @@ public class FilterHandler extends BaseDataHandler {
         }
     }
 
-    private boolean filter(Position p) {
+    private boolean filter(Position position) {
 
         boolean result = false;
         StringBuilder filterType = new StringBuilder(8 + 5 + 10 + 7 + 12 + 7 + 9);
 
-        if (filterInvalid(p)) {
+        if (filterInvalid(position)) {
             filterType.append("Invalid ");
         }
-        if (filterZero(p)) {
+        if (filterZero(position)) {
             filterType.append("Zero ");
         }
-        if (filterDuplicate(p)) {
+        if (filterDuplicate(position)) {
             filterType.append("Duplicate ");
         }
-        if (filterFuture(p)) {
+        if (filterFuture(position)) {
             filterType.append("Future ");
         }
-        if (filterApproximate(p)) {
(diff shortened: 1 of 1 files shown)


src/org/traccar/FilterHandler.java
```diff
@@ -150,16 +150,40 @@ private boolean filterLimit(Position position) {
 
     private boolean filter(Position p) {
 
+        String filterType = "";
```
tananaev: You should use `StringBuilder` instead.

src/org/traccar/FilterHandler.java
```diff
@@ -150,16 +150,40 @@ private boolean filterLimit(Position position) {
...
+
         boolean result = filterInvalid(p) || filterZero(p) || filterDuplicate(p)
                 || filterFuture(p) || filterApproximate(p) || filterStatic(p)
                 || filterDistance(p);
 
+        if (filterInvalid(p)) {
```
tananaev: You are checking every filter twice. I think it's not a good idea. Filter check should be combined with logging.

src/org/traccar/FilterHandler.java
```diff
@@ -150,16 +150,40 @@ private boolean filterLimit(Position position) {
...
+            filterType = "Invalid ";
+        }
+        if (filterZero(p)) {
+            filterType = (filterType + "Zero ");
```
tananaev: There is no need for parentheses here. Applies to all other cases as well.

src/org/traccar/FilterHandler.java
```diff
@@ -150,16 +150,40 @@ private boolean filterLimit(Position position) {
...
     private boolean filter(Position p) {
 
+        String filterType = "";
+
         boolean result = filterInvalid(p) || filterZero(p) || filterDuplicate(p)
                 || filterFuture(p) || filterApproximate(p) || filterStatic(p)
                 || filterDistance(p);
 
+        if (filterInvalid(p)) {
+            filterType = "Invalid ";
```
tananaev: This case shouldn't be different from others. For consistency and easier modifications in future.

src/org/traccar/FilterHandler.java
```diff
@@ -150,16 +150,40 @@ private boolean filterLimit(Position position) {
 
     private boolean filter(Position p) {
```
tananaev: If you break down filters into separate ifs, please rename the variable to "position".

src/org/traccar/FilterHandler.java
renaudallard: Yes, I don't like it either, but the filterLimit should be taken into account

src/org/traccar/FilterHandler.java
```diff
@@ -150,16 +150,37 @@ private boolean filterLimit(Position position) {
...
+            filterType.append("Approximate ");
+        }
+        if (filterStatic(p)) {
+            filterType.append("Static ");
+        }
+        if (filterDistance(p)) {
+            filterType.append("Distance ");
+        }
+
+        if (!"".equals(filterType.toString()) && !filterLimit(p)) {
```
tananaev: Please no. This is definitely not the right way.

renaudallard: But I don't like putting a result = true in every if either

tananaev: At least it should be `!filterType.isEmpty()`.

renaudallard:
Anything else is not OK for you?

tananaev:
I have some more comments, but I'll fix them myself.

PR #2451, opened 2016-10-17
Fixed Protocol Logic

Jaguar83:
Protocol decoder is incorrectly reading the third field as the IMEI, rather than the second field as specified in GlobalSat's TR-313 Documentation v0.3 and TR-206 Documentation v.7_110111.


tananaev:
Any comments why you closed this pull request?

Jaguar83:
Yeah, it was incorrect and caused errors. I seems that GlobalSat have a 
new, different version of the protocol that is incompatible with the 
implementation in Traccar. However, there is nothing in the data stream 
that identifies which version of the protocol to use. We're still 
testing with the unit we have.

PR #2461, opened 2016-10-19
Log device uniqueId in filters logging

renaudallard:
This combined with another commit to traccar-web makes it easier to see which device has been filtered.


test/org/traccar/FilterHandlerTest.java
```diff
@@ -6,10 +6,46 @@
 import static org.junit.Assert.assertNull;
 import org.junit.Before;
 import org.junit.Test;
+import org.traccar.database.IdentityManager;
 import org.traccar.model.Position;
+import org.traccar.model.Device;
 
 public class FilterHandlerTest {
 
+    static {
```
tananaev: We already have this somewhere. Just reuse the code.

src/org/traccar/FilterHandler.java
```diff
@@ -175,7 +175,16 @@ private boolean filter(Position position) {
         }
 
         if (filterType.length() > 0 && !filterLimit(position)) {
-            Log.info("Position filtered by " + filterType.toString() + "filters from " + position.getDeviceId());
+
+            StringBuilder logmsg = new StringBuilder();
```
tananaev: Please use proper name for variable. I think "message" would do.

test/org/traccar/FilterHandlerTest.java
```diff
@@ -6,15 +6,19 @@
...
 
 public class FilterHandlerTest {
 
+    private EventHandlerTest deviceHandler;
     private FilterHandler filtingHandler;
     private FilterHandler passingHandler;
 
     @Before
     public void setUp() {
+        deviceHandler = new EventHandlerTest();
```
tananaev: Why are you using `EventHandlerTest` here? It has nothing to do with filtering.

tananaev: You should reuse the code, but it's not the right way to do it. If the functionality is common between two modules, move it in some common place. Possibly a base class.

tananaev: Maybe `BaseTest`. Then `ProtocolTest` can extend it. Actually you can probably rename `EventHandlerTest` to `BaseTest`, and make it a base class of `ProtocolTest`.

test/org/traccar/EventHandlerTest.java
```diff
@@ -4,40 +4,6 @@
 import org.traccar.model.Device;
 import org.traccar.model.Position;
 
-public class EventHandlerTest {
+public class EventHandlerTest extends BaseTest {
```
tananaev: Why do we need an empty class?

tananaev:
Looks good now, thanks.

PR #2464, opened 2016-10-20
Fix export empty reports to Excel

Abyss777:
Also cleaned up templates metadata (author, path etc. Sometimes I hate MS programs)


tananaev:
Thanks.

Issue #2316, opened 2016-09-14
Feature Request : Device Management Info


jaimzj:
Dear All,

A small addition in device table I am listing as a feature request, as this will allow users with 100's of devices to better manage their inventory of devices on traccar.

In device table, two additional columns can be added to identify the following.

device->msisdn  (The sim card number/mobile number of sim card inside the device)
device->model/protocol - drop down list  (Helps identify the device model / protocol for sending commands etc)

tananaev:
I agree with protocol, but phone number/sim card is a device specific thing. Some devices don't use SIM cards at all.

nefertiti007:
Dear Anton,

correct me if am wrong,. I was under the impression, most of the commercial GPS Tracking devices worked on SIM Card and very few satellite based devices (expensive) ones do exist.

-Also the idea here is, if a user has lets say GT06 or TK103 (100's or 1000's of them), many a times the user would like to simply store the phone number of the device sim card as well, for easy retrieval while tracking, Otherwise end up having to maintain a separate record to cross refer. where lets say I want to send a command to a device (however the device does not support command from platform). If I can view the sim card number as a user, it makes the user/admins life easier.

- Similarly, With protocol/model number, what commands to send also can be easier to plan, for users who use multiple models of devices.

tananaev:
I guess most of devices do have SIM card. And it might be useful for SMS commands in future, so I guess it makes sense to add it as well.

Abyss777:
Hello
Are you talking about device attributes? They are already. Or something else?

tananaev:
Yes, but attributes are harder to access. I think it make sense to add some attributes directly to the device model.

khfrabbi:
I think you also can add drivers(vehicle user) mobile no to contact them or you can add a new table for drivers information and map driver to a device.

tananaev:
@khfrabbi, I think that would out of scope for Traccar. Traccar is used not only for vehicle tracking, but also for personal, animals, cargo tracking and other cases. Drivers would only make sense for vehicles.

khfrabbi:
@tananaev , actually I didn't meant drivers, it should be contact/responsible person's for that device whether it is personal, animals, cargo tracking or other cases. 

anyway thanks for your great job.

jaimzj:
@tananaev , @Abyss777 

I was thinking, maybe how about implementing a module based feature facility.? 

[...]

PR #2466, opened 2016-10-20
Extend device model with management info

Abyss777:
As discussed in #2316

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -168,8 +168,8 @@
     </entry>
 
     <entry key='database.insertDevice'>
-        INSERT INTO devices (name, uniqueId, groupId, attributes, phone, model, contact, type)
-        VALUES (:name, :uniqueId, :groupId, :attributes, :phone, :model, :contact, :type)
+        INSERT INTO devices (name, uniqueId, groupId, attributes, phone, model, contact, category)
+        VALUES (:name, :uniqueId, :groupId, :attributes, :phone, :model, :contact, :category)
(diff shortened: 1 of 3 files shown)


renaudallard:
What's the difference between model and type?

Abyss777:
Model is «Wonde Proud M7» or «Teltonika FM1100». Type is «truck» or «plane» or «yacht».

renaudallard:
Wouldn't something like "conveyor" be more explicit than "type"? I hardly find a better word to be honest.

renaudallard:
It would be nice to add the SIM number and SIM provider too.

tananaev:
I guess we can use "class" or "category" instead of "type", to make it more clear that it's the object type, not device type. @Abyss777, what do you think?

SIM number is the phone number, so it's already there.

renaudallard:
By SIM number, I meant the ICCID. Just like the phone number or contact or category, it's handy to have it there.

tananaev:
That can definitely go into device attributes. We don't need a separate column for it.

Abyss777:
I like "category". Renamed.

tananaev:
Looks good, thanks.

renaudallard:
Category is not visible/editable in the web interface, is that expected?

Abyss777:
@renaudallard 
Yes, it is. I'm going to use it to set images for devices. Therefore there will be additional dialog or something.

PR #2467, opened 2016-10-21
Copy some attributes from last position

### LINKED ISSUE #2325: Feature Request : Position : Sticky Informations (Key Values)
Dear All,

For few of the protocols I have tested lately, I have been noticing a situation. this includes GT06, meiligao, 	gps103 and such. I will take the example of GT06 in this case,.

In Gt06, KEY_POWER value is sent at set intervals, and new Messages from device supersede the said updates,.

Same for Ignition value as well (However, I do know Ignition can be now found in events) .

Whats your thought on this, Can a solution be found for this? where certain parameters / key values can be kept available for live mapping purposes.

- Specially for users who use traccar as a back-end.

Abyss777:
Base simple version without expiration
User can enable it by setting `<entry key='processing.copyAttributes.enable'>true</entry>`
and set list of attributes separated by space in `processing.copyAttributes` (device->group->config)

fix #2325

### DIFF
--- a/src/org/traccar/CopyAttributesHandler.java
+++ b/src/org/traccar/CopyAttributesHandler.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2016 Anton Tananaev (anton.tananaev@gmail.com)
+ * Copyright 2016 Andrey Kunitsyn (abyss@fox5.ru)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


src/org/traccar/CopyAttributesHandler.java
```diff
@@ -0,0 +1,45 @@
...
+        if (Context.getIdentityManager() != null) {
+            return Context.getIdentityManager().getLastPosition(deviceId);
+        }
+        return null;
+    }
+
+    @Override
+    protected Position handlePosition(Position position) {
+        String attributesString = Context.getDeviceManager().lookupConfigString(position.getDeviceId(),
+                "processing.copyAttributes", null);
```
tananaev: Please add an example to default config.

Abyss777:
Done

tananaev:
Merged, thanks.

PR #2470, opened 2016-10-22
Implement reset total distance API

Abyss777:
I have some doubts about API path, leaved `{id}` for consistency.
Only admins are allowed to reset `totalDistance`.
Just re-add  last position with corrected attribute. If there is no last position do nothing (and throw exception to notify user)

Will be thankful for advice where to place it to web-interface. I'm thinking of placing `textfield` and `button` in device edit dialog. Or make new dialog and call it from settings menu?

### DIFF
--- a/src/org/traccar/api/resource/DeviceResource.java
+++ b/src/org/traccar/api/resource/DeviceResource.java
@@ -18,6 +18,7 @@ package org.traccar.api.resource;
 import org.traccar.Context;
 import org.traccar.api.BaseResource;
 import org.traccar.model.Device;
+import org.traccar.model.DeviceTotalDistance;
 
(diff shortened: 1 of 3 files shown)


src/org/traccar/api/resource/DeviceResource.java
```diff
@@ -92,4 +93,12 @@ public Response remove(@PathParam("id") long id) throws SQLException {
         return Response.noContent().build();
     }
 
+    @Path("{id}/totaldistance")
```
tananaev: I think we can just call it distance. There won't be any other options anyway.

Abyss777:
Renamed path

tananaev:
Merged, thanks.

PR #2480, opened 2016-10-24
Add maximum coordinates error to CoordinatesHandler

Abyss777:
I've extended a bit `CoordinatesHandler`.

Some devices send wrong coordinates  when can not fix GPS ( 0.0/0.0 or with wrong sign like Granit) but send useful information about sensors or alarms that shouldn't be filtered.
Like "too small shift" I added "too big shift" that is replaced with last coordinates.

### DIFF
--- a/src/org/traccar/CoordinatesHandler.java
+++ b/src/org/traccar/CoordinatesHandler.java
@@ -20,11 +20,13 @@ import org.traccar.model.Position;
 
 public class CoordinatesHandler extends BaseDataHandler {
 
-    private final int coordinatesError;
+    private final int coordinatesMinError;
+    private final int coordinatesMaxError;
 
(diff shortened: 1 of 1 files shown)


tananaev:
Good option. The only problem I see with it is that it can work the opposite way. If first positions is incorrect (and in most cases it is), you will never get correct coordinates.

Abyss777:
You are right. It should be used carefully.
May be add exclusion if last has zero coordinates than do not replace them. It is workaround but can solve some time for user (change config, restart server)

tananaev:
We can check both "valid" flag and zero coordinates. Also, it would be nice to not have default values, so for example, if you don't want to use "maxError", you just don't specify it in the config. In current implementation, if you enable this handler, it would always have some values for both min and max.

PR #2481, opened 2016-10-24
Add exclusion for valid positions to CoordinatesHandler

Abyss777:
It should prevent problem you mention in #2480 
Valid position with "too big shift" should not be replaced with last coordinates.

### DIFF
--- a/src/org/traccar/CoordinatesHandler.java
+++ b/src/org/traccar/CoordinatesHandler.java
@@ -42,7 +42,7 @@ public class CoordinatesHandler extends BaseDataHandler {
         if (last != null) {
             double distance = DistanceCalculator.distance(
                     position.getLatitude(), position.getLongitude(), last.getLatitude(), last.getLongitude());
-            if (distance < coordinatesMinError || distance > coordinatesMaxError) {
+            if (distance < coordinatesMinError || distance > coordinatesMaxError && !position.getValid()) {
                 position.setLatitude(last.getLatitude());
                 position.setLongitude(last.getLongitude());
             }
--- a/src/org/traccar/CoordinatesHandler.java
+++ b/src/org/traccar/CoordinatesHandler.java
@@ -39,11 +39,12 @@ public class CoordinatesHandler extends BaseDataHandler {
     @Override
     protected Position handlePosition(Position position) {
         Position last = getLastPosition(position.getDeviceId());
-        if (last != null && last.getLatitude() != 0 && last.getLongitude() != 0) {
+        if (last != null && last.getValid() && last.getLatitude() != 0 && last.getLongitude() != 0) {
             double distance = DistanceCalculator.distance(
                     position.getLatitude(), position.getLongitude(), last.getLatitude(), last.getLongitude());
-            if (coordinatesMinError != 0 && distance < coordinatesMinError
-                    || coordinatesMaxError != 0 && distance > coordinatesMaxError && !position.getValid()) {
+            boolean satisfiesMin = coordinatesMinError == 0 || distance > coordinatesMinError;
+            boolean satisfiesMax = coordinatesMaxError == 0 || distance < coordinatesMaxError || position.getValid();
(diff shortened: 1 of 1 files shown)


tananaev:
Please see my last comment on #2480.

Abyss777:
The logic is:
- Any new position better than last with zeros
- New position with distance longer than `maxError` and not valid will be replaced. I think this combination will cover all possible wrong shifts. I suppose that if device send coordinates and mark them as valid that means it is really there (or broken).

Am I missed something?

src/org/traccar/CoordinatesHandler.java
```diff
@@ -39,10 +39,11 @@ private Position getLastPosition(long deviceId) {
     @Override
     protected Position handlePosition(Position position) {
         Position last = getLastPosition(position.getDeviceId());
-        if (last != null) {
+        if (last != null && last.getLatitude() != 0 && last.getLongitude() != 0) {
             double distance = DistanceCalculator.distance(
                     position.getLatitude(), position.getLongitude(), last.getLatitude(), last.getLongitude());
-            if (distance < coordinatesMinError || distance > coordinatesMaxError) {
+            if (coordinatesMinError != 0 && distance < coordinatesMinError
+                    || coordinatesMaxError != 0 && distance > coordinatesMaxError && !position.getValid()) {
```
tananaev: I would split the expression and put min and max conditions into variables. Something like this:
```java
boolean satisfiesMin = coordinatesMinError == 0 || distance > coordinatesMinError;
```

Also, I think it would make sense to make `position.getValid()` apply to both min and max.

Abyss777: For "min" we are avoiding fluctuations during the parking, I think it does not matter valid or not, just stabilize the coordinates.

tananaev: I know, but it looks like a bug. I think it's better to make it symmetrical so to say. It won't hurt.

Abyss777: I came to this, right?
```java
boolean satisfiesMin = coordinatesMinError == 0 || distance > coordinatesMinError || !position.getValid();
boolean satisfiesMax = coordinatesMaxError == 0 || distance < coordinatesMaxError || position.getValid();
if (!(satisfiesMin && satisfiesMax)) {
    position.setLatitude(last.getLatitude());
    position.setLongitude(last.getLongitude());
}
```

Abyss777: We should not replace coordinates for too big distance if they are valid, but we should do if not valid.

tananaev: Sorry, I think it should be `last.getValid()`, and we can probably move it to the first check.

Abyss777:
Commited changes.

src/org/traccar/CoordinatesHandler.java
```diff
@@ -39,10 +39,12 @@ private Position getLastPosition(long deviceId) {
...
     protected Position handlePosition(Position position) {
         Position last = getLastPosition(position.getDeviceId());
-        if (last != null) {
+        if (last != null && last.getValid() && last.getLatitude() != 0 && last.getLongitude() != 0) {
             double distance = DistanceCalculator.distance(
                     position.getLatitude(), position.getLongitude(), last.getLatitude(), last.getLongitude());
-            if (distance < coordinatesMinError || distance > coordinatesMaxError) {
+            boolean satisfiesMin = coordinatesMinError == 0 || distance > coordinatesMinError;
+            boolean satisfiesMax = coordinatesMaxError == 0 || distance < coordinatesMaxError || position.getValid();
+            if (!(satisfiesMin && satisfiesMax)) {
```
tananaev: Sorry for being so picky, but can we please expand this if expressions to be
```java
!satisfiesMin || !satisfiesMax
```
I think it would be more readable like this.

Abyss777:
It's OK. Done.

tananaev:
Merged, thanks.

PR #2491, opened 2016-10-27
Export positions to GPX

Abyss777:
Export only base values supported by GPX schema

- `latitude`
- `longitude`
- `fixTime` as `time`
- `altitude` as `geoidheight` (had some doubds, but think it fits better than `elevation`)
- `course`
- `speed`


tananaev:
Looks good, merged.

PR #2498, opened 2016-10-31
Update swagger.json with recent changes

Abyss777:
Looks like forget nothing since 3.7 release

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -159,6 +159,38 @@
                 }
             }
         },
+        "/devices/{id}/distance": {
+            "put": {
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "parameters": [
+                    {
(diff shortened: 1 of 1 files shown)


tananaev:
Merged, thanks.

Sorry for off-topic, but did you have any problems with submodule? Just want some feedback if it's easier to use than completely separate repo.

Abyss777:
I did not have any problems. But I do not see much benefits. They are only for newbie or people who just want to compile (not to do a pull-requests).

Man who want to contribute have to fork both repositories and add "remote" anyway. Even there is one additional action: rename your "origin" to "upstream". I mean contributors will treat them as two repositories anyway.

tananaev:
It's very handy with IntelliJ. It automatically keeps both repos in sync. With two separate repositories it's more complicated.

Abyss777:
I will work more with it. May be I did not noticed all benefits.

PR #2502, opened 2016-11-01
Remove explicit JSON conversion from reports

Abyss777:
There are two commits:

- Removed explicit conversion to JSON from reports
- Move `parseDate` to `DateUtil`

`JsonConverter` leaved as is.


tananaev:
Perfect. I forgot that we already have a `DateUtil` class.

PR #2512, opened 2016-11-07
Fix export templates

Abyss777:
Re-saved templates from MS Office.

Unfortunately removed summaries from trips report, there is no way (I have not found) to summarize values with units.


tananaev:
Thanks.

I hate that it doesn't show diffs. Excel documents are really just compressed XML files.

Abyss777:
Agree, really inconvenient.
We can extract content to subfolders and zip them on the build stage. Seems maven supports it.

tananaev:
I thought about that option as well, but then it's really painful to edit them.

PR #2535, opened 2016-11-15
Add Service traccar server for package_universal

CVSz:


tananaev:
I think it's better to put this file in a subfolder (e.g. "other").

PR #2539, opened 2016-11-16
Add lib for Microsoft JDBC Drivers

CVSz:
Test Server on
Windows Server 2012 R2
MSSQL 2014

### DIFF
--- a/setup/package.sh
+++ b/setup/package.sh
@@ -97,7 +97,7 @@ copy_wrapper () {
 copy_files () {
   cp ../target/tracker-server.jar out
   cp ../target/lib/* out/lib
-  cp ../sqljdbc/enu/sqljdbc*.jar out/lib
+  cp ../sqljdbc/sqljdbc*.jar out/lib
   cp ../schema/* out/schema
   cp -r ../templates/* out/templates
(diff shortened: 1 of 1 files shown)


tananaev:
I don't think it's a good idea to include proprietary driver.

Have you actually checked Microsoft terms and conditions to see if it's allowed.

Abyss777:
I'm not sure it is possible to distribute Microsoft JDBC driver in traccar installer at the moment.
According to license 

> Before the transfer, that party must agree that this agreement applies to the transfer and use of the software

user must agree its licence. We do not have such mechanism in installer at the moment.

tananaev:
@CVSz, please comment.

CVSz:
I am consulting friends programming. That can be used or not, and now I'm editing. Some defect

CVSz:
i see maven have repository for ms jdbc driver

        
        <dependency>
            <groupId>com.microsoft</groupId>
            <artifactId>sqljdbc4</artifactId>
            <version>3.0</version>
        </dependency>

CVSz:
i think we can use maven repo for mssql after build source

tananaev:
Agree, but it would be still nice to find out if it's possible to distribute it.

PR #2548, opened 2016-11-17
Add user expiration and device limit

tananaev:
@Abyss777, can you please review the changes.

Added:

- Flag to disable users
- Expiration date
- Device limit for user

Now to change any admin-related values user should be admin. It's easier logic, but it means that user won't be able to make itself readonly like before. Probably a good thing :)

I have hidden some field in the web UI for non-admins. Otherwise too many parameters.

### DIFF
--- a/schema/changelog-3.8.xml
+++ b/schema/changelog-3.8.xml
@@ -107,5 +107,15 @@
       <column name="category" type="VARCHAR(128)" />
     </addColumn>
 
+    <addColumn tableName="users">
+      <column name="disabled" type="BOOLEAN" defaultValueBoolean="false" />
+    </addColumn>
+    <addColumn tableName="users">
+      <column name="expirationtime" type="TIMESTAMP" />
+    </addColumn>
+    <addColumn tableName="users">
+      <column name="devicelimit" type="INT" defaultValueNumeric="0" />
+    </addColumn>
+
   </changeSet>
 </databaseChangeLog>
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -62,8 +62,8 @@
     </entry>
 
     <entry key='database.insertUser'>
-        INSERT INTO users (name, email, hashedPassword, salt, admin, map, distanceUnit, speedUnit, latitude, longitude, zoom, twelveHourFormat, attributes)
-        VALUES (:name, :email, :hashedPassword, :salt, :admin, :map, :distanceUnit, :speedUnit, :latitude, :longitude, :zoom, :twelveHourFormat, :attributes)
+        INSERT INTO users (name, email, hashedPassword, salt, admin, map, distanceUnit, speedUnit, latitude, longitude, zoom, twelveHourFormat, coordinateFormat, disabled, expirationTime, deviceLimit, attributes)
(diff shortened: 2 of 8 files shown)


Abyss777:
Looks good except one thing:

Disabled/expired user can update page, reopen browser until log out or server restart.
May be add some check to api/SecurityRequestFilter.java#81

tananaev:
Fixed the problem. Check user on every call.

Abyss777:
I think server part is OK now.

CVSz:
Good idea. @Abyss777  :)

Abyss777:
Or I'm too harried, may be better do 
`request.getSession().invalidate();` if account disabled

tananaev:
I think the problem is fixed now.

Abyss777:
I don't think so, the session is still cached on server and user will get browser authentication window until clear his cookies.
Here is my solution:
Make `checkUser` just boolean function
````java
public boolean checkUser(long userId) throws SecurityException {
        User user = getUser(userId);
        if (user.getDisabled() || (user.getExpirationTime() != null
                && System.currentTimeMillis() > user.getExpirationTime().getTime())) {
            return false;
        }
        return true;
    }
````
Throw exception in `login` function
````java
public User login(String email, String password) throws SQLException {
        User user = dataManager.login(email, password);
        if (user != null) {
            if (!checkUser(user.getId())) {
                throw new SecurityException("Account has disabled/expired");
            }
            return users.get(user.getId());
        }
        return null;
    }
````
Revert try-catch from `filter` and add session invalidating
````java
public void filter(ContainerRequestContext requestContext) {

        if (requestContext.getMethod().equals("OPTIONS")) {
            return;
        }

        SecurityContext securityContext = null;

        String authHeader = requestContext.getHeaderString(AUTHORIZATION_HEADER);
        if (authHeader != null) {

            try {
                String[] auth = decodeBasicAuth(authHeader);
                User user = Context.getPermissionsManager().login(auth[0], auth[1]);
                if (user != null) {
                    Context.getStatisticsManager().registerRequest(user.getId());
                    securityContext = new UserSecurityContext(new UserPrincipal(user.getId()));
                }
            } catch (SQLException e) {
                throw new WebApplicationException(e);
            }

        } else if (request.getSession() != null) {

            Long userId = (Long) request.getSession().getAttribute(SessionResource.USER_ID_KEY);
            if (userId != null) {
                if (Context.getPermissionsManager().checkUser(userId)) {
                    Context.getStatisticsManager().registerRequest(userId);
                    securityContext = new UserSecurityContext(new UserPrincipal(userId));
                } else {
                    request.getSession().invalidate();
                }
            }

        }

        if (securityContext != null) {
            requestContext.setSecurityContext(securityContext);
        } else {
            Method method = resourceInfo.getResourceMethod();
            if (!method.isAnnotationPresent(PermitAll.class)) {
                throw new WebApplicationException(
                        Response.status(Response.Status.UNAUTHORIZED).header(WWW_AUTHENTICATE, BASIC_REALM).build());
            }
        }

    }
````

Behavior will be next:
- If user is login and disabled after that then push F5 he will be thrown to login page
- If user is login and disabled after that then do some request, like edit or add something, he will see browser auth window, and if will pass his credential he will get auth error.

I think it is best behavior we can get and it is understandable for users.

tananaev:
A bit different solution. Works now and returns different errors for expired and disabled accounts. What do you think?

Abyss777:
Your solution is definitely more simple and has same result behavior.
Nothing to add. Everything is OK now for me.

PR #2552, opened 2016-11-18
Implement users tokens

Abyss777:
Here is nice addition to users disabling.
Main reason is to give link user to traccar site without username/password

User can have a token, that can be used to authorization via GET parameter instead of username/password:
`https://demo.traccar.org/api/session?token=s5XpfXKZ3zLi7X4A`

Only Admin can add/change token to user.

Also hardcoded minimal token length and symbols to use in token.

Security note: tokens should be used very carefully and mostly with read only users in combination with expiring. Only over HTTPS. Because `GET` parameters is absolutely unsecured and cached/logged everywhere.

### DIFF
--- a/src/org/traccar/api/resource/SessionResource.java
+++ b/src/org/traccar/api/resource/SessionResource.java
@@ -29,6 +29,7 @@ import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
@@ -48,7 +49,7 @@ public class SessionResource extends BaseResource {
 
     @PermitAll
     @GET
-    public User get() throws SQLException {
+    public User get(@QueryParam("token") String token) throws SQLException {
         Long userId = (Long) request.getSession().getAttribute(USER_ID_KEY);
         if (userId == null) {
             Cookie[] cookies = request.getCookies();
@@ -69,7 +70,7 @@ public class SessionResource extends BaseResource {
                     userId = user.getId();
                     request.getSession().setAttribute(USER_ID_KEY, userId);
                 }
-            } else if (request.getParameter("token") != null) {
+            } else if (token != null) {
                 User user = Context.getPermissionsManager().getUserByToken(request.getParameter("token"));
                 if (user != null) {
                     userId = user.getId();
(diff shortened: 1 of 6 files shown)


src/org/traccar/api/resource/SessionResource.java
```diff
@@ -69,6 +69,12 @@ public User get() throws SQLException {
                     userId = user.getId();
                     request.getSession().setAttribute(USER_ID_KEY, userId);
                 }
+            } else if (request.getParameter("token") != null) {
```
tananaev: Request parameter can be declared as a method argument. See position resource for example.

src/org/traccar/api/resource/UserResource.java
```diff
@@ -66,7 +66,8 @@ public Response update(@PathParam("id") long id, User entity) throws SQLExceptio
                 || old.getAdmin() != entity.getAdmin()
                 || old.getReadonly() != entity.getReadonly()
                 || old.getDisabled() != entity.getDisabled()
-                || old.getDeviceLimit() != entity.getDeviceLimit()) {
+                || old.getDeviceLimit() != entity.getDeviceLimit()
+                || !old.getToken().equals(entity.getToken())) {
```
tananaev: What if it's null? Is it possible?

src/org/traccar/database/PermissionsManager.java
```diff
@@ -39,6 +40,7 @@
     private volatile Server server;
 
     private final Map<Long, User> users = new ConcurrentHashMap<>();
+    private final Map<Long, String> usersTokens = new HashMap<>();
```
tananaev: I think you should flip the parameters. Lookup by token seems to be more frequent operation than creating new token.

src/org/traccar/model/User.java
```diff
@@ -180,6 +180,31 @@ public void setDeviceLimit(int deviceLimit) {
...
 
+    private String token;
+
+    public String getToken() {
+        return token;
+    }
+
+    public void setToken(String token) {
+        if (token != null && !token.isEmpty()) {
+            if (validateToken(token)) {
```
tananaev: I suggest to keep validation in the setter for now. Something like this:
```java
if (token != null && !token.isEmpty()) {
    if (!token.matches("^[a-zA-Z0-9]{16,}$")) {
        throw new IllegalArgumentException("Illegal token");
    }
    this.token = token;
}
```

Abyss777:
It's up to you.  I'm not sure I'll fix all comments before release.

Abyss777:
Sorry for missed null checks and `@QueryParam`.
`userId`-`token` map looked more simple to update for me.
Also combined validation to setter as you suggested.

tananaev:
I have a couple of small comments, but I'll fix them myself to completely this before release.

PR #2569, opened 2016-11-20
Increase minimal token length

Abyss777:
To be consistent with web part


tananaev:
I just though... what would happen to existing 16-character tokens?

Abyss777:
Hm, you are right. User object won't be created from DB.

tananaev:
I guess we should keep it as it is. There might be a way to deal with it,  but I don't want to make it too complicated.

Abyss777:
Move validation to API resource.
But OK. Leave it as is.

PR #2578, opened 2016-11-21
small fixes getting image build again

oudenaar:
Hi Anton,

I have fixed the stuff, I'm able to create an images and I can run the container.

I'm close, I see javascript spinning wheel trying to load login page

maybe you can inspect what I have done and to make it 100% working!

Thanks,
Bas

### DIFF
--- a/setup/docker/build.sh
+++ b/setup/docker/build.sh
@@ -13,7 +13,7 @@ tmp="./setup/docker/tmp"
 
 mkdir -p ${tmp}
 
-cat ./setup/traccar.xml | awk '/web.path/ && !modif { printf("    <entry key=\"web.debug\">true</entry>\n"); modif=1 } {print}' > ${tmp}/traccar.xml
+cat ./setup/traccar.xml | awk '/config.default/ && !modif { print;printf("    <entry key=\"web.debug\">true</entry>\n");next; modif=1 } {print}'  > ${tmp}/traccar.xml
 cp -rf ./setup/default.xml ${tmp}
 cp -rf ./schema ${tmp}/schema
 cp -rf ./target/tracker-server.jar ${tmp}/traccar-server.jar
--- a/setup/docker/build.sh
+++ b/setup/docker/build.sh
@@ -13,11 +13,12 @@ tmp="./setup/docker/tmp"
 
 mkdir -p ${tmp}
 
-cat ./setup/unix/traccar.xml | awk '/web.path/ && !modif { printf("    <entry key=\"web.debug\">true</entry>\n"); modif=1 } {print}' > ${tmp}/traccar.xml
+cat ./setup/traccar.xml | awk '/web.path/ && !modif { printf("    <entry key=\"web.debug\">true</entry>\n"); modif=1 } {print}' > ${tmp}/traccar.xml
+cp -rf ./setup/default.xml ${tmp}
 cp -rf ./schema ${tmp}/schema
 cp -rf ./target/tracker-server.jar ${tmp}/traccar-server.jar
 cp -rf ./target/lib ${tmp}/lib
-cp -rf ./web ${tmp}/web
+cp -rf ./traccar-web/web ${tmp}/web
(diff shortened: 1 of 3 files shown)


oudenaar:
INFO 11/21/16 1:51 PM: liquibase: Successfully released change log lock
[main] INFO org.eclipse.jetty.util.log - Logging initialized @5239ms
[main] INFO org.eclipse.jetty.server.Server - jetty-9.2.19.v20160908
[main] INFO org.eclipse.jetty.server.handler.ContextHandler - Started o.e.j.s.ServletContextHandler@108a46d6{/api,null,AVAILABLE}
[main] INFO org.eclipse.jetty.server.ServerConnector - Started ServerConnector@4d3ca6c7{HTTP/1.1}{0.0.0.0:8082}
[main] INFO org.eclipse.jetty.server.Server - Started @8225ms

But login screen is hanging in javascript, didn't minimize the javascript stuff, is this needed?

oudenaar:
Got it, http://127.0.0.1:8082/debug.html works, I saw with chrome dev tools, http://127.0.0.1:8082 is looking for minified version of app.js => app-min.js

Sencha is a commercial product?! So can't generate this :(

oudenaar:
little hack app.js  copy to app.min.js

tananaev:
Sencha is dual-licensed under their proprietary license and GPL as well.

The hack is not needed if docker build script worked fine. This line forces Traccar to use debug version:
```xml
<entry key="web.debug">true</entry>
```

You also have to copy templates folder for Excel reports.

setup/docker/build.sh
```diff
@@ -13,12 +13,13 @@ tmp="./setup/docker/tmp"
 
 mkdir -p ${tmp}
 
-cat ./setup/unix/traccar.xml | awk '/web.path/ && !modif { printf("    <entry key=\"web.debug\">true</entry>\n"); modif=1 } {print}' > ${tmp}/traccar.xml
+cat ./setup/traccar.xml | awk '/web.path/ && !modif { printf("    <entry key=\"web.debug\">true</entry>\n"); modif=1 } {print}' > ${tmp}/traccar.xml
```
sunhoww: I was about to make a PR for the same thing when I saw yours.

However, I think in order to properly set _web.debug=true_, you should operate on `./setup/default.xml` instead because you won't be able to find `/web.path/` on `traccar.xml`.

tananaev: We should just add it to the traccar.xml. Changing defaults is not a good idea.

sunhoww: Makes sense. In that case, I suggest changing this line to

`cat ./setup/traccar.xml | awk '/config.default/ && !modif { print;printf("    <entry key=\"web.debug\">true</entry>\n");next; modif=1 } {print}'  > ${tmp}/traccar.xml`

essentially adding the `web.debug` line after the `config.default` line

tananaev: Makes sense to me. @oudenaar, can you please fix the comment in the pull request.

oudenaar:
Ok  made changes according review comments @sunhoww

tananaev:
What about the other comment:
> You also have to copy templates folder for Excel reports.

sunhoww:
Maybe you could also
```
if [ -d ./traccar-web/web ]; then
  cp -rf ./traccar-web/web ${tmp}/web
else
  mkdir ${tmp}/web
fi
```
which I've been using to build without **traccar-web**.

tananaev:
Makes sense as well.

oudenaar:
@tananaev I have added all review comments, could you please merge the stuff

sunhoww:
Great. @oudenaar You haven't included `templates` folder though.

tananaev:
Added templates folder (see 6dc245d2082b7f0a4f440b817ce9c8455f441f93). Can someone please check if everything work now.

sunhoww:
Seems okay. Builds fine.

PR #2583, opened 2016-11-22
Maintenance events

Abyss777:
Here is implementation of "maintenance is needed" events.

Two configuration attributes:
- `device.maintenanceStart` zero if ommited 
- `device.maintenanceInterval` can't be zero
Both in meters

Example:
`device.maintenanceStart = 2000000`
`device.maintenanceInterval = 7500000`

Events will be generated once at the point then `totalDistance` overcome 2000 km, 9500 km, 17000 km, 24500 km and so on.

Also:
- Enabled `statusHandler` by default (or it was disabled intentionally? )
- Added mail templates for new event types

### DIFF
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -24,6 +24,8 @@
     <entry key='event.geofenceHandler'>true</entry>
     <entry key='event.alertHandler'>true</entry>
     <entry key='event.ignitionHandler'>true</entry>
+    <entry key='event.statusHandler'>true</entry>
+    <entry key='event.maintenanceHandler'>true</entry>
 
     <!-- DATABASE CONFIG -->
 
--- a/src/org/traccar/events/MaintenanceEventHandler.java
+++ b/src/org/traccar/events/MaintenanceEventHandler.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2016 Anton Tananaev (anton@traccar.org)
+ * Copyright 2016 Andrey Kunitsyn (andrey@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.events;
+
+import java.util.ArrayList;
+import java.util.Collection;
+
+import org.traccar.BaseEventHandler;
+import org.traccar.Context;
+import org.traccar.model.Device;
+import org.traccar.model.Event;
+import org.traccar.model.Position;
+
+public class MaintenanceEventHandler extends BaseEventHandler {
+
+    public static final String ATTRIBUTE_MAINTENANCE_START = "device.maintenanceStart";
+    public static final String ATTRIBUTE_MAINTENANCE_INTERVAL = "device.maintenanceInterval";
+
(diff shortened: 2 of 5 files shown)


setup/default.xml
```diff
@@ -24,6 +24,8 @@
     <entry key='event.geofenceHandler'>true</entry>
     <entry key='event.alertHandler'>true</entry>
     <entry key='event.ignitionHandler'>true</entry>
+    <entry key='event.statusHandler'>true</entry>
```
tananaev: Did we miss this?

src/org/traccar/events/MaintenanceEventHandler.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    protected Collection<Event> analyzePosition(Position position) {
+        Device device = Context.getIdentityManager().getDeviceById(position.getDeviceId());
+        if (device == null || !Context.getIdentityManager().isLatestPosition(position)) {
+            return null;
+        }
+
+        double maintenanceInterval = Context.getDeviceManager()
+                .lookupServerDouble(device.getId(), ATTRIBUTE_MAINTENANCE_INTERVAL, 0);
```
tananaev: Not really related to this future, but maybe we should have just one set of "lookup" methods? And always pass parameter to it? Do you remember why we decided to have two sets?

src/org/traccar/events/MaintenanceEventHandler.java
```diff
@@ -0,0 +1,70 @@
...
+        double maintenanceStart = Context.getDeviceManager()
+                .lookupServerDouble(device.getId(), ATTRIBUTE_MAINTENANCE_START, 0);
+
+        Collection<Event> events = new ArrayList<>();
+        double oldTotalDistance = 0.0;
+        double newTotalDistance = 0.0;
+
+        Position lastPosition = Context.getIdentityManager().getLastPosition(position.getDeviceId());
+        if (lastPosition != null && lastPosition.getAttributes().containsKey(Position.KEY_TOTAL_DISTANCE)) {
+            oldTotalDistance = ((Number) lastPosition.getAttributes().get(Position.KEY_TOTAL_DISTANCE)).doubleValue();
```
tananaev: We use it quite a lot, so maybe we should add some helper methods to Position model to get ints, string, double etc?

src/org/traccar/model/Event.java
```diff
@@ -56,6 +56,8 @@ public Event() {
     public static final String TYPE_IGNITION_ON = "ignitionOn";
     public static final String TYPE_IGNITION_OFF = "ignitionOff";
 
+    public static final String TYPE_MAINTENANCE_NEED = "maintenanceNeed";
```
tananaev: I think we should just name it "maintenance" or alternatively "maintenance needed" (note -ed).

tananaev:
Another tiny comment. We can probably rename "device.maintenanceStart" to "maintenance.start". I think it makes more sense.

src/org/traccar/events/MaintenanceEventHandler.java
Abyss777: It was here https://github.com/tananaev/traccar/pull/2348#discussion_r79599245
and here https://github.com/tananaev/traccar/pull/2348#discussion_r79766522

src/org/traccar/events/MaintenanceEventHandler.java
Abyss777: Yes, it must be useful. But It is better to make new pull request, and replace it everywhere.

tananaev: Sure.

src/org/traccar/events/MaintenanceEventHandler.java
tananaev: Can we please change it to be one set of methods with boolean variable indicating whether to lookup config file or server model. Let's do it in another pull request, I'll merge this.

tananaev:
Merged, thanks.

PR #2591, opened 2016-11-22
Update NotificationFormatter.java

Turbovix:
The URL that refers to the OpenStreetMaps website with the notifications received from the TRACCAR server now uses HTTPS, that is, a secure connection.


tananaev:
Merged, thanks.

PR #2592, opened 2016-11-23
Fix excel export if timezone of client and server are different

Abyss777:
Have to use second parser because `Ext.Date.format(from, 'c')` format without milliseconds
Parse parameters inside `getExcel()` functions.

Period cell become: `${from.toString("YYYY.MM.dd HH:mm:ss")+" - "+to.toString("YYYY.MM.dd HH:mm:ss")}`

Cells with times become:
`${new("org.joda.time.DateTime", event.serverTime, timezone).toString("YYYY.MM.dd HH:mm:ss")}`


tananaev:
Why does it not include milliseconds in the time?

Abyss777:
Good question. Frankly, I do not know.
http://docs.sencha.com/extjs/6.0.1/classic/Ext.Date.html it refers to http://fi2.php.net/manual/en/function.date.php that do not contains milliseconds in example.

tananaev:
Can we use `dateTimeParser()` instead of specific formatter? It should be able to handle any ISO format.

Abyss777:
I'll check.

Abyss777:
Used one parser and also reverted parsing to ReportResource.
Result changes should look more consistence.

Not sure, are new indentations in `ReportResource` right?

tananaev:
Merged, looks good to me.

PR #2596, opened 2016-11-24
Combine lookupAttribute helpers

Abyss777:
subj

### DIFF
--- a/src/org/traccar/CopyAttributesHandler.java
+++ b/src/org/traccar/CopyAttributesHandler.java
@@ -29,8 +29,8 @@ public class CopyAttributesHandler extends BaseDataHandler {
 
     @Override
     protected Position handlePosition(Position position) {
-        String attributesString = Context.getDeviceManager().lookupConfigString(position.getDeviceId(),
-                "processing.copyAttributes", null);
+        String attributesString = Context.getDeviceManager().lookupAttributeString(position.getDeviceId(),
+                "processing.copyAttributes", null, false);
         Position last = getLastPosition(position.getDeviceId());
         if (attributesString != null && last != null) {
             for (String attribute : attributesString.split(" ")) {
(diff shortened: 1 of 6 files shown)


src/org/traccar/database/DeviceManager.java
```diff
@@ -317,80 +317,43 @@ public void removeGroup(long groupId) throws SQLException {
         groupsById.remove(groupId);
     }
 
-    public boolean lookupServerBoolean(long deviceId, String attributeName, boolean defaultValue) {
-        String result = lookupAttribute(deviceId, attributeName, true);
+    public boolean lookupAttributeBoolean(long deviceId, String attributeName, boolean defaultValue,
+            boolean lookupServer) {
```
tananaev: I think we should change it to `lookupConfig` because server meaning is not that obvious.

src/org/traccar/CopyAttributesHandler.java
```diff
@@ -29,8 +29,8 @@ private Position getLastPosition(long deviceId) {
 
     @Override
     protected Position handlePosition(Position position) {
-        String attributesString = Context.getDeviceManager().lookupConfigString(position.getDeviceId(),
-                "processing.copyAttributes", null);
+        String attributesString = Context.getDeviceManager().lookupAttributeString(position.getDeviceId(),
```
tananaev: I think it's more readable if it doesn't fit on one line, move all parameters to the next one.

src/org/traccar/database/DeviceManager.java
```diff
@@ -317,80 +317,43 @@ public void removeGroup(long groupId) throws SQLException {
...
+        String result = lookupAttribute(deviceId, attributeName, lookupServer);
         if (result != null) {
             return Boolean.parseBoolean(result);
         }
         return defaultValue;
     }
 
-    public String lookupServerString(long deviceId, String attributeName, String defaultValue) {
-        String result = lookupAttribute(deviceId, attributeName, true);
+    public String lookupAttributeString(long deviceId, String attributeName, String defaultValue,
```
tananaev: Same here, it's better to move all parameters on the next line if it doesn't fit.

Abyss777:
Inverted variable to `lookupConfig` and fixed style

tananaev:
Perfect, thanks.

PR #2608, opened 2016-11-26
Implemented polyline geofence

### LINKED ISSUE #2600: [Feature Request] Polyline for Geofencing
Hi,

Please add Polyline for Geofencing.

Since currently there is no feature to draw a vehicle route on map, except using Polygon or Circle.

Thanks.

Abyss777:
fix  #2600 

Polyline Geofence consider to contains point if distance between point and at least one interval less than predefined

Default value for `geofence.polylineDistance` is 25 meters. It is with wide margins.

Also moved `Coordinate` class to parent for use it in both polyline and polygon.

### DIFF
--- a/src/org/traccar/geofence/GeofenceGeometry.java
+++ b/src/org/traccar/geofence/GeofenceGeometry.java
@@ -25,4 +25,33 @@ public abstract class GeofenceGeometry {
 
     public abstract void fromWkt(String wkt) throws ParseException;
 
+    public static class Coordinate {
+
+        public static final double DEGREE360 = 360;
+
+        private double lat;
+        private double lon;
+
+        public double getLat() {
+            return lat;
+        }
+
+        public void setLat(double lat) {
+            this.lat = lat;
+        }
+
+        public double getLon() {
+            return lon;
+        }
+
+        // Need not to confuse algorithm by the abrupt reset of longitude
+        public double getLon360() {
+            return lon + DEGREE360;
+        }
+
+        public void setLon(double lon) {
+            this.lon = lon;
+        }
+    }
+
 }
--- a/src/org/traccar/geofence/GeofenceGeometry.java
+++ b/src/org/traccar/geofence/GeofenceGeometry.java
@@ -27,8 +27,6 @@ public abstract class GeofenceGeometry {
 
     public static class Coordinate {
 
-        public static final double DEGREE360 = 360;
-
         private double lat;
         private double lon;
 
@@ -42,11 +40,6 @@ public abstract class GeofenceGeometry {
 
         public double getLon() {
             return lon;
-        }
-
-        // Need not to confuse algorithm by the abrupt reset of longitude
-        public double getLon360() {
-            return lon + DEGREE360;
         }
 
         public void setLon(double lon) {
--- a/src/org/traccar/helper/DistanceCalculator.java
+++ b/src/org/traccar/helper/DistanceCalculator.java
@@ -34,7 +34,7 @@ public final class DistanceCalculator {
         return d * 1000;
     }
 
-    public static double distanceToInterval(
+    public static double distanceToLine(
             double pointLat, double pointLon, double lat1, double lon1, double lat2, double lon2) {
         double d0 = distance(pointLat, pointLon, lat1, lon1);
         double d1 = distance(lat1, lon1, lat2, lon2);
--- a/src/org/traccar/helper/DistanceCalculator.java
+++ b/src/org/traccar/helper/DistanceCalculator.java
@@ -1,5 +1,6 @@
 /*
- * Copyright 2014 Anton Tananaev (anton@traccar.org)
+ * Copyright 2014 - 2016 Anton Tananaev (anton@traccar.org)
+ * Copyright 2016 Andrey Kunitsyn (andrey@traccar.org)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/org/traccar/helper/DistanceCalculator.java
+++ b/src/org/traccar/helper/DistanceCalculator.java
@@ -33,4 +33,20 @@ public final class DistanceCalculator {
         return d * 1000;
     }
 
+    public static double distanceToInterval(
+            double pointLat, double pointLon, double lat1, double lon1, double lat2, double lon2) {
+        double d0 = distance(pointLat, pointLon, lat1, lon1);
+        double d1 = distance(lat1, lon1, lat2, lon2);
+        double d2 = distance(lat2, lon2, pointLat, pointLon);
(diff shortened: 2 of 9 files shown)


src/org/traccar/helper/DistanceCalculator.java
```diff
@@ -33,4 +34,20 @@ public static double distance(double lat1, double lon1, double lat2, double lon2
         return d * 1000;
     }
 
+    public static double distanceToInterval(
```
tananaev: Maybe rename it to `distanceToLine`?

test/org/traccar/geofence/GeofencePolylineTest.java
```diff
@@ -0,0 +1,54 @@
...
+        Assert.assertTrue(geofenceGeometry.containsPoint(56.83801, 60.59748));
+
+        ((GeofencePolyline) geofenceGeometry).setDistance(15);
+        Assert.assertTrue(!geofenceGeometry.containsPoint(56.83801, 60.59748));
+
+        } catch (ParseException e){
+            Assert.assertTrue("ParseExceprion: " + e.getMessage(), true);
+        }
+    }
+
```
tananaev: Extra space

test/org/traccar/geofence/GeofencePolylineTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Test
+    public void testContainsPolyline3Intervals() {
+        String test = "LINESTRING (56.836 60.6126, 56.8393 60.6114, 56.83887 60.60811, 56.83782 60.5988)";
+
+        try {
+        GeofenceGeometry geofenceGeometry = new GeofencePolyline(test, 15);
```
tananaev: Please fix formatting in this method.

test/org/traccar/helper/DistanceCalculatorTest.java
```diff
@@ -10,5 +10,14 @@ public void testDistance() {
         Assert.assertEquals(
                 DistanceCalculator.distance(0.0, 0.0, 0.05, 0.05), 7863.0, 10.0);
     }
+    
+    @Test
+    public void testDistanceToInterval() {
```
tananaev: Don't forget to rename word "interval" in tests as well.

test/org/traccar/geofence/GeofencePolylineTest.java
tananaev: Formatting is this file is completely wrong, not only this method.

test/org/traccar/geofence/GeofencePolylineTest.java
```diff
@@ -0,0 +1,54 @@
...
+public class GeofencePolylineTest {
+
+    @Test
+    public void testPolylineWKT() {
+        String test = "LINESTRING (55.75474 37.61823, 55.75513 37.61888, 55.7535 37.6222, 55.75315 37.62165)";
+        GeofenceGeometry geofenceGeometry = new GeofencePolyline();
+        try {
+        geofenceGeometry.fromWkt(test);
+        } catch (ParseException e){
+            Assert.assertTrue("ParseExceprion: " + e.getMessage(), true);
```
tananaev: What are you trying to do here?

src/org/traccar/geofence/GeofenceGeometry.java
```diff
@@ -25,4 +25,33 @@
...
+            this.lat = lat;
+        }
+
+        public double getLon() {
+            return lon;
+        }
+
+        // Need not to confuse algorithm by the abrupt reset of longitude
+        public double getLon360() {
+            return lon + DEGREE360;
```
tananaev: Can you remind me why we are using this?

test/org/traccar/geofence/GeofencePolylineTest.java
Abyss777: I'm really sorry. Was so deep in testing and rely on checkstyle. Will fix it.

src/org/traccar/geofence/GeofenceGeometry.java
Abyss777: It is In case geofence is rather big and some points have +180 longitude and other -179.
Such jump will break the algorithms.

Abyss777:
Fixed all comments.

src/org/traccar/geofence/GeofenceGeometry.java
tananaev: So, how do we handle it? It seems like you always add 360 degrees. Won't it result in the same difference?

test/org/traccar/geofence/GeofencePolylineTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class GeofencePolylineTest {
+
+    @Test
+    public void testPolylineWKT() {
+        String test = "LINESTRING (55.75474 37.61823, 55.75513 37.61888, 55.7535 37.6222, 55.75315 37.62165)";
+        GeofenceGeometry geofenceGeometry = new GeofencePolyline();
+        try {
+            geofenceGeometry.fromWkt(test);
+        } catch (ParseException e){
```
tananaev: Space is missing before curly brace. Same for other catches.

test/org/traccar/geofence/GeofencePolylineTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.text.ParseException;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class GeofencePolylineTest {
+
+    @Test
+    public void testPolylineWKT() {
```
tananaev: Why is WKT uppercase here?

test/org/traccar/geofence/GeofencePolylineTest.java
```diff
@@ -0,0 +1,47 @@
...
+        String test = "LINESTRING (55.75474 37.61823, 55.75513 37.61888, 55.7535 37.6222, 55.75315 37.62165)";
+        GeofenceGeometry geofenceGeometry = new GeofencePolyline();
+        try {
+            geofenceGeometry.fromWkt(test);
+        } catch (ParseException e){
+            Assert.assertTrue("ParseExceprion: " + e.getMessage(), true);
```
tananaev: This still doesn't make much sense. It's always true. What's the point?

Abyss777:
You are absolutely right about all.
Try/catch is not necessary.

about +-180 longitude...
It is actual only for polygons, for polyline I used formula with distances that works correctly
I could not find math way of longitude normalization. (one work for 180 and broke near 0).
Just simple solution:
Pre find out if polygon has longitudes >90 and < -90 at the same time, and add 360 to negative longitudes.

Also added much more tests.

tananaev:
Merging, I think it looks good now. Thanks.

PR #2617, opened 2016-11-30
Add API to get positions by Id

Abyss777:
It is needed to show events on map

- Added request to get position by Id
- Added parameter to API

The only note is HTTP request length. It is depending on browser 300 - 1500 positions at once.
But GET is more natural for REST API.

### DIFF
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -171,6 +171,10 @@
         SELECT * FROM positions WHERE deviceId = :deviceId AND fixTime BETWEEN :from AND :to ORDER BY fixTime
     </entry>
 
+    <entry key='database.selectPosition'>
+        SELECT * FROM positions WHERE id = :id
+    </entry>
+
     <entry key='database.insertPosition'>
         INSERT INTO positions (deviceId, protocol, serverTime, deviceTime, fixTime, valid, latitude, longitude, altitude, speed, course, address, attributes)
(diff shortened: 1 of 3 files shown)


tananaev:
Looks good. Can you provide an example of the request URL here?

Abyss777:
It is how ExtJS store will build it:
`http://localhost:8082/api/positions?positionId=15661&positionId=15662&positionId=15663&positionId=15664&positionId=15665&positionId=15667&positionId=15673&positionId=15677&positionId=15680&page=1&start=0&limit=25`

tananaev:
Doesn't look very nice, but I guess we can keep it if it works fine on both ends.

Abyss777:
reports request with multiple devices looks the same.

At first I wanted to use POST, but did not find way post bare array `[1,2,3]` on ExtJS side or extract array from `{positionId:[1,2,3]}` on server side without implementing proxy objects.

tananaev:
I guess there no technical issues with it. It just looks a bit weird, especially because there can be many position ids. Anyway, let's keep it this way.

tananaev:
Just thought about one thing. We can just use "id" instead of "positionId" because we already know it's Position resource. This will shorten the request quite a bit.

Abyss777:
Good point! I will do a pull request.

PR #2618, opened 2016-11-30
Use "id" parameter instead of "positionId"

Abyss777:
It will short request
`http://localhost:8082/api/positions?id=15661&id=15661&id=15661&id=15662&id=15662&id=15663&id=15663&id=15663&id=15664&id=15664&id=15665&id=15667&id=15673&id=15677&id=15680&id=15680&page=1&start=0&limit=25`


tananaev:
Merged, thanks.

PR #2619, opened 2016-11-30
Implement API to test mail

Abyss777:
- added API point `api/users/notifications/testmail` that sends "unknown" event on GET request
- make `sendMailSync` public
- moved try/catch to runable

I decided to throw exceptions in resource. It will help debug without looking in log file.

### DIFF
--- a/src/org/traccar/api/resource/NotificationResource.java
+++ b/src/org/traccar/api/resource/NotificationResource.java
@@ -60,10 +60,10 @@ public class NotificationResource extends BaseResource {
         return Response.ok(entity).build();
     }
 
-    @Path("testmail")
-    @GET
+    @Path("test")
+    @POST
     public Response testMail() throws MessagingException {
-        NotificationMail.sendMailSync(getUserId(), new Event("unknown", 0), null);
+        NotificationMail.sendMailSync(getUserId(), new Event("test", 0), null);
         return Response.noContent().build();
     }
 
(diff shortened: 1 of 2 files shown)


src/org/traccar/api/resource/NotificationResource.java
```diff
@@ -57,4 +60,11 @@ public Response update(Notification entity) throws SQLException {
         return Response.ok(entity).build();
     }
 
+    @Path("testmail")
```
tananaev: I think we should use "POST" method, and we can probably name it just "test".

src/org/traccar/api/resource/NotificationResource.java
```diff
@@ -57,4 +60,11 @@ public Response update(Notification entity) throws SQLException {
...
+    @GET
+    public Response testMail() throws MessagingException {
+        NotificationMail.sendMailSync(getUserId(), new Event("unknown", 0), null);
```
tananaev: Maybe use "test" instead of "unknown"?

Abyss777: Frankly, it does not really mater. Just should disagree with existent types.
The email will come with "Unknown type" text.

Abyss777:
Renamed and used POST

tananaev:
Merged, thanks.

Abyss777:
Have no idea where to put button in web interface. Do you have?

tananaev:
I was thinking we can put it in the user dialog.

PR #2625, opened 2016-12-01
Use queryParam instead of pathParam

Abyss777:
It will fit more to REST

https://github.com/tananaev/traccar-web/pull/347 relies on this

### DIFF
--- a/src/org/traccar/api/resource/EventResource.java
+++ b/src/org/traccar/api/resource/EventResource.java
@@ -5,8 +5,8 @@ import java.sql.SQLException;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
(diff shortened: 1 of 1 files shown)


tananaev:
I don't agree with this. Path parameter fits REST much better. We use query param for positions only because we want to request multiple positions at the same time. Looks like we don't need it in this case.

Abyss777:
OK. I was confused that we retrieve objects with GET without id in path everywhere.

tananaev:
The idea is that GET without id gives a collection of objects. GET with id gives a single object. Ideally for positions we should be using path as well, but I guess to optimize it a little bit we decided to request all positions in one request, so we use query parameter. So, it's more an exception than a rule.

Abyss777:
I see. And yes, at first, I've tried do get every position in separate request. It was very inconvenient.

PR #2638, opened 2016-12-04
Use model instead of attributes to store Notification options

Abyss777:
- Extended model
- Copy values with help of DB changelog
- Remove notification only if everything is false and attributes is empty.
- Merge all types of notifications with actual user notifications on server side. 

Also tested on MSSQL, works fine.

### DIFF
--- a/src/org/traccar/api/resource/NotificationResource.java
+++ b/src/org/traccar/api/resource/NotificationResource.java
@@ -49,7 +49,7 @@ public class NotificationResource extends BaseResource {
             userId = getUserId();
         }
         Context.getPermissionsManager().checkUser(getUserId(), userId);
-        return Context.getNotificationManager().getUserNotifications(userId);
+        return Context.getNotificationManager().getAllUserNotifications(userId);
     }
 
     @POST
--- a/src/org/traccar/database/NotificationManager.java
+++ b/src/org/traccar/database/NotificationManager.java
@@ -59,10 +59,10 @@ public class NotificationManager {
                     && Context.getGeofenceManager().checkGeofence(userId, event.getGeofenceId())) {
                 Notification notification = getUserNotificationByType(userId, event.getType());
                 if (notification != null) {
-                    if (notification.getAttributes().containsKey("web")) {
+                    if (notification.getWeb()) {
                         Context.getConnectionManager().updateEvent(userId, event, position);
                     }
-                    if (notification.getAttributes().containsKey("mail")) {
+                    if (notification.getMail()) {
                         NotificationMail.sendMailAsync(userId, event, position);
                     }
                 }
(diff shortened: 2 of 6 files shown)


src/org/traccar/api/resource/NotificationResource.java
```diff
@@ -49,7 +49,7 @@
             userId = getUserId();
         }
         Context.getPermissionsManager().checkUser(getUserId(), userId);
-        return Context.getNotificationManager().getUserNotifications(userId);
+        return Context.getNotificationManager().getAllUserNotifications(userId);
```
tananaev: Do we still need "all" parameter in this method? Maybe we can also remove `getAllNotifications` or make it private?

src/org/traccar/database/NotificationManager.java
```diff
@@ -159,7 +164,7 @@ public void updateNotification(Notification notification) {
             } else {
                 notification.setId(cachedNotification.getId());
             }
-        } else if (!notification.getAttributes().isEmpty()) {
+        } else if (notification.getWeb() || notification.getMail() || !notification.getAttributes().isEmpty()) {
```
tananaev: Do we need to check for attributes here?

src/org/traccar/model/Notification.java
```diff
@@ -37,4 +37,25 @@ public void setType(String type) {
...
+    private boolean mail;
+
+    public boolean getMail() {
+        return mail;
+    }
+
+    public void setMail(boolean mail) {
+        this.mail = mail;
+    }
+
```
tananaev: Too make spaces. It should be just one.

src/org/traccar/database/NotificationManager.java
```diff
@@ -130,8 +130,11 @@ public Notification getUserNotificationByType(long userId, String type) {
     public void updateNotification(Notification notification) {
         Notification cachedNotification = getUserNotificationByType(notification.getUserId(), notification.getType());
         if (cachedNotification != null) {
-            if (!cachedNotification.getAttributes().equals(notification.getAttributes())) {
-                if (notification.getAttributes().isEmpty()) {
+            if (cachedNotification.getWeb() != notification.getWeb()
+                    || cachedNotification.getMail() != notification.getMail()
+                    || !cachedNotification.getAttributes().equals(notification.getAttributes())) {
```
tananaev: Do we need to check attributes here?

tananaev:
Have you tested the upgrade script?

src/org/traccar/api/resource/NotificationResource.java
Abyss777: We can't remove it. I remember, we use it to generate list of event types for `ReportConfig` dialog

src/org/traccar/database/NotificationManager.java
Abyss777: Let discuss it.
Actually we stop using attributes.
We need just decide if we need store notification like?
````
web=false
mail=false
attributes={ something }
````
 It can only happen if somebody use attributes for own tasks.

Abyss777:
Yes, I have tested it on MySQL and MS SQL.

src/org/traccar/api/resource/NotificationResource.java
tananaev: We can just use the list for the current user, but I guess we can keep it as well.

src/org/traccar/database/NotificationManager.java
tananaev: What is this code used for? We might use attributes for something later, but for now we are not using them.

Abyss777:
Removed attributes checks and extra spaces.

tananaev:
Merged, thanks.

PR #2650, opened 2016-12-05
Add configurable default values for group & category on on-the-fly created devices

duke2906:
Hello,

in my opinion a configurable default group and a default category would be usefull when using the optinal functionality to register devices automatically on incoming positions. These are the adjustments i would suggest to implement this enhancement.


tananaev:
Please fix the errors reported by the build system. Next time you should probably run all the checks locally before pushing any commits.

Abyss777:
Hello,

I think you also need update `groupsByName`  in add/update/removeGroup functions

src/org/traccar/BaseProtocolDecoder.java
```diff
@@ -34,8 +35,17 @@
 
     public long addUnknownDevice(String uniqueId) {
         Device device = new Device();
+        String defaultGroupName = Context.getConfig().getString("database.registerUnknown.defaultGroup", "NOTDEFINED");
```
tananaev: I don't think we need a default value here.

src/org/traccar/BaseProtocolDecoder.java
```diff
@@ -34,8 +35,17 @@
...
         device.setName(uniqueId);
         device.setUniqueId(uniqueId);
+        device.setCategory(Context.getConfig().getString("database.registerUnknown.defaultCategory", "default"));
```
tananaev: Same here. I don't think we need default value.

duke2906:
In removeGroup I implemented a "check" on the old group name to prevent the HashMap from getting bigger and bigger cause of old or doubled group-references when updating this name. Is that correct?

tananaev:
I just realized that we can have group duplicates. I think it's better to use id.

duke2906:
Oh, sorry. I just checked it out...ur right, I am able to create another Group with the same group name. Well ok, than I'll use the id and remove the getGroupByName. Unfortunately the only way to get the GroupID is to query the database, isn't it? Well, that shouldn't be a problem cause this is supposed to be set up once by the administrator, which ofcourse has access to the database...

tananaev:
Admin can always use API to get groups directly, so it shouldn't be a big problem. Plus presumably it needs to be set only once.

src/org/traccar/BaseProtocolDecoder.java
```diff
@@ -34,8 +34,14 @@
 
     public long addUnknownDevice(String uniqueId) {
         Device device = new Device();
+        long defaultGroupId = Context.getConfig().getLong("database.registerUnknown.defaultGroupId");
```
tananaev: Please move this declaration down just before you use the variable.

duke2906:
Whole declaration, or split declaration and initialisation and move
initialisation down right before its used?

Anton Tananaev <notifications@github.com> schrieb am Do., 8. Dez. 2016,
10:13:

tananaev:
Declaration. There is no point in declaring it earlier. It's not JavaScript.

tananaev:
Merged, thanks.

PR #2653, opened 2016-12-06
Add support for "mail.smtp.ssl.protocols" parameter

### LINKED ISSUE #2651: Notification emails cannot be sent to servers requiring TLSv1.2
As said in the title :)
Notification emails cannot be sent to servers requiring TLSv1.2
It seems mail.smtps.ssl.protocols has to be set to "TLSv1 TLSv1.1 TLSv1.2" for this to work, otherwise java uses TLSv1.0.

Abyss777:
fix #2651 
I don't think we should investigate why it does not work by default, may this parameter will help somebody else.


tananaev:
Merged. It won't hurt to have extra optional parameters.

PR #2655, opened 2016-12-07
Templates for mail notifications

Abyss777:
I used Velocity Engine.

Initiated engine in `Context` because we have to disable default velocity logger. Also set templates path.
It could be done in `formatMessage` function, but better do it once. I think `Context` is right place.

I think it is not good idea to have separate templates for body and subject. I made small trick.
Set variable `$Subject` in template (first line in every .vm file) and then retrieve it after merging.

[VTL](https://velocity.apache.org/engine/1.7/vtl-reference.html) is not so powerful as [JEXL](http://commons.apache.org/proper/commons-jexl/reference/syntax.html) but I think it will be able do all necessary things.

Templates leaved as was except links.

### DIFF
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -17,6 +17,9 @@ package org.traccar;
 
 import com.ning.http.client.AsyncHttpClient;
 
+import java.util.Properties;
+
+import org.apache.velocity.app.VelocityEngine;
 import org.traccar.database.AliasesManager;
 import org.traccar.database.ConnectionManager;
 import org.traccar.database.DataManager;
@@ -123,6 +126,12 @@ public final class Context {
 
     public static NotificationManager getNotificationManager() {
         return notificationManager;
+    }
+
+    private static VelocityEngine velocityEngine;
+
+    public static VelocityEngine getVelocityEngine() {
+        return velocityEngine;
     }
 
     private static final AsyncHttpClient ASYNC_HTTP_CLIENT = new AsyncHttpClient();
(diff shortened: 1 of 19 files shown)


src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -15,223 +15,65 @@
...
-                break;
+        Template template = null;
+        try {
+            template = Context.getVelocityEngine().getTemplate(event.getType() + ".vm");
+        } catch (ResourceNotFoundException error) {
+            Log.warning(error);
         }
-        return result;
+        if (template == null) {
+            template = Context.getVelocityEngine().getTemplate("unknown.vm");
```
tananaev: I think we can just move this to catch. Or is there a chance that `getTemplate` returns null?

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -15,223 +15,65 @@
...
+import org.traccar.helper.Log;
 import org.traccar.model.Device;
 import org.traccar.model.Event;
 import org.traccar.model.Position;
+import org.traccar.reports.ReportUtils;
 
 public final class NotificationFormatter {
 
-    private NotificationFormatter() {
+    public static class MailMessage {
```
tananaev: I think it's better to extract this class into a separate file.

src/org/traccar/Context.java
```diff
@@ -242,6 +251,13 @@ public static void init(String[] arguments) throws Exception {
 
         if (config.getBoolean("event.enable")) {
             notificationManager = new NotificationManager(dataManager);
+            Properties velocityProperties = new Properties();
+            velocityProperties.setProperty("file.resource.loader.path",
+                    Context.getConfig().getString("mail.templatesPath", "templates/mail") + "/");
+            velocityProperties.setProperty("runtime.log.logsystem.class",
+                    "org.apache.velocity.runtime.log.NullLogChute");
```
tananaev: What does this line mean?

tananaev: I agree that we don't want more files.

Abyss777:
Extracted class and moved getting "unknown.vm" to `catch`

tananaev:

PR #2668, opened 2016-12-09
Added links to server in mail templates

Abyss777:
- Used velocity engine properties to pass `web.url` to formatter. Checked documentation, it does not intersect with velocity properties.

About templates...
I called line "Link:" if event does not have assigned position, and "Point:" if has.

### DIFF
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -21,6 +21,7 @@ import java.net.InetAddress;
 import java.util.Properties;
 
 import org.apache.velocity.app.VelocityEngine;
+import org.eclipse.jetty.util.URIUtil;
 import org.traccar.database.AliasesManager;
 import org.traccar.database.ConnectionManager;
 import org.traccar.database.DataManager;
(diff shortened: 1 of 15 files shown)


src/org/traccar/Context.java
```diff
@@ -256,6 +257,13 @@ public static void init(String[] arguments) throws Exception {
                     Context.getConfig().getString("mail.templatesPath", "templates/mail") + "/");
             velocityProperties.setProperty("runtime.log.logsystem.class",
                     "org.apache.velocity.runtime.log.NullLogChute");
+
+            String address = config.getString("web.address", InetAddress.getLocalHost().getHostAddress());
+            String port = config.getString("web.port", "8082");
+            String webUrl = "http://" + address + ":" + port + "/";
```
tananaev: Probably we can use some URL builder to construct this. It will handle automatically cases like when port is 80, we don't need to add it.

templates/mail/geofenceExit.vm
```diff
@@ -4,7 +4,7 @@
 <body>
 Device: $device.name<br>
 Has exited geofence: $geofence.name<br>
-Point: <a href="https://www.openstreetmap.org/?mlat=$position.latitude&mlon=$position.longitude#map=16/$position.latitude/$position.longitude">#if($position.address)$position.address#{else}$position.latitude°, $position.longitude°#end</a><br>
 Time: $event.serverTime<br>
+Point: <a href="$webUrl?eventId=$event.id">#if($position.address)$position.address#{else}$position.latitude°, $position.longitude°#end</a><br>
```
tananaev: The syntax is a bit weird. Why "else" is in curly braces and "if" is not? Is that supposed to be like that?

tananaev: Can it be used for "if" as well to make it consistent? I think it would look better.

Abyss777: OK. Sure.

Abyss777:
Used URL builder from jetty and Curly braces for if/else/end.

tananaev:
Merged, thanks.

PR #2682, opened 2016-12-13
Fix Meiligao Protocol to work with tk228

joseant:
The tk228 send 3 lines of data, with this fix, all the data is saved in the database.


src/org/traccar/protocol/MeiligaoProtocolDecoder.java
```diff
@@ -87,17 +87,27 @@ public MeiligaoProtocolDecoder(MeiligaoProtocol protocol) {
             .number("(d+.d+),")                  // throttle
             .number("(d+.d+),")                  // engine load
             .number("(-?d+),")                   // coolant temp
-            .number("d+.d+,")                    // instantaneous fuel
+            .number("(d+.d+),")                    // instantaneous fuel
```
tananaev: Please make sure all pattern comments are aligned.

src/org/traccar/protocol/MeiligaoProtocolDecoder.java
```diff
@@ -87,17 +87,27 @@ public MeiligaoProtocolDecoder(MeiligaoProtocol protocol) {
...
+            .number("(d+.d+),")                  // Total driving time
+            .number("(d+.d+),")                  // Total idling time
+            .number("(d+),")                  // Average hot start time
+            .number("(d+),")                  // Average speed
+            .number("(d+),")                  // history hightest speed
+            .number("(d+),")                  // history hightest rpm
+            .number("(d+),")                  // total harsh acceleration
+            .number("(d+)")                  // total harsh break n0
             .compile();
-
```
tananaev: Add proper spacing between methods and definitions.

src/org/traccar/protocol/MeiligaoProtocolDecoder.java
```diff
@@ -284,11 +295,33 @@ private Position decodeObd(Position position, String sentence) {
...
+        position.set(Position.KEY_ODOMETER, parser.nextDouble());
         position.set("singleFuelConsumption", parser.nextDouble());
         position.set("totalFuelConsumption", parser.nextDouble());
+        position.set("currentError", parser.nextInt());
+        position.set("harshAcelerationNo", parser.nextInt());
+       position.set("harshBreakerNo", parser.nextInt());
+        return position;
+    }
+    private Position decodeObdA(Position position, String sentence) {
+            Parser parser = new Parser(PATTERN_OBDA, sentence);
```
tananaev: Fix indentations.

src/org/traccar/protocol/MeiligaoProtocolDecoder.java
```diff
@@ -284,11 +295,33 @@ private Position decodeObd(Position position, String sentence) {
...
 
+         position.set("totalIgnitionNo", parser.nextInt());
+        position.set("totalDrivingTime", parser.nextDouble());
+        position.set("totalIdlingTime", parser.nextDouble());
+        position.set("averageHotStartTime", parser.nextInt());
+        position.set("averageSpeed", parser.nextInt());
+        position.set("historyHightestSpeed", parser.nextInt());
+        position.set("historyHightestRPM", parser.nextInt());
+        position.set("totalHarshAccerleration", parser.nextInt());
+        position.set("totalHarsshBrake", parser.nextInt());
```
tananaev: Typo here.

test/org/traccar/protocol/MeiligaoProtocolDecoderTest.java
```diff
@@ -17,7 +17,7 @@ public void testDecode() throws Exception {
                 "2424007d0000000000000099553231303333302e3030302c562c343533342e333832342c532c30373230332e303630302c572c302e30302c302c3231313031362c2c2a31327c302e307c3332397c323030307c303030452c303030437c303244413030303145413634393541417c31307c30303030303030306e540d0a"));
 
         verifyAttributes(decoder, binary(
-                "4040005066104020094432990131302E312C302C3135362C302E30302C31392E36312C2D33342C33342E32362C32312E38332C372E39312C313033332C322E36392C362E35352C302C302C309DBF0D0A"));
+                "242400736610402421174399553130353033342e3937382c412c333933352e333638392c4e2c30303233382e313638342c452c303034382c3034322c3038313231362c2c2a31437c31312e357c3139347c313030317c303341362c303030307c30303130343030307c3030303030303cd00d0a2424004e66104024211743990131342e312c323638372c39302c32312e35372c342e37312c38352c372e31302c382e31362c342e32372c3130342c302e33342c392e33342c302c312c30b7160d0a2424003266104024211743990232352c322e34302c302e37392c32322c34332c3131392c333735362c37352c3132e4c90d0a"));
```
tananaev: Don't remove test cases. Add a new one instead.

joseant:
I have fix the style Issues.

The test, I have to change the verifyAttributes data, because, I think that data was wrong, if you chek the MeiligaoProtocolDecoderTest.java all the data begin for 242400 I have change and past the test.

tananaev:
Please don't change any test cases. Most of the data is collected directly from real devices, so it can't be "wrong".

1. Are you absolutely sure that data like `KEY_OBD_TOTAL_HARSH_BRAKE` comes directly from OBD? I doubt it.
2. I believe there are already keys for some elements. For example, `KEY_OBD_CURRENT_ERROR` seems to be the same as `KEY_DTCS`.
3. Don't just add new parameters to `Position` without thinking. Only commonly used keys go there.
4. What's the difference between `KEY_OBD_HARSH_ACCELERATION` and `KEY_OBD_TOTAL_HARSH_ACCELERATION`?

joseant:
The data in the test case, I had in the first commit, cames from the device. I have change in the new commit to the real data, from device.

1º- All the data is send to the server. I dont know if the data came directly from the canbus or the device.
2º Yes, It is the DTC, I have change in the commit
3º If I dont add parameters to the position, where do I get the typo?, 
4º 
KEY_OBD_HARSH_ACCELERATION -> It is the harsh acceleration betwen canbus reads.
KEY_OBD_TOTAL_HARSH_ACCELERATION -> It is the all time total.

joseant:
screenshot_1
screenshot_2

tananaev:
You should not add `KEY_OBD_` if it doesn't come from directly from OBD. Also, as I said, you shouldn't just add everything into the position attribute. Usually the rule is if it's used at least in 2 protocols, it should be added there. Otherwise just use string key.

joseant:
I have restore Position model. And I have used string key instead.

tananaev:
What you made in last commit doesn't make any sense for me anymore. Why did you add "MSG_" parameters? Those are intended for message types only. Please follow existing style and practices.

tananaev:
Those strings should be hard-coded.

joseant:
I have change to KEY_

And the string. Just let me know how do you want. Do I extract the string in a class in another file or I create that class inside MeiligaoProtocolDecoder

I have been reviewing other protocols decoder and all have or position object or directitly in the code:

screenshot_3
screenshot_4
screenshot_5

tananaev:
Sorry, I meant "KEY_". Don't do that, just hardcode values if they are used only in one place.

tananaev:
Wait... why you renamed them? As I said, don't create a KEY if it's used only in one place.

tananaev:
For example there was
```java
position.set("drivingRange", parser.nextDouble() * 1000);
```
but you replaced it with
```java
position.set(KEY_DRIVING_RANGE, parser.nextDouble());
```

Don't do that unless the key is used in more than one place.

joseant:
Ok, I have uploaded another version

tananaev:
Just one last thing. Please return back the original test case. I still don't understand why you removed it.

joseant:
I have restored the old one and I have added another one, with real data from device.

tananaev:
Merged, thanks.

PR #2686, opened 2016-12-14
Calendars implementation

Abyss777:
Implemented `Calendar` class that uses [iCal4j](https://github.com/ical4j/ical4j) to parse `ics` files.

Key changes:
- Simple permissions management (link calendar to user)
- Added `calendarId` to `Geofence` class
- `checkMoment` function makes simple trick. It filters calendar for period one second duration and check if there is any `VEVENT` after filtering.

Geofences are always calculated, but if geogence has linked calendar and `fixTime` is not in any VEVENT, Enter/Exit `Event` is not generated.

Also added support to read/save `byte[]` as `BLOB` to `QueryBuilder`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,6 +143,11 @@
             <artifactId>velocity</artifactId>
             <version>1.7</version>
         </dependency>
+        <dependency>
+	        <groupId>org.mnode.ical4j</groupId>
+	        <artifactId>ical4j</artifactId>
+	        <version>2.0.0</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/pom.xml
+++ b/pom.xml
@@ -144,9 +144,9 @@
             <version>1.7</version>
         </dependency>
         <dependency>
-	        <groupId>org.mnode.ical4j</groupId>
-	        <artifactId>ical4j</artifactId>
-	        <version>2.0.0</version>
+            <groupId>org.mnode.ical4j</groupId>
+            <artifactId>ical4j</artifactId>
+            <version>2.0.0</version>
         </dependency>
     </dependencies>
 
--- a/src/org/traccar/model/Calendar.java
+++ b/src/org/traccar/model/Calendar.java
@@ -55,7 +55,9 @@ public class Calendar extends Extensible {
     }
 
     public void setCalendarData(byte[] calendarData) throws IOException, ParserException, SQLException {
-        setCalendar(calendarData);
+        CalendarBuilder builder = new CalendarBuilder();
+        calendar = builder.build(new ByteArrayInputStream(calendarData));
+        this.calendarData = calendarData.clone();
     }
 
     private net.fortuna.ical4j.model.Calendar calendar;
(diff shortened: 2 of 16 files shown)


pom.xml
```diff
@@ -143,6 +143,11 @@
             <artifactId>velocity</artifactId>
             <version>1.7</version>
         </dependency>
+        <dependency>
+	        <groupId>org.mnode.ical4j</groupId>
```
tananaev: No tabs please.

src/org/traccar/model/Calendar.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public void setCalendar(byte[] calendarData) throws IOException, ParserException, SQLException {
+        CalendarBuilder builder = new CalendarBuilder();
+        calendar = builder.build(new ByteArrayInputStream(calendarData));
+        this.calendarData = calendarData.clone();
+    }
+
+    public boolean checkMoment(Date date) {
+        if (calendar != null) {
+            Period period = new Period(new DateTime(date), new Dur(0, 0, 0, 1));
```
tananaev: If I'm understanding correctly, this check a period, right? Is there a way to check a single point in time?

src/org/traccar/model/Calendar.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private net.fortuna.ical4j.model.Calendar calendar;
+
+    @JsonIgnore
+    public net.fortuna.ical4j.model.Calendar getCalendar() {
+        return calendar;
+    }
+
+    public void setCalendar(byte[] calendarData) throws IOException, ParserException, SQLException {
```
tananaev: This setter doesn't make sense. I think the code should be moved to `setCalendarData`.

src/org/traccar/model/Calendar.java
Abyss777: Yes, period of one second length.
rfc-2445 has paradigm of requests for events and free/busy states. Request has to have startTime and endTime, or startTime and duration. Filtering is some kind of request.

Do not think any implementation of rfc 2445 has functionality to check moment of time, not period.

Theoretically, we could reinvent it, iterate all vevents and check if moment in between start and end.
But firstly we have to request such vevents with some kind of request, that contain at least start date. It will be double work.

Abyss777: hm, but I checked and it works with `new Dur(0, 0, 0, 0)`

Abyss777:
Sorry for tabs, Eclipse like it.

Set duration to zero, removed setter, added couple test cases, fixed tabs.

tananaev:
Merged, thanks.

kirantpatil:
Hello @Abyss777, is there any guide/doc to know about Calender functionality ?

Abyss777:
@kirantpatil 

In two words:
Calendar is an wrapper-object on `ics` file. Now it is used to implement what is called "timed geofences"
You can create calendar in some calendar application like Mozilla Thunderbird, create repeated event, for example every day from 7:00 to 23:00

image

Export calendar to `ics` file
Create Calendar in traccar.
Link it to some geofence.

After that, the geofence will work only from 7:00 to 23:00. At night it won't generate any events.

I'll do my best to prepare better guide to next release.

kirantpatil:
Hello @Abyss777. 
Great and Thanks. :clap:

tananaev:
@Abyss777, I know we closed this pull request, but I thought of a couple small things we might want to fix.

First of all, I think we should rename "calendarData" to just "data". It's already in calendar class, there is no need to use longer name.

Another thing in "checkMoment" method. I think we can use diamond operation when creating "PeriodRule" and "Filter".

I'm also a bit confused with "CalendarComponent" and "Component". Do we need to use both or we can just use one of them?

tananaev:
Another small thing. Getter and setter method seems to throw unnecessary exceptions. Can you please clean that up.

netluca1:
Hello I saw that you are implementing the calendar function
It would be interesting if the maintenance menu there was the possibility of using also a date window.
I'll explain in your program currently is foreseen only  the possibility in the menu of maintenance of activating items  that are based on Km and not for on timetable.
I think with the addition of a timetable will be possile to integrate the following items:
stamp vehicle
vehicle insurance
vehicle review
Deadline Leasing vehicle
tax highway vehicle.
Finally it should be integrated into your web and e-mail notification system

tananaev:
@netluca1, if you have any suggestions, please create a new thread to discuss them. Also, you would have to provide a bit more details.

GOchispaGO:
@Abyss777
" ...I'll do my best to prepare better guide to next release... "
👏👏 👏
Where?? THANKS!!!

Abyss777:
@GOchispaGO 
Here it is https://www.traccar.org/documentation-calendars/

GOchispaGO:
Thank you very much @Abyss777 for your detailed explanation. My problems appear only using the Android App Traccar Manager:

https://github.com/tananaev/traccar-manager-android/issues/42

PR #2696, opened 2016-12-16
Put the html code for degree instead of the special character

renaudallard:
This causes display issues in some mail clients

### DIFF
--- a/templates/mail/alarm.vm
+++ b/templates/mail/alarm.vm
@@ -5,6 +5,6 @@
 Device: $device.name<br>
 Alarm:  $position.getString("alarm")<br>
 Time: $event.serverTime<br>
-Point: <a href="$webUrl?eventId=$event.id">#{if}($position.address)$position.address#{else}$position.latitude°, $position.longitude°#{end}</a><br>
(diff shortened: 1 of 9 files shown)


renaudallard:
Actually, putting the following header could probably also solve the issue, but I have not tested that.
`Content-Transfer-Encoding: 8bit`

I think the issue is because we currently have a charset of 8 bits with a transfer encoding of 7 bits.
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: 7bit

renaudallard:
Encoding in base64 might even be cleaner

tananaev:
Does this work will all email clients?

@Abyss777, can you have a quick look to see if there are any potential issues.

renaudallard:
It works with at least thunderbird, macos mail, rainloop.

Abyss777:
@tananaev checked it, no issues.
It is really more correct way for degree sign in html.

tananaev:
Merged, thanks @renaudallard for pull request and thanks @Abyss777 for reviewing the change.

PR #2712, opened 2016-12-20
Feature/2702 watch ext

amenk:


src/org/traccar/model/Command.java
```diff
@@ -41,6 +41,7 @@
     public static final String TYPE_OUTPUT_CONTROL = "outputControl";
     public static final String TYPE_VOICE_MONITORING = "voiceMonitoring";
     public static final String TYPE_SET_AGPS = "setAgps";
+    public static final String TYPE_SET_HEARTS = "setHearts";
```
tananaev: What does this command do? It doesn't sound generic enough.

tananaev: I guess something like that. Also, please use `KEY_DATA` for the value.

amenk:
All done I believe. Thank you for your review!

src/org/traccar/protocol/WatchProtocol.java
```diff
@@ -40,7 +41,10 @@ public WatchProtocol() {
                 Command.TYPE_SILENCE_TIME,
                 Command.TYPE_ALARM_CLOCK,
                 Command.TYPE_SET_PHONEBOOK,
-                Command.TYPE_VOICE_MESSAGE);
+                Command.TYPE_VOICE_MESSAGE,
+                Command.TYPE_SET_TIMEZONE,
+                Command.TYPE_SET_INDICATOR
+            );
```
tananaev: Please move this bracket to the previous line. Everything else looks good.

tananaev:
It looks like you changed it to minutes, not seconds.

amenk:
you are right :) - fixed

tananaev:

PR #2719, opened 2016-12-21
Add attributes to overspeed and maintenance events

### LINKED ISSUE #2701: Additional attributes for Events
If we started store some attributes in Events I have couple suggestions:

1. For OverspeedEventHandler - store effective `speedLimit` and device `speed`, Because it is impossible determine `speedLimit` for old events (it could be changed after). Also user can get information how much was overspeeding, like "Device exceeds the speed on 5 km/h"

2. For MaintenanceEventHandler  - store `totalDistance`, just for historical reports, without additional requests to positions table.

Abyss777:
fix #2701 

Should I declare "speed" constant somewhere?


tananaev:
I think it's fine to be hardcoded. Otherwise it might be confusing.

PR #2770, opened 2017-01-10
Cleanup calendar model

Abyss777:
Changes:
- Rename calendarData to data
- Use CalendarComponent only
- Use diamond operator in two places 

Also fixed TripsTest, I do not know why it was working for you, but I had the same issue as in #2750


tananaev:
Merged, thanks.

PR #2801, opened 2017-01-16
User management

Abyss777:
Here is implementation of user "manager"

"Manager" is user who has `userLimit > 0`
"Objects" are devices/groups/geofences/calendars.

Manager has access to all objects of all users linked to him without inheritance. That mean that if manager has linked manager, he can access only objects linked directly to "sub-manager"

Only admin can link users to manager
Only admin can change admin/deviceLimit/userLimit fields of other users.
Manager can edit readOnly/disabled/expirationTime/token only other users, not himself.

Admin does not respect device and user limits, managers and ordinary users must respect limits.
The only thing denied for admin is link user to himself.

I tried to make all checks in resources similar and look unified.


Abyss777:
Additional comment:

`getManaged*` functions return objects of manager and all managed users, it is made for simplify linking in web interface.

src/org/traccar/api/resource/UserPermissionResource.java
```diff
@@ -0,0 +1,57 @@
...
+@Path("permissions/users")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class UserPermissionResource extends BaseResource {
+
+    @POST
+    public Response add(UserPermission entity) throws SQLException {
+        Context.getPermissionsManager().checkAdmin(getUserId());
+        if (entity.getUserId() == entity.getOtherUserId()) {
+            throw new SecurityException("Selfmanagement prohibited");
```
tananaev: Maybe we can just do nothing in this case? User can always manage his own account.

schema/changelog-3.10.xml
```diff
@@ -48,5 +48,21 @@
...
 
+    <addColumn tableName="users">
+      <column name="userlimit" type="INT" defaultValueNumeric="0" />
+    </addColumn>
+
+    <createTable tableName="user_user">
+      <column name="userid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="otheruserid" type="INT">
```
tananaev: The name is a bit confusing. Maybe it's better to name it `manageduserid` or something else. Can't really think of any good name.

src/org/traccar/database/PermissionsManager.java
```diff
@@ -168,30 +225,52 @@ public void checkUserEnabled(long userId) throws SecurityException {
...
+        if (userId != otherUserId && !isAdmin(userId)) {
+            checkManager(userId, otherUserId);
         }
     }
 
     public void checkGroup(long userId, long groupId) throws SecurityException {
-        if (!getGroupPermissions(userId).contains(groupId)) {
-            throw new SecurityException("Group access denied");
+        if (!getGroupPermissions(userId).contains(groupId) && !isAdmin(userId)) {
+            Iterator<Long> iterator = getUserPermissions(userId).iterator();
```
tananaev: I think it would be easier to do something like this:
```java
            for (int userId : getUserPermissions(userId)) {
                if (getGroupPermissions(userId).contains(groupId)) {
                    return;
                }
            }
            throw new SecurityException("Group access denied");
```

tananaev: Are there any performance issues with going through all the users every time we need to check permissions?

Abyss777: I have not noticed, and theoretically there should not be any.
- first `if` in function is repeat previous behavior
- `getUserPermissions(userId)` will return empty collection for ordinary user.
- If i'm not mistaken `contains` works with hashed data.

The only longer than usual check will be for manager, but everything is cached and uses `HashSet`s, do not see how it can be optimized.

Only one thought is add 'isManager(userId)` before cycle.

Abyss777: sorry, `checkManager(userId)` of course

schema/changelog-3.10.xml
Abyss777: Agree, I can't think of too. Only may be `slaveUserId`, If it is not better then I'll use `managedUserId`

tananaev: Let's go with `managedUserId`. You already use term "managed" anyway.

Abyss777:
Done renaming and `getManaged*` functions optimization.

src/org/traccar/database/PermissionsManager.java
```diff
@@ -146,6 +169,39 @@ public void checkAdmin(long userId) throws SecurityException {
...
+        return users.containsKey(userId) && users.get(userId).getUserLimit() > 0;
+    }
+
+    public void checkManager(long userId) throws SecurityException {
+        if (!isManager(userId)) {
+            throw new SecurityException("Manager access required");
+        }
+    }
+
+    public void checkManager(long userId, long otherUserId) throws SecurityException {
```
tananaev: Make sure to rename "otherUserId" everywhere.

src/org/traccar/database/PermissionsManager.java
```diff
@@ -168,29 +224,49 @@ public void checkUserEnabled(long userId) throws SecurityException {
...
+            if (userId == after.getId()) {
+                checkAdmin(userId);
+            }
+            if (!isAdmin(userId)) {
+                checkManager(userId);
+            }
         }
     }
 
     public void checkUser(long userId, long otherUserId) throws SecurityException {
```
tananaev: Make sure to rename "otherUserId" everywhere.

src/org/traccar/database/PermissionsManager.java
```diff
@@ -75,10 +79,18 @@
         return groupDevices.get(groupId);
     }
 
+    public Set<Long> getUserPermissions(long userId) {
+        if (!userPermissions.containsKey(userId)) {
+            userPermissions.put(userId, new HashSet<Long>());
```
tananaev: Can we use diamond operator here?

src/org/traccar/api/resource/GeofenceResource.java
```diff
@@ -54,8 +54,12 @@
 
         Set<Long> result;
         if (all) {
-            Context.getPermissionsManager().checkAdmin(getUserId());
-            result = new HashSet<>(geofenceManager.getAllGeofencesIds());
+            if (Context.getPermissionsManager().isAdmin(getUserId())) {
+                result = new HashSet<>(geofenceManager.getAllGeofencesIds());
```
tananaev: I'm thinking we can create HashSet only once, and then just call `addAll`.

src/org/traccar/database/PermissionsManager.java
Abyss777: I do not think we could

image

src/org/traccar/database/PermissionsManager.java
Abyss777: I leaved it intentionally to consistence with `checkUser`.
Will rename both.

Abyss777:
Renamed `otherUserId` in both functions and optimized `HashSet` creation

tananaev:
Merged, thanks.

PR #2806, opened 2017-01-17
Add test mail template

### LINKED ISSUE #2792: Email notification testing
It seems like test event always produces a warning in the log file. Should we change its type to "unknown" instead of "test"?

@Abyss777

Abyss777:
fix #2792 

I think test template is more natural than "unknown"


tananaev:
Merged, thanks.

Issue #2702, opened 2016-12-18
Implement more commands for the watch protocol (Techkoo)


amenk:
What I did: Connected with telnet to the control server

    telnet 52.28.132.157 8001

Then I "authenticated" as a watch using the debug output from traccar.

Now I can send commands using the SeTracker android app.

Replace $ID$ with the ID.

Upload Interval in seconds 60 min

    [3G*$ID$*000b*UPLOAD,3600]

1 min
    [3G*$ID$*0009*UPLOAD,60]

SOS  Number 1, 2, 3

    [3G*$ID$*001e*SOS,$NUMBER1$,$NUMBER2$,$NUMBER3$]

Phone Book (00710071 seems to be an encoded name)
              `[3G*$ID$*0053*PHB,1111111,0066007200690065006e00640031,2222222,0066007200690065006e00640032,,,,,,]`

     1111111,0066007200690065006e00640031   # number1 = 111111  name2 = "friend1" (simply hex encoded string)
     2222222,0066007200690065006e00640032 # number2 = 2222222   name2 = $fiend2)

up to 10 friends from the app, but sends only 10 - probably it is waiting some kind of "okay"

     [3G*$ID$*006b*PHB,111111,0066007200690065006e00640031,2222222,00660032,333333,00660033,44444444,00660034,5555555,00660035]

Let the watch ring 

    [3G*$ID$*0004*FIND]

Set timezone: (GMT +1)

    [3G*$ID$*0007*LZ,0,+1]

Plus GMT+8:30  summer time

   [3G*$ID$*0009*LZ,0,+9.5]

Language + Timezone (5 = German)

    [3G*$ID$*0009*LZ,5,+11.5]

Seems to use 0007 for full-hour offset and 0009 for .5

Set silence time

    [3G*$ID$*002f*SILENCETIME,00:02-06:22,22:00-23:00,18:00-19:02]

Power off

    [3G*$ID$*0008*POWEROFF]

Monitor call back

    [3G*$ID$*0016*MONITOR,$NUMBER$]

Save-zone does not send anything to the app (but traccar can do the same)

Set Hearts

    [3G*$ID$*0008*FLOWER,2]

amenk:
I was able to set the timezone via traccar :-) awesome stuff

amenk:
Really interesting would be of course to be able to send and receive voice messages as well :-)

tananaev:
You don't need to reverse engineer it. There is official documentation available.

[...]

PR #2821, opened 2017-01-22
#2702 More comfortable custom commands (auto calc length + ID)

amenk:


src/org/traccar/protocol/WatchProtocolEncoder.java
```diff
@@ -123,6 +123,12 @@ protected Object encodeCommand(Command command) {
                 return formatCommand(command, "LZ,,{%s}", Command.KEY_TIMEZONE);
             case Command.TYPE_SET_INDICATOR:
                 return formatCommand(command, "FLOWER,{%s}", Command.KEY_DATA);
+            case Command.TYPE_CUSTOM:
+                String data = command.getString(Command.KEY_DATA);
+                if (data.startsWith("[")) {
```
tananaev: I think we should remove this check.

amenk:
okay, you can remove it if you want :-)​

tsmgeek:
This is now outdated and can be closed as seems code was added directly by @tananaev in 1c807549e315d04054b4803518d2624591976324

Issue #2810, opened 2017-01-18
Permissions polishing


Abyss777:
Agree, that permissions system became a bit too complicated, but not so much, we do not have user groups :)

I thing it has everything for most use cases now, except some small things should be polished:

1. As you mentioned, manager can set any expiration time for other user, it should be handled. If manager has expiration time, he should not be able set expiration time more then his one.

2. There is no protection from cycling management now. I mean if "manager" has user "sub-manager" who is manager of "manager". Behavior will be unpredictable.  Should we add something, or will rely on admins good sens? 

3. Could you remember, why you allowed readonly user edit his own account and notifications? I'm asking because totally readonly user fits more for public use case.

4. Users created by manager will have no device limit even manager has it, it is not good. But I do not have idea how to implement it correctly without backward user->manager confrontation. If you have, please share.

tananaev:
I agree that manager shouldn't be able to create a user with expiration date more than his own.

I don't think we should handle corner cases with cycles. We should make logic as simple as possible. I'm against handling those types of things.

Originally idea for readonly user was to not allow it to create and edit devices. I guess with device limit, it makes less sense. Maybe we should make readonly users into full readonly.

Maybe if manager has a device limit, we would just limit him to creating users with 0 device limit.

Abyss777:
Let's summarize 

1. Will implement it.
2. OK, leave it as is.
3. Will implement full readonly.
4. Will implement if manager has device limit then all users added/edited by him should has less or equal device limit. Not the best solution, but simple and rather close to optimal.

tananaev:
4. I was thinking to make device limit 0 for users, so that only manager can create and assign devices. This way we have total limit of devices limited to the number on manager's account.

Alternative option is to subtract the limit from the manager. For example, if manager has limit 5 and he creates a user with limit of 2 devices, he would lose those 2 from his limit. It might be a bit confusing, so I don't really like it.

Abyss777:
In that case manager can infinitely link device to user and unlink from himself.

Just subtraction is not enough, the ideal algorithm is following:

- Check user device limit (deny if reached)
- Get list of user managers (we need backward user->manager confrontation here)
- In cycle count managed devices for every manager and deny if count >= device limit of any appropriate manager.

It is too complicated and need additional HashSet user->manager

My suggestion is easier but in worth case company (manager and his users) will have userLimit * deviceLimit devices.

But agree with you we should implement some limitation in device creating as replacement for "not-full" readonly.

[...]

PR #2831, opened 2017-01-25
Permissions improvements

Abyss777:
Here is changes discussed #2810 

- Add "deviceReadonly" user field. Such user can not add/edit/delete devices.
- Allow users edit `token`s
- `deviceLimit = 0` means zero, not infinite 
- Managers can create users with `deviceLimit = 0` only
- `userLimit = 0` means ordinary user
- `userLimit = -1` means manager without user limit
- Manager can't create users with `expirationTime` later then their
- Only admin can unlink devices from himself
- Readonly user becomes totally readonly
- Other discussed permissions improvements

### DIFF
--- a/schema/changelog-3.10.xml
+++ b/schema/changelog-3.10.xml
@@ -64,5 +64,15 @@
     <addForeignKeyConstraint baseTableName="user_user" baseColumnNames="userid" constraintName="fk_user_user_userid" referencedTableName="users" referencedColumnNames="id" onDelete="CASCADE" />
     <addForeignKeyConstraint baseTableName="user_user" baseColumnNames="manageduserid" constraintName="fk_user_user_manageduserid" referencedTableName="users" referencedColumnNames="id" onDelete="CASCADE" />
 
+    <update tableName="users">
+        <column name="devicelimit" valueNumeric="-1" />
+        <where>devicelimit = 0</where>
+    </update>
+    <addDefaultValue tableName="users" columnName="devicelimit" defaultValueNumeric="-1" />
+
+    <addColumn tableName="users">
+      <column name="devicereadonly" type="BOOLEAN" defaultValueBoolean="false" />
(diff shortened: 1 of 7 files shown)


schema/changelog-3.10.xml
```diff
@@ -64,5 +64,15 @@
     <addForeignKeyConstraint baseTableName="user_user" baseColumnNames="userid" constraintName="fk_user_user_userid" referencedTableName="users" referencedColumnNames="id" onDelete="CASCADE" />
     <addForeignKeyConstraint baseTableName="user_user" baseColumnNames="manageduserid" constraintName="fk_user_user_manageduserid" referencedTableName="users" referencedColumnNames="id" onDelete="CASCADE" />
 
+    <update tableName="users">
+        <column name="devicelimit" valueNumeric="-1" />
+        <where>devicelimit = 0</where>
+    </update>
+    <addDefaultValue tableName="users" columnName="devicelimit" defaultValueNumeric="-1" />
```
tananaev: Why can't we set it above when we define column? What about `devicelimit` default value? Do we need to remove old default value before adding a new one?

Abyss777: We defined `devicelimit` column in 3.9 changelog.
update is just for update value.
We do not need remove old, it is equivalent of `ALTER TABLE` http://www.liquibase.org/documentation/changes/add_default_value.html
And it works correctly at least on MySQL

tananaev: What about `userlimit`? Shouldn't it be `-1` by default as well?

Abyss777: Ordinary user has 0 `userLimit`. I think 0 fits better for default value.

tananaev: Yes, I guess it makes sense.

tananaev:
Merged. Thanks.

Issue #2832, opened 2017-01-25
GT06 Bug on processing 7979 start bit packet, ignition state changes


jaimzj:
Hello,

I just came across a small problem, in traccar 3.9 when GT06 protocol,

Physical Devices Ignition = True
Traccar Shows and records it as = True

However the moment packet starting with start bit `7979` sub type `94` is received the Ignition status gets changed to = Off / false

-Then it gets into a infinite loop of ignition status change event.. (This happens because, everytime device sends igntion on via all other packets, and then in between when server receives the 7979 start bit packet it changes ignition state to off, then new other packets make it on)

The above is in version 3.9,   So I tested the same scenario in version 3.7 and here this problem did not occur in 3.7.

tananaev:
As far as I can tell, message type `94` doesn't have any ignition info, so you should probably use `processing.copyAttributes` config parameter to copy the ignition value over from previous record.

jaimzj:
You are right it message type 94 doesn't have ignition info, I am already using the processing.copyAttributes . with copyAttributes also this issue continues.

What I shared as a possible bug, above is not related to the attributes being copied from previous record.

Basically every time type 94 packet is received some change in the code probably is triggering a ingition off state in Traccar. (So if the packet 94 interval is 5 seconds) when the ignition value in actual is on (Every 5 second when the 94 packet comes in, there is an entry made as ignition off) then when the normal or any other packet with ignition on comes in which is the actual value it turns back to on. thus leading to a huge list of events on/off/on/off/on/off ........ (while the actual status is On only).

tananaev:
Message type `94` doesn't have any ignition value, so if you have `copyAttributes` configured properly, it won't cause notifications. You have to track down exact message that triggers it.

jaimzj:
Sure, I am going to enable copyAttributes test it now and share the results with you.
As for exact message, from the earlier tests I did I had noticed any message with start bit 7979 this gets triggered.

But allow me few minutes I Will update the copyAttributes test result.

jaimzj:
@tananaev I enabled copyAttributes, 

And with `copyAttributes` enabled for ignition, on receiving packet 94 or start bit 7979, while the ignition status was on, it did not update to off state anymore.

tananaev:
So, everything is working correctly. I thought you already tried it and it didn't work.

jaimzj:
No it was my mistake, I wrote I had copyAttributes enabled, but it was not enabled.

[...]

PR #2833, opened 2017-01-25
Rise ignition events only if both positions has ignition attribute

Abyss777:
As you suggested in #2832


tananaev:
Looks like test is failing now. Please fix.

Abyss777:
Oh, sorry.
Unfortunately I do not have an idea how to make `lastPosition` changeable in tests. I would be thankful if you point me. 
Just removed real check.

tananaev:
Merged, thanks.

PR #2835, opened 2017-01-26
Device readonly user shouldn't be able edit attribute aliases

Abyss777:


tananaev:
Merged. What about the web part?

PR #2837, opened 2017-01-26
New T580W protocol with GPS/LBS/WiFi positioning

ValeraVi:
... and couple useful fixes...

### DIFF
--- a/src/org/traccar/protocol/Ivt401ProtocolDecoder.java
+++ b/src/org/traccar/protocol/Ivt401ProtocolDecoder.java
@@ -38,8 +38,8 @@ public class Ivt401ProtocolDecoder extends BaseProtocolDecoder {
             .number("(d+),")                     // imei
             .number("(dd)(dd)(dd),")             // date (ddmmyy)
             .number("(dd)(dd)(dd),")             // time (hhmmss)
-            .number("([-+]d+.d+),")              // longitude
             .number("([-+]d+.d+),")              // latitude
+            .number("([-+]d+.d+),")              // longitude
             .number("(d+),")                     // speed
             .number("(d+),")                     // course
             .number("(-?d+.?d*),")               // altitude
@@ -74,8 +74,8 @@ public class Ivt401ProtocolDecoder extends BaseProtocolDecoder {
 
         position.setTime(parser.nextDateTime(Parser.DateTimeFormat.DMY_HMS));
 
-        position.setLongitude(parser.nextDouble());
         position.setLatitude(parser.nextDouble());
+        position.setLongitude(parser.nextDouble());
         position.setSpeed(UnitsConverter.knotsFromKph(parser.nextInt()));
         position.setCourse(parser.nextInt());
         position.setAltitude(parser.nextDouble());


src/org/traccar/geolocation/UniversalGeolocationProvider.java
```diff
@@ -37,7 +37,10 @@ public UniversalGeolocationProvider(String url, String key) {
     public void getLocation(Network network, final LocationProviderCallback callback) {
         try {
             String request = Context.getObjectMapper().writeValueAsString(network);
-            Context.getAsyncHttpClient().preparePost(url).setBody(request).execute(new AsyncCompletionHandler() {
+            Context.getAsyncHttpClient().preparePost(url)
+                    .setHeader("Content-Type", "application/json")
+                    .setHeader("Content-Length", "" + request.length())
```
tananaev: Please use `String.valueOf` instead of this conversion.

src/org/traccar/geolocation/UniversalGeolocationProvider.java
```diff
@@ -37,7 +37,10 @@ public UniversalGeolocationProvider(String url, String key) {
     public void getLocation(Network network, final LocationProviderCallback callback) {
         try {
             String request = Context.getObjectMapper().writeValueAsString(network);
-            Context.getAsyncHttpClient().preparePost(url).setBody(request).execute(new AsyncCompletionHandler() {
+            Context.getAsyncHttpClient().preparePost(url)
+                    .setHeader("Content-Type", "application/json")
```
tananaev: Try to use existing constants for this.

ValeraVi: done

src/org/traccar/geolocation/UniversalGeolocationProvider.java
ValeraVi: done

ValeraVi:
This is an implementation of protocol for devices of http://www.vjoycar.com/
Some description of protocol are here: http://www.vjoycar.com/File.asp?Id=119
also here: https://goo.gl/8DKadX

According to documentation supported devices are:
T16 T18 T18H TK05 TK10 Tk15 TK20 T12 T4400 T8800 T15400
with WiFi:
T510W, T580W, TK101W, T531W,TK05SDW,TK10SDW,TK15SDW,TK20SDW
List of commands (traccar -> Device) are from user guide from T580W: https://goo.gl/M4ancp
I suspect, that at least latest wifi devices also using same commands.

Protocol looks like tk103, but i am not sure that vjoycar and company behind tk103 protocol from https://www.traccar.org/protocols/ are in sync and develop it together.
I already found some differences even in GPS position message (+altitude, +rssi, -statuses, etc), LBS positioning (values is decimal for T580W), new WiFi positioning message, different alarm coding in messages.

Not sure that other manufacturers which uses tk103 like protocol are using completely the same messages for WiFi, LBS, Alarms and device commands.

tananaev:
Can you please provide some samples with the differences?

ValeraVi:
Diff betweed tk103 and t580w implementation: https://github.com/ValeraVi/traccar/commit/c94f965b88f83513725b3505fc32207351cf6f46

Messages samples for t580w are present for each message description in documentation here: https://goo.gl/8DKadX 

For example, GPS position:
**t580w:**
 `(013632651491,ZC07,040613,A,2234.0297N,11405.9101E,000.0,040137,178.48,1.21,8)`
**tk103** (accoording to https://drive.google.com/open?id=0B9QofkYKWsMuRjdmWFVrV2VrOUU - i do not have real device):
`(013612345678BP05000013612345678080524A2232.9806N11404.9355E000.1101241323.8700000000L000450AC)`

LBS/WiFi position:
**t580w:**
`(013632651491,DW50,460,1233,535,543,0,040613,040137)`
or
`(013632651491,DW50,460,1233,535,543,2,aa*bb*cc*dd*ee*ff:-8:0,aa*bb*cc*dd*ee*ff:-8:0,040613,040137)`
**tk103:**
I have not found any samples or documentation about such messages :(

So if comparing the messages samples or documentations i can't even tell that protocols are the same. But when comparing the tk103 protocol implementation in traccar it just support some messages of t580w protocol.

tananaev:
The point is that it already supports some messages, so it makes sense to extend it to support other messages as well.

Looking at the diff, I can see that there is very little change to the regular expressions. It's mostly just addition of some extra parameters at the end. I don't see why it can't be universal solution.

ValeraVi:
I do not have a devices, which existing implementations of tk103 supports, so i can't test resulted protocol after merge of tk103 and t580w. Hope that tests are covering all cases, so if they pass all real devices will also works ok.

Also seems like current implementation of tk103 in traccar goes far beyond which is described in doc's about tk103 here https://www.traccar.org/protocols/. For example i have not found doc's for LBS positioning of tk103.

My thinking was that protocols of different, independent manufacturers, should be tracked by different implementations as it will be hard to maintain one protocol when those different manufacturers will be implementing new features independently.

But from other side, i am agree that it is more easy for you to maintain less amount of code.

I will create new pull requests for other unrelated fixes from this pull request.

Then will try merge t580w to tk103. Or you prefer to do this yourself?

tananaev:
It would be great if you can do it.

ValeraVi:
Ok, i will try..

tsmgeek:
Where was this left, as its completely standalone protocol with no other affect on the project should it be merged in and closed.

tananaev:
Have you read comments? If you really want to help you should read the discussion first. Otherwise you just adding more work for me to reply to your comments. I think my opinion is pretty clear here.

tsmgeek:
Yes I read it all, this update is also by way to notify the PR creator that they may need to do more work as they did post update after your last comment but since have been quiet.

tananaev:
Then I don't understand why you are saying that it should be merged and closed. My suggestion clearly hasn't been implemented.

Issue #2793, opened 2017-01-13
Telic protocol improvement


argokask:

tananaev:
Can I have some message samples?

argokask:
Yes,

I would like to add two more event ID codes to that table which is not represented in the table i posted before.

98 distance  (configurable distance travelled since last message)
99 timer (configurable time period elapsed since last message)

Example message:

HEX message:

6500000033313230**3939xxxxxxxxxxxx**2c3137313131363036303230302c302c3137313131363036303230302c**xxxxxxxxxxxx**2c**xxxxxxxxxxxx**2c332c33372c3130362c31312c2c2c34332c353933313530382c2c303031302c30302c3134312c302c302c3031303800

Output message:

e3120**99xxxxxx**,171116060200,0,171116060200,**xxxxxx**,**xxxxxx**,3,37,106,11,,,43,5931508,,0010,00,141,0,0,0108

In that example 3939 (hex) means event ID 99 what according to table means that configurable time has elapsed since last message. The xxxx marks is not related with that message ID, i just replaced IMEI and coordinates with that.

tananaev:
I don't see 98 and 99 in your table.

argokask:
Yes, but i would like to if you can add these two event ID's also to traccar. I have checked them and these are also common to all telic tracking devices.

Descriptions:

98 distance (configurable distance travelled since last message)
99 timer (configurable time period elapsed since last message)

tananaev:
I would need documentation for those.

argokask:
Ok, i do not have documentation for those event codes but i can reference to one older issue in this forum [https://github.com/tananaev/traccar/issues/691](url) which also describes those event ID's.

tananaev:
So, which ones do you want me to add? We can only add alarms:

[Position.java#L71](https://github.com/tananaev/traccar/blob/master/src/org/traccar/model/Position.java#L71)

argokask:
Hi,

Atm in Traccar server under "State" section i see attribute "Type" with (for example) "Value" 99

My wish would be that instead of number 99 i see word "timer" because for most users that value 99 do not give any strait forward information. So to do that it is neccessary to make relations between Type values and descriptions in table (+ 98 and 99 types i sent you later).

Best regards,
Argo

tananaev:
Why would user need to know if it's a timer or distance report?

ssmipp:
[...]

PR #2841, opened 2017-01-27
#2793 Added some Telic events (alarm)

ssmipp:
Added Telic protocol events.

### DIFF
--- a/src/org/traccar/protocol/TelicProtocolDecoder.java
+++ b/src/org/traccar/protocol/TelicProtocolDecoder.java
@@ -26,6 +26,7 @@ import org.traccar.model.Position;
 
 import java.net.SocketAddress;
 import java.util.regex.Pattern;
+import org.traccar.helper.Log;
 
 public class TelicProtocolDecoder extends BaseProtocolDecoder {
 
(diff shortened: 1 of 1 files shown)


src/org/traccar/protocol/TelicProtocolDecoder.java
```diff
@@ -35,27 +36,27 @@ public TelicProtocolDecoder(TelicProtocol protocol) {
 
     private static final Pattern PATTERN = new PatternBuilder()
             .number("dddd")
-            .number("(d{6})")                    // device id
```
tananaev: Please don't change comment formatting here.

src/org/traccar/protocol/TelicProtocolDecoder.java
```diff
@@ -105,4 +110,57 @@ protected Object decode(
         return position;
     }
 
+    private String decodeAlarm(String event) {
+
+        int eventId = 0;
+
+        try {
+            eventId = Integer.parseInt(event);
+        } catch (Exception e) {
```
tananaev: It can't fail because we use regular expression, so I don't think we need to catch anything. We can just use `parser.nextInt()` and pass int value directly into `decodeAlarm` method.

src/org/traccar/protocol/TelicProtocolDecoder.java
```diff
@@ -105,4 +110,57 @@ protected Object decode(
...
+            Log.error("TelicProtocolDecoder: Error parsing event (" + eventId + "): " + e.toString());
+        }
+
+        switch (eventId) {
+            case 1:
+                return Position.ALARM_POWER_ON; // Power on
```
tananaev: Don't add obvious comments like this. Basically all known alarms don't require any extra comment.

src/org/traccar/protocol/TelicProtocolDecoder.java
```diff
@@ -105,4 +110,57 @@ protected Object decode(
...
+            case 6:
+                return null; // Course change
+            case 7:
+                return Position.ALARM_GEOFENCE_ENTER; // Geofence area enter
+            case 8:
+                return Position.ALARM_GEOFENCE_EXIT; // Geofence area exit
+            case 9:
+                return null; // GPS fix lost
+            case 10:
+                return null; // Periodic wakeup / Routine message
```
tananaev: Some of the ids are not alarms, so you shouldn't handle them. Most of the nulls seem to be just causes for the message, but not alarms. So we don't need cases for those ids here.

tananaev:
Merged, thanks.

PR #2874, opened 2017-02-06
Event forward header small improvement

Abyss777:
Removed user agent (default will be `User-Agent: AHC/1.0`)

Set content type to `application/json`

`event.forward.header` will override default values.


tananaev:
Merged, thanks.

PR #2878, opened 2017-02-07
Do not allow readonly users send commands

Abyss777:
It is can be controversial, but I think readonly and device readonly users should not be able send commands.

### DIFF
--- a/src/org/traccar/api/resource/CommandResource.java
+++ b/src/org/traccar/api/resource/CommandResource.java
@@ -33,6 +33,8 @@ public class CommandResource extends BaseResource {
 
     @POST
     public Response add(Command entity) {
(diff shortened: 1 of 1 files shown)


tananaev:
Merged.

tananaev:
What about client side?

djbo1982:
@Abyss777  @tananaev Do you mean about the point 3, right... because is the user is READ ONLY, suposed only can read reports or see the car in the map, but they can´t send command or modify the config or car.

This point are testing in other platform, and is very useful to persons how sell they platform to other persons, and this persons give access only to see the car.

In my case, i managed a float of 100 cars, and i give access to the Partner only to see the car, but as few days ago, i have a big accident with one car because, one partner send a engine cut off command to a devices and the driver lost the control of the car.

PR #2881, opened 2017-02-08
Fix MXT protocol response

gustavofcalmeida:
The last fix generated out of memory error because it was writing into the same buffer it was reading from.

I tested it online with a device and it's now working. The byte stuffing is ok too.

### DIFF
--- a/src/org/traccar/protocol/MxtProtocolDecoder.java
+++ b/src/org/traccar/protocol/MxtProtocolDecoder.java
@@ -47,7 +47,7 @@ public class MxtProtocolDecoder extends BaseProtocolDecoder {
             response.writeInt((int) id);
             response.writeShort(crc);
             response.writeShort(Checksum.crc16(
-                    Checksum.CRC16_XMODEM, response.toByteBuffer(0, response.readableBytes())));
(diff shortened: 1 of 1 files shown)


src/org/traccar/protocol/MxtProtocolDecoder.java
```diff
@@ -42,26 +42,25 @@ public MxtProtocolDecoder(MxtProtocol protocol) {
...
-            response.writeByte(0x01); // header
             response.writeByte(device);
             response.writeByte(MSG_ACK);
             response.writeInt((int) id);
             response.writeShort(crc);
             response.writeShort(Checksum.crc16(
-                    Checksum.CRC16_XMODEM, response.toByteBuffer(1, response.readableBytes() - 1)));
-            response.writeByte(0x04); // ending
-            channel.write(response);
+                    Checksum.CRC16_XMODEM, response.toByteBuffer(0, response.readableBytes())));
```
tananaev: You can just use `response.toByteBuffer()` without parameters here.

gustavofcalmeida: done

tananaev:

PR #2906, opened 2017-02-14
MiniFinder Updates

jon-stumpf:
I have been working with the EV-07S and EV-07W devices from the manufacturer, Eview.  After some investigation, I believe this to be the source technology for the MiniFinder Pico and the Defenstar GPS668.  I have made a number of updates to more closely conform with the Eview EV-07S documentation, collecting additional information, and fixed some typos.

In my update, I noticed there are two record types that are not documented: !A and !C.  I have used the data string found in the test class as a guide and implemented an equivalent parser for each.  In these parsers, I have changed the data collection such that the unknown fields are reported differently in the database.  Given that these records are undocumented, my first inclination was to delete them entirely.  However, I will leave that decision to you.

I have attached the Eview EV-07S documentation to be included on the <a href="https://www.traccar.org/protocols/">Protocols</a> page.  I have more changes coming but wanted to make sure the pull request and coding met your standards.  My last work here was v2.10.

Thanks for your consideration.

- jss

[EV-07S GPRS protocol (1).pdf](https://github.com/tananaev/traccar/files/772958/EV-07S.GPRS.protocol.1.pdf)


src/org/traccar/protocol/MiniFinderProtocolDecoder.java
```diff
@@ -34,44 +34,145 @@ public MiniFinderProtocolDecoder(MiniFinderProtocol protocol) {
...
+    // Eview, the manufacturer, indicated that resellers could implement new record types.
+    // Using the data in MiniFinderProtocolDecoderTest.java as a guide
+    private static final Pattern PATTERN_A = new PatternBuilder()
+            .expression("!A,")
             .number("(d+)/(d+)/(d+),")           // date
             .number("(d+):(d+):(d+),")           // time
             .number("(-?d+.d+),")                // latitude
             .number("(-?d+.d+),")                // longitude
-            .number("(d+.?d*),")                 // speed
+            .number("(d+.?d*),")                 // unknown
```
tananaev: I think instead of assuming that there will be 3 parameters, we should put `any()` here.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
```diff
@@ -34,44 +34,145 @@ public MiniFinderProtocolDecoder(MiniFinderProtocol protocol) {
...
+            .number("(d+)/(d+)/(d+),")           // date
+            .number("(d+):(d+):(d+),")           // time
+            .number("(-?d+.d+),")                // latitude
+            .number("(-?d+.d+),")                // longitude
+            .number("(d+.?d*),")                 // speed (km/h)
+            .number("(d+.?d*),")                 // course
+            .number("(x+),")                     // flags
+            .number("(-?d+.d+),")                // altitude (meters)
+            .number("(d+),")                     // battery (percentage)
+            .number("(d+.?d*),")                 // unknown
```
tananaev: Use `any()` here as well.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
```diff
@@ -34,44 +34,145 @@ public MiniFinderProtocolDecoder(MiniFinderProtocol protocol) {
         super(protocol);
     }
 
-    private static final Pattern PATTERN = new PatternBuilder()
-            .expression("![A-D],")
+    // The !A record is not in the documentation for MiniFinder or EV-07S.
```
tananaev: I think we should remove these comments. You can put them into the pull request description, so we can always go back and check why it was done like this, but I don't think we need to keep them in the code.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
```diff
@@ -34,44 +34,145 @@ public MiniFinderProtocolDecoder(MiniFinderProtocol protocol) {
...
-            .number("(d+),")                     // battery
+            .number("(-?d+.d+),")                // altitude (meters)
+            .number("(d+),")                     // battery (percentage)
             .number("(d+),")                     // satellites in use
             .number("(d+),")                     // satellites in view
-            .text("0")
-            .or()
-            .any()
-            .groupEnd()
+            .number("(d+.?d*)")                  // HDOP, confirmed with Eview manufacturer
```
tananaev: Comments are usually lower case. "Confirmed with Eview manufacturer" is unnecessary. You can mention it in the pull request comment for future reference.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
```diff
@@ -94,49 +195,46 @@ protected Object decode(
...
-                position.set(Position.KEY_SATELLITES, parser.next());
+            decodeFlags(position, parser.nextInt(16));
 
-            }
+            position.setAltitude(parser.nextDouble());
+
+            position.set(Position.KEY_BATTERY, parser.nextInt());
+
+            // Ignoring last three fields, unknown
+            position.set("unknown", parser.next() + "," + parser.next() + "," + parser.next());
```
tananaev: Remove this till we know what it is.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
```diff
@@ -94,49 +195,46 @@ protected Object decode(
...
+            DateBuilder dateBuilder = new DateBuilder()
+                    .setDateReverse(parser.nextInt(), parser.nextInt(), parser.nextInt())
+                    .setTime(parser.nextInt(), parser.nextInt(), parser.nextInt());
+            position.setTime(dateBuilder.getDate());
+
+            position.setLatitude(parser.nextDouble());
+            position.setLongitude(parser.nextDouble());
+
+            // Ignoring last three fields, unknown
+            position.set("unknown", parser.next() + "," + parser.next() + "," + parser.next());
```
tananaev: Same. Remove for now till we get proper documentation.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
```diff
@@ -94,49 +195,46 @@ protected Object decode(
...
+
+            return position;
         }
 
+        // If we get here, we didn't understand the message
```
tananaev: Unnecessary comment.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
```diff
@@ -94,49 +195,46 @@ protected Object decode(
...
+
         return null;
     }
-
```
tananaev: The space is here is left intentionally. It's consistent with all other classes.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
```diff
@@ -34,44 +34,145 @@ public MiniFinderProtocolDecoder(MiniFinderProtocol protocol) {
...
     }
 
-    private static final Pattern PATTERN = new PatternBuilder()
-            .expression("![A-D],")
+    // The !A record is not in the documentation for MiniFinder or EV-07S.
+    // Eview, the manufacturer, indicated that resellers could implement new record types.
+    // Using the data in MiniFinderProtocolDecoderTest.java as a guide
+    private static final Pattern PATTERN_A = new PatternBuilder()
+            .expression("!A,")
             .number("(d+)/(d+)/(d+),")           // date
```
tananaev: I think we should remove duplication in the patter. See `Gl200ProtocolDecoder.PATTERN_LOCATION` for example.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
```diff
@@ -34,44 +34,145 @@ public MiniFinderProtocolDecoder(MiniFinderProtocol protocol) {
...
+            if (!parser.matches()) {
                 return null;
             }
 
-            Parser parser = new Parser(PATTERN, sentence);
+            Position position = new Position();
+            position.setProtocol(getProtocolName());
+            position.setDeviceId(deviceSession.getDeviceId());
+
+            DateBuilder dateBuilder = new DateBuilder()
```
tananaev: It would be nice to remove code duplication for decoding as well. Beginning of the messages are the same for all 3 patterns.

src/org/traccar/model/Position.java
```diff
@@ -22,9 +22,11 @@
     public static final String KEY_ORIGINAL = "raw";
     public static final String KEY_INDEX = "index";
     public static final String KEY_HDOP = "hdop";
-    public static final String KEY_SATELLITES = "sat";
+    public static final String KEY_SATELLITES = "sat"; // in use
+    public static final String KEY_SATELLITES_VISIBLE = "satVisible";
```
tananaev: I don't think we need to save this value. I doubt anybody cares about it. Also, if attribute is used only one, don't put it into `Position` class.

src/org/traccar/model/Position.java
```diff
@@ -22,9 +22,11 @@
...
     public static final String KEY_RSSI = "rssi";
     public static final String KEY_GPS = "gps";
+    public static final String KEY_FIX = "fix"; // GPS, GSM, none
```
tananaev: What does this mean?

src/org/traccar/model/Position.java
```diff
@@ -98,6 +100,7 @@
     public static final String ALARM_FOOT_BRAKE = "footBrake";
     public static final String ALARM_OIL_LEAK = "oilLeak";
 
+
```
tananaev: Remove extra space.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: Agreed

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: Agreed

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: Agreed

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: Agreed

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: Agreed

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: In my initial attempt, I felt the code became too cumbersome and un-intuitive.  I will try again.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: Agreed

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: I would prefer to review the !A and !C records entirely as I was unable to find any documentation on the Internet or from the manufacturer.  Based on my discussions, these appear to be customizations in the firmware of a particular reseller (e.g., MiniFinder?).  However, they are not present in the MiniFinder documentation either.

Is this acceptable?

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: Agreed

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: Agreed

src/org/traccar/model/Position.java
jon-stumpf: Please explain, "if attribute is used only one, don't put it into Position class".

src/org/traccar/model/Position.java
jon-stumpf: The EV-07S device can provide location information (aka "a fix") either by GPS or by GSM triangulation.  Currently, the protocol will flag location information not provided by GPS as invalid.  I have maintained this behavior for now as I am testing the accuracy of the "GSM Locating" capability.  If the accuracy is decent, I will propose marking these locations as valid.

jon-stumpf: This KEY is to record what method was used to determine the location.

src/org/traccar/model/Position.java
jon-stumpf: Agreed

src/org/traccar/model/Position.java
tananaev: Typo.. it's should've been "once", not "one". To avoid polluting Position class with hundreds of keys, only ones that are used more than in one protocol are added here. If it's use only in a single place, just hardcode the key in that protocol.

src/org/traccar/model/Position.java
tananaev: This should probably be hardcoded in the protocol as well. Most likely you need to map this to position.valid flag.

src/org/traccar/model/Position.java
jon-stumpf: Understood.  I will review the documentation of some of the other protocols to determine if this data is available.  There is utility in this data.  If it is only used here, I will "hardcode" the key as you suggest.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: Do you agree that I can remove the !A and !C records from the protocol given that there is no documentation for such records?

src/org/traccar/model/Position.java
jon-stumpf: I will hardcode this into the protocol as suggested.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
tananaev: That's not an option. I would recommend to remove the "unknown" parameters, but please don't remove support for A and C records.

src/org/traccar/model/Position.java
jon-stumpf: In a quick search, I have found three protocols that have report visible satellites: Avl301 (no doc), MiniFinder, and Navigil.  I recommend keeping this new key: KEY_SATELLITES_VISIBLE.

tananaev: OK, I guess we can keep this one then. Please update those protocols as well, if you have time.

jon-stumpf: Already did.

src/org/traccar/protocol/MiniFinderProtocolDecoder.java
jon-stumpf: I will retain the A and C records and reduce the duplication of the patterns.

jon-stumpf:
Do you have the protocol documentation for Atrack?  It is not on the <a href="https://www.traccar.org/protocols/">Protocols</a> page.  I want to record CellTower information which it collects but I don't know how to separate the MCC and the MNC which is received as one Unsigned Int.

tananaev:
I have, but unfortunately I can't share it. They have a whole table for the MCC + MNC values, but basically first 3 digits is MCC and the last 2 digits is MNC.

For example:

MCC - 276
MNC - 01
MCC + MNC - 27601

jon-stumpf:
What is the difference between "accuracy" (i.e., position.setAccuracy()) and "hdop"?

jon-stumpf:
I notice that there is KEY_APPROXIMATE.  When is it appropriate to use this?

I assume if I have a means in which to GeoLocate (e.g., GPS, GSM triangulation, etc.), I use Position.setValid(true).  And, if the mechanism is not "precise" (e.g., GSM triangulation), I also use Position.set(Position.KEY_APPROXIMATE, true).  Is this correct?

jon-stumpf:
I have completed all commits for your review.  I have discovered anomalies elsewhere in the code and will be submitting a separate pull request once I have finished.  Please answer my last two questions above as these pertain to the future work I intend to submit.  Thanks.

tananaev:
Merged. Thanks for the pull request and fixing all the comments.

PR #2912, opened 2017-02-15
Remove check for valid in IgnitionEventHandler

### LINKED ISSUE #2911: Ignition Notifications
I have my [iTracMT-1V8](http://s.aliexpress.com/meMNzuEZ) (H02 port 5013) working "correctly" with your TraccarDemo4 server.

14 de febrero de 2017 141029 cet

Web/email notifications work correctly for example with the GeoFences, ...but I do not receive Ignition ON/OFF changes which are displayed ([reversed](https://www.traccar.org/forums/topic/mistake-in-ignition/)) on the device screen.

Do I have to configure ignition notifications somewhere else?
THANKS

Abyss777:
In some cases we need to get ignition event even if position is not valid. For example if car starts in underground parking.

fix #2911

### DIFF
--- a/src/org/traccar/events/IgnitionEventHandler.java
+++ b/src/org/traccar/events/IgnitionEventHandler.java
@@ -30,10 +30,7 @@ public class IgnitionEventHandler extends BaseEventHandler {
     @Override
     protected Collection<Event> analyzePosition(Position position) {
         Device device = Context.getIdentityManager().getDeviceById(position.getDeviceId());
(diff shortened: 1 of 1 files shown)


tananaev:
Merged, thanks.

Arturo2017:
Thanks @Abyss777 and @tananaev !!!

That could be my problem. Not by parking the car underground, I was doing tests of first configuration (connected to an auxiliary battery inside my house)

PR #2960, opened 2017-02-27
Server-wide "device readonly" flag

Abyss777:
Also allowed "device readonly" users send commands

It is server part for this https://github.com/tananaev/traccar-web/issues/411


tananaev:
Merged, thanks.

PR #2971, opened 2017-03-02
Implement SMS commands

Abyss777:
Unfortunately a lot of changes at one pull-request...

Key changes:

- Added map `protocol name` -> `protocol` for all protocols in `ServerManager`, we need protocol object access at any time. `deviceId` -> `lastPosition.getProtocol()` -> `protocol`
- Extended `BaseProtocol` to set list of supported sms commands and send sms command function and the main is store sms encoder
- `BaseProtocolSmsEncoder` extends `StringProtocolEncoder` there are few benefits:

  - We can use `initDevicePassword` and `formatCommand` functions for sms
  -  We can combine gprs and sms commands in one encoder in case they are the same or similar

- Implemented `sendCommand` and `getSupportedCommands` functions in `DeviceManager`
- Faced with one problem, my device (and I think most of other) understand commands only in ASCII or GSM encoding, but I think most of users want get notifications in their language (unicode UCS2). I had to introduce separate charset config parameters for notifications and for commands in `SmppClient`
- There are two example implementations:
  - Wondex has the same SMS format and encoder combined in one
  - Granit has different set of SMS commnds and encoders are split 

Other comments in the code...


src/org/traccar/api/resource/CommandTypeResource.java
```diff
@@ -34,9 +34,9 @@
 public class CommandTypeResource extends BaseResource {
 
     @GET
-    public Collection<CommandType> get(@QueryParam("deviceId") long deviceId) throws SQLException {
+    public Collection<CommandType> get(@QueryParam("deviceId") long deviceId, @QueryParam("sms") boolean sms) {
```
Abyss777: The same request for sms and gprs commands

src/org/traccar/database/ActiveDevice.java
```diff
@@ -47,16 +43,6 @@ public long getDeviceId() {
         return deviceId;
     }
 
-    public Collection<CommandType> getCommandTypes() {
```
Abyss777: It is unified in `DeviceManager.getSupportedCommands`

src/org/traccar/database/DeviceManager.java
```diff
@@ -420,4 +427,46 @@ public void resetTotalDistance(DeviceTotalDistance deviceTotalDistance) throws S
...
+                        command.getString(Command.KEY_DATA), true);
+            } else {
+                throw new RuntimeException("Command " + command.getType() + " is not supported");
+            }
+        } else {
+            ActiveDevice activeDevice = Context.getConnectionManager().getActiveDevice(command.getDeviceId());
+            if (activeDevice != null) {
+                activeDevice.sendCommand(command);
+            } else {
+                throw new RuntimeException("Device is not online");
```
Abyss777: Add human-readable exception in case send gprs command to offline device

src/org/traccar/database/DeviceManager.java
```diff
@@ -420,4 +427,46 @@ public void resetTotalDistance(DeviceTotalDistance deviceTotalDistance) throws S
...
+            if (sms) {
+                commands = Context.getServerManager().getProtocol(lastPosition.getProtocol()).getSupportedSmsCommands();
+            } else {
+                commands = Context.getServerManager().getProtocol(lastPosition.getProtocol()).getSupportedCommands();
+            }
+            for (String commandKey : commands) {
+                result.add(new CommandType(commandKey));
+            }
+        } else {
+            result.add(new CommandType(Command.TYPE_CUSTOM));
```
Abyss777: If device was never online, we still will able send custom sms command (for initial configuration for example)

src/org/traccar/protocol/WondexProtocolEncoder.java
```diff
@@ -40,4 +40,9 @@ protected Object encodeCommand(Command command) {
         return null;
     }
 
+    @Override
+    protected String encodeSmsCommand(Command command) {
+        return (String) encodeCommand(command);
```
Abyss777: Both formats are the same

src/org/traccar/smpp/ClientSmppSessionHandler.java
```diff
@@ -44,14 +44,14 @@ public PduResponse firePduRequestReceived(PduRequest request) {
...
             if (request instanceof DeliverSm) {
                 if (request.getOptionalParameters() != null) {
-                    Log.debug("Message Delivered: "
+                    Log.debug("SMS Message Delivered: "
                             + request.getOptionalParameter(SmppConstants.TAG_RECEIPTED_MSG_ID).getValueAsString()
                             + ", State: "
                             + request.getOptionalParameter(SmppConstants.TAG_MSG_STATE).getValueAsByte());
                 } else {
-                    Log.debug("Message Received: "
+                    Log.debug("SMS Message Received: "
```
Abyss777: Small visual improvement, to understand what is message received.

src/org/traccar/smpp/ClientSmppSessionHandler.java
```diff
@@ -44,14 +44,14 @@ public PduResponse firePduRequestReceived(PduRequest request) {
...
                             + CharsetUtil.decode(((DeliverSm) request).getShortMessage(),
-                            smppClient.mapDataCodingToCharset(((DeliverSm) request).getDataCoding()))
+                            smppClient.mapDataCodingToCharset(((DeliverSm) request).getDataCoding())).trim()
```
Abyss777: My wondex device add new line at the end of message, log looks not very nice.

src/org/traccar/smpp/SmppClient.java
```diff
@@ -184,31 +193,41 @@ private void destroySession() {
         }
     }
 
-    public synchronized void sendMessageSync(String destAddress, String message) throws RecoverablePduException,
-            UnrecoverablePduException, SmppTimeoutException, SmppChannelException, InterruptedException {
+    public synchronized void sendMessageSync(String destAddress, String message, boolean command)
```
Abyss777: `command` flag to choose charset

src/org/traccar/smpp/SmppClient.java
```diff
@@ -184,31 +193,41 @@ private void destroySession() {
...
             submit.setSourceAddress(new Address(sourceTon, sourceNpi, sourceAddress));
             submit.setDestAddress(new Address(destTon, destNpi, destAddress));
-            submit.setDataCoding(dataCoding);
-            submit.setShortMessage(textBytes);
             SubmitSmResp submitResponce = getSession().submit(submit, submitTimeout);
-            Log.debug("SMS submited, msg_id: " + submitResponce.getMessageId());
+            if (submitResponce.getCommandStatus() == SmppConstants.STATUS_OK) {
+                Log.debug("SMS submited, msg_id: " + submitResponce.getMessageId());
+            } else {
+                throw new IllegalStateException(submitResponce.getResultMessage());
```
Abyss777: In some cases communication with smpp server can be successful, but really something wrong. For example destination address. We need to log this.

src/org/traccar/api/resource/CommandResource.java
```diff
@@ -32,10 +37,11 @@
 public class CommandResource extends BaseResource {
 
     @POST
-    public Response add(Command entity) {
+    public Response add(Command entity) throws RecoverablePduException, UnrecoverablePduException,
```
tananaev: Maybe we should just declare generic exception here.

src/org/traccar/database/DeviceManager.java
```diff
@@ -420,4 +427,46 @@ public void resetTotalDistance(DeviceTotalDistance deviceTotalDistance) throws S
             throw new IllegalArgumentException();
         }
     }
+
+    public void sendCommand(Command command) throws RecoverablePduException, UnrecoverablePduException,
```
tananaev: Same. I think it's better to throw generic exception.

src/org/traccar/database/DeviceManager.java
```diff
@@ -420,4 +427,46 @@ public void resetTotalDistance(DeviceTotalDistance deviceTotalDistance) throws S
...
+        }
+    }
+
+    public Collection<CommandType> getCommandTypes(long deviceId, boolean sms) {
+        List<CommandType> result = new ArrayList<>();
+        Position lastPosition = Context.getDeviceManager().getLastPosition(deviceId);
+        if (lastPosition != null) {
+            Collection<String> commands;
+            if (sms) {
+                commands = Context.getServerManager().getProtocol(lastPosition.getProtocol()).getSupportedSmsCommands();
```
tananaev: I think the code will be more readable if you get protocol once and then use ternary operator to get corresponding commands list.

src/org/traccar/smpp/SmppClient.java
```diff
@@ -184,31 +193,41 @@ private void destroySession() {
...
-            UnrecoverablePduException, SmppTimeoutException, SmppChannelException, InterruptedException {
+    public synchronized void sendMessageSync(String destAddress, String message, boolean command)
+            throws RecoverablePduException, UnrecoverablePduException, SmppTimeoutException, SmppChannelException,
+            InterruptedException, IllegalStateException {
         if (getSession() != null && getSession().isBound()) {
-            byte[] textBytes = CharsetUtil.encode(message, charsetName);
-
             SubmitSm submit = new SubmitSm();
+            byte[] textBytes;
+            if (command) {
```
tananaev: I think ternary operator would be easier here as well.

src/org/traccar/database/DeviceManager.java
```diff
@@ -420,4 +427,46 @@ public void resetTotalDistance(DeviceTotalDistance deviceTotalDistance) throws S
             throw new IllegalArgumentException();
         }
     }
+
+    public void sendCommand(Command command) throws RecoverablePduException, UnrecoverablePduException,
+            SmppTimeoutException, SmppChannelException, InterruptedException {
+        if (command.getSms()) {
```
tananaev: Should we automatically try SMS command if device is not online? Possibly a configuration option.

src/org/traccar/protocol/WondexProtocolEncoder.java
tananaev: I like that it can be reused.

src/org/traccar/BaseProtocolSmsEncoder.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar;
+
+import org.traccar.model.Command;
+
+public abstract class BaseProtocolSmsEncoder extends StringProtocolEncoder {
+
+    protected abstract String encodeSmsCommand(Command command);
```
tananaev: Do we need a separate method? I think we never share encoders anyway, right?

src/org/traccar/database/DeviceManager.java
```diff
@@ -420,4 +426,48 @@ public void resetTotalDistance(DeviceTotalDistance deviceTotalDistance) throws S
...
+                throw new RuntimeException("Command " + command.getType() + " is not supported");
+            }
+        } else {
+            ActiveDevice activeDevice = Context.getConnectionManager().getActiveDevice(deviceId);
+            if (activeDevice != null) {
+                activeDevice.sendCommand(command);
+            } else {
+                if (fallbackToSms) {
+                    command.setSms(true);
+                    sendCommand(command);
```
Abyss777: Don't you mind the recursion?

Abyss777:
`BaseProtocolSmsEncoder` is not needed now.
Fixed all other comments.
Added `command.fallbackToSms` config to allow fallback to sms command if device is not online

src/org/traccar/smpp/SmppClient.java
```diff
@@ -184,31 +193,36 @@ private void destroySession() {
...
+            submit.setDataCoding(command ? commandsDataCoding : notificationsDataCoding);
+            submit.setShortMessage(textBytes);
             submit.setSourceAddress(new Address(sourceTon, sourceNpi, sourceAddress));
             submit.setDestAddress(new Address(destTon, destNpi, destAddress));
-            submit.setDataCoding(dataCoding);
-            submit.setShortMessage(textBytes);
             SubmitSmResp submitResponce = getSession().submit(submit, submitTimeout);
-            Log.debug("SMS submited, msg_id: " + submitResponce.getMessageId());
+            if (submitResponce.getCommandStatus() == SmppConstants.STATUS_OK) {
+                Log.debug("SMS submited, msg_id: " + submitResponce.getMessageId());
```
tananaev: There is a typo here. Should be "submitted". Also, I think `msg_id` should be removed or converted to something more readable, like "id" or "message id".

src/org/traccar/BaseProtocol.java
```diff
@@ -67,4 +86,25 @@ public void sendCommand(ActiveDevice activeDevice, Command command) {
...
+            if (command.getType().equals(Command.TYPE_CUSTOM)) {
+                Context.getSmppManager().sendMessageSync(phone, command.getString(Command.KEY_DATA), true);
+            } else if (supportedSmsCommands.contains(command.getType()) && smsEncoder != null) {
+                Context.getSmppManager().sendMessageSync(phone, (String) smsEncoder.encodeCommand(command), true);
+            } else {
+                throw new RuntimeException(
+                        "Command " + command.getType() + " is not supported in protocol " + getName());
+            }
+        } else {
+            throw new RuntimeException("Smpp client is not enabled");
```
tananaev: Change "Smpp" to "SMPP" here. Camel-case naming convention applies to Java code, but log messages should follow English language rules. SMPP is an abbreviation, so should be all upper case, same as SMS.

src/org/traccar/smpp/SmppClient.java
```diff
@@ -184,31 +193,36 @@ private void destroySession() {
...
+            } else {
+                throw new IllegalStateException(submitResponce.getResultMessage());
+            }
         } else {
             throw new SmppChannelException("Smpp session is not connected");
```
tananaev: Change "Smpp" to "SMPP" here. In the log message, not Java class name.

src/org/traccar/Protocol.java
```diff
@@ -16,4 +21,9 @@
 
     void initTrackerServers(List<TrackerServer> serverList);
 
+    Collection<String> getSupportedSmsCommands();
+
+    void sendSmsCommand(String phone, Command command) throws RecoverablePduException, UnrecoverablePduException,
```
tananaev: I think it's better to declare generic `Exception` here as well. Otherwise too many things to catch. Maybe we should create an issue in cloudhopper repository to suggest a base cloudhopper exception for all of their exceptions.

Abyss777:
OK. Sorry.
Fixed log messages and combined exceptions in `Protocol`

Abyss777:
Am I missing the comment? Do not see it

src/org/traccar/BaseProtocol.java
```diff
@@ -67,4 +86,25 @@ public void sendCommand(ActiveDevice activeDevice, Command command) {
         }
     }
 
+    public void setSmsEncoder(StringProtocolEncoder smsEncoder) {
+        this.smsEncoder = smsEncoder;
+    }
+
+    @Override
+    public void sendSmsCommand(String phone, Command command) throws RecoverablePduException, UnrecoverablePduException,
```
tananaev: The comment was here. Please use generic Exception here as well.

Abyss777:
It's done.

tananaev:
Merged, thanks.

Abyss777:
What do you think about renaming some functions in `Protocol` and `BaseProtocol` ?
For example call usual commands `GprsCommands` or `ChannelCommands` 
`Protocol.sendCommand()` -> `Protocol.sendChannelCommand()`
`BaseProtocol.setSupportedCommands` -> `BaseProtocol.setSupportedChannelCommands`

If protocol has the same set of commands for GPRS and SMS then we can have `setSupportedCommands` to set both types.

The only thing is that renaming will touch all protocols.

tananaev:
Maybe we should use terms "data" and "text" instead of "GPRS" and "SMS" correspondingly?

Abyss777:
Good idea, we won't be locked on transport

Abyss777:
Should we also rename `Command` model filed `boolean sms` to `boolean text`? It might be confusing for API users.

tananaev:
We can rename the field to something like `textChannel` in the `Command`.

Abyss777:
Should we rename all protocol encoders from `*ProtocolEncoder` to `*ProtocolDataEncoder` ?
And use `*ProtocolTextEncoder` for SMS encoders

tananaev:
I would say no... there would be too many changes. Maybe at some point in future when we have many text encoders.

PR #2973, opened 2017-03-03
Wait cookie credentials in base64

Abyss777:
We need it to correctly pass emails and passwords with symbols and international

Both parameters must be "escaped" "base64" "UTF8" string

`admin` -> `YWRtaW4=` -> `YWRtaW4%3D`
`pa$$word` -> `cGEkJHdvcmQ=` -> `cGEkJHdvcmQ%3D`
`пароль` -> `0L/QsNGA0L7Qu9GM` -> `0L/QsNGA0L7Qu9GM`


tananaev:
Merged, thanks.

PR #2974, opened 2017-03-03
Schedule smpp reconnection only if current task stopped

Abyss777:
Schedule reconnection task only if previous finished
Fix infinite re-connection in some cases


tananaev:
Merged.

PR #2982, opened 2017-03-06
Introduce terms "data" and "text" commands

Abyss777:
Mostly renaming functions and variables
The only thing was added is `setSupportedCommands` to set both types if they are the same.

Used `destAddress` instead of `phone` in `sendTextCommand` function, just not to lock on SMS as transport.

### DIFF
--- a/src/org/traccar/database/DeviceManager.java
+++ b/src/org/traccar/database/DeviceManager.java
@@ -56,14 +56,14 @@ public class DeviceManager implements IdentityManager {
 
     private final Map<Long, Position> positions = new ConcurrentHashMap<>();
 
-    private boolean fallbackToSms;
+    private boolean fallbackToText;
 
     public DeviceManager(DataManager dataManager) {
         this.dataManager = dataManager;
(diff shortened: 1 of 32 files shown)


src/org/traccar/database/DeviceManager.java
```diff
@@ -56,14 +56,14 @@
...
-    private boolean fallbackToSms;
+    private boolean fallbackToText;
 
     public DeviceManager(DataManager dataManager) {
         this.dataManager = dataManager;
         this.config = Context.getConfig();
         dataRefreshDelay = config.getLong("database.refreshDelay", DEFAULT_REFRESH_DELAY) * 1000;
         lookupGroupsAttribute = config.getBoolean("deviceManager.lookupGroupsAttribute");
-        fallbackToSms = config.getBoolean("command.fallbackToSms");
+        fallbackToText = config.getBoolean("command.fallbackToSms");
```
Abyss777: Decided to leave config name to be more clear for users

tananaev:

PR #2986, opened 2017-03-09
SMPP: Allow set different source address for commands

Abyss777:
There is a case:
Some users may want send notifications from short or alphanumeric number, but commands should be send from "routable" international or national number to catch answer from device.

Also there may be different price for SMSs from short number and from "routable" number. Commands are not so frequent as notifications and it may allow save some money.

Added 3 new optional config parameters in this way: if `commandSourceAddress` omitted used `sourceAddress`. Same for TON and NPI.


tananaev:
Makes sense. Merged.

PR #2987, opened 2017-03-09
Removed execute permissions from source code files;

jon-stumpf:
No code changes.  Just some permissions cleanup.


tananaev:
Merged, thanks. I also noticed this problem recently. I guess those files were originally created on Windows.

PR #3001, opened 2017-03-13
Timezone field to User and Server model

### LINKED ISSUE #2969: Feature Request - SMS - Time Zone As per User set zone
Hello,

Currently when a SMS is sent, the time displayed in the Message us 'UTC' which could be hard for some end-users to understand, and also confuse them.

Web-UI already displays the time as per the client time zone,. similarly for emails/sms etc can the admin and or user set his preferred time zone for the server or per user, based on which the time/date would be displayed in notifications etc.

### LINKED ISSUE #2989: TimeZone in Email Notification Is Wrong
Hi. I'm using Traccar with GT02 Gps tracker. I installed th server to my VPS and it works perfect.
But date and times in e-mail notifications are wrong.

Tracker's time is correct, my server's time is correct. And they are same. It's OK. I can see the times in log file and web-ui. I'm in GMT+3 but email says GMT+5

what is wrong?
I'm using Traccar3.10-Win

Abyss777:
Key changes:
- Added fields to `User` and `Server` models
- Pass Velocity `DateTool` to templates as `$dateTool` variable
- Adjusted templates to use `timezone`

fix #2969 and fix #2989

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -144,6 +144,11 @@
             <version>1.7</version>
         </dependency>
         <dependency>
+		    <groupId>org.apache.velocity</groupId>
+		    <artifactId>velocity-tools</artifactId>
+		    <version>2.0</version>
+		</dependency>
+        <dependency>
             <groupId>org.mnode.ical4j</groupId>
             <artifactId>ical4j</artifactId>
             <version>2.0.0</version>
--- a/pom.xml
+++ b/pom.xml
@@ -144,10 +144,10 @@
             <version>1.7</version>
         </dependency>
         <dependency>
-		    <groupId>org.apache.velocity</groupId>
-		    <artifactId>velocity-tools</artifactId>
-		    <version>2.0</version>
-		</dependency>
+            <groupId>org.apache.velocity</groupId>
+            <artifactId>velocity-tools</artifactId>
+            <version>2.0</version>
+        </dependency>
         <dependency>
             <groupId>org.mnode.ical4j</groupId>
             <artifactId>ical4j</artifactId>
--- a/src/org/traccar/database/PermissionsManager.java
+++ b/src/org/traccar/database/PermissionsManager.java
@@ -25,7 +25,6 @@ import org.traccar.model.Server;
 import org.traccar.model.User;
 import org.traccar.model.UserPermission;
 
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.sql.SQLException;
 import java.util.ArrayList;
@@ -413,8 +412,7 @@ public class PermissionsManager {
             if (method != null) {
                 serverPreference = method.invoke(server, (Object[]) null);
             }
-        } catch (NoSuchMethodException | SecurityException | IllegalAccessException
-                | IllegalArgumentException | InvocationTargetException exception) {
+        } catch (ReflectiveOperationException | SecurityException | IllegalArgumentException exception) {
             return defaultValue;
         }
         if (server.getForceSettings()) {
--- a/src/org/traccar/database/PermissionsManager.java
+++ b/src/org/traccar/database/PermissionsManager.java
@@ -25,6 +25,8 @@ import org.traccar.model.Server;
 import org.traccar.model.User;
 import org.traccar.model.UserPermission;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -393,6 +395,34 @@ public class PermissionsManager {
 
     public User getUserByToken(String token) {
         return users.get(usersTokens.get(token));
+    }
+
+    public Object lookupPreference(long userId, String key, Object defaultValue) {
+        String methodName = "get" + key.substring(0, 1).toUpperCase() + key.substring(1);
+        Object preference;
+        Object serverPreference = null;
+        Object userPreference = null;
+        try {
+            Method method = null;
+            method = User.class.getMethod(methodName, (Class<?>[]) null);
+            if (method != null) {
+                userPreference = method.invoke(users.get(userId), (Object[]) null);
+            }
+            method = null;
+            method = Server.class.getMethod(methodName, (Class<?>[]) null);
+            if (method != null) {
+                serverPreference = method.invoke(server, (Object[]) null);
+            }
+        } catch (NoSuchMethodException | SecurityException | IllegalAccessException
+                | IllegalArgumentException | InvocationTargetException exception) {
+          return defaultValue;
+        }
+        if (server.getForceSettings()) {
+            preference = serverPreference != null ? serverPreference : userPreference;
+        } else {
+            preference = userPreference != null ? userPreference : serverPreference;
+        }
+        return preference != null ? preference : defaultValue;
     }
 
 }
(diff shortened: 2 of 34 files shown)


src/org/traccar/model/Server.java
```diff
@@ -166,4 +168,17 @@ public void setCoordinateFormat(String coordinateFormat) {
         this.coordinateFormat = coordinateFormat;
     }
 
+    private String timezone;
+
+    public void setTimezone(String timezone) {
+        if (timezone != null) {
+            this.timezone = TimeZone.getTimeZone(timezone).getID();
```
Abyss777: Added validation here. 
`TimeZone.getTimeZone()` return "GMT" if not recognizes timezone and we get always valid timezone ID in DB

src/org/traccar/model/User.java
```diff
@@ -265,4 +266,17 @@ public boolean isPasswordValid(String password) {
         return Hashing.validatePassword(password, hashedPassword, salt);
     }
 
+    private String timezone;
+
+    public void setTimezone(String timezone) {
+        if (timezone != null) {
+            this.timezone = TimeZone.getTimeZone(timezone).getID();
```
Abyss777: The same validation.

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -48,6 +50,9 @@ private static VelocityContext prepareContext(long userId, Event event, Position
             velocityContext.put("geofence", Context.getGeofenceManager().getGeofence(event.getGeofenceId()));
         }
         velocityContext.put("webUrl", Context.getVelocityEngine().getProperty("web.url"));
+        velocityContext.put("dateTool", new DateTool());
+        velocityContext.put("timezone", ReportUtils.getTimezone(userId));
+        velocityContext.put("locale", Locale.getDefault());
```
Abyss777: There is no variant of `DateTool.format()` without locale and with timezone, just pass default one.
The side effect is that default locale may be overridden via JVM parameters. And user can format date in desired locale.

src/org/traccar/reports/ReportUtils.java
```diff
@@ -46,6 +47,14 @@ public static String getSpeedUnit(long userId) {
         return unit != null ? unit : "kn";
     }
 
+    public static TimeZone getTimezone(long userId) {
+        String timezone = Context.getPermissionsManager().getUser(userId).getTimezone();
+        if (timezone == null) {
```
Abyss777: Should we apply `server.forceSettings` here?

src/org/traccar/notification/NotificationFormatter.java
Abyss777: Also may be in future we will store locale for users...

pom.xml
```diff
@@ -144,6 +144,11 @@
             <version>1.7</version>
         </dependency>
         <dependency>
+		    <groupId>org.apache.velocity</groupId>
```
tananaev: Please use only spaces in this file.

src/org/traccar/model/Server.java
tananaev: Use ternary operator here. It should make code easier to read in this case.

src/org/traccar/model/User.java
tananaev: Same as above.

src/org/traccar/notification/NotificationFormatter.java
tananaev: Looks good to me.

src/org/traccar/reports/ReportUtils.java
tananaev: Yes, we should. Don't we have a helper method to automatically do this?

templates/mail/alarm.vm
```diff
@@ -4,7 +4,7 @@
 <body>
 Device: $device.name<br>
 Alarm:  $position.getString("alarm")<br>
-Time: $event.serverTime<br>
+Time: $dateTool.format("YYYY.MM.dd HH:mm:ss", $event.serverTime, $locale, $timezone)<br>
```
tananaev: I think hyphens instead of dot would be more universal in this case. Applies for all templates.

src/org/traccar/reports/ReportUtils.java
Abyss777: We have never used `server.forceSettings` on the server side.

tananaev: I was mostly talking about handling user/server configuration, like we do in the web app. If we never used `forceSettings`, we should probably start doing it. I recommend creating another issue to fix all the places.

src/org/traccar/reports/ReportUtils.java
```diff
@@ -33,23 +33,23 @@ private ReportUtils() {
 
     public static String getDistanceUnit(long userId) {
         String unit = Context.getPermissionsManager().getUser(userId).getDistanceUnit();
-        if (unit == null) {
+        if (unit == null || Context.getPermissionsManager().getServer().getForceSettings()) {
```
tananaev: There is actually problem with this code. If server value is not set, we still want to use user value. Web app handles this situation.

Abyss777:
Sorry for tabs in pom.xml.
Applied `forceSettings` and fixed it
Used ternary operator
Changed date formatting in templates

tananaev:
Now it seems like too much code duplication. Can we share the logic of value lookup? Possibly utilising reflection.

Abyss777:
OK. I'll try.

tananaev:
Let me know if you need any help with it.

Abyss777:
Implemented helper
I placed it in `PermissionManager` because `Sever` and `users` are there.

src/org/traccar/database/PermissionsManager.java
```diff
@@ -395,4 +397,32 @@ public User getUserByToken(String token) {
...
+                userPreference = method.invoke(users.get(userId), (Object[]) null);
+            }
+            method = null;
+            method = Server.class.getMethod(methodName, (Class<?>[]) null);
+            if (method != null) {
+                serverPreference = method.invoke(server, (Object[]) null);
+            }
+        } catch (NoSuchMethodException | SecurityException | IllegalAccessException
+                | IllegalArgumentException | InvocationTargetException exception) {
+          return defaultValue;
```
Abyss777: Should we log the exception here?

src/org/traccar/database/PermissionsManager.java
```diff
@@ -395,4 +397,32 @@ public User getUserByToken(String token) {
...
+            method = User.class.getMethod(methodName, (Class<?>[]) null);
+            if (method != null) {
+                userPreference = method.invoke(users.get(userId), (Object[]) null);
+            }
+            method = null;
+            method = Server.class.getMethod(methodName, (Class<?>[]) null);
+            if (method != null) {
+                serverPreference = method.invoke(server, (Object[]) null);
+            }
+        } catch (NoSuchMethodException | SecurityException | IllegalAccessException
```
tananaev: Most of the exceptions can be combined into `ReflectiveOperationException`. Everything else looks fine.

Abyss777:
Combined, thanks for the tip.

tananaev:
Merged, thanks.

Abyss777:
Will we apply this to Excel reports? Or leave as is?

tananaev:
Makes sense to apply to Excel reports as well.

PR #3011, opened 2017-03-15
Use timezone from preferences for excel reports

Abyss777:
Used the same way as for Velocity templates

Just in case pass `webUrl`, now in excel files used openstreetmap. Not sure if excel files should link to traccar server. Notifications are sent to traccar users, but reports may be resent to not-traccar people.


src/org/traccar/reports/Summary.java
```diff
@@ -90,7 +91,10 @@ public static void getExcel(OutputStream outputStream,
             jxlsContext.putVar("to", to);
             jxlsContext.putVar("distanceUnit", ReportUtils.getDistanceUnit(userId));
             jxlsContext.putVar("speedUnit", ReportUtils.getSpeedUnit(userId));
-            jxlsContext.putVar("timezone", from.getZone());
+            jxlsContext.putVar("webUrl", Context.getVelocityEngine().getProperty("web.url"));
+            jxlsContext.putVar("dateTool", new DateTool());
+            jxlsContext.putVar("timezone", ReportUtils.getTimezone(userId));
+            jxlsContext.putVar("locale", Locale.getDefault());
```
tananaev: Is it possible to share context initialization code? It keeps growing, so it would be nice to remove some code duplication.

tananaev:
As for OSM link, I think we can keep it for now.

Abyss777:
You are right. It should be optimized.

Implemented two functions:
`ReportUtils.initializeContext()` - initializes context with common variables
`ReportUtils.processTemplateWithSheets()` - processes template and delete first(template) sheet after that.

tananaev:
Merged, thanks.

PR #3022, opened 2017-03-17
Handle incoming SMS from devices as Event

Abyss777:
Key changes:

- Schema changes applies to "non-mssql" DB engines only. It "nulls" all "phone" duplicates and add unique constraint
- Added map phone -> devices
- Added new type of Event - `textMessage`

If DB is MS then behavior will be next: In case phone has duplicates last added device will be mapped.

Frankly, I do not like the changes very much, especially null phone if it is `""` in `addDevice` and  `updateDevice` methods. Only multiple nulls are possible with unique constraint and it comes as `""` from web-client by default.
But I do not have other more beautiful solution may be you have another idea...

### DIFF
--- a/src/org/traccar/database/DeviceManager.java
+++ b/src/org/traccar/database/DeviceManager.java
@@ -34,6 +34,7 @@ import org.traccar.model.Command;
 import org.traccar.model.CommandType;
 import org.traccar.model.Device;
 import org.traccar.model.DeviceTotalDistance;
+import org.traccar.model.Event;
 import org.traccar.model.Group;
 import org.traccar.model.Position;
 import org.traccar.model.Server;
@@ -49,6 +50,7 @@ public class DeviceManager implements IdentityManager {
 
     private Map<Long, Device> devicesById;
     private Map<String, Device> devicesByUniqueId;
+    private Map<String, Device> devicesByPhone;
     private AtomicLong devicesLastUpdate = new AtomicLong();
 
     private Map<Long, Group> groupsById;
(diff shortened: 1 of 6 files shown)


tananaev:
Maybe we should just allow phone to be non-unique? Then either use first/last found device, like you do for MS SQL already anyway.

Abyss777:
I have replaced the commit.
Do not require uniqueness for `phone` now.
Also map only not empty string.

Behavior will be the next: last added/updated device phone is actual.

src/org/traccar/database/DeviceManager.java
```diff
@@ -470,4 +507,13 @@ public void sendCommand(Command command) throws Exception {
         }
         return result;
     }
+
+    public void handleTextMessage(String phone, String message) {
```
tananaev: I think this method belongs to connection manager or something. Where do we throw most of the other events?

src/org/traccar/database/DeviceManager.java
```diff
@@ -127,8 +144,14 @@ private void updateDeviceCache(boolean force) throws SQLException {
...
                 }
             }
+            for (String cachedDevicePhone : devicesByPhone.keySet()) {
+                if (!databaseDevicesPhones.contains(cachedDevicePhone)) {
+                    devicesByPhone.remove(cachedDevicePhone);
+                }
+            }
             databaseDevicesIds.clear();
             databaseDevicesUniqueIds.clear();
+            databaseDevicesPhones.clear();
```
tananaev: Why do we clear these 3 sets? I think they will be garbage-collected anyway.

src/org/traccar/database/DeviceManager.java
```diff
@@ -127,8 +144,14 @@ private void updateDeviceCache(boolean force) throws SQLException {
                     devicesByUniqueId.remove(cachedDeviceUniqId);
```
tananaev: Can you please rename `cachedDeviceUniqId` to `cachedDeviceUniqueId`. Not related to this code review, but I just noticed it.

src/org/traccar/database/DeviceManager.java
Abyss777: We throw them in different event handlers. I will create separate EventHandler for that.

Abyss777:
Done
Moved function
Renamed variable
Removed set clearing

src/org/traccar/database/DeviceManager.java
```diff
@@ -122,13 +138,16 @@ private void updateDeviceCache(boolean force) throws SQLException {
                     devicesById.remove(cachedDeviceId);
                 }
             }
-            for (String cachedDeviceUniqId : devicesByUniqueId.keySet()) {
-                if (!databaseDevicesUniqueIds.contains(cachedDeviceUniqId)) {
-                    devicesByUniqueId.remove(cachedDeviceUniqId);
+            for (String cachedDeviceUniqueId : devicesByUniqueId.keySet()) {
+                if (!databaseDevicesUniqueIds.contains(cachedDeviceUniqueId)) {
+                    devicesByUniqueId.remove(cachedDeviceUniqueId);
```
tananaev: Just notices another problem here.

I know it was here before, but I think if this condition happens, there would be an exception. You can't iterate through a collection and remove elements from it at the same time. Can you please double check. I think instead you should be using iterators which allow to remove elements.

Abyss777: You are right, iterator should be used here. I'll fix it
We did not get any exceptions because it is very unlikely to remove something. Only if something changed in DB directly.

Abyss777:
Replaced with Iterator

tananaev:
Merged, thanks.

Issue #3025, opened 2017-03-18
Feature request - Visibility of devices


Arturo2017:
In devices windows, would be great to have a checkbox by devices showing on the map (header default ALL)

![18 de marzo de 2017 21018 cet](https://cloud.githubusercontent.com/assets/25770031/24067731/b9fb40f4-0b80-11e7-8b3b-d4c51350b08b.jpg)

conorwhyman:
Good idea. Would be also useful if we had an option that only the online devices would be visible.

Abyss777:
@tananaev 
Where will we store state of this switches between sessions?

tananaev:
I don't think we should.

@Arturo2017, what if we just had an option to switch between all/online only devices? Would that work for you? If not, please explain the use case.

Abyss777:
Just looking for icons for switcher "All/Online"...
One suggestion is
All - http://fontawesome.io/icon/eye/
Online - http://fontawesome.io/icon/low-vision/

Or

All - http://fontawesome.io/icon/chain-broken
Online - http://fontawesome.io/icon/link

May be some other?

tananaev:
I like the first one.

Arturo2017:
Explanation to my example of the upper image: 1&3 devices are from one group while devices 2,4,5,6 are from another group.

http://traccar.litvak.su/features/visibility.html

tananaev:
I see. We already have filtering by groups and search. Maybe we can use that for filtering devices on map as well. Also, we can add another filter by online/offline status. I think it would cover most of the cases.

Abyss777:
We can add Status column and use this ExtJS feature http://examples.sencha.com/extjs/6.2.0/examples/kitchensink/#grid-filtering

And reflect devices filtering on the map.

tananaev:
Sounds even better than my idea. Then we can remove regular search field.

Abyss777:
The only doubt I have is performance of map markers update.
We will have to add/remove three type of markers from layers. (device, accuracy, liveRoute)

tananaev:
I think if we do it properly, performance impact should be minimal.

Arturo2017:
![img_20170321_105507_691](https://cloud.githubusercontent.com/assets/25770031/24141962/65e76fc2-0e25-11e7-9711-911b6f480bc8.jpg)

Anton&Andrey you are very big!!! THANKS

tananaev:
[...]

PR #3034, opened 2017-03-21
Set status to offline for new device

Abyss777:
I'm working on filtering in web-part discussed in #3025 

By default new device will have `""` as status and handle fourth value (empty string) is very very inconvenient and break logic.

### DIFF
--- a/src/org/traccar/database/DeviceManager.java
+++ b/src/org/traccar/database/DeviceManager.java
@@ -131,7 +131,6 @@ public class DeviceManager implements IdentityManager {
                             device.setGeofenceIds(geofenceManager.getCurrentDeviceGeofences(lastPosition));
                         }
                     }
-                    device.setStatus(Device.STATUS_OFFLINE);
                 }
             }
             for (Iterator<Long> iterator = devicesById.keySet().iterator(); iterator.hasNext();) {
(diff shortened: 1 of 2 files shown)


tananaev:
I think it's better to initialize it in the Device model.

Abyss777:
You are right. Much better. Looks inverted but just not check for `null`

tananaev:
Actually that's not what I had in mind. I thought just making it a default value for the instance variable would be enough.

Abyss777:
Not really, it still be `""` when create device from web client

tananaev:
Are you sure? Does the web app send empty status?

Abyss777:
Yes.
Request
image

Response 
image

tananaev:
This doesn't seem right. Same for other empty strings as well. Maybe there is a way to tell extjs to not send or send `null` if it's not set?

Abyss777:
Agree, that it is not right, I'll try to find the solution from ExtJS side.
I was thinking... It is still not good to allow change Status via API...

Abyss777:
It did not help.
Initialized on server side:
````java
private String status = STATUS_OFFLINE;
````
I used http://docs.sencha.com/extjs/6.2.0/classic/Ext.data.field.String.html#cfg-allowNull on the web side
And got the `null` in response
image

Looks like API explicitly set all incoming values then constructs entity

**Comment by tananaev** [comment]
OK, then I think instead of changing setter, we should change getter to something like this:
```java
return status != null : status : STATUS_OFFLINE;
```

tananaev:
Merged, thanks.

PR #3050, opened 2017-03-24
Format speed in notification templates

### LINKED ISSUE #3049: Format numbers in templates
We need tool to format numbers in templates https://www.traccar.org/forums/topic/overspeed-email-template/
We can pass https://velocity.apache.org/tools/2.0/apidocs/org/apache/velocity/tools/generic/NumberTool.html object in templates.

Abyss777:
Pass `NumberTool` object to templates (also to reports for consistence and case if somebody wants use decimal format)

Also adjusted notification templates to format one digit after point. Used intermediate variable `$speedValue`, because we can not use math everywhere, just in `#set`.
Did not change report templates, because it is already formatted with help of `String.format()`

fix #3049


tananaev:
Merged.

PR #3069, opened 2017-04-02
make methods accessible to ther classes

ivanfmartinez:
Other classes can use the formatter, if this two methods are public.

### DIFF
--- a/src/org/traccar/notification/NotificationFormatter.java
+++ b/src/org/traccar/notification/NotificationFormatter.java
@@ -37,7 +37,7 @@ public final class NotificationFormatter {
     private NotificationFormatter() {
     }
 
-    private static VelocityContext prepareContext(long userId, Event event, Position position) {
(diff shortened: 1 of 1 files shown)


tananaev:
What other classes?

ivanfmartinez:
I'm migrating my classes  that used an PositionUtil  class that does not have merge accepted. #1601 

Making this methods public any class can use this code for template processing ( in my case classes called from extra.handlers configuration ).

I'm moving my code to use the new features that are included in traccar base code like this template processing and geofences.

tananaev:
If we want to use those methods in some other classes, it would probably make sense to move them out of "notification" class.

ivanfmartinez:
my code is doing notification, just using different methods, and using the same template process will be good and will make easier to move later my code directly inside traccar.

I think that maybe the sms method can be split to be more generic and permit other notification types that are just single string template processing like SMS.

tananaev:
Merged, thanks.

PR #3095, opened 2017-04-19
Update MeitrackProtocolDecoder.java

lalamper:
To show correct values for Battery and Power in case of device TC68S it needs some changes in code.
If parameter "meitrack.devicetype" with value "tc68s" exists in global configuration, values are calculated in another way according to official GPRS protocol.

### DIFF
--- a/src/org/traccar/protocol/MeitrackProtocolDecoder.java
+++ b/src/org/traccar/protocol/MeitrackProtocolDecoder.java
@@ -155,7 +155,7 @@ public class MeitrackProtocolDecoder extends BaseProtocolDecoder {
                 position.set(Position.PREFIX_ADC + i, parser.nextHexInt(0));
             }
         }
-        if (!Context.getIdentityManager().getDeviceById(deviceSession.getDeviceId()).getModel()equals("")) {
+        if (!Context.getIdentityManager().getDeviceById(deviceSession.getDeviceId()).getModel().equals("")) {
             switch (Context.getIdentityManager().getDeviceById(deviceSession.getDeviceId()).getModel()) {
                 case "MVT340":
                     position.set(Position.KEY_BATTERY, parser.nextHexInt(0) * 3.0 * 2.0 / 1024.0);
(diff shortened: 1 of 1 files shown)


src/org/traccar/protocol/MeitrackProtocolDecoder.java
```diff
@@ -155,9 +155,14 @@ private Position decodeRegularMessage(Channel channel, SocketAddress remoteAddre
                 position.set(Position.PREFIX_ADC + i, parser.nextHexInt(0));
             }
         }
-
-        position.set(Position.KEY_BATTERY, parser.nextHexInt(0));
-        position.set(Position.KEY_POWER, parser.nextHexInt(0));
+        
+        if (Context.getConfig().hasKey(getProtocolName() + ".devicetype") == "tc68s") {
```
tananaev: We have a special model field in the device model. I think we should use that instead.

lalamper:
I hope I used the correct call to determine model.

src/org/traccar/protocol/MeitrackProtocolDecoder.java
```diff
@@ -155,9 +155,13 @@ private Position decodeRegularMessage(Channel channel, SocketAddress remoteAddre
                 position.set(Position.PREFIX_ADC + i, parser.nextHexInt(0));
             }
         }
-
-        position.set(Position.KEY_BATTERY, parser.nextHexInt(0));
-        position.set(Position.KEY_POWER, parser.nextHexInt(0));
+        if (Context.getIdentityManager().getDeviceById(deviceSession.getDeviceId()).getModel() == "TC68S") {
```
Abyss777: I believe it is better to use `.getModel().equals("TC68S")` to compare strings.

lalamper:
Added logic for devices MVT340/MVT380/MT90/T1/T3/MVT100/MVT600/MVT800/TC68/TC68S/T311/T322X/T333/T355.

tananaev:
There is a lot of code duplications and too many commits. Please remove duplication and squash commits.

lalamper:
Sorry, this blows my head up! How the fuck I can squash commits??? I really do not understand github. There is no possiblity in web. From shell I have no idea about branches, patches, whatever.. I have no overview sorry.

PR #3099, opened 2017-04-20
Show proper Battery and Power values - Update MeitrackProtocolDecoder.java

lalamper:
In case of Meitrack devices to show the correct Battery and Power values some additional calculation required.
I had to recreate the whole request, github drived me crazy...


tananaev:
Merged, thanks.

PR #3100, opened 2017-04-20
Adding and correcting event numbers - Update MeitrackProtocolDecoder.java

lalamper:
Adding and correcting event numbers reported by device:

17 - ALARM_LOW_BATTERY: when device's internal battery is low
18 - ALARM_LOW_POWER: when external power measure by the device is under limit (limit configured in device)
22 - ALARM_POWER_CUT: when external power disconnected
23 - ALARM_POWER_RESTORED: when external power reconnected (this alarm needed to be created in other files)


tananaev:
What is `ALARM_POWER_RESTORED`? Does this code actually compile?

lalamper:
There is an event code for this in Meitrack protocol: Event 22.
I think it would be nice if it can generate an alarm.
As I wrote, maybe other changes are requed somewhere else in the source (model/Position.java??)

PR #3118, opened 2017-04-25
Adding event handler for detecting engine start or stop using car's power voltage

lalamper:
This event handler is for detecting engine start or stop using car's power voltage.

ATTRIBUTE_ENGINE_START_STOP_POWER_THRESHOLD can be defined with a specific voltage level of car power.
According to my measures 14.00V can be working level.

If this voltage drops below the threshold and previously it was higher, it means engine has stopped (car's power generator does not operate)
If voltage gets higher than threshold and previously it was lower, it means engine has started (car's power generator operates)

May need to change code in other files as well.

### DIFF
--- a/src/org/traccar/events/EngineStartStopEventHandler.java
+++ b/src/org/traccar/events/EngineStartStopEventHandler.java
@@ -51,7 +51,7 @@ public class EngineStartStopEventHandler extends BaseEventHandler {
 
         // Getting current power
         double power = position.getAttributes().containsKey(Position.KEY_POWER);
-        
+
         // Getting old power
         double oldPower = 0;
         Position lastPosition = Context.getIdentityManager().getLastPosition(position.getDeviceId());
--- a/src/org/traccar/events/EngineStartStopEventHandler.java
+++ b/src/org/traccar/events/EngineStartStopEventHandler.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2017 Anton Tananaev (anton@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.events;
+
+import org.traccar.BaseEventHandler;
+import org.traccar.Context;
+import org.traccar.model.Device;
+import org.traccar.model.Event;
+import org.traccar.model.Position;
+
+import java.util.Collection;
+import java.util.Collections;
+
+public class EngineStartStopEventHandler extends BaseEventHandler {
+
+    // Define new attribute
+    public static final String ATTRIBUTE_ENGINE_START_STOP_POWER_THRESHOLD = "engineStartStopPowerThreshold";
+
+    @Override
+    protected Collection<Event> analyzePosition(Position position) {
+
+        Device device = Context.getIdentityManager().getDeviceById(position.getDeviceId());
+        if (device == null) {
+            return null;
+        }
+        if (!Context.getIdentityManager().isLatestPosition(position)) {
+            return null;
+        }
+
+        // Getting ATTRIBUTE_ENGINE_START_STOP_POWER_THRESHOLD value
+        double engineStartStopPowerThreshold = Context.getDeviceManager()
+                .lookupAttributeDouble(device.getId(), ATTRIBUTE_ENGINE_START_STOP_POWER_THRESHOLD, 0, false);
+
+        // If no power threshold specified for current device using attributes
+        if (engineStartStopPowerThreshold == 0) {
+            return null;
+       }
+
+        // Getting current power
+        double power = position.getAttributes().containsKey(Position.KEY_POWER);
+        
+        // Getting old power
+        double oldPower = 0;
+        Position lastPosition = Context.getIdentityManager().getLastPosition(position.getDeviceId());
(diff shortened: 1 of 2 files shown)


lalamper:
Hi Anton,
Can you please help me out with this type error?
Thanks!

double power = position.getAttributes().containsKey(Position.KEY_POWER);
[ERROR] /home/travis/build/tananaev/traccar/src/org/traccar/events/EngineStartStopEventHandler.java:[53,60] incompatible types
[ERROR] required: double
[ERROR] found:    boolean

oldPower = lastPosition.getAttributes().containsKey(Position.KEY_POWER);
[ERROR] /home/travis/build/tananaev/traccar/src/org/traccar/events/EngineStartStopEventHandler.java:[59,64] incompatible types
[ERROR] required: double
[ERROR] found:    boolean

tananaev:
There is already an ignition event handler that does this. Why do we need another one?

lalamper:
Current ignition handler only looks for KEY_IGNITION value, which needed to be set by the protocol decoder implementation. It works only if the currently used protocol and device support this information (most of the devices do not provide it)

My event handler works different, independently from this flag in the protocol. It follows different logic, based on voltage level, it can be used with all protocol which provides car's voltage.

On the other hand, ignition and starting/stopping the engine are not the same exactly. In modern cars ignition is a status without running engine, but with all electric part prepared operating.

I was thinking to modify the current event handler or create a new, I decided to set up a new.
What do you think?

tananaev:
I think the right thing would probably be to set "ignition" flag based on the voltage level instead of implementing completely separate event.

lalamper:
Did you mean to implement it into each protocol one by one, or into the current IgnitionEventHandler.java with a condition to check for this voltage level attribute?

tananaev:
No, you can implement a pipeline handler for that as well, so it would work for any protocol.

lalamper:
Sorry, I do not know how pipeline handler works and how to do that.
Is there any documentation for that?

tananaev:
You've implemented one.

lalamper:
"You've implemented one." - this does not help me, sorry. I am not a programmer. Maybe a little bit more explanation would be fine.

tananaev:
So you don't even know what you implemented? I would recommend to read Netty 3 documentation for more explanation.

lalamper:
I will try to implement this feature into existing IgnitionEventHandler.java. If voltage level attribute exists it will be used to detect ignition.
Are you OK with this?

tananaev:
No, I think it should be another separate handler.

lalamper:
Ok, I will try to finalize EngineStartStopEventHandler.java in the next few days.

tananaev:
Well, I think you still misunderstand what I proposed.

lalamper:
Maybe if you could put a little bit more effort into pointing me to the right direction, I would be more happy. Sorry, but your answer really does not help.

tananaev:
I am proposing a handler that would add ignition flag to the position based on the power value, if it's not reported by the device.

lalamper:
Which file needs to be modified?

tananaev:
You need to create a new class.

lalamper:
Hi Anton,

I think we should reconsider the whole approach. There are two ways to implement this.

1. Creating a separate event handler (EngineStartStopEventHandler.java) for detecting engine start/stop based on the voltage level if threshold set.
This would work independently from the existing ignition event handler. In this case a new event handler java needed.

I think this is the better solution, because ignition on/off and engine start/stop are completely different actions in real world.

2. Or we can use the existing ignition event handler to manage engine start/stop. For this, existing IgnitionEventHandler.java should be extended to check for the threshold attribute (if no KEY_IGNITION flag set by the protocol) and generate ignition event based on voltage level.
In this case there is no need a separate handler java file, because it has no sense.

tananaev:
Both options are different from what I'm proposing. I don't like either of them because you are assuming that "ignition" flag is used only for ignition notification.

lalamper:
Not really see what else ignition flag can indicate? Let's see a simple case.

You insert your car key and turn it into stage 1. Now you have ignition, but the engine does not run. If a device can report this, ignition on event generated. If you turn your key to stage 2, you still have ignition and a running engine. Now you turn backwards your key, back to stage 1: your engine is stopped, but you still have ignition. In next step your turn your key backwards again: ignition is off.

If this does not OK with you, then what is your proposal (logic)? I do not still see what you mention.

If I understand well, ignition flag is set (or not) by the protocol at first. Then come the event handlers into the picture, one by one. They are generating events based on some logic.

tananaev:
I suggest to calculate ignition flag based on the power level if protocol doesn't include it. Similar to distance handler.

Ignition flag can potentially be used for trips report or something else in future.

lalamper:
Ok, I will try to implement soon.

ermalbaj:
Lalamper did you implemented this feature ?
Im interested too...

lalamper:
Hi, unfortunatelly I have very few free time in the last month, but I did not give up.
It would be nice to finish it. I will try to fix it in the next weeks.
Any help appropriated. :)

tsmgeek:
@tananaev device may report ignition flag as it could be in protocol and/or user decided to not wire that up, so it may be best if this was to be implemented that it can handle that possibility.

@lalamper  were you able to make the suggested changes and fix build errors?

tananaev:
I believe this can be implemented using computed attributes.

tsmgeek:
Based on current implementation of Computed Attributes
Event
`ignition ? (ignition == true && power >= 13 ? 'engineStarted':'engineStopped') : power >= 13 ? 'engineStarted':'engineStopped'`
This would cause a constant engineStopped event to be fired.

I still do not think it could work as it needs to compare the last position to the current one to see the transition happen. This could go wrong for protocol decoders like eelink that only partially populate the attributes when they get event from tracker.

leifnel:
This won't work with cars that automatically stop the engine at standstills like red lights and congestions.

But of cause that is not a reason for not building this feature, as long as it is optional.

tsmgeek:
@lalamper have you been able to fix the build issues and update

lalamper:
@tsmgeek 
Hi,
Sorry, no, I had no time for this project, I was alone without any useful support.

PR #3136, opened 2017-05-02
Environment variables

gahissy:
When working with Docker, passing environment variables is more practical than using a configuration file.
This PR changed the `Config` class to look for an environment variable before checking the pre-configured configuration files.
There is an UPPERCASE_SNAKE rule applied to properties names.
Ex:
forward.url --> FORWARD_URL
database.driver --> DATABASE_DRIVER

### DIFF
--- a/src/org/traccar/Config.java
+++ b/src/org/traccar/Config.java
@@ -15,17 +15,19 @@
  */
 package org.traccar;
 
+import org.apache.commons.lang3.StringUtils;
+
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Enumeration;
 import java.util.Properties;
 
 public class Config {
 
     private final Properties properties = new Properties();
-    private Properties defaultProperties;
 
-    public void load(String file) throws IOException {
+    void load(String file) throws IOException {
         try (InputStream inputStream = new FileInputStream(file)) {
             properties.loadFromXML(inputStream);
         }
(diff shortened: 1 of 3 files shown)


src/org/traccar/Config.java
```diff
@@ -15,48 +15,55 @@
  */
 package org.traccar;
 
+import org.apache.commons.lang3.StringUtils;
```
tananaev: I think we should avoid using this indirect dependency, especially if it's easy to use standard Java.

src/org/traccar/Config.java
```diff
@@ -15,48 +15,55 @@
...
 
         String defaultConfigFile = properties.getProperty("config.default");
         if (defaultConfigFile != null) {
             try (InputStream inputStream = new FileInputStream(defaultConfigFile)) {
-                defaultProperties = new Properties();
+                Properties defaultProperties = new Properties();
                 defaultProperties.loadFromXML(inputStream);
+
+                Enumeration props = defaultProperties.propertyNames();
+                while (props.hasMoreElements()) {
```
tananaev: Do we really need this? I don't like that there are two ways of setting up environment variables.

src/org/traccar/Config.java
```diff
@@ -15,48 +15,55 @@
...
-        return properties.containsKey(key) || defaultProperties != null && defaultProperties.containsKey(key);
+        return properties.containsKey(key);
     }
 
     public String getString(String key) {
-        if (properties.containsKey(key) || defaultProperties == null) {
-            return properties.getProperty(key);
-        } else {
-            return defaultProperties.getProperty(key);
+        String envName = key.toUpperCase().replaceAll("\\.", "_");
```
tananaev: Replacing using regular expressions is a pretty heavy operation. I think we should have some global flag to tell server to consider environment variables. Otherwise just use config.

Also, it would be nice to add underscore between words in camel-case parameters. For example `database.ignoreUnknown` should be converted to `DATABASE_IGNORE_UNKNOWN`.

src/org/traccar/Config.java
gahissy: Actually that part is not related to environment variables.
Instead of holding properties into two separate variables `properties` and `defaultProperties` the idea was to merge everything and use only `properties`. That way

tananaev: Sorry, my bad. Totally missed that. Then I think it's a good idea to keep it this way. Is there a better way to merge properties? Properties is a hashtable, so you should be able to just call `putAll` or something like that.

gahissy:
PR updated @tananaev 
- `database.ignoreUnknown` is translated to `DATABASE_IGNORE_UNKNOWN` now
- Test added
- `config.useEnv` (property) and `CONFIG_USE_ENV` (env var) can be used to enable environment variables interpolation
- Default properties and main properties are merged with a cleaner code

tananaev:
I have fixed a problem and did some cleanup including renaming. The problem was the you used properties before loading them.

PR #3139, opened 2017-05-03
Change GT06 command result to result key from command key.

AnshulJain1985:
Hi Anton 

Currently GT06 command response is not creating CommandResult event as it is been treated as command and not result. 
This change will help to handle custom command results as events.

Thanks


AnshulJain1985:
Thanks Anton.

PR #3156, opened 2017-05-10
Implement computed attributes

Abyss777:
Here is implementation.

Structure are very similar to Attribute Aliases. ComputedAttributes are linked to device. Manager and api.resource almost copy.

`ComputedAttribute` model has following fields:
`id`
`deviceId`
`expression` - JEXL expression to compute
`attribute` - name of attribute, where result will be placed
`type` - is type to which result will be cast (number/boolean/string)

Manager is always working, but processing itself can be disabled (in case of performance issues)
I placed `ComputedAttributesHandler` after `CopyAttributes` and before Events handlers.

Also I've moved `CopyAttributesHandler` to `processing` package.

### DIFF
--- a/schema/changelog-3.12.xml
+++ b/schema/changelog-3.12.xml
@@ -22,6 +22,60 @@
         <constraints nullable="false" />
       </column>
     </addColumn>
+    
+    <createTable tableName="attributes">
+      <column name="id" type="INT" autoIncrement="true">
+        <constraints primaryKey="true" />
+      </column>
+      <column name="description" type="VARCHAR(4000)">
+        <constraints nullable="false" />
+      </column>
+      <column name="type" type="VARCHAR(128)">
+        <constraints nullable="false" />
+      </column>
+      <column name="attribute" type="VARCHAR(128)">
+        <constraints nullable="false" />
+      </column>
+      <column name="expression" type="VARCHAR(4000)">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+    
+    <createTable tableName="user_attribute">
+      <column name="userid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="attributeid" type="INT">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <addForeignKeyConstraint baseTableName="user_attribute" baseColumnNames="userid" constraintName="fk_user_attribute_userid" referencedTableName="users" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="user_attribute" baseColumnNames="attributeid" constraintName="fk_user_attribute_attributeid" referencedTableName="attributes" referencedColumnNames="id" onDelete="CASCADE" />
+
+    <createTable tableName="group_attribute">
+      <column name="groupid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="attributeid" type="INT">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <addForeignKeyConstraint baseTableName="group_attribute" baseColumnNames="groupid" constraintName="fk_group_attribute_groupid" referencedTableName="groups" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="group_attribute" baseColumnNames="attributeid" constraintName="fk_group_attribute_attributeid" referencedTableName="attributes" referencedColumnNames="id" onDelete="CASCADE" />
+
+    <createTable tableName="device_attribute">
+      <column name="deviceid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="attributeid" type="INT">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <addForeignKeyConstraint baseTableName="device_attribute" baseColumnNames="deviceid" constraintName="fk_device_attribute_deviceid" referencedTableName="devices" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="device_attribute" baseColumnNames="attributeid" constraintName="fk_device_attribute_attributeid" referencedTableName="attributes" referencedColumnNames="id" onDelete="CASCADE" />
 
   </changeSet>
 
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -18,6 +18,7 @@
     <entry key='logger.file'>./logs/tracker-server.log</entry>
 
     <entry key='event.enable'>true</entry>
+    <entry key='processing.computedAttributes.enable'>true</entry>
 
     <!-- DATABASE CONFIG -->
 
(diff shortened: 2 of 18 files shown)


schema/changelog-3.12.xml
```diff
@@ -22,6 +22,30 @@
...
+      <column name="type" type="VARCHAR(128)">
+        <constraints nullable="false" />
+      </column>
+      <column name="attribute" type="VARCHAR(128)">
+        <constraints nullable="false" />
+      </column>
+      <column name="expression" type="VARCHAR(128)">
+        <constraints nullable="false" />
+      </column>
+      <column name="attributes" type="VARCHAR(4000)">
```
tananaev: Do we really need attributes here?

src/org/traccar/database/ComputedAttributesManager.java
```diff
@@ -0,0 +1,104 @@
...
+
+import org.traccar.Context;
+import org.traccar.helper.Log;
+import org.traccar.model.ComputedAttribute;
+
+public class ComputedAttributesManager {
+
+    private final DataManager dataManager;
+
+    private final Map<Long, Set<ComputedAttribute>> deviceComputedAttributes = new ConcurrentHashMap<>();
```
tananaev: I think we need to think of a way of applying attributes for all devices and/or specific protocol and/or groups. I think in most cases users would have some number of devices of the same model and similar configuration.

Probably applying to a group would be the most universal solution, but then it might interfere with some other grouping that user might want to do.

What do you think?

Abyss777: Agree, it is reasonable.
Also, agree about using groups. It is most universal. In most cases users will have to rearrange their groups, but I think in most cases it is solvable.

Will we just add `groupId` field or totally split on `device_computed_attributes` and `group_computed_attributes` ?

schema/changelog-3.12.xml
Abyss777: I do not think so, just to consistence with other models.

tananaev: I guess it's not such a big deal. It will be empty by default anyway.

src/org/traccar/database/ComputedAttributesManager.java
tananaev: I guess it would make sense to make it similar to geofences. It would add some extra code, but I think it's the right approach.

Abyss777: Agree, It will be most flexible, but Manager will grow in size. I think it is fine.

tananaev: I guess it's OK for now. I'm thinking if there is a way to better manage all the permissions, but can't think of anything yet.

Abyss777: As for me, linkage simple objects like `device`<->`geofence` is good enough.
Unfortunately, I did not come across with any framework that can manage such permissions, only users, roles and something like that.
The only think I would improve is "multi" groups for groups and devices. Like `group_group` and `group_device`.

May be we could standardize the way how we work with it. Some universal linkage manager...

tananaev: Maybe. Anyway, we can discuss it separately. For now we can continue with current approach.

Abyss777: Sure.
One difference with geofence I see that user will see result of applying `computedAttribute` any way, even if he do not have access to `computedAttribute` directly.

Basically, `user`<->`computedAttribute` link will mean that user can edit it. But if someone else linked `computedAttribute` to `device` everybody who have any access to device will see result of applying. I think it is fine.

By the way, may be you have idea how to call it in other way? `groupComputedAttributes` looks long...

tananaev: Maybe we should just rename "computedAttribute" to "attribute" everywhere? We are not planning to have any other attribute-related tables, and we'll remove aliases at some point as well.

Abyss777: One more question...
After "geofence-like" linking we could not control unique device-> attribute, I mean user can create two `computedAttributes` which will compute the same attribute and link them both to the same device. Moreover user can create 3 attributes and link them on three levels groups.

As for me, logically we should apply attributes continuously from device to top level group. Applying function won't be so simple.

But other logic also is possible...
What do you think?

tananaev: I don't think that we should really cater for that use case. I think whatever is easier to implement. It can be random order, from device to group, or even throw exception if there are two attributes with the same key. I can't think of any real case when user would need multiple computed attributes on the same field.

Abyss777: I thought of why multiple `computedAttributes` needed.
For example `alarm` field. It can be set from different conditions...
And it will be really more simple to have a few `alarm` `computedAttribute` with one "if" than one "5-floor" if-else if- else if ... else.

tananaev: Agree. Then exception is probably not an option, but the order of processing probably doesn't matter much in case of alarm.

Abyss777: Yes. I will implement it without any order, just bare `Set` of all applicable attributes for device.

Abyss777:
Looks like done. Replaced commit.

Implemented the same permissions model as for geofences.
Everything named just `attribute` except `ComputedAttributesHandler`
Added field `description` to model.

tananaev:
Looks good, merged.

Abyss777:
Thanks. Web part is almost ready. Will send a PR after weekend.

I'm thinking over some way of testing expressions. For example some API point that accept `attribute` entity and whole `position` object or just `deviceId` and used `lastPosition` for compute.

tananaev:
Good idea, but I guess we can start without testing API and add it later.

Abyss777:
I just thought.
If "Device Readonly" user link/unlink geofence to device it can not influence on device object, data from device or other users who have access to the device.
But!
If "Device Readonly" user link/unlink `ComputedAttribute` to a device, it will influence on data from the device, for example alarm can be blocked at all, or ignition status overridden.
I think that it is not what we and users wait from "Device Readonly" user.

I would suggest disable access to `ComputedAttributes` for "Device Readonly" users. What do you think?

tananaev:
Not sure. Main reason for using "Device Readonly" is restricting user to specific devices. I think use case is usually that companies sell their devices and provide tracking service and they don't want people to add any other devices or modifying device id or anything like that. They still want people to be able to use all features, so allowing computed attributes might be a good idea. What do you think?

Abyss777:
There are some sensible attributes, that should not be overridden by not very trusted user ("Device Readonly") for example `fuel`, `totalDistance`. Such "not-very-trusted" user can steal fuel or something.

May be we will not allow compute some list of such sensible attributes?

tananaev:
The point is that user owns the device, so he should be able to do anything he wants. The only thing we want to restrict is basically creating new devices and editing unique id, so that he can't use the service with some unauthorised devices. It also makes sense to disable removing, so he can't accidentally remove device and not able to add it back. Maybe we should even allow editing device except unique id.

Abyss777:
I got your point and agree with you from the point of view of service provider. But if we look at it from the point of company manager, he might want restrict manipulation incoming data from devices.
But OK, lets see if will be some feature requests about it.

tananaev:
I think for that case we can use full readonly. Those are two main real use cases.

Abyss777:
Now I'm try to solve another problem.
`position` object can be modified inside the JEXL script, any fields. (just call setter in script)

I see the simplest solution is pass the clone of `position`. I tried to use `org.apache.commons.lang.SerializationUtils.clone()`. It works, but a bit decreases the performance about 10 positions less in 5 seconds ( 380 -> 370)

Also it looks not very good, make `Position` `Serializable` and clone every attribute in loop.

Another option I tried is https://commons.apache.org/proper/commons-jexl/javadocs/apidocs-2.1/org/apache/commons/jexl2/introspection/Uberspect.html just not allow call setters, and some other methods like `add`, `put` etc. This solution does not influence on performance. But there might still be some methods that allow modify the `position` object.

Any ideas appreciated.

tananaev:
I was thinking about making it easier to write the expression. I don't like that you need to call `position.getXxx`. It would be nice if you could just use attribute directly. Do you think it's possible to just pass all the attributes directly into the `expressionContext`? This way we will be able to write something like `adc1 * 0.1 + 10` directly. This will also solve the problem of modifying the position because we just won't pass the full position object. What do you think?

Abyss777:
Now it is possible to use `position.attributes.adc1` directly. I would not limit context to attributes only, It will limit the flexibility, some position fields might be useful for scripting, like `speed`, `valid`, `protocol`, `xxxTime`.

But ... It might be the solution...
Map all fields and attributes as separate variables... The only problem might be is intersection of position field names and attributes names ...

PR #3172, opened 2017-05-18
Unwrap position fields and attributes to computedAttribute context

Abyss777:
Map all position fields and attributes to jexl context with help of reflection.

Expressions look more simple now.

Did not notice any changes in performance.

Added some tests to check if scripts can influence on the original position.

### DIFF
--- a/src/org/traccar/processing/ComputedAttributesHandler.java
+++ b/src/org/traccar/processing/ComputedAttributesHandler.java
@@ -18,8 +18,11 @@ package org.traccar.processing;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.commons.jexl2.JexlEngine;
 import org.apache.commons.jexl2.JexlException;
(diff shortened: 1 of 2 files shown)


src/org/traccar/processing/ComputedAttributesHandler.java
```diff
@@ -36,10 +39,32 @@ public ComputedAttributesHandler() {
         engine.setStrict(true);
     }
 
+    private MapContext prepareContext(Position position) {
+        MapContext result = new MapContext();
+        Method[] methods = position.getClass().getMethods();
+        for (Method method : methods) {
+            if (method.getName().startsWith("get") && method.getParameterTypes().length == 0
+                    && !method.getName().equals("getClass")) {
+                String name = method.getName().substring(3, 4).toLowerCase() + method.getName().substring(4);
```
tananaev: You can do a small optimization here and use `Character.toLowerCase(...)` method on a single `char`.

src/org/traccar/processing/ComputedAttributesHandler.java
```diff
@@ -36,10 +39,32 @@ public ComputedAttributesHandler() {
         engine.setStrict(true);
     }
 
+    private MapContext prepareContext(Position position) {
+        MapContext result = new MapContext();
+        Method[] methods = position.getClass().getMethods();
+        for (Method method : methods) {
+            if (method.getName().startsWith("get") && method.getParameterTypes().length == 0
+                    && !method.getName().equals("getClass")) {
```
tananaev: I suggest some more generic solution for excluding methods here. Something like this:
```java
Set<Method> methods = new HashSet<>(Arrays.asList(position.getClass().getMethods()));
methods.removeAll(Arrays.asList(Object.class.getMethods()));
for (Method method : methods) {
    ...
```

Abyss777:
Thank you for tips. Fixed.

tananaev:
Merged, thanks.

PR #3174, opened 2017-05-19
Adjust battery and batteryLevel type

Abyss777:
Here is part of big common position attributes improvement.
`KEY_BATTERY` and `KEY_BATTERY_LEVEL` today

Added test for number and percents.

One half changed according to documentation, second half according to intuition and logic.
I supposed that `(d+.d+)` usually voltage and `(dd)` usually percents.

The only protocol I do not know how interpret is `taip`, there are only three values in tests: `229`, `4416` and `4190` and no documentation. I have no idea how handle it. Just converted to integer to not fail tests.

### DIFF
--- a/src/org/traccar/protocol/Ardi01ProtocolDecoder.java
+++ b/src/org/traccar/protocol/Ardi01ProtocolDecoder.java
@@ -78,7 +78,7 @@ public class Ardi01ProtocolDecoder extends BaseProtocolDecoder {
         position.set(Position.KEY_SATELLITES, satellites);
 
         position.set(Position.KEY_EVENT, parser.next());
-        position.set(Position.KEY_BATTERY, parser.next());
+        position.set(Position.KEY_BATTERY_LEVEL, parser.nextInt(0));
         position.set(Position.PREFIX_TEMP + 1, parser.next());
 
         return position;
--- a/src/org/traccar/protocol/WondexProtocolDecoder.java
+++ b/src/org/traccar/protocol/WondexProtocolDecoder.java
@@ -110,7 +110,7 @@ public class WondexProtocolDecoder extends BaseProtocolDecoder {
             position.set(Position.KEY_SATELLITES, satellites);
 
             position.set(Position.KEY_EVENT, parser.next());
-            position.set(Position.KEY_BATTERY, parser.nextDouble(0));
+            position.set(Position.KEY_BATTERY, parser.nextDouble());
             if (parser.hasNext()) {
                 position.set(Position.KEY_ODOMETER, parser.nextDouble(0) * 1000);
             }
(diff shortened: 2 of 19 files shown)


test/org/traccar/ProtocolTest.java
```diff
@@ -192,6 +192,16 @@ private void verifyDecodedPosition(Object decodedObject, boolean checkLocation,
             Assert.assertTrue(attributes.get(Position.KEY_FUEL_LEVEL) instanceof Number);
         }
 
+        if (attributes.containsKey(Position.KEY_BATTERY)) {
+            Assert.assertTrue(attributes.get(Position.KEY_BATTERY) instanceof Number);
+        }
+
+        if (attributes.containsKey(Position.KEY_BATTERY_LEVEL)) {
+            Object batteryLevel = attributes.get(Position.KEY_BATTERY_LEVEL);
```
tananaev: I think you can just downcast to Number and get int value here. Instanceof check is not necessary because it will fail with pretty obvious error anyway.

test/org/traccar/ProtocolTest.java
```diff
@@ -192,6 +192,16 @@ private void verifyDecodedPosition(Object decodedObject, boolean checkLocation,
...
+            Assert.assertTrue(batteryLevel instanceof Number);
+            Assert.assertTrue(((Number) batteryLevel).intValue() <=100 && ((Number) batteryLevel).intValue() >= 0);
```
tananaev: Space is missing. Having int variable should simplify this expression.

src/org/traccar/protocol/Ardi01ProtocolDecoder.java
```diff
@@ -78,7 +78,7 @@ protected Object decode(
         position.set(Position.KEY_SATELLITES, satellites);
 
         position.set(Position.KEY_EVENT, parser.next());
-        position.set(Position.KEY_BATTERY, parser.next());
+        position.set(Position.KEY_BATTERY_LEVEL, parser.nextInt(0));
```
tananaev: In this case value is not optional, so it's not a problem, but can you please check that it's case for all the changed protocols. If the value is optional, it's better to use `nextInt` without argument, so the value is not included in the attributes instead of including zero.

src/org/traccar/protocol/WondexProtocolDecoder.java
```diff
@@ -110,7 +110,7 @@ protected Object decode(
             position.set(Position.KEY_SATELLITES, satellites);
 
             position.set(Position.KEY_EVENT, parser.next());
-            position.set(Position.KEY_BATTERY, parser.next());
+            position.set(Position.KEY_BATTERY, parser.nextDouble(0));
```
tananaev: For example here is seems optional, so it's better to use `parser.nextDouble()`. Please check all the places.

Abyss777:
Sorry. You are right, in some places it is optional. Especially in very complicated GL200.
Also fixed `ProtocolTest`.

tananaev:
Merged, thanks.

tananaev:
About the `nextInt` and other similar functions. I have done refactoring and included options with default value to not break any existing code, but in most cases we should be using the ones without default parameter.

PR #3178, opened 2017-05-22
Organize KEY_POWER and KEY_CHARGE

Abyss777:
Organized `KEY_POWER` and `KEY_CHARGE` using in few places

- Sorry, reverted `KEY_POWER` without default value in GoSafe (from last PR)
- "chargerPressure" has nothing to `KEY_CHARGE` in Intellitrac, it is about turbine or compressor or something
- I'm not sure in BlackKite, there is no documentation and any description, but it looks like `KEY_CHARGE`
- And I'm pretty sure that it is `KEY_CHARGE` in Visiontek


tananaev:
Merged, thanks.

PR #3181, opened 2017-05-23
Organize satellites and RSSI values

Abyss777:
Next portion...

Satellites was most easier.
Supposed that RSSI is float dBm value or 0..31 signal strength.

V680 is not RSSI at all, it can contain lbs data

As for Megastek... checked again documentation and fixed battery and power. Also according to documentation satellites can be only number, but some test cases has "0/3" or "0/8" values, I believe that it is used and visible number of satellites.

I checked `KEY_GPS` using for next changes and become upset. It used for very different data.
Frankly, it is very thankless work, organize attributes without documentation. I'm going to make a pause.


tananaev:
Merged, thanks.

Yes, it's hard to organize them. In many cases documentation does not specify format or units for the values.

PR #3183, opened 2017-05-23
Implement testing API for computedAttribute

Abyss777:
Did not create a new API point, just used POST with query parameter.

Pass `deviceId` as query param and try to apply `computedAttribute` to last position, if `deviceId` omitted will be default "add" behavior.

Check both, expression and type casting.


tananaev:
Can we separate it into two methods? Will Java automatically route to the correct one based on query parameters?

Abyss777:
I think jax-rs can not do it by default, some tricks are possible, but it will be more complex than just two functions.

tananaev:
OK, thanks.

PR #3226, opened 2017-06-07
Fix calendar.data type for postgresql

### LINKED ISSUE #3223: Import calendar error
When using traccar with postgres I get this error when I try to import an ics file.
```
  ERROR: column "data" is of type oid but expression is of type bytea
  Hint: You will need to rewrite or cast the expression.
  Position: 68 - PSQLException (... < QueryBuilder:475 < DataManager:508 < CalendarManager:100 < ...)
```
I have tried files generated from iCalendar, Google Calendar and Thunderbird.

Abyss777:
fix #3223 

We need specify type `bytea` for data column for postgresql

### DIFF
--- a/schema/changelog-3.12.xml
+++ b/schema/changelog-3.12.xml
@@ -108,4 +108,19 @@
 
   </changeSet>
 
+  <changeSet author="author" id="changelog-3.12-pgsql">
+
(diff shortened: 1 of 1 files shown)


tananaev:
Merged, thanks.

PR #3235, opened 2017-06-11
MaxError despite Valid mark

5of9:
I get every day jumps over 4000km because the tracker sends valid mark despite it sends useless coordinates, I already set coordinatesMaxError, but now  I see this only works when the position is not valid, isn't it? So I think it would be better to have another way to see if its a tracker error or a tracker flown in a plane or so!? What do you think?

### DIFF
--- a/src/org/traccar/CoordinatesHandler.java
+++ b/src/org/traccar/CoordinatesHandler.java
@@ -43,7 +43,7 @@ public class CoordinatesHandler extends BaseDataHandler {
             double distance = DistanceCalculator.distance(
                     position.getLatitude(), position.getLongitude(), last.getLatitude(), last.getLongitude());
             boolean satisfiesMin = coordinatesMinError == 0 || distance > coordinatesMinError;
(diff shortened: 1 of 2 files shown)


5of9:
Maybe they should also implement a check, if the coordinates are valid and set to invalid by server when it gets coordinates with 300 degrees or so... Server should be more intelligent than those stupid trackers... :)

tananaev:
There is already `filter.maxSpeed` option available to filter out jumps like that.

5of9:
Thank you, I was so fixed on the distance, so I didn't even read the Speedfilter in the source. But what do you think about a filter.nonEarth which filters positions over 90 and 180 degrees, which doesn't exist on earth?
How can I help you with the https://www.traccar.org/configuration-file/ to add speedfilter and some other things?

tananaev:
I don't think it's necessary. Do you have a device that sends data over 90/180 degress?

5of9:
Yes. The TK5000 does sometimes...

tananaev:
Then it can be a part of "valid" filtering.

5of9:
I'll do so. Closed because this is not nessesary

PR #3239, opened 2017-06-12
Added new filter for invalid positions

5of9:
Added a new Filtertype called "NotEarth". It could be used if they have a tracker which submits "valid" marked positions with latitudes or longitudes which doesn't exist on earth... TK5000 does for example


tananaev:
Why do you have 8 commits? Can you please squash your history. Also, you have to fix build issues.

5of9:
Sorry wasnt planned to do so. I close and open the correct one

tananaev:
Can you please not spam in a pull request. I'm closing it. Please create a new one when everything is fixed.

PR #3240, opened 2017-06-13
Fix managed devices enumerating

### LINKED ISSUE #3238: Inconsistent device limit reached
If the same device is also linked to more than one user, this device counts as more than one.
Note: The user must be managed by the primary user. In my case “instalador_sc”

Example:
User manager: “instalador_sc”
Devices: device01; Device02
Device Limit: 3
Managed users:
Client01 (linked – device01 and device02)
Client02 (linked – device01 and device02)
. Total devices associated with user “instalador_sc”: 4
. The user will not be able to create new devices.
. Although the user can no longer create devices he can make new links which further increases the total device limit count

I believe the best solution would be to add user-manager devices to devices created by 'normal' users that are not linked to the user manager.
The total would be analyzed by the field: devices limit

what do you think?

Abyss777:
Enumerate managed devices as `Set` to exclude duplicates

fix #3238

### DIFF
--- a/src/org/traccar/database/DeviceManager.java
+++ b/src/org/traccar/database/DeviceManager.java
@@ -190,7 +190,7 @@ public class DeviceManager implements IdentityManager {
     }
 
     public Collection<Device> getManagedDevices(long userId) throws SQLException {
-        Collection<Device> devices = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


tananaev:
Merged, thanks.

PR #3245, opened 2017-06-13
Added a new Filtertype called "NotEarth".

5of9:
It could be used if they have a tracker which submits "valid" marked positions with latitudes or
longitudes which doesn't exist on earth... TK5000 does for example


tananaev:
I think it should be part of `filterInvalid` filtering and not a separate one.

5of9:
In the Wondex protocol answers to messages are invalid because they don't have a position. If we filter the invalids they would loose these. Maybe in other protocols there could be also a sense for not filter invalids!? That was the reason I thought a new filter would be better.

tananaev:
They should not be invalid. That sounds like an issue in the protocol decoder.

PR #3247, opened 2017-06-14
Imlement Stops report

Abyss777:
About stops:
- "Stop" is a gap between trips
- It's minimal duration can't be less than `report.trip.minimalParkingDuration` except one case: it is in the end of period and after a trip
- Stop address and coordinates are taken from the first position
- Also it has following fields: `duration`, `engineHours`, `spentFuel`

image

About code:
- Moved trips detector to `ReportUtils` and adapted it to generate stops
- Moved `spentFuel` to `BaseReport`
- Implemented `TripsConfig` object just because of limit in function parameters count
- Extended tests for some edge cases

More comments in the code


src/org/traccar/reports/ReportUtils.java
```diff
@@ -123,4 +127,186 @@ public static void processTemplateWithSheets(InputStream templateStream, OutputS
...
+                            if (trips) {
+                                result.add(calculateTrip(
+                                        positions, previousEndParkingIndex, startParkingIndex, ignoreOdometer));
+                            } else {
+                                if (result.isEmpty() && previousEndParkingIndex > previousStartParkingIndex) {
+                                    long previousParkingDuration = positions.get(previousEndParkingIndex)
+                                            .getFixTime().getTime() - positions.get(previousStartParkingIndex)
+                                            .getFixTime().getTime();
+                                    if (previousParkingDuration >= tripsConfig.getMinimalParkingDuration()) {
+                                        result.add(calculateStop(positions, previousStartParkingIndex,
```
Abyss777: Add Stop before first trip if it is long enough

src/org/traccar/reports/ReportUtils.java
```diff
@@ -123,4 +127,186 @@ public static void processTemplateWithSheets(InputStream templateStream, OutputS
...
+                        }
+                        previousEndParkingIndex = endParkingIndex;
+                        skipped = false;
+                    } else {
+                        skipped = true;
+                    }
+                    startParkingIndex = -1;
+                }
+            }
+            if (result.isEmpty() && !trips) {
```
Abyss777: If we did not detected any trips then whole(or piece) period might be a stop

test/org/traccar/reports/TripsAndStopsTest.java
```diff
@@ -0,0 +1,199 @@
...
+import java.util.Collection;
+import java.util.Date;
+import java.util.TimeZone;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+public class TripsAndStopsTest extends BaseTest {
```
tananaev: Name of the test file should exactly match the class that you are testing (and have `Test` suffix).

src/org/traccar/reports/ReportUtils.java
```diff
@@ -123,4 +127,186 @@ public static void processTemplateWithSheets(InputStream templateStream, OutputS
...
+            }
+        }
+        stop.setEngineHours(engineHours);
+
+        return stop;
+
+    }
+
+    public static Collection<BaseReport> detectTripsAndStops(TripsConfig tripsConfig,
+             boolean ignoreOdometer, double speedThreshold,
```
tananaev: There is indentation problem somewhere here.

Abyss777:
Fixed indentation and moved tests to correct place `ReportUtilsTest`

Agree, it is unreadable and agree do not think there is a way to make it clear. I can try add comments to algorithm, but I think make it looks worse.

tananaev:
Merged, thanks. Is web part coming as well?

kirantpatil:
Does it also shows stop points on the map ?

Abyss777:
@kirantpatil The question is more related to this PR https://github.com/tananaev/traccar-web/pull/504
But yes, stop points are shown on the map.

kirantpatil:
Great and Thanks @Abyss777

PR #3249, opened 2017-06-14
Solidify OwnTracks protocol decoder

jpmens:
You kindly merged, the new [OwnTracks Protocol Decoder](https://github.com/tananaev/traccar/pull/3242), and this pull request attempts to address some things I omitted the first time around. In particular:

1. Ensures most attributes in [OwnTracks JSON](http://owntracks.org/booklet/tech/json/#_typelocation) are optional: only `lat`, `lon`, and `_type` are mandatory.
2. Checks for payloads of `"_type" : "location"` and uses only those. (OwnTracks has other payload types, e.g. "transition" which probably shouldn't be used in Traccar.)
3. Addresses an originally too simplistic method of creating a Traccar _deviceId_ using our _tracking ID_ (`tid`). Starting now, if the payload contains our `topic` (e.g. `owntracks/jpmens/iphone5s`) that will be used in preference to avoid collisions.
4. Adds another unit test.

I hope this is OK.

We're updating our documentation to [include Traccar](http://owntracks.org/booklet/features/traccar/), and that could be the landing for "device type OwnTracks" on your [https://www.traccar.org/devices/](https://www.traccar.org/devices/) page, for example. Just a thought.

### DIFF
--- a/src/org/traccar/protocol/OwnTracksProtocolDecoder.java
+++ b/src/org/traccar/protocol/OwnTracksProtocolDecoder.java
@@ -94,15 +94,17 @@ public class OwnTracksProtocolDecoder extends BaseProtocolDecoder {
 
         position.setTime(new Date(root.getJsonNumber("tst").longValue() * 1000));
 
-        String uniqueId = "";
-        if (root.containsKey("tid")) {
-            uniqueId = root.getString("tid");
-        }
-        if (root.containsKey("topic")) {
+        Boolean haveTopic = root.containsKey("topic");
+        Boolean haveTid = root.containsKey("tid");
+        String uniqueId = null;
+
+        if (haveTopic) {
             uniqueId = root.getString("topic");
-            if (root.containsKey("tid")) {
+            if (haveTid) {
                 position.set("tid", root.getString("tid"));
             }
+        } else {
+            uniqueId = root.getString("tid");
         }
 
         DeviceSession deviceSession = getDeviceSession(channel, remoteAddress, uniqueId);
@@ -115,5 +117,4 @@ public class OwnTracksProtocolDecoder extends BaseProtocolDecoder {
         sendResponse(channel, HttpResponseStatus.OK);
         return position;
     }
-
 }
--- a/src/org/traccar/protocol/OwnTracksProtocolDecoder.java
+++ b/src/org/traccar/protocol/OwnTracksProtocolDecoder.java
@@ -58,6 +58,13 @@ public class OwnTracksProtocolDecoder extends BaseProtocolDecoder {
         JsonObject root = Json.createReader(
                 new StringReader(request.getContent().toString(StandardCharsets.US_ASCII))).readObject();
 
+        if (!root.containsKey("_type") || !root.getString("_type").equals("location")) {
+            sendResponse(channel, HttpResponseStatus.BAD_REQUEST);
+            return null;
+        }
+
+        long timestamp;
+        String deviceId = new String();
         Position position = new Position();
         position.setProtocol(getProtocolName());
         position.setValid(true);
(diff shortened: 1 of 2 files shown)


tananaev:
Please fix merge issues. I did some cleanup.

jpmens:
I think you added the line 

```
position.setTime(new Date(root.getJsonNumber("tst").longValue() * 1000));
```

which I've removed again because that would cause the code to fail in the (unlikely) event that the JSON doesn't contain `tst`; that's the reason for the block above it.

test/org/traccar/protocol/OwnTracksProtocolDecoderTest.java
```diff
@@ -17,6 +17,8 @@ public void testDecode() throws Exception {
         verifyPosition(decoder, request(HttpMethod.POST, "/",
                 buffer("{\"cog\":271,\"lon\":2.29513,\"acc\":5,\"vel\":61,\"vac\":21,\"lat\":48.85833,\"tst\":1497349316,\"alt\":167,\"_type\":\"location\",\"tid\":\"JJ\",\"t\":\"u\",\"batt\":67}")));
 
+        verifyPosition(decoder, request(HttpMethod.POST, "/",
+                buffer(text("{\"lat\":48.85,\"lon\":2.295,\"_type\":\"location\",\"tid\":\"JJ\"}"))));
```
tananaev: Please remove `text` here.

src/org/traccar/protocol/OwnTracksProtocolDecoder.java
```diff
@@ -58,6 +58,13 @@ protected Object decode(
         JsonObject root = Json.createReader(
                 new StringReader(request.getContent().toString(StandardCharsets.US_ASCII))).readObject();
 
+        if (!root.containsKey("_type") || !root.getString("_type").equals("location")) {
+            sendResponse(channel, HttpResponseStatus.BAD_REQUEST);
+            return null;
+        }
+
+        long timestamp;
+        String deviceId = new String();
```
tananaev: Why is there a `new String()` here. That's just wrong. Also, the name should be `uniqueId`. We use `deviceId` for internal device identifier. Also, why are these variables at the top?

src/org/traccar/protocol/OwnTracksProtocolDecoder.java
```diff
@@ -80,16 +87,34 @@ protected Object decode(
...
+        }
         if (root.containsKey("batt")) {
             position.set(Position.KEY_BATTERY, root.getInt("batt"));
         }
-        if (root.containsKey("topic")) {
-            position.set("topic", root.getString("topic"));
+
+        if (root.containsKey("tst")) {
+            timestamp = root.getJsonNumber("tst").longValue();
+            if (timestamp < Integer.MAX_VALUE) {
```
tananaev: I have removed this code. I believe it's unnecessary.

src/org/traccar/protocol/OwnTracksProtocolDecoder.java
```diff
@@ -80,16 +87,34 @@ protected Object decode(
...
         }
+        position.setTime(new Date(timestamp));
 
-        position.setTime(new Date(root.getJsonNumber("tst").longValue() * 1000));
+        if (root.containsKey("tid")) {
+            deviceId = root.getString("tid");
+        }
+        if (root.containsKey("topic")) {
+            deviceId = root.getString("topic");
+            if (root.containsKey("tid")) {
```
tananaev: We already do this check. You can move it to the code above.

jpmens: Not really. If `topic` exists, we want `tid` as an attribute, in addition to uniqueId set to `topic`.

tananaev: I'm sure there is a better way to write this code without checking twice.

jpmens:
I have, I hope, addressed all your comments.

src/org/traccar/protocol/OwnTracksProtocolDecoder.java
```diff
@@ -80,16 +85,29 @@ protected Object decode(
...
             position.set(Position.KEY_BATTERY, root.getInt("batt"));
         }
-        if (root.containsKey("topic")) {
-            position.set("topic", root.getString("topic"));
-        }
 
         position.setTime(new Date(root.getJsonNumber("tst").longValue() * 1000));
 
-        DeviceSession deviceSession = getDeviceSession(channel, remoteAddress, root.getString("tid"));
+        Boolean haveTopic = root.containsKey("topic");
```
tananaev: You shouldn't use boxed primitives. Also, there is no need for variables here at all.

src/org/traccar/protocol/OwnTracksProtocolDecoder.java
```diff
@@ -80,16 +85,29 @@ protected Object decode(
...
+        Boolean haveTid = root.containsKey("tid");
+        String uniqueId = null;
```
tananaev: You don't need to assign `null`. here.

tananaev: You don't need to assign `null`. here.

src/org/traccar/protocol/OwnTracksProtocolDecoder.java
```diff
@@ -80,16 +85,29 @@ protected Object decode(
         if (root.containsKey("t")) {
             position.set("t", root.getString("t"));
         }
+        if (root.containsKey("p")) {
+            position.set("pb", root.getInt("p"));
```
tananaev: Wait, I just noticed this. What is "pb"?

jpmens: `pb` was atmospheric pressure; it is reported by our iOS app on iPhone >= 6. I've removed it for the time being; it's not so important.

tananaev:
It seems like you haven't read me comments carefully enough. Why do you need variables for something that you are using once?

jpmens:
Apologies: I thought the code looked clearer.

tananaev:
Merged, thanks.

jpmens:
Thank you.

PR #3250, opened 2017-06-14
Filter invalid positions and handle messages in wondex as valid

5of9:


src/org/traccar/protocol/WondexProtocolDecoder.java
```diff
@@ -78,7 +78,7 @@ protected Object decode(
             position.setProtocol(getProtocolName());
             position.setDeviceId(deviceSession.getDeviceId());
             getLastLocation(position, new Date());
-            position.setValid(false);
+            position.setValid(true);
```
tananaev: This should not be here.

src/org/traccar/protocol/WondexProtocolDecoder.java
```diff
@@ -107,7 +107,7 @@ protected Object decode(
             position.setAltitude(parser.nextDouble(0));
 
             int satellites = parser.nextInt(0);
-            position.setValid(satellites >= 3);
+            position.setValid(satellites >= 3 && position.getLatitude() < 90.0 && position.getLongitude() < 180.0);
```
tananaev: Here "valid" should be just set to "true".

src/org/traccar/FilterHandler.java
```diff
@@ -84,7 +84,7 @@ public FilterHandler() {
     }
 
     private boolean filterInvalid(Position position) {
-        return filterInvalid && !position.getValid();
+        return filterInvalid && !position.getValid() || position.getLatitude() > 90 || position.getLongitude() > 180;
```
tananaev: It would make sense to check minimum values as well.

src/org/traccar/FilterHandler.java
```diff
@@ -84,7 +84,9 @@ public FilterHandler() {
     }
 
     private boolean filterInvalid(Position position) {
-        return filterInvalid && !position.getValid();
+        return filterInvalid && !position.getValid()
+               || position.getLatitude() > 90 || position.getLongitude() > 180
```
tananaev: You should only filter if the filter is enabled. Please be very careful and test your code before submitting anything.

tananaev:
I'm not sure what's happening here, but can you please write proper git commits in English and clean up your commit history. Also, you have to fix all the comments that I had before we can merge this code.

tananaev:
I'm closing this because I get to many notifications about each push. Please re-open when everything is fixed.

PR #3252, opened 2017-06-15
Detect Trips and Stops by gaps in data

Abyss777:
Introduced `report.trip.minimalNoDataDuration` config parameter. It is 0 (disabled) by default.

Check time between current and "next" position if gap is more than config, then this is start of parking period.


tananaev:
I think we should pick some reasonable default value. One hour sounds good to me.

Also, if you have time it would be nice if you can write some documentation for the website. I was thinking about some diagram of how trips are calculated and where different configuration parameters fit in. So people can understand how trips work and why they might not work in their case. What do you think?

Abyss777:
OK, set default value to one hour

Sure, I'll try to prepare some visual documentation.

tananaev:
Merged, thanks.

PR #3253, opened 2017-06-15
Filter Invalid upgrade and Wondex fix

5of9:
"filter.invalid" filters now impossible positions. 
In Wondex protocol answers to messages are no longer marked as invalid, even without position.
Sat over 2 check removed.
Wondex decoder test changed to pass because of the removed Sat over 2 check

### DIFF
--- a/src/org/traccar/FilterHandler.java
+++ b/src/org/traccar/FilterHandler.java
@@ -84,12 +84,9 @@ public class FilterHandler extends BaseDataHandler {
     }
 
     private boolean filterInvalid(Position position) {
-        if (filterInvalid) {
-           return !position.getValid() || position.getLatitude() > 90
-           || position.getLongitude() > 180 || position.getLatitude() < -90
-           || position.getLongitude() < -180;
-        }
-        return false;
+        return filterInvalid && (!position.getValid()
+           || position.getLatitude() > 90 || position.getLongitude() > 180
+           || position.getLatitude() < -90 || position.getLongitude() < -180);
     }
 
     private boolean filterZero(Position position) {
--- a/src/org/traccar/FilterHandler.java
+++ b/src/org/traccar/FilterHandler.java
@@ -84,7 +84,12 @@ public class FilterHandler extends BaseDataHandler {
     }
 
     private boolean filterInvalid(Position position) {
-        return filterInvalid && !position.getValid();
+        if (filterInvalid) {
+           return !position.getValid() || position.getLatitude() > 90
+           || position.getLongitude() > 180 || position.getLatitude() < -90
+           || position.getLongitude() < -180;
+        }
+        return false;
     }
 
     private boolean filterZero(Position position) {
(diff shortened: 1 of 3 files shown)


src/org/traccar/protocol/WondexProtocolDecoder.java
```diff
@@ -107,7 +106,7 @@ protected Object decode(
             position.setAltitude(parser.nextDouble(0));
 
             int satellites = parser.nextInt(0);
```
tananaev: Please remove this variable. It's no longer needed.

src/org/traccar/FilterHandler.java
```diff
@@ -84,7 +84,12 @@ public FilterHandler() {
     }
 
     private boolean filterInvalid(Position position) {
-        return filterInvalid && !position.getValid();
+        if (filterInvalid) {
+           return !position.getValid() || position.getLatitude() > 90
```
tananaev: Indentation is completely wrong here on all 3 lines.

src/org/traccar/FilterHandler.java
```diff
@@ -84,7 +84,12 @@ public FilterHandler() {
     }
 
     private boolean filterInvalid(Position position) {
-        return filterInvalid && !position.getValid();
```
tananaev: I think you can still use return expression straight away. You just need to groups the rest in parentheses.

Also, I liked the split of latitude and longitude checks on separate lines as you had before. I think it makes it more readable.

tananaev:
Merged, thanks.

Abyss777:
Hello
Frankly I think it is not good idea to make all positions valid for wondex protocol. It is not true.
here is quote from documentation

> 1) The device returns the last valid GPS information upon request regardless the
GPS reception. The parameter of “Number of Satellites” is ‘0’ if there is no GPS
reception or GPS is not fixed. Thus the parameter of “number of satellite” could be
a reference to check whether there is GPS reception or not.

tananaev:
That's true for many protocol. It returns outdated, but valid location. It's still valid, so should be marked as valid. Timestamp would indicate that it's outdated.

Abyss777:
Unfortunately, wondex devices do not report fixTime separately.
Do we have any attribute to mark it "outdated" ?

Abyss777:
Oh, sorry. We have ordinary field for this.

tananaev:
That field is only for marking positions that didn't have any location data. Also, it's not saved in the database.

Are you sure time returned by the device is not "fix time"?

Abyss777:
Yes, first position after morning wake up has previous coordinates, satellites = 0 and current time, not yesterday.

tananaev:
OK, maybe we should check if satellites number is zero.

Abyss777:
And mark such position as invalid?

5of9:
@Abyss777, will you change this, or should I do it? If you want to fix this we could also change jobs, I could fix this, because I would need help with the "duplicate.filter issue" I opened...

Abyss777:
@5of9 It's already implemented and merged in #3258

PR #3255, opened 2017-06-15
Duplicate Filter filtered to much

5of9:
For example commands may not have a new fix time and get filtered out, if the duplicate filter only checks the FixTime. It should also check if the attributes differs, to prevent something like that.

### DIFF
--- a/src/org/traccar/FilterHandler.java
+++ b/src/org/traccar/FilterHandler.java
@@ -15,6 +15,8 @@
  */
 package org.traccar;
 
+import java.util.Iterator;
+
 import org.traccar.helper.DistanceCalculator;
 import org.traccar.helper.Log;
 import org.traccar.helper.UnitsConverter;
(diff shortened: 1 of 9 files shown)


tananaev:
I don't think this would work. Last position might have some attributes that were calculated by the server after filtering. I think we should check if last position contains at least all the attributes from current position or possibly more. Also, for performance reasons I would recommend to just compare keys, not values.

5of9:
Please give me a little hint. I find no way to check all the keys at once, should I make a list of keys and compare last with actual? I think only a few would be nessesary, example:
position.getAttributes().containsKey(Position.KEY_IGNITION).equals(lastPosition.getAttributes().containsKey(Position.KEY_IGNITION))
Thank you

tananaev:
Just go through the keys in current position and check if they exist in the last one.

5of9:
I really tried a lot with iterates and so on, but I don't get it to work, please can you give me a code snippet or a better hint how you would do it, because I think this should get fixed for the next version, but I'm not the java specialist like you are...

tananaev:
You are basically asking to write the whole for you. I'm not sure what the difficulty with iterating through keys and checking if they are available in the other position.

tananaev:
I'm closing this. Please re-open when it's done. For the future I would recommend that you don't push your changes until you are finished. Everyone subscribed to the repo gets a notification about every push.

tananaev:
@5of9, do you plan to fix the issues or you are no longer interested in this?

5of9:
I'm not able to do so. Sorry

tananaev:
Implemented in 9a597a7f902eea46405a345d4398139c13f69655.

PR #3256, opened 2017-06-16
Centralized motion detection

Abyss777:
Implemented motion handler in pipeline and use `Position_KEY_MOTION` for event generation and trips detection.

### DIFF
--- a/src/org/traccar/MotionHandler.java
+++ b/src/org/traccar/MotionHandler.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2017 Anton Tananaev (anton@traccar.org)
+ * Copyright 2017 Andrey Kunitsyn (andrey@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar;
+
+import org.traccar.model.Position;
+
+public class MotionHandler extends BaseDataHandler {
+
+    private double speedThreshold;
+
+    public MotionHandler(double speedThreshold) {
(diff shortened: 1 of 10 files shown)


src/org/traccar/protocol/Gl200ProtocolDecoder.java
```diff
@@ -716,7 +716,7 @@ private Object decodeOther(Channel channel, SocketAddress remoteAddress, String
 
         int reportType = parser.nextInt(0);
         if (type.equals("NMR")) {
-            position.set(Position.KEY_MOTION, reportType);
+            position.set(Position.KEY_MOTION, reportType == 1);
```
Abyss777: According to documentation it can be 1 or 0 only

src/org/traccar/reports/ReportUtils.java
```diff
@@ -220,7 +220,12 @@ private static boolean isMoving(ArrayList<Position> positions, int index,
                 >= tripsConfig.getMinimalNoDataDuration()) {
             return false;
         }
-        return positions.get(index).getSpeed() > speedThreshold;
+        if (positions.get(index).getAttributes().containsKey(Position.KEY_MOTION)
```
Abyss777: We have to leave this complex check to backward compatibility with old data

Abyss777:
Sorry, added unrelated to PR one small fix to Stops detector.

src/org/traccar/MotionHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class MotionHandler extends BaseDataHandler {
+
+    private double speedThreshold;
+
+    public MotionHandler(double speedThreshold) {
+        this.speedThreshold = speedThreshold;
+    }
+
+    public Position calculateMotion(Position position) {
```
tananaev: I don't think we need a separate method here. Just put it into `handlePosition`.

src/org/traccar/reports/ReportUtils.java
tananaev: We should keep it backward compatible for about a year and then fully migrate to the new version. Please create an issue so we don't forget about it.

src/org/traccar/MotionHandler.java
Abyss777: May be also do the same for `DistanceHandler` ?
https://github.com/tananaev/traccar/blob/master/src/org/traccar/DistanceHandler.java#L34

tananaev: Yes please.

tananaev:
Merged, thanks.

PR #3258, opened 2017-06-16
Revert satellites check in wondex and add 'archive' in wialon and granit

Abyss777:
- Test 'archive' attribute for boolean
- Add 'archive' attribute for wialon and granit
- Reverted check satellites in wondex
- Fixed typo in comment


tananaev:
Merged, thanks.

PR #3260, opened 2017-06-16
Correctly ignore OwnTracks publishes of type != location

jpmens:
OwnTracks apps can publish payloads with distinct `_type`s in the JSON payload, but Traccar should process only `_type: location`. However, if the apps publish, say, a `_type: transition` and Traccar returns 400, the app will continue to attempt delivery for ever.

This PR fixes that by having Traccar return HTTP status 200 to the apps, indicating "ok" but otherwise i gnoring that particular data.


tananaev:
Merged, thanks.

Issue #3268, opened 2017-06-20
Pipeline optimization


Abyss777:
I see two places for optimization:
1. Distance calculation. We calculate it 4 times in the worst case
https://github.com/tananaev/traccar/blob/master/src/org/traccar/FilterHandler.java#L114
https://github.com/tananaev/traccar/blob/master/src/org/traccar/FilterHandler.java#L124
https://github.com/tananaev/traccar/blob/master/src/org/traccar/CoordinatesHandler.java#L43
https://github.com/tananaev/traccar/blob/master/src/org/traccar/DistanceHandler.java#L45

We can move `DistanceHandler` before `FilterHandler` and `CoordinatesHandler`, but there is one problem: If `CoordinatesHandler` get worked we need reset distance and undo total distance.
One option is split on `DistanceHandler` (before filter and coordinates) and `TotalDistanceHandler` (leave on current place)

2. Move reverse geocoder after filter and coordinates filter https://www.traccar.org/forums/topic/reverse-geocode-still-attempted-for-filtered-positions/

It will prevent unnecessary reverse geocoder requests(that usually cost money) and make address more consistence with coordinates if `CoordinatesHandler` get worked.

tananaev:
Maybe we should just combine `DistanceHandler` and `CoordinatesHandler` into one handler?

Actually reverse geocoder also calculates distance to avoid extra requests, so we can probably just use `distance` there is `DistanceHandler` is first in the pipeline.

Abyss777:
About second one...
https://github.com/tananaev/traccar/blob/master/src/org/traccar/BasePipelineFactory.java#L201
What is better? Move `filter` up before `geocoder`
````java
        if (distanceHandler != null) {
            pipeline.addLast("distance", distanceHandler);
        }

        if (filterHandler != null) {
            pipeline.addLast("filter", filterHandler);
        }

        if (geocoderHandler != null) {
            pipeline.addLast("geocoder", geocoderHandler);
        }
        pipeline.addLast("remoteAddress", new RemoteAddressHandler());

        addDynamicHandlers(pipeline);

        if (motionHandler != null) {
            pipeline.addLast("motion", motionHandler);
        }
````
or move `geocoder` down after filter?
````java
        if (distanceHandler != null) {
            pipeline.addLast("distance", distanceHandler);
        }

        pipeline.addLast("remoteAddress", new RemoteAddressHandler());

        addDynamicHandlers(pipeline);

        if (filterHandler != null) {
[...]

PR #3269, opened 2017-06-20
Distance calculation optimization

Abyss777:
Made optimization discussed in #3268 

- Combine 'DistanceHandler' and 'CoordinatesHandler'
- Move 'DistanceHandler' closer to beginning of pipeline (after coordinates set, but before geocoder)
- Use KEY_DISTANCE instead of distance recalculation

One difference with previous is that `distance` provided by device used for everything.


tananaev:
Merged, thanks.

PR #3271, opened 2017-06-20
Do reverse geocoding after filtering

Abyss777:
Moved `geocoder` down in pipeline after `filter` to prevent unnecessary requests.


tananaev:
Merged, thanks.

PR #3275, opened 2017-06-21
Osmand: Store 'batt' parameter as 'batteryLevel'

### LINKED ISSUE #3274: Battery level for osmand
Traccar client report phone battery level as `&batt=` and it is stored as `KEY_BATTERY`. That is wrong, it should use `KEY_BATTERY_LEVEL`.

I'm not sure what is right, change server to store `&batt=` as `KEY_BATTERY_LEVEL` or change Traccar Client to send phone battery level as `&batteryLevel=`

What do you think?

Abyss777:
fix #3274


tananaev:
Merged, thanks.

PR #3293, opened 2017-06-26
Fix SMPP reconnection on enquire link failed

Abyss777:
We need schedule Reconnection Task here, not just only one attempt to reconnect


tananaev:
Merged.

PR #3309, opened 2017-06-30
[NEW] add GET device to API for external monitoring

ckrey:


tananaev:
I would prefer it to be consistent with the way PositionResource works. The `id` is passed through query parameters and it accepts an array.

src/org/traccar/api/resource/DeviceResource.java
```diff
@@ -42,20 +44,31 @@
...
+                Context.getPermissionsManager().checkUser(getUserId(), userId);
+                return Context.getDeviceManager().getDevices(userId);
             }
         } else {
-            if (userId == 0) {
-                userId = getUserId();
+            ArrayList<Device> devices = new ArrayList<>();
+            for (Long deviceId : deviceIds) {
+                Device device = Context.getDeviceManager().getDeviceById(deviceId);
+                Context.getPermissionsManager().checkDevice(getUserId(), device.getId());
```
tananaev: We can do permission check first before getting device. Also, there is no need for `device` variable.

tananaev:
Merged, thanks.

PR #3348, opened 2017-07-13
Adding ARNAVI 4 binary protocol's implementation

binakot:
This is refreshed version of [deprecated pull-request](https://github.com/tananaev/traccar/pull/2503) :octocat:

The implementation of [ARNAVI 4 binary protocol](https://docs.google.com/spreadsheets/d/15s-2ZbqOQ1bZvAtFFm9sIEuKy3jbJzxdeynp72sjoYU/pubhtml) with basic tags support:
* Latitude
* Longitude
* Additional telemetry data (speed, satellites, altitude, course)

ARNAVI version 4 is not compatible with more early versions of this protocol.


tananaev:
Please fix build issues.

binakot:
Already working on it! 👍

binakot:
Done! 🍹

src/org/traccar/helper/Checksum.java
```diff
@@ -243,4 +243,12 @@ public static long luhn(long imei) {
         return (10 - (checksum % 10)) % 10;
     }
 
+    public static int modulo256(byte... bytes) {
```
tananaev: I think it should be array, not vararg.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
```diff
@@ -0,0 +1,51 @@
+package org.traccar.protocol;
+
+import org.jboss.netty.buffer.ChannelBuffer;
+import org.jboss.netty.channel.Channel;
+import org.jboss.netty.channel.ChannelHandlerContext;
+import org.jboss.netty.handler.codec.frame.FrameDecoder;
+
+/**
+ * Created by Ivan Muratov @binakot on 12.07.2017.
```
tananaev: Please remove this comment. Copyright can be added at the top.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
```diff
@@ -0,0 +1,51 @@
...
+    protected Object decode(
+            ChannelHandlerContext ctx,
+            Channel channel,
+            ChannelBuffer buf) throws Exception {
+
+        if (buf.readableBytes() == 0) {
+            return null;
+        }
+
+        byte[] bytes = new byte[buf.readableBytes()];
```
tananaev: Why do we need to copy it into array? I think it's unnecessary.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
```diff
@@ -0,0 +1,51 @@
...
+        buf.getBytes(0, bytes);
+
+        if (bytes[0] == HEADER_START_SIGN
+                && bytes.length == HEADER_LENGTH
+                && (bytes[1] == HEADER_VERSION_1 || bytes[1] == HEADER_VERSION_2)) {
```
tananaev: Check length.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
```diff
@@ -0,0 +1,51 @@
...
+            return buf.readBytes(HEADER_LENGTH);
+        }
+
+        int parcelNumber = bytes[1] & 0xFF;
+        if (bytes[0] == PACKAGE_START_SIGN && bytes[bytes.length - 1] == PACKAGE_END_SIGN
```
tananaev: Also, check length before accessing data.

src/org/traccar/protocol/Arnavi4Protocol.java
```diff
@@ -0,0 +1,33 @@
...
+import org.jboss.netty.bootstrap.ServerBootstrap;
+import org.jboss.netty.channel.ChannelPipeline;
+import org.traccar.BaseProtocol;
+import org.traccar.TrackerServer;
+
+import java.nio.ByteOrder;
+import java.util.List;
+
+/**
+ * Created by Ivan Muratov @binakot on 11.07.2017.
```
tananaev: Same.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.List;
+
+import static org.traccar.protocol.Arnavi4FrameDecoder.HEADER_START_SIGN;
+import static org.traccar.protocol.Arnavi4FrameDecoder.HEADER_VERSION_1;
+import static org.traccar.protocol.Arnavi4FrameDecoder.HEADER_VERSION_2;
+import static org.traccar.protocol.Arnavi4FrameDecoder.PACKAGE_START_SIGN;
+import static org.traccar.protocol.Arnavi4FrameDecoder.PACKAGE_END_SIGN;
+
+/**
+ * Created by Ivan Muratov @binakot on 11.07.2017.
```
tananaev: Same.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,170 @@
...
+            byte recordStartSign = buf.readByte();
+            while (recordStartSign != PACKAGE_END_SIGN) {
+                switch (recordStartSign) {
+                    case RECORD_PING:
+                    case RECORD_DATA:
+                    case RECORD_TEXT:
+                    case RECORD_FILE:
+                    case RECORD_BINARY:
+                        int length = buf.readUnsignedShort();
+                        long timestamp = buf.readUnsignedInt() * 1000;
```
tananaev: We can create `Date` object straight away.

test/org/traccar/helper/ChecksumTest.java
```diff
@@ -4,6 +4,7 @@
 import org.jboss.netty.buffer.ChannelBuffers;
 import org.junit.Assert;
 import org.junit.Test;
+import org.traccar.protocol.Arnavi4ProtocolDecoder;
```
tananaev: Why do we need this here?

test/org/traccar/protocol/Arnavi4FrameDecoderTest.java
```diff
@@ -0,0 +1,37 @@
+package org.traccar.protocol;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.traccar.ProtocolTest;
+
+import java.nio.ByteOrder;
+
+/**
+ * Created by Ivan Muratov @binakot on 13.07.2017.
```
tananaev: Remove.

test/org/traccar/protocol/Arnavi4ProtocolDecoderTest.java
```diff
@@ -0,0 +1,37 @@
+package org.traccar.protocol;
+
+import org.junit.Test;
+import org.traccar.ProtocolTest;
+
+import java.nio.ByteOrder;
+
+/**
+ * Created by Ivan Muratov @binakot on 11.07.2017.
```
tananaev: Remove.

test/org/traccar/protocol/Arnavi4ProtocolDecoderTest.java
```diff
@@ -0,0 +1,37 @@
...
+        decoder = new Arnavi4ProtocolDecoder(new Arnavi4Protocol());
+
+        verifyNull(decoder, binary(ByteOrder.LITTLE_ENDIAN, // Valid HEADER v1 packet with IMEI
+                "ff22f30c45f5c90f0300"));
+
+        verifyPositions(decoder, binary(ByteOrder.LITTLE_ENDIAN, // Valid PACKAGE packet with one DATA packet
+                "5b01012800a3175f5903513934420447221c42055402781E0900f0c5215b4e0084005c00007c005d0000a300fa37010000295d"),
+                position("2017-07-07 05:09:55.000", true, 45.05597, 39.03347));
+
+        decoder = new Arnavi4ProtocolDecoder(new Arnavi4Protocol());
```
tananaev: Split into separate test case.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,170 @@
...
+
+                        if (recordStartSign == RECORD_DATA) {
+                            positions.add(decodePosition(deviceSession, recordBuf, timestamp));
+                        }
+
+                        buf.readUnsignedByte(); // crc
+                        break;
+
+                    default:
+                        throw new IllegalArgumentException("unsupported record type");
```
tananaev: Should we throw here? Usually we ignore messages if type is unknown. Is it possible?

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,170 @@
...
+            while (recordStartSign != PACKAGE_END_SIGN) {
+                switch (recordStartSign) {
+                    case RECORD_PING:
+                    case RECORD_DATA:
+                    case RECORD_TEXT:
+                    case RECORD_FILE:
+                    case RECORD_BINARY:
+                        int length = buf.readUnsignedShort();
+                        long timestamp = buf.readUnsignedInt() * 1000;
+                        ChannelBuffer recordBuf = buf.readBytes(length);
```
tananaev: Probably we don't need a variable. Pass directly into the function.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,170 @@
...
+        DeviceSession deviceSession = getDeviceSession(channel, remoteAddress);
+        if (deviceSession == null) {
+            return null;
+        }
+
+        if (startSign == PACKAGE_START_SIGN) {
+
+            List<Position> positions = new LinkedList<>();
+
+            int parcelNumber = buf.readUnsignedByte();
```
tananaev: We usually call this `index`.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,170 @@
...
+
+            byte recordStartSign = buf.readByte();
```
tananaev: The name is weird. Probably needs renaming.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,170 @@
...
+    private Position decodePosition(DeviceSession deviceSession, ChannelBuffer buf, long timestamp) {
+
+        final Position position = new Position();
+        position.setProtocol(getProtocolName());
+        position.setDeviceId(deviceSession.getDeviceId());
+
+        position.setTime(new Date(timestamp));
+
+        while (buf.readableBytes() > 0) {
+            short tagId = buf.readUnsignedByte();
```
tananaev: This can just be called `tag`.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,170 @@
...
+            int tagValue = buf.readInt();
```
tananaev: I think it's better to decode value depending on the tag type inside cases.

src/org/traccar/helper/Checksum.java
binakot: Done.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: Done. Copyright replaced to the top.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: I did it, because it's necessary to check the last byte in a buffer, or there is the way to check the last byte without reading all bytes to the end?
```java
bytes[bytes.length - 1] == PACKAGE_END_SIGN
```

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: Sorry, I don't understand. What does you mean? There is the check...

```java
bytes.length == HEADER_LENGTH
```

src/org/traccar/protocol/Arnavi4FrameDecoder.java
tananaev: Yes, you can use `getByte(...)` method. It takes index and it doesn't change reader index.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
tananaev: What if your buffer is one byte long? This code will crash.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: Added the minimal packet length in ARNAVI 4 protocol. It's 4 bytes. Now it's impossible to get out-of-range exception.

src/org/traccar/protocol/Arnavi4Protocol.java
binakot: Done.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Done.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Done.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Done.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Renamed to `recordType`. Is it better?

tananaev: Sounds good.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Done.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: We can't just ingore it, because we won't know how much bytes to read. Removed throwing and added `return null`.

test/org/traccar/helper/ChecksumTest.java
binakot: Yeah, sorry, my bad. Forget to optimize the imports. Fixed.

test/org/traccar/protocol/Arnavi4FrameDecoderTest.java
binakot: Done.

test/org/traccar/protocol/Arnavi4ProtocolDecoderTest.java
binakot: Done.

test/org/traccar/protocol/Arnavi4ProtocolDecoderTest.java
binakot: I did it like in similar tests. Okay, splitted.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: It won't. I use &&, but not &. This code will not try to get bytes[1], if condition before is false.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: Great, thanks for help. Done.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Done.

src/org/traccar/helper/Checksum.java
```diff
@@ -243,4 +243,12 @@ public static long luhn(long imei) {
         return (10 - (checksum % 10)) % 10;
     }
 
+    public static int modulo256(byte[] bytes) {
```
tananaev: I though about the name and I think it makes sense to rename it to `modulo8`. Other functions here are named with number of bits.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+package org.traccar.protocol;
+
+import org.jboss.netty.buffer.ChannelBuffer;
+import org.jboss.netty.channel.Channel;
+import org.jboss.netty.channel.ChannelHandlerContext;
+import org.jboss.netty.handler.codec.frame.FrameDecoder;
+
+public class Arnavi4FrameDecoder extends FrameDecoder {
```
tananaev: I have read the code once again and I believe it's completely wrong.

Biggest problem here is that you should never rely on buffer size. The whole purpose of frame decoder is to split stream of data into frames.

1. Am I understanding correctly that first message is always header? If the size is fixed, you should just read first 10 bytes as a header and save a flag somewhere to indicate that header has been read.

2. I believe regular packages should be split using length field. Never return original buffer from frame decoder. You need to extract frame using `readBytes` and send it upstream.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,187 @@
...
+                    position.setLongitude(buf.readFloat());
+                    position.setValid(true);
+                    break;
+
+                case TAG_COORD_PARAMS:
+                    position.setCourse(buf.readUnsignedByte() * 2.0);
+                    position.setAltitude(buf.readUnsignedByte() * 10.0);
+                    byte satellites = buf.readByte();
+                    position.set(Position.KEY_SATELLITES, satellites & 0x0F + (satellites >> 4) & 0x0F); // gps+glonass
+                    position.setSpeed(buf.readUnsignedByte() * 1.852);
```
tananaev: There is a utility class that you can use to convert units.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,187 @@
...
+                    position.setValid(true);
+                    break;
+
+                case TAG_LONGITUDE:
+                    position.setLongitude(buf.readFloat());
+                    position.setValid(true);
+                    break;
+
+                case TAG_COORD_PARAMS:
+                    position.setCourse(buf.readUnsignedByte() * 2.0);
```
tananaev: For these two lines use integer multiplication. It's faster.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,187 @@
...
+            String imei = String.valueOf(buf.readLong());
+            DeviceSession deviceSession = getDeviceSession(channel, remoteAddress, imei);
+
+            if (deviceSession != null && channel != null) {
+
+                final ChannelBuffer response;
+
+                if (version == HEADER_VERSION_1) {
+                    response = ChannelBuffers.dynamicBuffer(ByteOrder.LITTLE_ENDIAN, 4);
+                    response.writeBytes(new byte[]{0x7B, 0x00, 0x00, 0x7D});
```
tananaev: I think we should create a separate method to send response. Please check GT06 or some other binary protocol for an example.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,187 @@
...
+                } else if (version == HEADER_VERSION_2) {
+                    response = ChannelBuffers.dynamicBuffer(ByteOrder.LITTLE_ENDIAN, 9);
+                    response.writeBytes(new byte[]{0x7B, 0x04, 0x00});
+                    byte[] timeBytes = ByteBuffer.allocate(4).putInt((int) (System.currentTimeMillis() / 1000)).array();
+                    response.writeByte(Checksum.modulo256(timeBytes));
+                    response.writeBytes(timeBytes);
+                    response.writeByte(0x7D);
+
+                } else {
+                    throw new IllegalArgumentException("unsupported header version");
```
tananaev: I think this is unnecessary. We usually avoid throwing exceptions from decoder. This way there is a good chance that server would still work even if protocol is updated.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,187 @@
...
+
+            return null;
+        }
+
+        DeviceSession deviceSession = getDeviceSession(channel, remoteAddress);
+        if (deviceSession == null) {
+            return null;
+        }
+
+        if (startSign == PACKAGE_START_SIGN) {
```
tananaev: I think this check is unnecessary as well.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,187 @@
...
+
+            List<Position> positions = new LinkedList<>();
+
+            int index = buf.readUnsignedByte();
+
+            byte recordType = buf.readByte();
+            while (recordType != PACKAGE_END_SIGN) {
```
tananaev: This is incorrect check. What if binary data happen to contain same byte as end sign? I think we should check length.

src/org/traccar/helper/Checksum.java
binakot: `modulo 256 checksum` is official name of this alghoritm. But I can rename this one, if you really want it.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: 1. Yes, the first packet is always header. I will add the flag for that.
2. There is no length value in beginning of the packet. We never know where is the real end of packet. We only know that in the end must be end sign byte 0x5D. But I can write some smart check for that. Working on it :octocat:

src/org/traccar/helper/Checksum.java
tananaev: No, if it's official name keep it.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Fixed, thanks.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
tananaev: Wait, but what if binary data in between happen to have 0x5D in it? Documentation says that there is a length field.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Do you store the speed in knots or smth else? ARNAVI4 send the speed in knots by default.

tananaev: Traccar stores value in knots.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Done.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Done.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: You cannot be 100% sure of anythin 😄 But okay, I will remove it.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Okay, I will do some additional check.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: We have package with packets inside. Every packet has own length, but total length of package is unknown. I will add additional calculation for that, but it's required read every header of internal packets inside package.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
tananaev: I don't think it should be additional. I think we should just replace it with length check.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: As a matter of fact it's impossible that the end sign byte will part of data section. On this position in a buffer may be the end sign or packet type bytes. Data blocks have strongly fixed size.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: Okay, I re-write the frame decoder and change buffer reading while its over. Let's see...

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,185 @@
...
+
+    private static final byte TAG_LATITUDE = 3;
+    private static final byte TAG_LONGITUDE = 4;
+    private static final byte TAG_COORD_PARAMS = 5;
+
+    public Arnavi4ProtocolDecoder(Arnavi4Protocol protocol) {
+        super(protocol);
+    }
+
+    private void sendHeaderResponse(Channel channel, byte version) {
```
tananaev: The idea was to unify all responses, not just extract pieces of code into separate methods. Please avoid using array. Check any protocol encoder for example on how we usually construct a message.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,185 @@
...
+        if (channel != null) {
+            final ChannelBuffer response;
+            if (version == HEADER_VERSION_1) {
+                response = ChannelBuffers.dynamicBuffer(ByteOrder.LITTLE_ENDIAN, 4);
+                response.writeBytes(new byte[]{0x7B, 0x00, 0x00, 0x7D});
+            } else if (version == HEADER_VERSION_2) {
+                response = ChannelBuffers.dynamicBuffer(ByteOrder.LITTLE_ENDIAN, 9);
+                response.writeBytes(new byte[]{0x7B, 0x04, 0x00});
+                byte[] timeBytes = ByteBuffer.allocate(4).putInt((int) (System.currentTimeMillis() / 1000)).array();
```
tananaev: I'm sure there is a better way to write 4 bytes. Can we use `writeInt`?

src/org/traccar/protocol/Arnavi4FrameDecoder.java
```diff
@@ -0,0 +1,70 @@
...
+            return null;
+        }
+
+        if (buf.getByte(0) == HEADER_START_SIGN
+                && buf.readableBytes() == HEADER_LENGTH
+                && (buf.getByte(1) == HEADER_VERSION_1 || buf.getByte(1) == HEADER_VERSION_2)) {
+            return buf.readBytes(HEADER_LENGTH);
+        }
+
+        int index = buf.getByte(1) & 0xFF; // parcel number
```
tananaev: Why do we need mask here?

src/org/traccar/protocol/Arnavi4FrameDecoder.java
```diff
@@ -0,0 +1,70 @@
...
+        if (buf.getByte(0) == PACKAGE_START_SIGN
```
tananaev: Why do we need all these checks? We just need to calculate length and extract frame.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
```diff
@@ -0,0 +1,70 @@
...
+                && index >= PACKAGE_MIN_PARCEL_NUMBER && index <= PACKAGE_MAX_PARCEL_NUMBER) {
+
+            // package: 1 byte start sign, 1 byte parcel number, packets, 1 byte end sign
+            int packetStartIndex = 2;
+            while (packetStartIndex + 8 < buf.readableBytes() && buf.getByte(packetStartIndex) != PACKAGE_END_SIGN) {
```
tananaev: What would happen if we haven't received full package yet?

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: Yeah, `getUnsignedByte()` is better. Fixed.

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: Yes, u're right! I tried to re-write the code of frame decoder looking at At2000, Gt06 and Galileo. Let's see...

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: How can I calculate CRC for 4 bytes of timestamp using `writeInt` before?

binakot: How can I write an integer to buffer if I need to calculated and write CRC of it before?

src/org/traccar/protocol/Arnavi4FrameDecoder.java
binakot: Need to know which type of packet was received. Device can send an auth packet, than few data packets, and auth packet again. What's problem with types checking? I found similar code in a lot of other protocols (autofon, adm, at2000, etc).

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
tananaev: You can calculate CRC directly from buffer. Check what other protocols do.

binakot:
Okay, I did my best. If you have to do some changes, just do it 😄
Or say me what I should to do exactly 👨‍🎓

src/org/traccar/protocol/Arnavi4FrameDecoder.java
tananaev: Concerns from this comment weren't fixed. I think you have missed some other comments as well.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,182 @@
...
+    private void sendResponse(Channel channel, byte version, int index) {
+        if (channel != null) {
+            final ChannelBuffer response;
+            if (version == HEADER_VERSION_1) {
+                response = ChannelBuffers.dynamicBuffer(ByteOrder.LITTLE_ENDIAN, 4);
+                response.writeByte(0x7b);
+                response.writeByte(0x00);
+                response.writeByte((byte) index);
+                response.writeByte(0x7d);
+            } else if (version == HEADER_VERSION_2) {
```
tananaev: Please combine same code. That was the whole purpose of combining two methods into one.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
```diff
@@ -0,0 +1,182 @@
...
+                response = ChannelBuffers.dynamicBuffer(ByteOrder.LITTLE_ENDIAN, 9);
+                response.writeByte(0x7b);
+                response.writeByte(0x04);
+                response.writeByte(0x00);
+                byte[] timeBytes = ByteBuffer.allocate(4).putInt((int) (System.currentTimeMillis() / 1000)).array();
```
tananaev: Still not fixed.

test/org/traccar/protocol/Arnavi4FrameDecoderTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+        Assert.assertEquals( // Valid HEADER v1 packet with IMEI
+                binary(ByteOrder.LITTLE_ENDIAN, "ff22f30c45f5c90f0300"),
+                decoder.decode(null, null, binary(ByteOrder.LITTLE_ENDIAN, "ff22f30c45f5c90f0300")));
+
+        Assert.assertEquals( // Valid HEADER v2 packet with IMEI
+                binary(ByteOrder.LITTLE_ENDIAN, "ff23f30c45f5c90f0300"),
+                decoder.decode(null, null, binary(ByteOrder.LITTLE_ENDIAN, "ff23f30c45f5c90f0300")));
+
+        Assert.assertEquals( // Valid PACKAGE with answer to server on file transfer
```
tananaev: I just noticed this... why are valid packages decoded into "null". That should not be the case.

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
binakot: What's exactly I should to fix this? I need bytes of timestamp to calculate CRC and put it before. What is the right way for that?

test/org/traccar/protocol/Arnavi4FrameDecoderTest.java
binakot: Because this is the packet with answer on file transfer data. At this moment protocol doesn't support file transfering. Why do I need to implement answer on unsupported feature?

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
tananaev: You can use `setByte` to write CRC at a specific index.

test/org/traccar/protocol/Arnavi4FrameDecoderTest.java
tananaev: Frame decoder should be able to decode any data. In protocol decoder you can choose to ignore it.

What would happen if you server receives this or any other unsupported message?

src/org/traccar/protocol/Arnavi4ProtocolDecoder.java
tananaev: You can use `setByte` to write CRC at a specific index.

binakot:
I'll try to finish this PR asap. Sorry for long delay.

binakot:
Oh, really? :octocat: 
It's finally merged :)

tananaev:
Yes. Just FIY, I don't receive notification for every commit. You need to comment that your PR is ready for review.

PR #3349, opened 2017-07-13
Drivers implementation

### LINKED ISSUE #3332: Drivers functionality
Want to discuss the functionality for Drivers
My vision is:

- Implement new model of Driver
  - `name`
  - `rfid` (or `uniqueId`)
  - `company`, `department` etc.

- Implement permissions like for Geofences and Computed Attributes
Driver linked to device mean that he can operate (drive) it
Driver linked to group mean that he can operate (drive) all devices from this group
Driver linked to User mean that User can edit/delete the Driver.

- Extend Device model for one field:
  - `boolean limitDrivers`

- Implement DriversHandler that will generate alarm if Device has `limitDrivers = true` and `position.attributes.rfid` not in list of linked Drivers

- Extend Trip report to set Driver name for every trip

Or it is too complex?

Abyss777:
Huge PR again, but I split it on commits, hope it'll help in review. Nothing complex, just much much code.

Drivers management is almost the same as for Computed Attributes and for Geofences.

Driver model is simple, did not add any organization fields, do not think we can cover all, just added attributes.

Extended Trip report on two fields `driverUniqueId` and `driverName`

The main things related to Drivers:
- `PopulateDriverHandler` - set `driverUniqueId` attribute from `rfid` if present or copy `driverUniqueId` from last position.
- `DriverEventHandler` - generate `driverUnauthorized` event if driver not in list of linked Drivers. If Device Driver list is empty, event will not be generated.

Other changes
- Map User object to notification template, might be used for "Dear, $user.name" or like that.
- Adjusted `Position.KEY_RFID` type in some protocols
- Add `rfid` parameter to `test-generator.py`

I believe we should give some mechanism to stop copying `driverUniqueId` from position to position, but do not have good idea how.

fix #3332


setup/default.xml
```diff
@@ -431,6 +431,63 @@
...
+        uniqueId = :uniqueId,
+        attributes = :attributes
+        WHERE id = :id
+    </entry>
+
+    <entry key='database.deleteDriver'>
+        DELETE FROM drivers WHERE id = :id
+    </entry>
+    
+        <entry key='database.selectDriverPermissions'>
```
tananaev: Indentation problem.

src/org/traccar/processing/PopulateDriverHandler.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.processing;
+
+import org.traccar.BaseDataHandler;
+import org.traccar.Context;
+import org.traccar.model.Position;
+
+public class PopulateDriverHandler extends BaseDataHandler {
```
tananaev: I think this is unnecessary. RFID is the unique id.

Abyss777:
A few questions:
- I should replace `KEY_RFID` on `KEY_DRIVER_UNIQUE_ID` in all protocols, right?
- Does this https://github.com/tananaev/traccar/blob/4e084ebd790286392225e44ce462f987a2c3046a/src/org/traccar/protocol/AtrackProtocolDecoder.java#L330 have same meaning?
- Where is best place to copy `driverUniqueId` from one position to another?

tananaev:
1. Yes, I think we should replace `RFID`.
2. I think so.
3. I think we don't need any extra code for that. We already have attribute copying.

Abyss777:
3. It can be disabled in config, is that OK?

tananaev:
Do you mean copying attributes can be disabled? That's up to admin to configure everything.

Abyss777:
- Used `driverUniqueId` instead of `rfid` in all protocols
- Removed `PopulateDriverHandler`
- Implemented `driverChanged` event instead of `driverUnauthorized`

src/org/traccar/processing/CopyAttributesHandler.java
```diff
@@ -32,9 +32,9 @@ private Position getLastPosition(long deviceId) {
     @Override
     protected Position handlePosition(Position position) {
         String attributesString = Context.getDeviceManager().lookupAttributeString(
-                position.getDeviceId(), "processing.copyAttributes", null, true);
+                position.getDeviceId(), "processing.copyAttributes", Position.KEY_DRIVER_UNIQUE_ID + ",", true);
```
tananaev: Why do we need extra comma here?

Abyss777: To `attributesString.split("[ ,]")` later works

Abyss777: Oh, sorry here is a problem. I'll fix it

tananaev: But split will still work. It will return single parameter, as we want it to. In your case it will return 2 string with empty string as a second one.

Abyss777:
I've replaced last commit, now it works correctly without extra iterations.

tools/test-generator.py
```diff
@@ -46,6 +47,8 @@ def send(conn, lat, lon, course, speed, alarm, ignition, accuracy, rpm, fuel):
         params = params + (('rpm', rpm),)
     if fuel:
         params = params + (('fuel', fuel),)
+    if rfid:
+        params = params + (('rfid', rfid),)
```
tananaev: Probably this script should be updated as well. By the way, have you tested everything after making changes?

src/org/traccar/protocol/OsmAndProtocolDecoder.java
```diff
@@ -128,6 +128,9 @@ protected Object decode(
                 case "batt":
                     position.set(Position.KEY_BATTERY_LEVEL, Double.parseDouble(value));
                     break;
+                case "driverUniqueId":
```
tananaev: Sorry, I just realized that we don't need this in OsmAnd. Decoder automatically saves all extra parameters in attributes.

Abyss777: We need it, because it saves as double, but we are waiting String.

tools/test-generator.py
Abyss777: Yes, I test almost after every changes, sorry sometime superficially. 
I tested it more carefully last time.

src/org/traccar/protocol/OsmAndProtocolDecoder.java
tananaev: You are right. If driver id is digits-only, then it would try to save it as Double.

tananaev:
Merged, thanks.

src/org/traccar/processing/CopyAttributesHandler.java
```diff
@@ -32,10 +32,14 @@ private Position getLastPosition(long deviceId) {
     @Override
     protected Position handlePosition(Position position) {
         String attributesString = Context.getDeviceManager().lookupAttributeString(
-                position.getDeviceId(), "processing.copyAttributes", null, true);
+                position.getDeviceId(), "processing.copyAttributes", "", true);
         Position last = getLastPosition(position.getDeviceId());
-        if (attributesString != null && last != null) {
-            attributesString += Position.KEY_DRIVER_UNIQUE_ID + ",";
+        if (attributesString.isEmpty()) {
```
jcardus, line 37: Why are we forcing driver unique id to be copied?

jcardus (commented):
Why are we forcing driver unique id to be copied?

src/org/traccar/processing/CopyAttributesHandler.java
```diff
@@ -32,10 +32,14 @@ private Position getLastPosition(long deviceId) {
...
+            attributesString = Position.KEY_DRIVER_UNIQUE_ID;
```
jcardus, line 38: Why are we forcing driver unique id to be copied?

PR #3354, opened 2017-07-15
Fixed <version> in pom.xml;

jon-stumpf:


tananaev:
Actually I deliberately left version as 3.12 because there were no (or very minor) changes to the back-end. Version 3.13 was a fix for the web app issue. Otherwise it's exactly the same as 3.12.

jon-stumpf:
Ah.  Then, I will retract my pull request.

socstur:
Thanks

PR #3360, opened 2017-07-17
Map Device attributes to Computed Attributes

### LINKED ISSUE #3333: Map Device fields and attributes to Computed Attributes
It would be helpful if we implement mapping of device fields and attributes in Computed Attributes.

We should do it carefully to not decrease some security aspects.

### LINKED ISSUE #3347: Use of device specific attributes in computed attributes
Is it possible to allow the use of parameters created in device attributes in computed attributes?

Abyss777:
Mapped attributes only and as is. If position has attributes with same names, they will replace device attributes.

fix #3347 
fix #3333

### DIFF
--- a/src/org/traccar/processing/ComputedAttributesHandler.java
+++ b/src/org/traccar/processing/ComputedAttributesHandler.java
@@ -31,19 +31,34 @@ import org.traccar.BaseDataHandler;
 import org.traccar.Context;
 import org.traccar.helper.Log;
 import org.traccar.model.Attribute;
+import org.traccar.model.Device;
 import org.traccar.model.Position;
 
 public class ComputedAttributesHandler extends BaseDataHandler {
 
     private JexlEngine engine;
 
+    private boolean mapDeviceAttributes;
+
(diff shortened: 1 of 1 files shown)


src/org/traccar/processing/ComputedAttributesHandler.java
```diff
@@ -31,19 +31,34 @@
...
     private JexlEngine engine;
 
+    private boolean mapDeviceAttributes;
+
     public ComputedAttributesHandler() {
         engine = new JexlEngine();
         engine.setStrict(true);
+        if (Context.getConfig() != null) {
+            mapDeviceAttributes = Context.getConfig().getBoolean(
+                    "processing.computedAttributes.mapDeviceAttributes.enable");
```
tananaev: Seems too long. Can we just name it `processing.computedAttributes.deviceAttributes`?

Abyss777:
Sure. Renamed.

tananaev:
Merged, thanks.

socstur:
wow you guys are fast with this stuff.

PR #3369, opened 2017-07-19
[NEW] add GET with list of uniqueId to API

jpmens:
This pull request adds the possibility to GET a device with its `uniqueId`. Modelled after `Id`, we also accept a list of `uniqueId`. Basically, if `all` is false, we collect devices by `Id` and `uniqueId` and return those.

Hence, the following GET request is possible:

```
http://127.0.0.1:8082/api/devices?uniqueId=9876543210987654&id=11
```

### DIFF
--- a/src/org/traccar/api/resource/DeviceResource.java
+++ b/src/org/traccar/api/resource/DeviceResource.java
@@ -45,30 +45,39 @@ public class DeviceResource extends BaseResource {
     @GET
     public Collection<Device> get(
             @QueryParam("all") boolean all, @QueryParam("userId") long userId,
+            @QueryParam("uniqueId") List<String> uniqueIds,
             @QueryParam("id") List<Long> deviceIds) throws SQLException {
-        if (deviceIds.isEmpty()) {
-            if (all) {
-                if (Context.getPermissionsManager().isAdmin(getUserId())) {
-                    return Context.getDeviceManager().getAllDevices();
-                } else {
-                    Context.getPermissionsManager().checkManager(getUserId());
-                    return Context.getDeviceManager().getManagedDevices(getUserId());
-                }
+        if (all) {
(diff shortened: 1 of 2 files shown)


src/org/traccar/api/resource/DeviceResource.java
```diff
@@ -45,30 +45,39 @@
...
+        if (uniqueIds.isEmpty() && deviceIds.isEmpty()) {
+            if (userId == 0) {
+                userId = getUserId();
+            }
+            Context.getPermissionsManager().checkUser(getUserId(), userId);
+            return Context.getDeviceManager().getDevices(userId);
+        }
+
+        ArrayList<Device> devices = new ArrayList<>();
+        if (!uniqueIds.isEmpty()) {
```
tananaev: Redundant check.

src/org/traccar/api/resource/DeviceResource.java
```diff
@@ -45,30 +45,39 @@
...
+            for (String uniqueId : uniqueIds) {
+                Device device = Context.getDeviceManager().getDeviceByUniqueId(uniqueId);
+                Context.getPermissionsManager().checkDevice(getUserId(), device.getId());
+                devices.add(device);
+            }
+        }
+        if (!deviceIds.isEmpty()) {
```
tananaev: Same.

jpmens:
Addressed

tananaev:
Merged, thanks. What the use case for this by the way?

jpmens:
We manage a list of devices and store their uniqueIds (because the internal database id doesn't mean much to us). In order to be able to effectively get a an ID, we GET uniqueid.

tananaev:
I'm not sure if official app will ever use it, but it might be useful for some integrations.

PR #3373, opened 2017-07-20
Refactor managers, permissions etc.

Abyss777:
It is extremely big change. I'm not sure if it is possible to review well. Ask me about any doubts.

Main changes:
- Implemented `SimpleObjectManager` that serves set of objects and links to users (Calendars)
- Implemented `ExtendedObjectManager` that also serves links to devices and groups (Geofences, Drivers, Attributes)
- Implemented `getObjects`, `addObject`, `updateObject`, `removeObject` database functions, they used Class name to select request string.
- Some `database.*` names was changed to consistence
- Changed permissions model hierarchy to fit new managers model. Added interface with `getSlaveId` function.
- Combined permissions resources. Unfortunately had to combine devices and groups permissions in its main resource, because `@Delete` regex matching failed if it is in different resources.
- Implemented smart functions to check and refresh permissions

PS: I believe new managers model might be used to optimize users, devices and groups management.


src/org/traccar/database/PermissionsManager.java
```diff
@@ -298,52 +298,85 @@ public void checkRegistration(long userId) {
...
+
+        if (manager != null) {
+            if (manager.checkItemPermission(userId, objectId) && !isAdmin(userId)) {
+                checkManager(userId);
+                for (long managedUserId : getUserPermissions(userId)) {
+                    if (manager.checkItemPermission(managedUserId, objectId)) {
+                        return;
+                    }
                 }
+                throw new SecurityException(object.substring(0, 1).toUpperCase() + object.substring(1)
```
tananaev: I would prefer less fiddling with names. Maybe just use "Type xxx access denied".

src/org/traccar/model/AttributePermission.java
```diff
@@ -38,4 +28,9 @@ public void setAttributeId(long attributeId) {
         this.attributeId = attributeId;
     }
 
+    @Override
+    public long getSlaveId() {
```
tananaev: Preferably rename. See earlier comment.

src/org/traccar/model/CalendarPermission.java
```diff
@@ -16,17 +16,7 @@
...
-    private long userId;
-
-    public long getUserId() {
-        return userId;
-    }
-
-    public void setUserId(long userId) {
-        this.userId = userId;
-    }
+public class CalendarPermission extends BaseUserPermission {
```
tananaev: Another thought... can we get rid of permission classes? Can we just use Map with keys-value pairs? I believe it will be serialized into same JSON object?

setup/default.xml
```diff
@@ -60,7 +60,7 @@
         WHERE email = :email
     </entry>
 
-    <entry key='database.selectUsersAll'>
+    <entry key='database.selectUsers'>
```
tananaev: Another thought for improvement. Probably better to implement separately. Generate SQL automatically, with option to override in the config file.

src/org/traccar/api/resource/DevicePermissionResource.java
```diff
@@ -1,66 +0,0 @@
...
-        if (Context.getGeofenceManager() != null) {
-            Context.getGeofenceManager().refresh();
-        }
-        return Response.ok(entity).build();
-    }
-
-    @DELETE
-    public Response remove(DevicePermission entity) throws SQLException {
-        Context.getPermissionsManager().checkReadonly(getUserId());
-        if (getUserId() != entity.getUserId()) {
```
Abyss777: Oh, yes, missed that. Will fix.

src/org/traccar/api/resource/DeviceResource.java
```diff
@@ -122,4 +121,34 @@ public Response updateTotalDistance(DeviceTotalDistance entity) throws SQLExcept
         return Response.noContent().build();
     }
 
+    @Path("/{slave : (geofences|drivers|attributes)}")
```
Abyss777: Is it OK if we change API in such way?

tananaev: I think it's OK because it's a major refactoring. It will be a really small change for anyone who uses API directly. Just need to remind people about it with release notes.

src/org/traccar/api/resource/DeviceResource.java
```diff
@@ -122,4 +121,34 @@ public Response updateTotalDistance(DeviceTotalDistance entity) throws SQLExcept
...
+        }
+        for (String key : entity.keySet()) {
+            Context.getPermissionsManager().checkPermission(key.replace("Id", ""), getUserId(), entity.get(key));
+        }
+        Context.getDataManager().linkObject(entity, true);
+        Context.getPermissionsManager().refreshPermissions(entity);
+        return Response.noContent().build();
+    }
+
+    @Path("/{slave : (geofences|drivers|attributes)}")
```
Abyss777: Lets make agreement about the naming then. I think `target` is more proper for first object (user/device/group) and for second object offer `accessory`. What do you think?

src/org/traccar/api/resource/DeviceResource.java
```diff
@@ -122,4 +121,34 @@ public Response updateTotalDistance(DeviceTotalDistance entity) throws SQLExcept
...
+    @DELETE
+    public Response remove(Map<String, Long> entity) throws SQLException {
+        Context.getPermissionsManager().checkReadonly(getUserId());
+        for (String key : entity.keySet()) {
+            Context.getPermissionsManager().checkPermission(key.replace("Id", ""), getUserId(), entity.get(key));
+        }
+        if (entity.size() != 2) {
```
Abyss777: Oh, looks like I misplaced it, it should be upper before the loop.
Just security and DOS, it is possible that someone pass JSON with a lot of object we should not loop over it.

src/org/traccar/api/resource/DriverResource.java
```diff
@@ -50,62 +51,65 @@
...
     }
 
     @POST
     public Response add(Driver entity) throws SQLException {
         Context.getPermissionsManager().checkReadonly(getUserId());
-        Context.getDriversManager().addDriver(entity);
-        Context.getDataManager().linkDriver(getUserId(), entity.getId());
-        Context.getDriversManager().refreshUserDrivers();
+        Context.getDriversManager().addItem(entity);
+        LinkedHashMap<String, Long> link = new LinkedHashMap<>();
```
Abyss777: Good idea.

src/org/traccar/api/resource/GroupResource.java
```diff
@@ -75,22 +78,48 @@ public Response update(Group entity) throws SQLException {
...
+            Context.getPermissionsManager().checkPermission(key.replace("Id", ""), getUserId(), entity.get(key));
+        }
+        Context.getDataManager().linkObject(entity, true);
+        Context.getPermissionsManager().refreshPermissions(entity);
+        return Response.noContent().build();
+    }
+
+    @Path("/{slave : (geofences|drivers|attributes)}")
+    @DELETE
+    public Response remove(Map<String, Long> entity) throws SQLException {
```
Abyss777: Why not. I'm always for optimization.

src/org/traccar/database/AttributesManager.java
```diff
@@ -17,183 +17,30 @@
...
-
-    public Set<Long> getDeviceAttributes(long deviceId) {
-        return getDeviceAttributes(deviceAttributes, deviceId);
-    }
-
-    public Set<Long> getAllDeviceAttributes(long deviceId) {
-        return getDeviceAttributes(deviceAttributesWithGroups, deviceId);
+        super(dataManager, Attribute.class, AttributePermission.class, DeviceAttribute.class, GroupAttribute.class);
+        refreshItems();
+        refreshExtendedPermissions();
```
Abyss777: We need to do it in constructor for initialization.
But for the reason of performance we have 3 functions to refresh: `refreshItems`, `refreshUserItems` and `refreshExtendedPermissions`.
I've tried to implement smart functions to refresh and update only necessary in appropriate caches.

src/org/traccar/database/DriversManager.java
```diff
@@ -17,147 +17,53 @@
...
+    public void refreshItems() {
+        if (getDataManager() != null) {
             try {
-                drivers.clear();
-                driversByUniqueId.clear();
-                for (Driver driver : dataManager.getDrivers()) {
-                    drivers.put(driver.getId(), driver);
-                    driversByUniqueId.put(driver.getUniqueId(), driver);
+                clearItems();
+                for (Identifiable item : getDataManager().getObjects(getBaseClass())) {
```
Abyss777: The meaning of words are perfectly reflects the idea. One has ID, second has attributes and can be easily extended.
But agree nouns fits more for class naming.

src/org/traccar/model/CalendarPermission.java
Abyss777: I'll check if can get Map from database.

src/org/traccar/api/resource/DeviceResource.java
tananaev: How about `owner` and `property`?

src/org/traccar/api/resource/DeviceResource.java
tananaev: Maybe we should make some helper method in super class for it as well...

src/org/traccar/database/AttributesManager.java
tananaev: Can't we just pass argument with the type that we want to update? If it complicates things probably is not worth it.

src/org/traccar/model/CalendarPermission.java
tananaev: We might need to adjust `QueryBuilder`, but that's fine if it helps to get rid of like 10 classes.

src/org/traccar/api/resource/DeviceResource.java
Abyss777: Yes, it is good.

src/org/traccar/api/resource/DeviceResource.java
Abyss777: Yes, pieces of this functions used at least 6 times, it is worth to be detached in some function.

src/org/traccar/api/resource/DeviceResource.java
Abyss777: I would suggest use one path for all permissions (`/permissions`), we have all needed information inside request payload.

Abyss777:
Hope did not missed any of your comments.

- Totally removed any permissions classes, works with them as Map
- Combined link/unlink API to one `/api/permissions`
- Make type constants and helpers to work with type names.

As consequence `checkPermissions(object,... ` and `linkObject` become tolerate to names. They correctly react on `device`/`Device`/`deviceId` names.

src/org/traccar/api/BaseResource.java
```diff
@@ -29,4 +37,35 @@ protected long getUserId() {
...
+            }
+        } else {
+            Context.getPermissionsManager().checkPermission(owner, getUserId(), ownerId);
+        }
+        Context.getPermissionsManager().checkPermission(property, getUserId(), propertyId);
+
+        Context.getDataManager().linkObject(owner, ownerId, property, propertyId, link);
+    }
+
+    protected void linkNew(BaseModel entity) throws SQLException {
```
tananaev: I would recommend to rename this method to `linkNewEntity` to make it more clear in places where we use it.

src/org/traccar/api/BaseResource.java
```diff
@@ -29,4 +37,35 @@ protected long getUserId() {
         }
         return 0;
     }
+
+    protected void handlePermission(Map<String, Long> entity, boolean link) throws SQLException {
+        if (entity.size() != 2) {
```
tananaev: I think we can remove this check. If someone misuses the API, he will either get an error or extra parameters will be ignored.

src/org/traccar/api/BaseResource.java
```diff
@@ -29,4 +37,35 @@ protected long getUserId() {
...
+            throw new IllegalArgumentException();
+        }
+        Iterator<String> iterator = entity.keySet().iterator();
```
tananaev: Are you sure that this map preserves order? Please double check.

src/org/traccar/api/resource/PermissionsResource.java
```diff
@@ -28,28 +29,25 @@
...
     }
 
     @DELETE
-    public Response remove(UserPermission entity) throws SQLException {
-        Context.getPermissionsManager().checkAdmin(getUserId());
-        Context.getDataManager().unlinkUser(entity.getUserId(), entity.getManagedUserId());
-        Context.getPermissionsManager().refreshUserPermissions();
+    public Response remove(Map<String, Long> entity) throws SQLException {
+        Context.getPermissionsManager().checkReadonly(getUserId());
+        handlePermission(entity, false);
```
tananaev: I think we should rename this method to be more descriptive as well.

src/org/traccar/database/ExtendedObjectManager.java
```diff
@@ -0,0 +1,118 @@
...
+        super.removeItem(itemId);
+        refreshExtendedPermissions();
+    }
+
+    public void refreshExtendedPermissions() {
+        if (getDataManager() != null) {
+            try {
+
+                Collection<Map<String, Long>> databaseGroupPermissions =
+                        getDataManager().getPermissions("Group", getBaseClassName());
```
tananaev: Can we use same constants for all these names as well?

src/org/traccar/database/DataManager.java
```diff
@@ -501,224 +268,52 @@ public void addStatistics(Statistics statistics) throws SQLException {
...
-                .setLong("driverId", driverId)
-                .executeUpdate();
-    }
-
-    public void unlinkDeviceDriver(long deviceId, long driverId) throws SQLException {
-        QueryBuilder.create(dataSource, getQuery("database.unlinkDeviceDriver"))
-                .setLong("deviceId", deviceId)
-                .setLong("driverId", driverId)
+    public void removeObject(Class<? extends BaseModel> clazz, long entityId) throws SQLException {
+        String query = "database.delete" + clazz.getSimpleName();
```
tananaev: Why in some places we use strings, in others classes, and in others constants? I think class names would be the best option. Is there any reason for not using them?

src/org/traccar/database/DataManager.java
```diff
@@ -501,224 +268,52 @@ public void addStatistics(Statistics statistics) throws SQLException {
...
+        return object.substring(0, 1).toLowerCase() + object.substring(1) + (object.indexOf("Id") == -1 ? "Id" : "");
     }
 
-    public Collection<CalendarPermission> getCalendarPermissions() throws SQLException {
-        return QueryBuilder.create(dataSource, getQuery("database.selectCalendarPermissions"))
-                .executeQuery(CalendarPermission.class);
-    }
+    public void linkObject(String owner, long ownerId, String property, long propertyId,
+            boolean link) throws SQLException {
+        String query = "database." + (!link ? "un" : "") + "link" + makeName(owner) + makeName(property);
```
tananaev: I think it would be better to use `(link ? "link" : "unlink")` here. It's shorter and much more readable.

src/org/traccar/database/DataManager.java
```diff
@@ -501,224 +268,52 @@ public void addStatistics(Statistics statistics) throws SQLException {
...
-        QueryBuilder.create(dataSource, getQuery("database.updateCalendar"))
-                .setObject(calendar)
-                .executeUpdate();
-    }
-
-    public void removeCalendar(long calendarId) throws SQLException {
-        QueryBuilder.create(dataSource, getQuery("database.deleteCalendar"))
-                .setLong("id", calendarId)
-                .executeUpdate();
+    public static String makeNameId(String object) {
```
tananaev: Do we need these methods to be public?

src/org/traccar/database/DataManager.java
Abyss777: I like class names or constants, but for `Simple`/`ExtendedObjectManager` we need class itself to pass to database  `getObjects`, `addObject`, `updateObject`. I think it is easier to get name from class than get class by name.

Here class was used just for consistence with `add`/`updateObject`

src/org/traccar/api/BaseResource.java
Abyss777: I'm pretty sure. Can't find the proof, but I get `LinkedHashMap` from Jersey.

src/org/traccar/api/BaseResource.java
Abyss777: Yes. Without loop it is possible.

src/org/traccar/database/DataManager.java
Abyss777: Oh. I've written some mess.

Again.

We need to know class to work `getObjects` correctly, to `executeQuery` returned correct collection of objects.

In `addObject` and `updateObject` I used class name just not to pass additional parameter.
For `removeObject` does not matter what to pass. Used class for some consistence, but will replace it.

tananaev: So, can we use classes everywhere?

src/org/traccar/api/BaseResource.java
tananaev: `LinkedHashMap` should be fine.

src/org/traccar/database/DataManager.java
Abyss777: OK, I'll try. I need a tip. What is best way to convert `"deviceId"` -> `Device.class` ? 

Also, don't you mind renaming Permissions configs to User for consistence ( `linkDriverPermission` -> `linkUserDriver` ) ?

tananaev: Probably better to have some switch statement to convert name to class... somewhere in a helper method. But maybe you should tell me the place where we need it and I'll think of a better solution.

I don't mind renaming methods. Makes sense.

Abyss777: Please, take a look at `BaseResource.handlePermission()` function, it accepts Map {"deviceId:2", "driverId":4}
We need to call `PermissionManager.checkPermission(object, userId, objectId)` where object should be replaced to class and call `Database.linkObject(owner, ownerId, property, propertyId, link)` where owner and property should be replaced to class.

tananaev: Of course it's possible to use reflection to construct name and get class by name, but I'm not sure it's a good idea. Maybe it's OK because we only get class and don't do any iterations.

Abyss777: Something like `Class.forName("org.traccar.model."  + name)` ?

tananaev: Yes, but make sure that code is isolated in a single place, so we can change it to switch if we want to.

Abyss777:
Used classes everywhere and fix other comments

src/org/traccar/api/BaseResource.java
```diff
@@ -29,4 +39,34 @@ protected long getUserId() {
...
+        String owner = iterator.next();
+        Class<?> ownerClass = DataManager.getClassByName(owner);
+        String property = iterator.next();
+        Class<?> propertyClass = DataManager.getClassByName(property);
+
+        long ownerId = entity.get(owner);
+        long propertyId = entity.get(property);
+
+        if (!link && ownerClass.equals(User.class)
+                && propertyClass.equals(Device.class)) {
```
tananaev: Can this be on the same line?

src/org/traccar/api/BaseResource.java
```diff
@@ -29,4 +39,34 @@ protected long getUserId() {
...
+        Iterator<String> iterator = entity.keySet().iterator();
+        String owner = iterator.next();
+        Class<?> ownerClass = DataManager.getClassByName(owner);
+        String property = iterator.next();
+        Class<?> propertyClass = DataManager.getClassByName(property);
+
+        long ownerId = entity.get(owner);
+        long propertyId = entity.get(property);
+
+        if (!link && ownerClass.equals(User.class)
```
tananaev: Why are we only checking permissions on un-linking?

src/org/traccar/api/BaseResource.java
```diff
@@ -29,4 +39,34 @@ protected long getUserId() {
         }
         return 0;
     }
+
+    protected void checkAndLinkPermission(LinkedHashMap<String, Long> entity, boolean link)
```
tananaev: I'm still not completely satisfied with the name. Maybe we should split it into check permissions?.. and then `linkObject` method is already there.

src/org/traccar/api/BaseResource.java
```diff
@@ -29,4 +39,34 @@ protected long getUserId() {
...
+            }
+        } else {
+            Context.getPermissionsManager().checkPermission(ownerClass, getUserId(), ownerId);
+        }
+        Context.getPermissionsManager().checkPermission(propertyClass, getUserId(), propertyId);
+
+        Context.getDataManager().linkObject(ownerClass, ownerId, propertyClass, propertyId, link);
+    }
+
+    protected void linkNewEntity(BaseModel entity) throws SQLException {
```
tananaev: Do we actually need this method? It just calls one other method?

src/org/traccar/api/resource/GroupResource.java
```diff
@@ -75,22 +73,18 @@ public Response update(Group entity) throws SQLException {
...
         Context.getDeviceManager().updateGroup(entity);
-        if (Context.getGeofenceManager() != null) {
-            Context.getGeofenceManager().refresh();
-        }
+        Context.getPermissionsManager().refreshAllExtendedPermissions();
         return Response.ok(entity).build();
     }
 
-    @Path("{id}")
+    @Path("{id : \\d+}")
```
tananaev: Do we still need this?

src/org/traccar/api/resource/PermissionsResource.java
```diff
@@ -28,30 +29,25 @@
...
-@Path("groups/drivers")
+@Path("permissions")
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
-public class GroupDriverResource extends BaseResource {
+public class PermissionsResource  extends BaseResource {
 
     @POST
-    public Response add(GroupDriver entity) throws SQLException {
+    public Response add(LinkedHashMap<String, Long> entity) throws SQLException, ClassNotFoundException {
```
tananaev: Idea for improvement. Have some helper class that will have methods to get owner and property keys and values. It can extend `LinkedHashMap`, so it might even work automatically as an argument here. I believe this will simplify some places where we "parse" this entity.

src/org/traccar/events/GeofenceEventHandler.java
```diff
@@ -57,18 +59,20 @@ public GeofenceEventHandler() {
 
         Collection<Event> events = new ArrayList<>();
         for (long geofenceId : newGeofences) {
-            long calendarId = geofenceManager.getGeofence(geofenceId).getCalendarId();
-            if (calendarId == 0 || Context.getCalendarManager().getCalendar(calendarId) == null
-                    || Context.getCalendarManager().getCalendar(calendarId).checkMoment(position.getFixTime())) {
+            long calendarId = ((Geofence) geofenceManager.getById(geofenceId)).getCalendarId();
+            if (calendarId == 0 || Context.getCalendarManager().getById(calendarId) == null
```
tananaev: I think we should avoid calling `getById` multiple times.

src/org/traccar/events/GeofenceEventHandler.java
```diff
@@ -57,18 +59,20 @@ public GeofenceEventHandler() {
...
                 events.add(event);
             }
         }
         for (long geofenceId : oldGeofences) {
-            long calendarId = geofenceManager.getGeofence(geofenceId).getCalendarId();
-            if (calendarId == 0 || Context.getCalendarManager().getCalendar(calendarId) == null
-                    || Context.getCalendarManager().getCalendar(calendarId).checkMoment(position.getFixTime())) {
+            long calendarId = ((Geofence) geofenceManager.getById(geofenceId)).getCalendarId();
+            if (calendarId == 0 || Context.getCalendarManager().getById(calendarId) == null
+                    || ((Calendar) Context.getCalendarManager().getById(calendarId))
```
tananaev: Same here.

src/org/traccar/api/BaseResource.java
Abyss777: Its exclusion came from `DevicePermissionResource` Only admin can unlink devices from himself.

src/org/traccar/api/BaseResource.java
Abyss777: But this method substitute 3 common values.  It is called many times, and only `entity` is changed

Abyss777:
Added permission helper class and reorganized checking

src/org/traccar/events/GeofenceEventHandler.java
```diff
@@ -57,18 +59,24 @@ public GeofenceEventHandler() {
 
         Collection<Event> events = new ArrayList<>();
         for (long geofenceId : newGeofences) {
-            long calendarId = geofenceManager.getGeofence(geofenceId).getCalendarId();
-            if (calendarId == 0 || Context.getCalendarManager().getCalendar(calendarId) == null
-                    || Context.getCalendarManager().getCalendar(calendarId).checkMoment(position.getFixTime())) {
+            long calendarId = ((Geofence) geofenceManager.getById(geofenceId)).getCalendarId();
+            Calendar calendar = null;
+            if (calendarId != 0) {
```
tananaev: Can we use ternary expression here? If it's too long to fit on one line, please leave it as it is.

src/org/traccar/events/GeofenceEventHandler.java
```diff
@@ -57,18 +59,24 @@ public GeofenceEventHandler() {
...
                 events.add(event);
             }
         }
         for (long geofenceId : oldGeofences) {
-            long calendarId = geofenceManager.getGeofence(geofenceId).getCalendarId();
-            if (calendarId == 0 || Context.getCalendarManager().getCalendar(calendarId) == null
-                    || Context.getCalendarManager().getCalendar(calendarId).checkMoment(position.getFixTime())) {
+            long calendarId = ((Geofence) geofenceManager.getById(geofenceId)).getCalendarId();
+            Calendar calendar = null;
+            if (calendarId != 0) {
```
tananaev: Same here.

src/org/traccar/api/resource/PermissionsResource.java
```diff
@@ -0,0 +1,79 @@
...
+        } else {
+            Context.getPermissionsManager().checkPermission(
+                    permission.getOwnerClass(), getUserId(), permission.getOwnerId());
+        }
+        Context.getPermissionsManager().checkPermission(
+                permission.getPropertyClass(), getUserId(), permission.getPropertyId());
+    }
+
+    @POST
+    public Response add(LinkedHashMap<String, Long> entity) throws SQLException, ClassNotFoundException {
```
tananaev: Is it possible to make Java automatically convert request to `Permission`, so we can use it directly in the argument?

Abyss777: It is possible but not automatically, Because field names are different.
We can implement own Entity Provider with custom `MessageBodyReader` http://docs.oracle.com/javaee/7/api/javax/ws/rs/ext/MessageBodyReader.html

Or we can change API and require something like:
````json
{
"owner":"driver",
"ownerId": 5,
"property":"geofence",
"propertyId":8
}
````
instead of simple
````json
{
"deviceId":5,
"geofenceId":8
}
````

Comment (Abyss777) [inline src/org/traccar/api/resource/PermissionsResource.java]: Hm, we use jackson, then it should be implemented in other way https://jersey.github.io/documentation/latest/media.html#json.jackson

Comment (tananaev) [inline src/org/traccar/api/resource/PermissionsResource.java]: Let's leave it as it is for now.

File: src/org/traccar/api/BaseResource.java [hunk 32]
```diff
@@ -29,4 +35,8 @@ protected long getUserId() {
         }
         return 0;
     }
+
+    protected void linkNewEntity(BaseModel entity) throws SQLException {
```
tananaev: I had this comment before, but why do we need this method? It just calls one other method. I think we should remove it. It would make it easier to follow the code.

src/org/traccar/database/ExtendedObjectManager.java
```diff
@@ -0,0 +1,122 @@
...
+import org.traccar.model.Group;
+import org.traccar.model.BaseModel;
+
+public abstract class ExtendedObjectManager extends SimpleObjectManager {
+
+    private final Map<Long, Set<Long>> deviceItems = new ConcurrentHashMap<>();
+    private final Map<Long, Set<Long>> deviceItemsWithGroups = new ConcurrentHashMap<>();
+    private final Map<Long, Set<Long>> groupItems = new ConcurrentHashMap<>();
+
+    protected ExtendedObjectManager(DataManager dataManager,
```
tananaev: Minor issue, but I usually prefer to move all parameters on the next line if it doesn't fit. I think it makes it easier to read like this:
```java
protected ExtendedObjectManager(
        DataManager dataManager, Class<? extends BaseModel> baseClass) {
```

Not a big deal. Just a comment for future code so we make styling more consistent across the codebase.

src/org/traccar/database/PermissionsManager.java
```diff
@@ -298,52 +305,79 @@ public void checkRegistration(long userId) {
...
     }
 
-
-    public void checkCalendar(long userId, long calendarId) throws SecurityException {
-        if (!Context.getCalendarManager().checkCalendar(userId, calendarId) && !isAdmin(userId)) {
-            checkManager(userId);
-            for (long managedUserId : getUserPermissions(userId)) {
-                if (Context.getCalendarManager().checkCalendar(managedUserId, calendarId)) {
-                    return;
+    public void refreshPermissions(Map<String, Long> entity) {
```
tananaev: I think I had this comment already, but I think it would be better to take `Permission` as an argument here. So, we can compare actual classes which should be faster than constructing many string and comparing them.

Abyss777:
Sorry missed that. Also made some clean up after PMD report.

tananaev:
Merged, looks good. I hope you thoroughly tested it ;)

tananaev:
Am I understanding correctly that we need some modifications to the web app because API paths changes? Also, can you please update swagger file.

Abyss777:
I'll try to describe it in swagger, But  not sure if it is possible to do it clear.

tananaev:
Worst case you can just have some long comment describing the functionality.

PR #3385, opened 2017-07-24
Few managers optimizations

Abyss777:
Polished few things:

- Items cache:
  - Reorganized cache update, now it is similar to how we do it for Devices and Groups (do not clear items map, but update every item)
  - Used "overridable" functions `addNewItem`/`updateCachedItem` for two purposed update cache and update items. By default it just `put` but might be used to track additional maps (like uniqueIds) or update item by fields.
  - It is not necessary to override whole functions `addItem`/`updateItem`/`removeItem` and especially `refreshItems` just put manager specific code in `addNewItem`/`updateCachedItem`/`removeCachedItem` if necessary
  - Spread refresh calls by appropriate constructors

- Used `Permission` class also for retrieving permissions from database
  - Do not think we will use common function `executeMapQuery` in near future, replace it to particular `executePermissionsQuery`
  - Updating permission cache now looks much much clear.

Also fixed small issue in group cache update.

### DIFF
--- a/src/org/traccar/database/SimpleObjectManager.java
+++ b/src/org/traccar/database/SimpleObjectManager.java
@@ -27,6 +27,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import org.traccar.Context;
 import org.traccar.helper.Log;
 import org.traccar.model.BaseModel;
+import org.traccar.model.Permission;
 import org.traccar.model.User;
 
 public abstract class SimpleObjectManager {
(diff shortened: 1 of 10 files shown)


src/org/traccar/database/SimpleObjectManager.java
```diff
@@ -65,11 +67,7 @@ protected final void clearItems() {
         items.clear();
     }
 
-    protected final void putItem(long itemId, BaseModel item) {
-        items.put(itemId, item);
-    }
-
-    protected final void removeCachedItem(long itemId) {
+    protected void removeCachedItem(long itemId) {
```
tananaev: I think it makes sense to keep `updateCachedItem` and `removeCachedItem` next to each other.

src/org/traccar/database/SimpleObjectManager.java
```diff
@@ -91,9 +89,23 @@ public final boolean checkItemPermission(long userId, long itemId) {
...
+                    databaseItemIds.add(item.getId());
+                    if (items.containsKey(item.getId())) {
+                        updateCachedItem(item);
+                    } else {
+                        addNewItem(item);
+                    }
+                }
+                for (Long cachedItemId : items.keySet()) {
+                    if (!databaseItemIds.contains(cachedItemId)) {
+                        items.remove(cachedItemId);
```
tananaev: Should we use `removeCachedItem` here?

Abyss777:
Thanks for the tip. Did not think about it. Now it is perfect.

tananaev:
Merged, thanks.

PR #3388, opened 2017-07-25
Move users management to separate manager

Abyss777:
Next step
On the one hand it is an extra class, but on the other we are reusing code and unloading `PermissionsManager`

Also I hope we will optimize objects resources at least add/update/remove functions, separate managers should help for that.

### DIFF
--- a/src/org/traccar/database/PermissionsManager.java
+++ b/src/org/traccar/database/PermissionsManager.java
@@ -30,29 +30,33 @@ import org.traccar.model.User;
 
 import java.lang.reflect.Method;
 import java.sql.SQLException;
-import java.util.ArrayList;
-import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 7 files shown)


src/org/traccar/database/PermissionsManager.java
```diff
@@ -235,6 +210,9 @@ public void checkDeviceReadonly(long userId) throws SecurityException {
 
     public void checkUserEnabled(long userId) throws SecurityException {
         User user = getUser(userId);
+        if (user == null) {
+            throw new SecurityException("Unknown account");
```
Abyss777: Noticed next thing: if user was logged in and was deleted from server, after refreshing page he will get `NullPointerException`. But now he will get login page.

tananaev:
Looks good, merged.

PR #3391, opened 2017-07-26
Split Drivers and Groups management

Abyss777:
The main problem is that devices and groups relationship is complex, we use `GroupTree` to trace it.
The next problem is that we have to have list of devices and groups before creating the tree.
I decided to add new layer in object managers structure and it looks next:

- `BaseObjectManager` class that only stores objects cache and do add/update/remove
- `ManagableObjects` (probably need to be renamed) interface that implements two functions `getUserItems` and `getManagedItems`. `getManagedItems` are the same in 3 classes, but it is impossible to use `default` in Java 7. This interface used only to strict the code.
- `SimpleObjectManager` class that implements `ManagableObjects` and store User -> Object permissions
- `ExtendedObjectManager` is not changed, it stores Device -> Object and Group -> Object permissions

I leaved this complex devices and groups tree in `PermissionsManager` and just apply to it from Device and Group managers.

Now objects API resources look very similar and this will allow combine them on some future stage.

Some comments in code.

### DIFF
--- a/src/org/traccar/database/AttributesManager.java
+++ b/src/org/traccar/database/AttributesManager.java
@@ -26,9 +26,14 @@ public class AttributesManager extends ExtendedObjectManager {
     }
 
     @Override
+    public Attribute getById(long calendarId) {
+        return (Attribute) super.getById(calendarId);
+    }
+
+    @Override
     public void updateCachedItem(BaseModel item) {
         Attribute attribute = (Attribute) item;
-        Attribute cachedAttribute = (Attribute) getById(item.getId());
+        Attribute cachedAttribute = getById(item.getId());
         cachedAttribute.setDescription(attribute.getDescription());
         cachedAttribute.setAttribute(attribute.getAttribute());
         cachedAttribute.setExpression(attribute.getExpression());
--- a/src/org/traccar/database/AttributesManager.java
+++ b/src/org/traccar/database/AttributesManager.java
@@ -17,23 +17,16 @@
 package org.traccar.database;
 
 import org.traccar.model.Attribute;
-import org.traccar.model.BaseModel;
 
-public class AttributesManager extends ExtendedObjectManager {
+public class AttributesManager extends ExtendedObjectManager<Attribute> {
 
     public AttributesManager(DataManager dataManager) {
         super(dataManager, Attribute.class);
     }
 
     @Override
-    public Attribute getById(long calendarId) {
-        return (Attribute) super.getById(calendarId);
-    }
-
-    @Override
-    public void updateCachedItem(BaseModel item) {
-        Attribute attribute = (Attribute) item;
-        Attribute cachedAttribute = getById(item.getId());
+    public void updateCachedItem(Attribute attribute) {
+        Attribute cachedAttribute = getById(attribute.getId());
         cachedAttribute.setDescription(attribute.getDescription());
         cachedAttribute.setAttribute(attribute.getAttribute());
         cachedAttribute.setExpression(attribute.getExpression());
--- a/src/org/traccar/database/BaseObjectManager.java
+++ b/src/org/traccar/database/BaseObjectManager.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2017 Anton Tananaev (anton@traccar.org)
+ * Copyright 2017 Andrey Kunitsyn (andrey@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.database;
+
+import java.sql.SQLException;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.traccar.helper.Log;
+import org.traccar.model.BaseModel;
+
+public class BaseObjectManager {
+
+    private final DataManager dataManager;
+
+    private Map<Long, BaseModel> items;
+    private Class<? extends BaseModel> baseClass;
+
+    protected BaseObjectManager(DataManager dataManager, Class<? extends BaseModel> baseClass) {
+        this.dataManager = dataManager;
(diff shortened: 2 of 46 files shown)


src/org/traccar/database/PermissionsManager.java
```diff
@@ -94,11 +94,12 @@ public void refreshServer() {
         }
     }
 
-    public final void refreshPermissions() {
+    public final void refreshDeviceAndGroupPermissions() {
```
Abyss777: Just renamed to reflect what it does.

src/org/traccar/database/PermissionsManager.java
```diff
@@ -176,9 +177,9 @@ public void checkDeviceLimit(long userId) throws SecurityException, SQLException
         if (deviceLimit != -1) {
             int deviceCount = 0;
             if (isManager(userId)) {
-                deviceCount = Context.getDeviceManager().getManagedDevices(userId).size();
+                deviceCount = Context.getDeviceManager().getManagedItems(userId).size();
             } else {
-                deviceCount = getDevicePermissions(userId).size();
+                deviceCount = Context.getDeviceManager().getUserItems(userId).size();
```
Abyss777: Basically nothing changed here, just a couple function calls, but it looks more consistent.

src/org/traccar/reports/Events.java
```diff
@@ -92,7 +92,7 @@ public static void getExcel(OutputStream outputStream,
             deviceEvents.setDeviceName(device.getName());
             sheetNames.add(WorkbookUtil.createSafeSheetName(deviceEvents.getDeviceName()));
             if (device.getGroupId() != 0) {
-                Group group = Context.getDeviceManager().getGroupById(device.getGroupId());
+                Group group = (Group) Context.getGroupsManager().getById(device.getGroupId());
```
tananaev: Can we have a generic function that will automatically return correct object type?

src/org/traccar/database/DeviceManager.java
```diff
@@ -169,76 +90,129 @@ public Device getDeviceByPhone(String phone) {
...
+        if (devicesByPhone == null) {
+            devicesByPhone = new ConcurrentHashMap<>(getAllItems().size());
+        }
+        devicesByPhone.put(device.getPhone(), device);
+    }
 
-        devicesById.put(device.getId(), device);
-        devicesByUniqueId.put(device.getUniqueId(), device);
+    @Override
+    protected void addNewItem(BaseModel item) {
```
tananaev: Can we use `Device` directly in the arguments here? Same probably applies to other classes as well.

src/org/traccar/database/DeviceManager.java
```diff
@@ -169,76 +90,129 @@ public Device getDeviceByPhone(String phone) {
...
+        Device device = (Device) item;
+        super.addNewItem(item);
```
tananaev: Please move super up above variable.

src/org/traccar/reports/Events.java
Abyss777: Similar to `<T> Collection<T> getItems(Class<T> clazz, Set<Long> itemIds)` but for the single object?

tananaev: Yes. Will it work? I think Java should be able to infer the type automatically.

Abyss777: It will. But it might look longer
````java
Group group = (Group) Context.getGroupsManager().getById(device.getGroupId());
````
vs
````java
Group group = Context.getGroupsManager().getById(Group.class, device.getGroupId());
````

File: src/org/traccar/database/DeviceManager.java (hunk 4 above)
Comment (Abyss777) [inline src/org/traccar/database/DeviceManager.java]: I'm afraid we can't. Overridden method have to have same set of parameters

File: src/org/traccar/reports/Events.java (hunk 3 above)
Comment (tananaev) [inline src/org/traccar/reports/Events.java]: I was thinking just
```java
Group group = Context.getGroupsManager().getById(device.getGroupId());
```

src/org/traccar/database/DeviceManager.java
tananaev: Right. We can only change return parameter. Then ignore my comment.

Abyss777:
Overrode `getById` in all managers, except `DeviceManager` because we already have `getDeviceById` for `IdentityManager`

tananaev:
Should we just rename methods in `IdentityManager` to match?

src/org/traccar/database/AttributesManager.java
```diff
@@ -26,9 +26,14 @@ public AttributesManager(DataManager dataManager) {
     }
 
     @Override
+    public Attribute getById(long calendarId) {
```
tananaev: I'm thinking now... can't we just change  `ExtendedObjectManager` to be generic?

src/org/traccar/database/BaseObjectManager.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.traccar.helper.Log;
+import org.traccar.model.BaseModel;
+
+public class BaseObjectManager {
```
tananaev: Probably same here. We can make it generic to avoid all those downcasts. What do you think?

Abyss777:
We have
````
Device getDeviceById(long id);

Device getDeviceByUniqueId(String uniqueId) throws Exception;
````
in `IdentityManager` probably we should also rename to `getByUniqueId`

**Comment by tananaev** [comment]
Yes, makes sense to rename both.

File: src/org/traccar/database/AttributesManager.java (hunk 6 above)
Comment (Abyss777) [inline src/org/traccar/database/AttributesManager.java]: It is mistake here, variable should be named `attributeId` of course.

I also thought about it, but I'm not familiar with it, It will take time.

Comment (tananaev) [inline src/org/traccar/database/AttributesManager.java]: It's really easy:

https://docs.oracle.com/javase/tutorial/java/generics/types.html

**Comment by Abyss777** [comment]
Generics are awesome, they perfectly fit to us.  Looks more ideal now. No type casts at all. Hope I've removed them from everywhere.

Also renamed `getDeviceById` to just `getById` a lot of files touched.

File: src/org/traccar/database/GroupsManager.java [hunk 8]
```diff
@@ -0,0 +1,104 @@
...
+    private AtomicLong groupsLastUpdate = new AtomicLong();
+    private final long dataRefreshDelay;
+
+    public GroupsManager(DataManager dataManager) {
+        super(dataManager, Group.class);
+        dataRefreshDelay = Context.getConfig().getLong("database.refreshDelay",
+                DeviceManager.DEFAULT_REFRESH_DELAY) * 1000;
+    }
+
+    private void checkGroupCycles(BaseModel group) {
```
tananaev: I think this needs to be updated as well.

Abyss777:
Fixed

tananaev:
Merged, thanks.

PR #3396, opened 2017-07-27
Combine PUT/POST/DELETE functions in object resources

Abyss777:
Implemented generic `BaseObjectResource` and combined all checks and updates in 3 functions.
They look rather complex, but really simple:
- Check before action
- Action: add/update/remove
- Refresh after action

Leaved `add` in users resource as is because it is too different.

I'm not very satisfied with `getManagers` function because of unchecked type casting warning, but it works correctly.

`@GET` functions not optimized for now, set of parameters and logic are different. Probably I'll implement some intermediate subclasses with common GET functions for `SimpleObjectManager`s and for `ExtendedObjectManager`s. They look similar.

### DIFF
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -378,7 +378,7 @@ public final class Context {
         identityManager = testIdentityManager;
     }
 
-    public static <T extends BaseModel> BaseObjectManager<T> getManager(Class<? extends BaseModel> clazz) {
+    public static <T extends BaseModel> BaseObjectManager<T> getManager(Class<T> clazz) {
         if (clazz.equals(Device.class)) {
             return (BaseObjectManager<T>) deviceManager;
         } else if (clazz.equals(Group.class)) {
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -28,6 +28,7 @@ import org.eclipse.jetty.util.URIUtil;
 import org.traccar.database.AliasesManager;
 import org.traccar.database.CalendarManager;
 import org.traccar.database.AttributesManager;
+import org.traccar.database.BaseObjectManager;
 import org.traccar.database.ConnectionManager;
 import org.traccar.database.DataManager;
 import org.traccar.database.DeviceManager;
@@ -51,6 +52,14 @@ import org.traccar.geocoder.OpenCageGeocoder;
 import org.traccar.geocoder.Geocoder;
 import org.traccar.geolocation.UnwiredGeolocationProvider;
 import org.traccar.helper.Log;
+import org.traccar.model.Attribute;
+import org.traccar.model.BaseModel;
+import org.traccar.model.Calendar;
+import org.traccar.model.Device;
+import org.traccar.model.Driver;
+import org.traccar.model.Geofence;
+import org.traccar.model.Group;
+import org.traccar.model.User;
 import org.traccar.geolocation.GoogleGeolocationProvider;
 import org.traccar.geolocation.GeolocationProvider;
 import org.traccar.geolocation.MozillaGeolocationProvider;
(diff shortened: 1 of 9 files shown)


src/org/traccar/api/resource/AttributeResource.java
```diff
@@ -85,15 +85,11 @@
 
     }
 
-    private Response add(Attribute entity) throws SQLException {
+    @POST
+    @Path("test")
```
Abyss777: Had to move test functionality in sub-path (actually it is more consistent with others) because we can't remove parent `@POST` annotation in child and can't override `add` with additional parameter.

src/org/traccar/Context.java
```diff
@@ -369,4 +378,23 @@ public static void init(IdentityManager testIdentityManager) {
         identityManager = testIdentityManager;
     }
 
+    public static <T extends BaseModel> BaseObjectManager<T> getManager(Class<? extends BaseModel> clazz) {
```
tananaev: Shouldn't it be `Class<T>`?

src/org/traccar/api/BaseObjectResource.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+        if (baseClass.equals(Group.class) || baseClass.equals(Device.class) || baseClass.equals(User.class)) {
+            Context.getPermissionsManager().refreshDeviceAndGroupPermissions();
+            if (baseClass.equals(User.class)) {
+                Context.getPermissionsManager().refreshAllUsersPermissions();
+            } else {
+                Context.getPermissionsManager().refreshAllExtendedPermissions();
+            }
+        }
+        // deprecated
```
tananaev: What does this mean?

src/org/traccar/api/BaseObjectResource.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+        Context.getPermissionsManager().checkPermission(baseClass, getUserId(), id);
+
+        BaseObjectManager<T> manager = Context.getManager(baseClass);
+        manager.removeItem(id);
+
+        if (manager instanceof SimpleObjectManager) {
+            ((SimpleObjectManager<T>) manager).refreshUserItems();
+        }
+        if (manager instanceof ExtendedObjectManager) {
```
tananaev: Can we use else?

src/org/traccar/api/BaseObjectResource.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @Path("{id}")
+    @PUT
+    public Response update(T entity) throws SQLException {
+        Context.getPermissionsManager().checkReadonly(getUserId());
+        if (baseClass.equals(Device.class)) {
+            Context.getPermissionsManager().checkDeviceReadonly(getUserId());
+        }
+        if (baseClass.equals(User.class)) {
```
tananaev: Please use `else` where possible to shorten and optimise code.

src/org/traccar/api/BaseObjectResource.java
```diff
@@ -0,0 +1,126 @@
...
+import org.traccar.Context;
+import org.traccar.database.BaseObjectManager;
+import org.traccar.database.ExtendedObjectManager;
+import org.traccar.database.SimpleObjectManager;
+import org.traccar.model.BaseModel;
+import org.traccar.model.Device;
+import org.traccar.model.Group;
+import org.traccar.model.User;
+
+public class BaseObjectResource<T extends BaseModel> extends BaseResource {
```
tananaev: Should we make this class abstract to indicate that it shouldn't be used directly?

src/org/traccar/api/BaseObjectResource.java
Abyss777: I did not combine next `if` with others to easily remove it when we remove Attribute Aliases.

tananaev: OK, can you please make a more detailed comment so we don't forget about it. By the way, when do you think it will happen?

Abyss777: You have warned users in 3.12 release post. I suggest after 3.14 release. To leave time in two releases for migration (if somebody used it). Also a have some cons removing it: Computed Attributes do not allow just translate attribute name, it allow duplicate some attribute with other (may be localized) name.

tananaev: Agree. By the way, it would be nice to make another release soon after all this refactoring. Maybe in about a month?

Abyss777: Yes, sure. I hope I'll finish everything and have at least a week for final testing and polishing.

src/org/traccar/api/BaseObjectResource.java
Abyss777: We can not, `refreshUserItems` we should do for both simple and extended.

Abyss777:
Looks like combined conditions that can be. Changed parameter declaration and comment.

src/org/traccar/api/BaseObjectResource.java
tananaev: `ExtendedObjectManager` extends `SimpleObjectManager`, so it can be inside the first `if`.

Abyss777:
You are right. Moved it inside.

tananaev:
Merged, thanks.

PR #3397, opened 2017-07-27
Initialize some maps, because they might be accessed before initialized.

Abyss777:
fix for https://www.traccar.org/forums/topic/build-source-code-error/
This maps might be not initialized if objects list is empty, but called `getXxxByUniqueId` function.


Abyss777:
Looks like some databases and engines do not save column name case. I've tried to construct name with `.toLowerCase` or something, but `ManagedUser` is out of list. Decided to use just switch.

tananaev:
Merged.

PR #3399, opened 2017-07-28
Combine some GET functions

Abyss777:
Put `getSimpleManagerItems` function to Base because it is impossible make strict inheritance Base->Simple->Extended because it is impossible override `@GET` and `get` with different set of parameters.

`ManagableObjects` interface become useful.


tananaev:
Merged. Thanks.

PR #3401, opened 2017-07-30
Round Robin Geocoders with DB Cache  loading

eriansv:
RRCacheGeocoder: Round Robin Geocoders  with DB Cache loading from the database.rrcache query.
database.rrcache: it must to return the follow fields -> regs,latitude,longitude,address

Configuration Example:

```xml
    <entry key='geocoder.cacheSize'>3000</entry>
    <entry key='geocoder.rrcache'>2</entry>
    <entry key='geocoder.type'>google</entry>
    <entry key='geocoder.limit'>2500</entry>
    <entry key='geocoder.type.1'>nominatim</entry>
    <entry key='geocoder.url.1'>https://locationiq.org/v1/reverse.php</entry>
    <entry key='geocoder.limit.1'>10000</entry>
    <entry key='geocoder.key.1'>xxxxx_your_key_xxxxx</entry>
    <entry key='geocoder.type.2'>opencage</entry>
    <entry key='geocoder.url.2'>http://api.opencagedata.com/geocode/v1</entry>
    <entry key='geocoder.limit.2'>2500</entry>
    <entry key='geocoder.key.2'>xxxxx_your_key_xxxxx</entry>

    <entry key='database.rrcache'>
	SELECT 
		COUNT(*) as regs,
		CAST(p.latitude AS NUMERIC(8, 5)) AS latitude,
		CAST(p.longitude AS NUMERIC(8, 5)) AS longitude,
		p.address
	FROM	
		positions p
	INNER JOIN
	(
		SELECT
			COUNT(*) AS ad_regs,
			address
		FROM
			positions
		WHERE
			address IS NOT NULL
		GROUP BY
			address
		HAVING
			COUNT(*) &gt; 100
		ORDER BY
			ad_regs DESC
		LIMIT 1000
	) as ad
	ON
		p.address = ad.address
	GROUP BY 
		latitude,longitude,p.address
	HAVING
		COUNT(*) &gt; 5
	ORDER BY
		regs desc
	LIMIT 2000
    </entry>
```

### DIFF
--- a/src/org/traccar/geocoder/RRCacheGeocoder.java
+++ b/src/org/traccar/geocoder/RRCacheGeocoder.java
@@ -1,0 +1,284 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.traccar.geocoder;
+
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.sql.ResultSet;
+import java.util.AbstractMap;
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.sql.DataSource;
+import org.traccar.Config;
+import org.traccar.Context;
+import org.traccar.helper.Log;
+
+/**
+ *
+ * @author eriansv@gmail.com
+ */
+
+public class RRCacheGeocoder implements Geocoder {
+
+    private final List<Provider> providers;
+    private Map<Map.Entry<Double, Double>, String> cache;
+    private int pointer = 0;
+    private int cacheCounter = 0;
+    private int cacheDay = 0;
+
+    private class Provider implements Geocoder {
+
+        private final String name;
+        private final Geocoder geocoder;
+        private final int limit;
+        private int counter = 0;
+        private int day = 0;
+
+        Provider(String name, Geocoder geocoder, int limit) {
+
+            this.name = name;
+            this.geocoder = geocoder;
+
+            if (limit > 0) {
+                this.limit = limit;
+            } else {
+                this.limit = 2000;
+            }
+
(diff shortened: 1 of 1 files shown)


tananaev:
Thank for pull request, but I don't think we can include it in Traccar. Three main reasons:

1. Cache SQL query is not universal, so it means it won't work on some databases
2. Cache loading will be very very slow if `positions` table is big
3. I am not sure that it's very ethical to encourage people to abuse free geocoding quota

I feel like I should explain last point a bit more. Most services that provide free quota do it for testing purposes, not for production use. We already have problems with Google, Nominatim and Gisgraphy where they seem to completely block Traccar requests because people abused their free quota in the past. It means that now I can't even use free services to test Traccar code.

eriansv:
I'm agree with your reasons.
Most important the last reason. Because the first and second could be fixed.

What about a Round Robin email sender?
I'll get my email account with servermx.com. But, I need to improve the limits on sending email and I was thinking get 2-4 email account and implement a  Round Robin email sender.

Regards,

tananaev:
Email services are really cheap comparing to reverse geocoding, so I don't see much point. For example, you can use Amazon SES for $0.10 per 1000 emails. Or even configure your own SMTP server.

tsmgeek:
What I thought that could be useful under this would be a way so that if the location only moves by a couple of meters it does not geocode and uses just last geocode. When stopped trackers can wonder a little and it then gives different addresses from next door etc, this would be a geo lookup each time unless there is an internal location table which prevents this based on distance.

tananaev:
@tsmgeek, such option is already available. Not sure how it's related to this pull request though.

tsmgeek:
I misunderstood the cache part on this PR.

The round Robin code though could be useful if altered to be used as backup, ie primary coder goes down or fails it will switch to backup for a while.

eriansv:
That version has reload issue.
I could provide the fixed version.

By the way, my version use locations with 5 decimal digits

tananaev:
Closing as I don't think we're going to merge this as is.

PR #3402, opened 2017-07-31
Implement SQL requests construction

Abyss777:
It does not cover all request, but most standard covers.

- Two objects Device and User have two update requests, implemented `@QueryAdditional` to split fields on standard and additional
- Decided not to add new requests cache, just add them to `Context.properties` if absent.
- Renamed "server" table to "servers" for consistence

More comments in code


src/org/traccar/database/DataManager.java
```diff
@@ -115,6 +124,73 @@ private void initDatabase() throws Exception {
...
+                StringBuilder values = new StringBuilder();
+
+                Set<Method> methods = new HashSet<>(Arrays.asList(clazz.getMethods()));
+                methods.removeAll(Arrays.asList(Object.class.getMethods()));
+                methods.removeAll(Arrays.asList(BaseModel.class.getMethods()));
+                for (Method method : methods) {
+                    if (method.getName().startsWith("get") && method.getParameterTypes().length == 0
+                            && (additional ? method.isAnnotationPresent(QueryAdditional.class)
+                                : !method.isAnnotationPresent(QueryIgnore.class)
+                                && !method.isAnnotationPresent(QueryAdditional.class))) {
```
Abyss777: This ternary operator (3 lines) distinguishes between standard and additional fields

src/org/traccar/model/Position.java
```diff
@@ -294,4 +297,10 @@ public void setNetwork(Network network) {
         this.network = network;
     }
 
+    @Override
+    @QueryIgnore
+    public String getType() {
```
Abyss777: It is inherited from `Message` but not used in Position.

src/org/traccar/database/DataManager.java
```diff
@@ -302,44 +402,50 @@ public static String makeNameId(Class<?> clazz) {
...
                 .executeUpdate());
     }
 
     public void updateObject(BaseModel entity) throws SQLException {
-        String query = "database.update" + entity.getClass().getSimpleName();
-        QueryBuilder.create(dataSource, getQuery(query))
+        QueryBuilder.create(dataSource, getQuery(ACTION_UPDATE, entity.getClass()))
                 .setObject(entity)
                 .executeUpdate();
+        if (entity instanceof User && ((User) entity).getHashedPassword() != null) {
```
Abyss777: Fixed password change missed in previous optimizations

src/org/traccar/database/DataManager.java
```diff
@@ -123,6 +199,59 @@ private String getQuery(String key) {
         return query;
     }
 
+    public String getQuery(String action, Class<?> clazz) {
```
Abyss777: It is just for short, because most calls with `additional = false`

src/org/traccar/database/DataManager.java
```diff
@@ -115,6 +124,73 @@ private void initDatabase() throws Exception {
...
+                            && (additional ? method.isAnnotationPresent(QueryAdditional.class)
+                                : !method.isAnnotationPresent(QueryIgnore.class)
+                                && !method.isAnnotationPresent(QueryAdditional.class))) {
+                        String name = method.getName().substring(3, 4).toLowerCase()
+                                + method.getName().substring(4);
+                        if (action.equals(ACTION_INSERT)) {
+                            fields.append(name).append(", ");
+                            values.append(":").append(name).append(", ");
+                        } else {
+                            fields.append(name).append(" = :").append(name).append(",\n");
```
tananaev: Do we need new lines in queries? I'm thinking it's just better to use spaces. New lines helped when we needed to construct queries ourselves, but for automatically generated ones I think spaces are better.

src/org/traccar/database/DataManager.java
tananaev: This if statement is really hard to read. I think you should split it. Also, why do we only check for ignore annotation only if additional is false?

src/org/traccar/database/DataManager.java
```diff
@@ -115,6 +124,73 @@ private void initDatabase() throws Exception {
...
+
+                Set<Method> methods = new HashSet<>(Arrays.asList(clazz.getMethods()));
+                methods.removeAll(Arrays.asList(Object.class.getMethods()));
+                methods.removeAll(Arrays.asList(BaseModel.class.getMethods()));
+                for (Method method : methods) {
+                    if (method.getName().startsWith("get") && method.getParameterTypes().length == 0
+                            && (additional ? method.isAnnotationPresent(QueryAdditional.class)
+                                : !method.isAnnotationPresent(QueryIgnore.class)
+                                && !method.isAnnotationPresent(QueryAdditional.class))) {
+                        String name = method.getName().substring(3, 4).toLowerCase()
```
tananaev: I think you can use `Introspector.decapitalize` here... maybe `Introspector` can be used in some other places as well (managers?).

src/org/traccar/database/DataManager.java
```diff
@@ -115,6 +124,73 @@ private void initDatabase() throws Exception {
...
+            + " (" + makeNameId(owner) + ", " + makeNameId(property) + ") VALUES (:"
+            + makeNameId(owner) + ", :" + makeNameId(property) + ")";
+        case ACTION_DELETE:
+            return "DELETE FROM " + getPermissionsTableName(owner, property)
+            + " WHERE " + makeNameId(owner) + " = :" + makeNameId(owner)
+            + " AND " + makeNameId(property) + " = :" + makeNameId(property);
+        default:
+            throw new IllegalArgumentException("Unknown action");
+    }
+    }
```
tananaev: There seems to be some problem with indentation here.

src/org/traccar/database/DataManager.java
```diff
@@ -115,6 +124,73 @@ private void initDatabase() throws Exception {
...
+                return result.toString();
+            case ACTION_SELECT_ALL:
+                return "SELECT * FROM " + getObjectsTableName(clazz);
+            case ACTION_SELECT:
+                return "SELECT * FROM " + getObjectsTableName(clazz) + " WHERE id = :id";
+            case ACTION_DELETE:
+                return "DELETE FROM " + getObjectsTableName(clazz) + " WHERE id = :id";
+            default:
+                throw new IllegalArgumentException("Unknown action");
+            }
```
tananaev: Indentation problem?

src/org/traccar/database/DataManager.java
```diff
@@ -123,6 +199,59 @@ private String getQuery(String key) {
...
+        return query;
+    }
+
+    private static String getPermissionsTableName(Class<?> owner, Class<?> property) {
+        String ownerName = owner.getSimpleName();
+        String propertyName = property.getSimpleName();
+        if (propertyName.equals("ManagedUser")) {
+            propertyName = "User";
+        }
+        return ownerName.substring(0, 1).toLowerCase() + ownerName.substring(1) + "_"
```
tananaev: `Introspector.decapitalize`?

src/org/traccar/database/DataManager.java
```diff
@@ -115,6 +124,73 @@ private void initDatabase() throws Exception {
         }
     }
 
+    public static String constructObjectQuery(String action, Class<?> clazz, boolean additional) {
```
tananaev: Can you please explain what `additional` mean? Ideally with some examples. Also, maybe `extended` is better name (depends on what it's used for).

src/org/traccar/database/DataManager.java
Abyss777: User has two stages of update, `hashedPassword`/`salt` is additional other fields are standard.
Device also have two stages, `lastUpdate` is additional, other fields are standard.

I did not check `@QueryIgnore` if additional is true, because I hope we will not use them in combination, it is meaningless, moreover it will cause problems because `QueryBuilder` will not set such fields in statement.

It was only way that I could invent. If you have another idea, please share.

src/org/traccar/database/DataManager.java
```diff
@@ -115,6 +125,76 @@ private void initDatabase() throws Exception {
...
+    }
+
+    public static String constructPermissionQuery(String action, Class<?> owner, Class<?> property) {
+        switch (action) {
+        case ACTION_SELECT_ALL:
+            return "SELECT " + makeNameId(owner) + ", " + makeNameId(property) + " FROM "
+                    + getPermissionsTableName(owner, property);
+        case ACTION_INSERT:
+            return "INSERT INTO " + getPermissionsTableName(owner, property)
+            + " (" + makeNameId(owner) + ", " + makeNameId(property) + ") VALUES (:"
```
tananaev: There is an indentation problem here and below.

src/org/traccar/database/DataManager.java
```diff
@@ -123,6 +203,56 @@ private String getQuery(String key) {
...
+    public String getQuery(String action, Class<?> clazz) {
+        return getQuery(action, clazz, false);
+    }
+
+    public String getQuery(String action, Class<?> clazz, boolean extended) {
+        String queryName;
+        if (action.equals(ACTION_SELECT_ALL)) {
+            queryName = "database.select" + clazz.getSimpleName() + "s";
+        } else {
+            queryName = "database." + action.toLowerCase() + clazz.getSimpleName() + (extended ? "Extended" : "");
```
tananaev: I would split the `extended` check into a new line with a regular `if` statement.

Abyss777:
Done, also fix travis-ci build, do not know why my Eclipse decided that `putIfAbsent` present in Java 7

tananaev:
Merged, thanks.

I forgot about one important thing here. In some cases we don't want some queries. For example, OpenGTS integration only requires two queries:

https://www.traccar.org/opengts/

My suggestion is to add a config parameter `database.generateQueries` which will be `false` by default (for backward compatibility), and we set it to `true` in the official release. What do you think?

Abyss777:
Yes, good solution.

Also want to note, to fix it in our memory (need to change documentation) we renamed some requests:
- `database.select[Object]sAll` -> `database.select[Object]s`
- `database.updateUserPassword` -> `database.updateUserExtended`
- `database.updateDeviceStatus` -> `database.updateDeviceExtended`

tananaev:
Yes, the first is the most important one because everyone is using at least `selectDevicesAll`.

PR #3406, opened 2017-08-01
Fix retrieving objects by unknown tokens/uniqueId

Abyss777:
There was a `NullPointerException` if we try to get object by unknown uniqueId or token.
I suppose there is no difference what to map String->Long or String->User, these are only links.
Replaced map to objects itself and now we correctly get `null` in `getByXXX` functions without additional checks. Also it is consistent with `DeviceManager`


tananaev:
Merged, thanks.

tananaev:
@gulsal123, if you want to ask a question create a new forum topic. Don't comment on random unrelated threads.

PR #3407, opened 2017-08-02
Avoid unneeded log record and fix table name constructions

Abyss777:
Two fixes:
- IF condition was wrong, log record was made if `query != null`. fixed that
- noticed incorrect table name construction
````
2017-08-02 00:10:12  WARN: Table 'traccar.statisticss' doesn't exist - MySQLSyntaxErrorException (... < QueryBuilder:477 < DataManager:445 < StatisticsManager:64 < ...)
````
There was two options: correct table name construction or rename `Statistics` to `Statistic` for consistence, decide first one. If it is better second, i'm ready to rename.

### DIFF
--- a/src/org/traccar/database/DataManager.java
+++ b/src/org/traccar/database/DataManager.java
@@ -222,11 +222,13 @@ public class DataManager {
             }
         }
         String query = config.getString(queryName);
-        if (query == null && generateQueries) {
-            query = constructObjectQuery(action, clazz, extended);
-            config.setString(queryName, query);
-        } else {
-            Log.info("Query not provided: " + queryName);
+        if (query == null) {
+            if (generateQueries) {
+                query = constructObjectQuery(action, clazz, extended);
+                config.setString(queryName, query);
+            } else {
+                Log.info("Query not provided: " + queryName);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/org/traccar/database/DataManager.java [hunk 1]
```diff
@@ -261,7 +266,11 @@ private static String getPermissionsTableName(Class<?> owner, Class<?> property)
     }
 
     private static String getObjectsTableName(Class<?> clazz) {
-        return Introspector.decapitalize(clazz.getSimpleName()) + "s";
+        String result = Introspector.decapitalize(clazz.getSimpleName());
+        if (!result.endsWith("s")) {
+            result += "s";
```
tananaev: Please add some comment here.

Abyss777: It is in case object name already plural, like `statistics`

tananaev: I meant to put a comment in the code, so we don't forget why it's here.

Abyss777:
Oh, sorry, added.

tananaev:
Merged, thanks.

PR #3410, opened 2017-08-04
Implement delayed stop/motion events

Abyss777:
Here is long-awaited feature: delayed stop/motion events. It may help avoid false events in traffic jams or fluctuations on parking.

It was inspired by timers used in `ConnectionManager`

If config parameters are not defined, old scheme is working (in case some resources troubles or for tests)

### DIFF
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -35,6 +35,7 @@ import org.traccar.database.DeviceManager;
 import org.traccar.database.DriversManager;
 import org.traccar.database.IdentityManager;
 import org.traccar.database.MediaManager;
+import org.traccar.database.MotionManager;
 import org.traccar.database.NotificationManager;
 import org.traccar.database.PermissionsManager;
 import org.traccar.database.GeofenceManager;
@@ -229,6 +230,12 @@ public final class Context {
         return smppClient;
     }
 
+    private static MotionManager motionManager;
+
+    public static MotionManager getMotionManager() {
+        return motionManager;
+    }
+
     public static void init(String[] arguments) throws Exception {
 
         config = new Config();
@@ -368,6 +375,11 @@ public final class Context {
 
         if (config.getBoolean("sms.smpp.enable")) {
             smppClient = new SmppClient();
+        }
+
+        if (config.getLong("event.motion.stopDelay") > 0 || config.getLong("event.motion.motionDelay") > 0) {
+            motionManager = new MotionManager(config.getLong("event.motion.stopDelay"),
+                    config.getLong("event.motion.motionDelay"));
         }
 
     }
--- a/src/org/traccar/database/MotionManager.java
+++ b/src/org/traccar/database/MotionManager.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2017 Anton Tananaev (anton@traccar.org)
+ * Copyright 2017 Andrey Kunitsyn (andrey@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.database;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+
+import org.jboss.netty.util.Timeout;
+import org.jboss.netty.util.TimerTask;
+import org.traccar.Context;
+import org.traccar.GlobalTimer;
+import org.traccar.model.Event;
+import org.traccar.model.Position;
+
+public class MotionManager {
+
+    private final long stopTimeout;
+    private final long motionTimeout;
+
+    private final Map<Long, Timeout> stopTimeouts = new ConcurrentHashMap<>();
+    private final Map<Long, Timeout> motionTimeouts = new ConcurrentHashMap<>();
+
+    public MotionManager(long stopDelay, long motionDelay) {
+        stopTimeout = stopDelay * 1000;
+        motionTimeout = motionDelay * 1000;
+    }
+
+    public void updateDeviceMotion(final Position position, boolean oldMotion) {
+        boolean motion = position.getBoolean(Position.KEY_MOTION);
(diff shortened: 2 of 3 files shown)


Abyss777:
What if it is last position before large gap, we will get this event only after this gap.

tananaev:
If there is a gap, then you don't know what actual state device is in. Your code just assumes that it's the same as last message, right? It's not necessarily a correct. Maybe for large gaps we just not show any events. We can have some threshold.

Abyss777:
Yes, but without this modification it also assumes as last, just generate event instant.

By the way, what is wrong with timers?

tananaev:
If there is a gap in data, we don't need to generate any event. We can assume state from the first position after the gap, but notify user only after threshold has passed.

As for reasons, there is a performance concern. It also makes code more complicated. It's harder to reason about the code with asynchronous timeout. I would prefer if we only use timers if absolutely necessary, like in case when device hasn't report any data for a long time and we want to generate an event for that.

Abyss777:
OK. here is an example. Speed values come every minute:

...10-5-1-0- "device goes sleep because of no movement and gap in data 10 hours" - 0-1-5...

If we have stopDelay 5 minutes, user will get event "Device stopped" after 10 hours or never. It absolutely wrong, no one user will understand this behavior.

tananaev:
No, if we implement it the way I'm suggesting there will be no event.

If there is a gap, we assume status on the first position after the gap, which is 0. Of course there might be GPS fluke which would result in a wrong status, but I think it's a very corner case.

Abyss777:
It is wrong, Event should be. Because device really stopped and we have information about it, speed is zero.

tananaev:
That's an assumption. Imagine sequence:

`...10-5-1-0-1-5-5-5...`

...and devices loses connection after first zero. It is not a stop.

Abyss777:
It was not moving before connection lost for us it is stopped. And as for me I would prefer get this stop event. (as it is currently implemented)

I think if device stopped and gone offline it is more likely stopped than lost connection, but if it was moving and gone offline it is more likely lost connection.

And here in this pull-request protection from such false events. If device come back online and report moving before timeout expired event will not be generated.

tananaev:
I don't like the idea of adding timers. That's my main concern. The server is already pretty heavy and I don't want to make it worse just because of this relatively small feature.

I think the gap in data is a common case only for devices that don't have back-up battery and die when ignition is turned off. For this particular case user can easily utilise "unknown" status event. I don't see much reason of adding another timer which will be used in exactly the same case as the "unknown" status event.

Abyss777:
OK. I understand your opinion.

Also gaps common case for magnet/pocket devices with sleep functionality, otherwise they could not live long from internal battery.

What do you think if we implement some event queues (for all type of events) is it worth to be implemented with help of timers?

tananaev:
Possibly If there is a valid use case for it, but it sounds like even more code complication.

Abyss777:
Most of all new features is code complication ...

tananaev:
Of course every new feature adds complexity, but we need to think about the balance and value that we get out of it. In this case I don't see much value on top of what can already be achieved with "unknown" event timer.

Abyss777:
I see. Will try invent something using just positions flow and device status for correction.

I think stop/motion events should be more reliable than now, in most cases they just annoying, especially in traffic jams and because of GPS fluctuations on parking. I mean it is not so small feature and rather demanded.
https://github.com/tananaev/traccar/issues/3408
https://github.com/tananaev/traccar/issues/2644
https://github.com/tananaev/traccar/issues/3204

tananaev:
Totally agree. My opinion is that we should save somewhere (possibly temporary in memory or device attributes?) current state and last potential state change timestamp. Then we analyse positions and see if time is more than threshold and state is opposite of the current one, then it means state has changed.

For example we have data like this (assuming threshold is 2 minutes):

00:01 - 0 km/h - no state before, so assuming that device is in stopped state
00:02 - 0 km/h - no changes
00:03 - 5 km/h - current state is stopped, potential state change at 00:03
00:04 - 0 km/h - it was a fluke, so keep as stopped and reset to timestamp to null
00:05 - 5 km/h - current state is stopped, potential state change at 00:05
00:06 - 10 km/h - current state is stopped, potential state change at 00:05
00:07 - 10 km/h - state changed to moving because it's more than 2 minutes with speed

Exactly the same for transition from movement to stop.

What do you think?

Abyss777:
Yes, like this. Also make some config switch that will enable offline/unknown status interpretation of potential state to event.

About place... device attributes will allow not move this code to external manager (like `MotionManager`) but they might be stored in database or passed to API. May be some additional device complex field with potential state and timestamp that will be `JsonIgnore` and `QueryIgnore`.
Or just make some `StateManager` that will trace motion state and timestamp as maps. Do the same for overspeed in future, may be some other states.

tananaev:
We can probably store it in the same place where we store last position. I believe it's in `DeviceManager`.

Another thought I has this morning. If there was a gap in data or we just have a new device without known state, we can assign state to "unknown" (not the same as connection state), and use the same timing logic to establish initial state. Probably in this case we don't want to send user a notification.

tananaev:
Another thought related to this... make logic universal between trips/stops and motion events.

Abyss777:
I do not think it is possible, we look too much in future. Basically, stop is detected only after next trip ended.
But I'll check if we can use more simple logic from motion events in trips/stops.

tananaev:
Sorry, I'm not saying that we should implement it straight away. For example we can start by unifying attributes. Later we can think of combining logic somehow.

Abyss777:
I also was thinking about unifying attributes. On the one hand it will be more clear for users, but on the other somebody might want to get events more faster than trip/stop minimal duration, for example in security use case.

tananaev:
I think making it more clear and consistent is a higher priority here.

Abyss777:
How will we handle next situation:
parkingDuration = 5 minutes, we get positions every 2 minutes:
00:00 0 kn
00:02 0 kn
00:04 0 kn
00:06 5 kn
... 
Difference between  1 and 3 position is not enough to trigger event, but on the 4 position we are switching state, theoretically we should throw event, what do you think?

tananaev:
I think we shouldn't, but it's really 50/50. You never know if device started moving before or after 5 minutes because there is not enough information. I think we should only throw event when it's guaranteed that threshold has passed.

Abyss777:
Logically you are right, but it will be more consistent with trips/stop. According to algorithm, parking end time is time of the first position following trip.

tananaev:
What about following example:

parkingDuration = 5 minutes
00:00 0 kn
00:06 5 kn

Abyss777:
According to trips detector parking will be last from 00:00 to 00:06

Abyss777:
Mostly because it was trips detector initially, and following trip started from 00:06.
I just added the logic that everything between trips is parking.

tananaev:
I guess if it works like this in reports, we should use the same logic in the alerts as well. We can always change it later if we want to.

PR #3416, opened 2017-08-04
Write all fields on insert action

### LINKED ISSUE #3411: Error creating users
@Abyss777, there seems to be some problem with creating new users after our changes to SQL scripts. If you try to register a new user from login or as an admin, you get following error:
```
NULL not allowed for column "HASHEDPASSWORD"; SQL statement:
INSERT INTO users (expirationTime, phone, zoom, timezone, token, disabled, attributes, longitude, latitude, userLimit, deviceReadonly, readonly, distanceUnit, deviceLimit, name, coordinateFormat, email, admin, twelveHourFormat, speedUnit, map) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) [23502-196] - JdbcSQLException (... < QueryBuilder:477 < DataManager:455 < BaseObjectManager:87 < ...)
```

Abyss777:
Write both standard and extended fields on insert action
fix #3411

### DIFF
--- a/src/org/traccar/database/DataManager.java
+++ b/src/org/traccar/database/DataManager.java
@@ -146,7 +146,7 @@ public class DataManager {
                         skip = !method.isAnnotationPresent(QueryExtended.class);
                     } else {
                         skip = method.isAnnotationPresent(QueryIgnore.class)
-                                || method.isAnnotationPresent(QueryExtended.class);
+                                || method.isAnnotationPresent(QueryExtended.class) && !action.equals(ACTION_INSERT);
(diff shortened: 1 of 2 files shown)


tananaev:
Merged, thanks for the fix. Do you think there are any other corner cases that we need to test to make sure queries are working correctly?

Abyss777:
There might be, but now I hope unlikely.
But testing is never too much.

PR #3420, opened 2017-08-06
Update to EElink Decoder/Encoder

tsmgeek:
- update EElink protocol as per v1.8.5 specs
- add support for processing SMS messages received though tracker that are forwarded to server
- add support to process DOWNLINK messages received
- fix issues in parsing of various messages as several have GPS data but not STATE data, these have now been split into separate functions
- validate header of packet starts per spec
- validate length of packet per spec

### DIFF
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -256,7 +256,8 @@ public final class Context {
                 config.getLong("report.trip.minimalTripDuration", 300) * 1000,
                 config.getLong("report.trip.minimalParkingDuration", 300) * 1000,
                 config.getBoolean("report.trip.greedyParking"),
-                config.getLong("report.trip.minimalNoDataDuration", 3600) * 1000);
+                config.getLong("report.trip.minimalNoDataDuration", 3600) * 1000,
+                config.getBoolean("report.trip.useIgnition"));
     }
 
     public static void init(String[] arguments) throws Exception {
--- a/src/org/traccar/events/MotionEventHandler.java
+++ b/src/org/traccar/events/MotionEventHandler.java
@@ -78,13 +78,19 @@ public class MotionEventHandler extends BaseEventHandler {
         if (motionPosition != null) {
             long motionTime = motionPosition.getFixTime().getTime();
             double distance = ReportUtils.calculateDistance(motionPosition, position, false);
+            Boolean ignition = null;
+            if (tripsConfig.getUseIgnition()
+                    && position.getAttributes().containsKey(Position.KEY_IGNITION)) {
+                ignition = position.getBoolean(Position.KEY_IGNITION);
+            }
             if (newMotion) {
                 if (motionTime + tripsConfig.getMinimalTripDuration() <= currentTime
                         || distance >= tripsConfig.getMinimalTripDistance()) {
                     result = newEvent(deviceState, newMotion);
                 }
             } else {
-                if (motionTime + tripsConfig.getMinimalParkingDuration() <= currentTime) {
+                if (motionTime + tripsConfig.getMinimalParkingDuration() <= currentTime
+                        || ignition != null && !ignition) {
                     result = newEvent(deviceState, newMotion);
                 }
             }
--- a/src/org/traccar/reports/model/TripsConfig.java
+++ b/src/org/traccar/reports/model/TripsConfig.java
@@ -21,13 +21,14 @@ public class TripsConfig {
     public TripsConfig() {
     }
 
-    public TripsConfig(double minimalTripDistance, long minimalTripDuration,
-            long minimalParkingDuration, boolean greedyParking, long minimalNoDataDuration) {
+    public TripsConfig(double minimalTripDistance, long minimalTripDuration, long minimalParkingDuration,
+            boolean greedyParking, long minimalNoDataDuration, boolean useIgnition) {
         this.minimalTripDistance = minimalTripDistance;
         this.minimalTripDuration = minimalTripDuration;
         this.minimalParkingDuration = minimalParkingDuration;
         this.greedyParking = greedyParking;
         this.minimalNoDataDuration = minimalNoDataDuration;
+        this.useIgnition = useIgnition;
     }
 
     private double minimalTripDistance;
@@ -78,6 +79,16 @@ public class TripsConfig {
 
     public void setMinimalNoDataDuration(long minimalNoDataDuration) {
         this.minimalNoDataDuration = minimalNoDataDuration;
+    }
+
+    private boolean useIgnition;
+
+    public boolean getUseIgnition() {
+        return useIgnition;
+    }
+
+    public void setUseIgnition(boolean useIgnition) {
+        this.useIgnition = useIgnition;
     }
 
 }
--- a/test/org/traccar/events/MotionEventHandlerTest.java
+++ b/test/org/traccar/events/MotionEventHandlerTest.java
@@ -1,6 +1,7 @@
 package org.traccar.events;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
@@ -29,7 +30,7 @@ public class MotionEventHandlerTest extends BaseTest {
     @Test
     public void testMotionWithPosition() throws Exception {
         MotionEventHandler motionEventHandler = new MotionEventHandler(
-                new TripsConfig(500, 300 * 1000, 300 * 1000, false, 0));
+                new TripsConfig(500, 300 * 1000, 300 * 1000, false, 0, false));
 
         Position position = new Position();
         position.setTime(date("2017-01-01 00:00:00"));
@@ -68,7 +69,7 @@ public class MotionEventHandlerTest extends BaseTest {
     @Test
     public void testMotionWithStatus() throws Exception {
         MotionEventHandler motionEventHandler = new MotionEventHandler(
-                new TripsConfig(500, 300 * 1000, 300 * 1000, false, 0));
+                new TripsConfig(500, 300 * 1000, 300 * 1000, false, 0, false));
 
         Position position = new Position();
         position.setTime(new Date(System.currentTimeMillis() - 360000));
@@ -82,6 +83,31 @@ public class MotionEventHandlerTest extends BaseTest {
         assertNotNull(event);
         assertEquals(Event.TYPE_DEVICE_MOVING, event.getType());
         assertTrue(deviceState.getMotionState());
+        assertNull(deviceState.getMotionPosition());
+    }
+
+    @Test
+    public void testStopWithPositionIgnition() throws Exception {
+        MotionEventHandler motionEventHandler = new MotionEventHandler(
+                new TripsConfig(500, 300 * 1000, 300 * 1000, false, 0, true));
+
+        Position position = new Position();
+        position.setTime(date("2017-01-01 00:00:00"));
+        position.set(Position.KEY_MOTION, false);
+        position.set(Position.KEY_IGNITION, true);
+        DeviceState deviceState = new DeviceState();
+        deviceState.setMotionState(true);
+        deviceState.setMotionPosition(position);
+
+        Position nextPosition = new Position();
+        nextPosition.setTime(date("2017-01-01 00:02:00"));
+        nextPosition.set(Position.KEY_MOTION, false);
+        nextPosition.set(Position.KEY_IGNITION, false);
+
+        Event event = motionEventHandler.updateMotionState(deviceState, nextPosition);
+        assertNotNull(event);
+        assertEquals(Event.TYPE_DEVICE_STOPPED, event.getType());
+        assertFalse(deviceState.getMotionState());
         assertNull(deviceState.getMotionPosition());
     }
 
--- a/test/org/traccar/reports/ReportUtilsTest.java
+++ b/test/org/traccar/reports/ReportUtilsTest.java
@@ -79,7 +79,7 @@ public class ReportUtilsTest extends BaseTest {
                 position("2016-01-01 00:06:00.000", 0, 3000),
                 position("2016-01-01 00:07:00.000", 0, 3000));
 
-        TripsConfig tripsConfig = new TripsConfig(500, 300000, 180000, false, 900000);
+        TripsConfig tripsConfig = new TripsConfig(500, 300000, 180000, false, 900000, false);
 
         Collection<? extends BaseReport> result = ReportUtils.detectTripsAndStops(tripsConfig, false, 0.01, data, true);
 
@@ -119,7 +119,7 @@ public class ReportUtilsTest extends BaseTest {
                 position("2016-01-01 00:04:00.000", 1, 0),
                 position("2016-01-01 00:05:00.000", 0, 0));
 
-        TripsConfig tripsConfig = new TripsConfig(500, 300000, 200000, false, 900000);
+        TripsConfig tripsConfig = new TripsConfig(500, 300000, 200000, false, 900000, false);
 
         Collection<? extends BaseReport> result = ReportUtils.detectTripsAndStops(tripsConfig, false, 0.01, data, false);
 
@@ -145,7 +145,7 @@ public class ReportUtilsTest extends BaseTest {
                 position("2016-01-01 00:04:00.000", 1, 0),
                 position("2016-01-01 00:05:00.000", 2, 0));
 
-        TripsConfig tripsConfig = new TripsConfig(500, 300000, 200000, false, 900000);
+        TripsConfig tripsConfig = new TripsConfig(500, 300000, 200000, false, 900000, false);
 
         Collection<? extends BaseReport> result = ReportUtils.detectTripsAndStops(tripsConfig, false, 0.01, data, false);
 
@@ -184,7 +184,7 @@ public class ReportUtilsTest extends BaseTest {
                 position("2016-01-01 00:04:00.000", 0, 0),
                 position("2016-01-01 00:05:00.000", 0, 0));
 
-        TripsConfig tripsConfig = new TripsConfig(500, 300000, 200000, false, 900000);
+        TripsConfig tripsConfig = new TripsConfig(500, 300000, 200000, false, 900000, false);
 
         Collection<? extends BaseReport> result = ReportUtils.detectTripsAndStops(tripsConfig, false, 0.01, data, false);
 
@@ -210,7 +210,7 @@ public class ReportUtilsTest extends BaseTest {
                 position("2016-01-01 00:04:00.000", 5, 0),
                 position("2016-01-01 00:05:00.000", 5, 0));
 
-        TripsConfig tripsConfig = new TripsConfig(500, 300000, 200000, false, 900000);
+        TripsConfig tripsConfig = new TripsConfig(500, 300000, 200000, false, 900000, false);
 
         Collection<? extends BaseReport> result = ReportUtils.detectTripsAndStops(tripsConfig, false, 0.01, data, false);
 
@@ -232,7 +232,7 @@ public class ReportUtilsTest extends BaseTest {
                 position("2016-01-01 00:24:00.000", 5, 800),
                 position("2016-01-01 00:25:00.000", 5, 900));
 
-        TripsConfig tripsConfig = new TripsConfig(500, 200000, 200000, false, 900000);
+        TripsConfig tripsConfig = new TripsConfig(500, 200000, 200000, false, 900000, false);
 
         Collection<? extends BaseReport> result = ReportUtils.detectTripsAndStops(tripsConfig, false, 0.01, data, true);
 


tsmgeek:
CI fails due to requirement on JDK 1.7 but I cant download this version.

pom.xml
```diff
@@ -309,6 +309,14 @@
                 <artifactId>maven-pmd-plugin</artifactId>
                 <version>3.8</version>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <configuration>
+                    <configLocation>config/sun_checks.xml</configLocation>
```
tananaev: First of all, it's a wrong config file. Secondly, we already do `checkstyle` on the build step.

src/org/traccar/model/ExtendedModel.java
```diff
@@ -122,4 +122,10 @@ public long getLong(String key) {
         }
     }
 
+    public void remove(String key) {
```
tananaev: Why do you need this?

src/org/traccar/protocol/EelinkProtocolEncoder.java
```diff
@@ -54,6 +54,17 @@ protected Object encodeCommand(Command command) {
                 return encodeContent("RELAY,0#");
             case Command.TYPE_REBOOT_DEVICE:
                 return encodeContent("RESET#");
+            case Command.TYPE_GET_VERSION:
+                return encodeContent("VERSION#");
+            case Command.TYPE_GET_DEVICE_STATUS:
+                return encodeContent("STATUS#");
+            case Command.TYPE_SOS_NUMBER:
+                String num = command.getString(Command.KEY_DATA);
+                if (!num.isEmpty()) {
```
tananaev: I think we should check for `null` here as well.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -25,40 +25,55 @@
...
     public static final int MSG_REPORT = 0x15;
     public static final int MSG_COMMAND = 0x16;
+
     public static final int MSG_OBD_DATA = 0x17;
     public static final int MSG_OBD_BODY = 0x18;
     public static final int MSG_OBD_CODE = 0x19;
-    public static final int MSG_CAMERA_INFO = 0x1E;
-    public static final int MSG_CAMERA_DATA = 0x1F;
 
-    private void sendResponse(Channel channel, int type, int index) {
```
tananaev: Why did you rename it? That's a standard name we use in all other decoders.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -25,40 +25,55 @@
...
+    private int index = 0x0000;
+
     public static final int MSG_LOGIN = 0x01;
     public static final int MSG_GPS = 0x02;
     public static final int MSG_HEARTBEAT = 0x03;
     public static final int MSG_ALARM = 0x04;
     public static final int MSG_STATE = 0x05;
     public static final int MSG_SMS = 0x06;
     public static final int MSG_OBD = 0x07;
+    public static final int MSG_OBDFAULT = 0x09;
```
tananaev: Words should be separated by underscore.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -25,40 +25,55 @@
 import org.traccar.model.CellTower;
 import org.traccar.model.Network;
 import org.traccar.model.Position;
+import org.traccar.helper.Log;
 
 import java.net.SocketAddress;
 import java.util.Date;
 
+import java.nio.charset.StandardCharsets;
+//import org.apache.commons.lang.StringUtils;
```
tananaev: Please remove any commented code.

src/org/traccar/model/Position.java
```diff
@@ -80,7 +80,8 @@
     public static final String KEY_OBD_SPEED = "obdSpeed"; // knots
     public static final String KEY_OBD_ODOMETER = "obdOdometer"; // meters
 
-    public static final String KEY_RESULT = "result";
+    public static final String KEY_RESULT = "result"; // command results
+    public static final String KEY_INFO = "message";
```
tananaev: It would make sense to name it `KEY_MESSAGE`.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -97,29 +112,61 @@ private String decodeAlarm(Short value) {
     }
 
     private void decodeStatus(Position position, int status) {
+        /**
```
tananaev: Please don't use multiline comments, especially with double-asterisk (javadoc comment).

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -97,29 +112,61 @@ private String decodeAlarm(Short value) {
...
+         * 16 bits, read lowest 0 to 15
+         */
+
+        // GPS - 0 bit
```
tananaev: I think you should remove all these unnecessary comments. Code is pretty clear.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tananaev: Also, the comment seems unnecessary.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -97,29 +112,61 @@ private String decodeAlarm(Short value) {
...
+        // Digital input 2 - 13 bit - def 1 high
+        position.set(Position.PREFIX_IN + 2, BitUtil.check(status, 13));
+
+        // Digital input 3 - 14 bit - def 0 low
+        position.set(Position.PREFIX_IN + 3, BitUtil.check(status, 14));
+
+        // Digital input 4 - 15 bit - def 1 high
+        position.set(Position.PREFIX_IN + 4, BitUtil.check(status, 15));
+
+        // log the whole status packet
```
tananaev: Completely unnecessary.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -97,29 +112,61 @@ private String decodeAlarm(Short value) {
...
         position.set(Position.KEY_STATUS, status);
     }
 
-    private Position decodeOld(DeviceSession deviceSession, ChannelBuffer buf, int type, int index) {
-
+    private Position decodeInit(DeviceSession deviceSession, ChannelBuffer buf, int type, int index) {
```
tananaev: Why is this separated into a method? It seems to be used only once.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -131,38 +178,120 @@ private Position decodeOld(DeviceSession deviceSession, ChannelBuffer buf, int t
...
+        position.setValid(false);
 
-            position.set(Position.KEY_RSSI, buf.readUnsignedShort());
+        switch (buf.readUnsignedByte()) {
+            case MSGSIGN_COMMAND:
+                buf.skipBytes(4); //server flag
 
-            position.set(Position.PREFIX_ADC + 1, buf.readUnsignedShort());
-            position.set(Position.PREFIX_ADC + 2, buf.readUnsignedShort());
+                int remainingbytes = buf.readableBytes();
```
tananaev: Why do you need a variable here?

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -209,41 +338,68 @@ protected Object decode(
...
-
                 getLastLocation(position, null);
 
                 decodeStatus(position, buf.readUnsignedShort());
 
+                sendAck(channel, type, index);
+
                 return position;
+            } else {
+                Log.warning(new UnsupportedOperationException());
```
tananaev: We don't usually log these.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -209,41 +338,68 @@ protected Object decode(
...
 
+                // send default ack for unknown data just incase
```
tananaev: Unnecessary comment. Also typo.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -25,40 +25,55 @@
...
     public static final int MSG_OBD_BODY = 0x18;
     public static final int MSG_OBD_CODE = 0x19;
-    public static final int MSG_CAMERA_INFO = 0x1E;
-    public static final int MSG_CAMERA_DATA = 0x1F;
 
-    private void sendResponse(Channel channel, int type, int index) {
+    public static final int MSG_CAMERA_INFO = 0x0E;
+    public static final int MSG_CAMERA_DATA = 0x0F;
+
+    public static final int MSGSIGN_COMMAND = 0x01;
```
tananaev: What does `MSGSIGN` mean?

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -131,38 +178,120 @@ private Position decodeOld(DeviceSession deviceSession, ChannelBuffer buf, int t
...
+                    response.writeByte(0x67);
+                    response.writeByte(0x67); // header
+                    response.writeByte(type);
+                    response.writeShort(2 + 21 + responsemessage.getBytes().length); // length
+                    response.writeShort(index); // index
+                    response.writeBytes(ppp); // phone number
+                    response.writeBytes(responsemessage.getBytes(StandardCharsets.UTF_8)); // message
+                    channel.write(response);
+                    //System.out.println(response.readableBytes());
+                    //System.out.println(Hex.encodeHexString(response.readBytes(response.readableBytes()).array()));
```
tananaev: Remove any commented code.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -131,38 +178,120 @@ private Position decodeOld(DeviceSession deviceSession, ChannelBuffer buf, int t
...
+                        responsemessage = "PONG";
+                        break;
+                    default:
+                        responsemessage = "";
+                        break;
+                }
+
+                // pad number to 21 bytes
+                byte[] pp = phone.getBytes(StandardCharsets.US_ASCII);
+                ChannelBuffer ppp = ChannelBuffers.buffer(21);
```
tananaev: Please use sensible names for all variables. This is not acceptable.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -131,38 +178,120 @@ private Position decodeOld(DeviceSession deviceSession, ChannelBuffer buf, int t
...
+                    position.set(Position.PREFIX_ADC + 1, buf.readUnsignedShort());
+                    position.set(Position.PREFIX_ADC + 2, buf.readUnsignedShort());
+                }
+                break;
+            case MSG_SMS:
+                String phone = buf.readBytes(21).toString(StandardCharsets.US_ASCII);
+                String realnumber = phone.replaceAll("^(.*)\\+", "+").trim();
+                int remainingbytes = buf.readableBytes();
+                String message = buf.readBytes(remainingbytes).toString(StandardCharsets.UTF_8);
+                position.set(Position.KEY_RESULT, "From: " + realnumber + "\n" + message);
```
tananaev: Why is there a new line? Result should contain raw data from the device as it is, not a formatted string. If you want to save a phone number, I suggest using separate key.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -131,38 +178,120 @@ private Position decodeOld(DeviceSession deviceSession, ChannelBuffer buf, int t
...
+                    position.set(Position.KEY_RSSI, buf.readUnsignedShort());
+
+                    position.set(Position.PREFIX_ADC + 1, buf.readUnsignedShort());
+                    position.set(Position.PREFIX_ADC + 2, buf.readUnsignedShort());
+                }
+                break;
+            case MSG_SMS:
+                String phone = buf.readBytes(21).toString(StandardCharsets.US_ASCII);
+                String realnumber = phone.replaceAll("^(.*)\\+", "+").trim();
+                int remainingbytes = buf.readableBytes();
```
tananaev: Not following naming convention, and seems unnecessary.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -131,38 +178,120 @@ private Position decodeOld(DeviceSession deviceSession, ChannelBuffer buf, int t
...
+
+                    position.set(Position.KEY_RSSI, buf.readUnsignedShort());
+
+                    position.set(Position.PREFIX_ADC + 1, buf.readUnsignedShort());
+                    position.set(Position.PREFIX_ADC + 2, buf.readUnsignedShort());
+                }
+                break;
+            case MSG_SMS:
+                String phone = buf.readBytes(21).toString(StandardCharsets.US_ASCII);
+                String realnumber = phone.replaceAll("^(.*)\\+", "+").trim();
```
tananaev: Not following naming convention. Also, why are you doing this? Please provide examples.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -131,38 +178,120 @@ private Position decodeOld(DeviceSession deviceSession, ChannelBuffer buf, int t
...
+                int remainingbytes = buf.readableBytes();
+                String message = buf.readBytes(remainingbytes).toString(StandardCharsets.UTF_8);
+                position.set(Position.KEY_RESULT, "From: " + realnumber + "\n" + message);
+
+                String responsemessage;
```
tananaev: Not following naming convention.

pom.xml
tsmgeek: Sorry, it was just throwing up some warning on my NB build that version param was missing.

src/org/traccar/model/ExtendedModel.java
tsmgeek: I first thought that after calling getLastPosition() that the attributes all came across, but could not find a way to remove a specific attribute. I dont use it now but cant see it being detrimental having this option.

pom.xml
tananaev: How does this solve a warning? I'm confused.

src/org/traccar/protocol/EelinkProtocolEncoder.java
tsmgeek: Agreed

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: Ok

src/org/traccar/protocol/EelinkProtocolEncoder.java
tananaev: Please don't reply to every comment. Fix everything that you can/agree, and then we can discuss other ones.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: It is not so much a response as an ACK, sending a response is slightly different and there is more data that a few functions require.
I thought about refactoring this so that either an ACK or response could be in the same function.
I can revert if you want.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: They just refer to Message Sign in the spec documents. I suspect they mean Message Type.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: For some reason the device puts 20 before the phone number and this is to remove it for display, its not documented in their spec but its also not part of the number at all that I am sending the message from. + will always be the start of the number and nothing should precede it.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: For some reason the device puts 20 before the phone number and this is to remove it for display, its not documented in their spec but its also not part of the number at all that I am sending the message from. + will always be the start of the number and nothing should precede it.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: I wanted the number that sent the SMS message to the GPS device to be seen.
If I used a separate value it would not be.
I could store the raw message and number in separate keys and the merged one in the result, or implement another way to display a rendered message to user.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tananaev: ACK is a response. By the way, you should probably combine ACK with some other code where you send response as well.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: Changed line as follows.

`String content = buf.readBytes(buf.readableBytes()).toString(StandardCharsets.UTF_8);`

src/org/traccar/protocol/EelinkProtocolDecoder.java
tananaev: Then I would recommend to use `substring` and `indexOf` functions instead of regex.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: Leftover, its been removed.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tananaev: Maybe something like "phone_no: message" will be acceptable. Formatting should be minimal and message should be as close to the original as possible.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tananaev: OK, then it should be `MSG_SIGN_`.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -112,9 +118,7 @@ private String decodeAlarm(Short value) {
     }
 
     private void decodeStatus(Position position, int status) {
-        /**
-         * 16 bits, read lowest 0 to 15
-         */
+        /* 16 bits, read lowest 0 to 15 */
```
tananaev: I don't understand the purpose of this comment. What value does it provide? Also, as I said, don't use multi-line comments (`/* */`).

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -241,45 +235,41 @@ private Position decodeOld(
...
                 }
 
-                // pad number to 21 bytes
-                byte[] pp = phone.getBytes(StandardCharsets.US_ASCII);
-                ChannelBuffer ppp = ChannelBuffers.buffer(21);
-                ppp.writeBytes(pp);
-                for (int j = pp.length; j < 21; j++) {
-                    ppp.writeByte(0x00);
-                }
+                // send back response
```
tananaev: Please remove any unnecessary comments like this.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -241,45 +235,41 @@ private Position decodeOld(
...
+                if (responseMessage != null && !responseMessage.isEmpty()) {
+                    // pad number to 21 bytes
+                    ChannelBuffer paddedPhone = ChannelBuffers.buffer(21);
+                    paddedPhone.writeBytes(phone.getBytes(StandardCharsets.US_ASCII));
+                    for (int j = paddedPhone.readableBytes(); j < 21; j++) {
```
tananaev: Use `buffer.capacity()` to avoid hardcoding same number twice.

tananaev: Also, you can probably just set `writerIndex` instead of doing a loop.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -241,45 +235,41 @@ private Position decodeOld(
...
-                    response.writeByte(0x67); // header
-                    response.writeByte(type);
-                    response.writeShort(2 + 21 + responsemessage.getBytes().length); // length
-                    response.writeShort(index); // index
-                    response.writeBytes(ppp); // phone number
-                    response.writeBytes(responsemessage.getBytes(StandardCharsets.UTF_8)); // message
-                    channel.write(response);
-                    //System.out.println(response.readableBytes());
-                    //System.out.println(Hex.encodeHexString(response.readBytes(response.readableBytes()).array()));
+                    response.writeBytes(paddedPhone); // phone number
```
tananaev: Please remove unnecessary comments. You can already see that it's a phone number. What's the point of the comment?

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -379,7 +369,12 @@ protected Object decode(
                 return null;
             }
 
-            Position position = decodeInit(deviceSession, buf, type, index);
+            /* Init position */
```
tananaev: Please don't use multi-line comments. Also, this comment is unnecessary.

src/org/traccar/protocol/EelinkProtocolEncoder.java
```diff
@@ -60,7 +60,7 @@ protected Object encodeCommand(Command command) {
                 return encodeContent("STATUS#");
             case Command.TYPE_SOS_NUMBER:
                 String num = command.getString(Command.KEY_DATA);
```
tananaev: Please rename this variable to a full word.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: capacity() - will tell me its 21 bytes long, readableBytes() tells me how many bytes has been written so far.

Are you saying that putting writerIndex = capacity() it should essentially zero pad vs me doing the for loop? If thats the case then yes I would guess it works. Ile test.

tananaev: Yes, I am saying set `writerIndex` to `capacity()`. Java always initializes arrays with zeroes, so you don't need to worry about it.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -209,41 +325,72 @@ protected Object decode(
...
+        if (buf.readShort() != HEADER_KEY) {
+            Log.warning(new IllegalArgumentException("Invalid Header"));
+            return null;
+        }
+
+        int type = buf.readUnsignedByte(); // protocol number
 
-        if (type != MSG_GPS && type != MSG_DATA) {
-            sendResponse(channel, type, index);
+        short length = buf.readShort(); // length
```
tananaev: Please remove unnecessary comments. Please remove it in ALL your code, not just this line.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -209,41 +325,72 @@ protected Object decode(
...
+        if (buf.readableBytes() != length) { // check remaining length
+            Log.warning(new IllegalArgumentException("Packet length error"));
+            return null;
         }
 
-        if (type == MSG_LOGIN) {
+        index = buf.readUnsignedShort();
+
+        // check if we have received an old packet
+        /*int newIndex = buf.readUnsignedShort(); // current message index
```
tananaev: Remove commented code. Applies to all commented code, not just this part.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -131,38 +172,113 @@ private Position decodeOld(DeviceSession deviceSession, ChannelBuffer buf, int t
...
+                    case "PING?":
+                    case "PING#":
+                        responseMessage = "PONG";
+                        break;
+                    default:
+                        responseMessage = null;
+                        break;
+                }
+
+
```
tananaev: Remove extra space.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: This part just needs a little more testing. This was why I asked a question about if the decoder was inited once for the connection or for each packet. If per connection this check would work just fine except I need to handle when it rolls over back to 0 again.

tsmgeek: Would this be a valid check for rollover in a short?

		`short index = 10; //max 32767
		short counter = Short.MAX_VALUE-10;
		
		if (index < counter && (counter - index) < 32700) {
			System.out.println("old");
		} else {
			System.out.println("new");
			counter = index;
		}`

tananaev: Can you please explain what the problem is, why we need to store index in the decoder, and why we need a rollover.

tsmgeek: The unit sends a incrementing message number in every packet, it increments, replies send back the same index it received.
The protection would be for the possibility that the server receives either the same packet twice from the tracker (because an ACK was not received, this is not clear in the API spec) or if received out of order for some reason.

tananaev: We don't do such things in Traccar. Just respond with received index. I don't see much harm in occasional duplicate message, but it's also possible to filter it out if user doesn't want duplicates in the database.

tsmgeek: But that may still end up doing something, ie say someone did an action based on a received status change, but if they saw it twice when the second was just a duplicate it could cause a problem.
Personally I cant see the harm if the option to have it in the decoder if index variable persists across each packet received.

ie this unit can forward SMS received to the server, you only want to process these once if that action was designed say to top up the credit.

tananaev: It is not consistent with other protocols. And if you really want to implement some filtering, it shouldn't be in the decoder.

You example seems like a rare corner case, and most likely if device didn't receive confirmation, it will reconnect and you will lose your index anyway.

tsmgeek:
Is there a way to have a 'message' which is not an alert?

With heartbeats, I have to getLastLocation(), does this add additional row of data? If it does, would it not be better to have a method to update the device last update timestamp rather than add an additional row in DB.

tananaev:
You don't need to do anything about heartbeat in the decoder. Traccar will automatically update device status.

tsmgeek:
Just found that part in BaseProtocolDecoder where even returning null marks device as online.

It seems that heartbeats on eelink proto also passes status flags which seems to be the reason that a Position is returned with the last location data else there seems to be no way to record these states otherwise.

src/org/traccar/model/ExtendedModel.java
tananaev: I don't see any reason for having it. Less code is better.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -25,43 +25,60 @@
...
-    public static final int MSG_CAMERA_DATA = 0x1F;
 
-    private void sendResponse(Channel channel, int type, int index) {
+    public static final int MSG_CAMERA_INFO = 0x0E;
+    public static final int MSG_CAMERA_DATA = 0x0F;
+
+    public static final int MSG_SIGN_COMMAND = 0x01;
+    public static final int MSG_SIGN_NEWS = 0x02;
+
+    private void sendResponse(Channel channel, int type, int index, ChannelBuffer buf) {
```
tananaev: Please rename `buf` to `content`, so it's clear that it's not the incoming message buffer.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -25,43 +25,60 @@
...
+            ChannelBuffer response = ChannelBuffers.dynamicBuffer();
+            response.writeShort(HEADER_KEY);
             response.writeByte(type);
-            response.writeShort(2); // length
+            response.writeShort(2); // total packet length
             response.writeShort(index);
+
+            if (buf != null && buf.readableBytes() > 0) {
+                response.writeBytes(buf);
+                response.setShort(3, 2 + buf.writerIndex()); // change packet length including buf
```
tananaev: Why do we need to set it twice? Also comments seems unnecessary.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -97,28 +114,72 @@ private String decodeAlarm(Short value) {
     }
 
     private void decodeStatus(Position position, int status) {
+
+        // GPS - 0 bit
```
tananaev: Please remove unnecessary comments EVERYWHERE!

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -97,28 +114,72 @@ private String decodeAlarm(Short value) {
...
+        // log the whole status packet
         position.set(Position.KEY_STATUS, status);
     }
 
-    private Position decodeOld(DeviceSession deviceSession, ChannelBuffer buf, int type, int index) {
+    private Position decodeDownlink(
+            DeviceSession deviceSession, ChannelBuffer buf, Channel channel, Position position, int type, int index) {
+
+        getLastLocation(position, new Date());
+        position.setValid(false);
```
tananaev: Why is it set to not valid? It's incorrect.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -97,28 +114,72 @@ private String decodeAlarm(Short value) {
...
+
+        // Digital input 4 - 15 bit - def 1 high
+        position.set(Position.PREFIX_IN + 4, BitUtil.check(status, 15));
+
+        // log the whole status packet
         position.set(Position.KEY_STATUS, status);
     }
 
-    private Position decodeOld(DeviceSession deviceSession, ChannelBuffer buf, int type, int index) {
+    private Position decodeDownlink(
```
tananaev: I don't think it makes sense to split this into a separate method. Also, looks like you have parameters that you are not using.

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -97,28 +114,72 @@ private String decodeAlarm(Short value) {
...
+            case MSG_SIGN_COMMAND:
+                buf.skipBytes(4); // server flag
 
-        position.set(Position.KEY_INDEX, index);
+                String content = buf.readBytes(buf.readableBytes()).toString(StandardCharsets.UTF_8);
+
+                position.set(Position.KEY_RESULT, content);
+                break;
+            default:
+                position = null;
```
tananaev: Why can't we just return null here?

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -97,28 +114,72 @@ private String decodeAlarm(Short value) {
...
+
+        getLastLocation(position, new Date());
+        position.setValid(false);
 
-        Position position = new Position();
-        position.setDeviceId(deviceSession.getDeviceId());
-        position.setProtocol(getProtocolName());
+        switch (buf.readUnsignedByte()) {
+            case MSG_SIGN_COMMAND:
+                buf.skipBytes(4); // server flag
```
tananaev: Are we saving empty position here?

src/org/traccar/protocol/EelinkProtocolDecoder.java
```diff
@@ -209,41 +318,61 @@ protected Object decode(
...
+        short length = buf.readShort();
+
+        if (buf.readableBytes() != length) { // check remaining length
+            Log.warning(new IllegalArgumentException("Packet length error"));
+            return null;
         }
 
+        int index = buf.readUnsignedShort();
+
+        // process packet data by type
```
tananaev: Remove.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: Sometimes its just an ACK packet so length would be 2bytes, and the if statement is not run, but if there is content we need to change it to be the length of the content + 2 bytes.
Comment is to clarify this point.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: They are necessary for clarification to all people, no point in people having to refer back to specs to know what the provider refer to the parts, you use KEY_IGNITION they refer to ACC.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tananaev: I don't understand why can't you just write full length from the beginning. Something like this:
```java
response.writeShort(2 + buf != null ? content.readableBytes() : 0);
```

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: It is cleaner, downlink does not need to process any GPS/Network data.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tananaev: Here is an example of completely unnecessary comment:
```java
          // Charge - 7,8 bits
          if (BitUtil.check(status, 7)) {
              position.set(Position.KEY_CHARGE, BitUtil.check(status, 8));
          }
```

Actually, it's easier to understand the code than your comment.

We always save ACC as ignition value. We are not going to put the same comment in 100 places to explain it. You can put some comment in position table to clarify it, if you want.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: Downlink data does not come with GPS data, it could be delayed response for some reason and the last location could be incorrect. This was in part why I was asking if there was a way to log a message without Position.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: There is no position data provided on downlink. I have used getLastLocation() before to prevent errors, can I just post an empty Position?

tananaev:
There are some established practices that we use in the code. There are more than 100 different supported protocols. If you don't agree with something, we can discuss it separately and you can fix it in all decoders, but we are not going to make one protocol decoder different from the others.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tananaev: I don't see any reason for storing empty position.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: I could but but would need to do the same when writing out the content too which can be null on ACK packets;
It would still end up with a if statement when writing the content so you dont try and pass null to writeBytes.

tananaev: Checking for null twice is better than writing a variable twice.

src/org/traccar/protocol/EelinkProtocolDecoder.java
tsmgeek: Why write unnecessary comments into the position table that no one needs to refer to unless they look at the decoder when they are trying to work out what is happening.
I am using the words that are in their spec file to clarify. You are nitpicking on the this.

tsmgeek:
Ive just noticed an issue when looking at my tracks.
Im going to implement a check on heartbeat, currently it adds a position into the DB just to store all the status info even if its not changed.

> Position last = Context.getIdentityManager().getLastPosition(position.getDeviceId());

Was thinking to get the whole of the last position, check the new status !=last status int and only if its changed log a new position with updated status, else just ignore as nothing has changed, as you said previously traccar handles device online anyway.

If I do not do this I am ending up with MSG_GPS packet which has GPS/RSSI/ADC params but intermin MSG_HEARTBEAT do not have these, they come in and there is less param data. No point really, its all fluff.

tsmgeek:
Do you have the document for the new eelink protocol?

> Old: SYSTEM:M6100_V1.9.9 VERSION:MXAPP_V2.0.2 BUILD:Oct 19 2016 16:31:00
> New: SYSTEM:M6100_V2.0.2 VERSION:M6100_V2.0.2 BUILD:2016/10/19 17:11

I got another tk116 tracker and the firmware is newer and now seems to follow decoderNew but there are issues such as while it sends old 0x03 heartbeats it also sends 0x15 updates which have no location data but due to the way the decoder was done previously it does not getLastLocation when it puts info into the DB so a location jump back to 0 happens.

Im putting in some fixes.

tananaev:
Not sure if I have it.

tsmgeek:
Think I have now sorted the status and alarm flags.
Last bit is this, I have split the packet into its parts.

`6767 14 0023 02cb - 598f0b17 - 0e - 00ea 000a 56d8 000078b0 0f - 56d800006e0b0a 56d80000870409 - 01 - 206a`

This is an alarm message, on any message that seems to pass cell data this is decoded from `00ea 000a 56d8 000078b0 0f` but there is also this `56d800006e0b0a 56d80000870409` which someone wrote in as BSID1 & 2 but nothing is being done with it.
I am having a stab in the dark that is is decoded as the following based on the first two bytes being the same as the 3rd short in the previous block.
>56d800006e0b0a  ->  Lac:56d8  CellId: 00006e0b RSSI: 0a 
>56d80000870409 -> Lac: 56d8 CellId: 00008704 RSSI: 09

What do you think? 
Ive checked on https://unwiredlabs.com/dashboard?ref=opencellid and seems quite close.

tananaev:
I have found protocol 2.0. Not sure if it's going to help:

[EELINK Device Protocol V2.0 (A4).pdf](https://www.dropbox.com/s/eud7mdiyi5j144w/EELINK%20Device%20Protocol%20V2.0%20%28A4%29.pdf?dl=0)

tsmgeek:
Thanks, it does clarify some bits, yup those unknown blocks are WiFi bssid macs that can be used for location. 
Seems the tracker I am using does not have the WiFi chip though. 

At what point does traccar decide to use geolocation via google/cell based on cell/WiFi points?

tananaev:
It uses it when GPS location is not available.

PR #3421, opened 2017-08-07
Remove "mail.smtp.auth" parameter

### LINKED ISSUE #3417: Email SMTP auth default not set
https://github.com/tananaev/traccar/blob/master/src/org/traccar/notification/NotificationMail.java#L71

Seems that if you have not got mail.smtp.auth set it will throw exception, why not default this to false?

Abyss777:
fix #3417 

We do not need it, because it automatically enabled if username and password provided.


tananaev:
Merged, thanks.

PR #3426, opened 2017-08-08
Implement motion detection similar to trips detection

Abyss777:
Another try...
Implemented delayed motion events without timers, as you suggested.

`updateMotionState` looks complex only because I combined in one update (with new position) and correct (without position).

Positions-Events flow become not linear, because of that we need update events directly and some modifications.

More comments in code...

### DIFF
--- a/src/org/traccar/database/ConnectionManager.java
+++ b/src/org/traccar/database/ConnectionManager.java
@@ -23,8 +23,11 @@ import org.traccar.GlobalTimer;
 import org.traccar.Protocol;
 import org.traccar.helper.Log;
 import org.traccar.model.Device;
+import org.traccar.model.DeviceState;
 import org.traccar.model.Event;
 import org.traccar.model.Position;
+import org.traccar.reports.ReportUtils;
+import org.traccar.reports.model.TripsConfig;
 
 import java.net.SocketAddress;
 import java.sql.SQLException;
@@ -41,6 +44,7 @@ public class ConnectionManager {
 
     private final long deviceTimeout;
     private final boolean enableStatusEvents;
+    private TripsConfig tripsConfig = null;
 
     private final Map<Long, ActiveDevice> activeDevices = new ConcurrentHashMap<>();
     private final Map<Long, Set<UpdateListener>> listeners = new ConcurrentHashMap<>();
@@ -49,6 +53,9 @@ public class ConnectionManager {
     public ConnectionManager() {
         deviceTimeout = Context.getConfig().getLong("status.timeout", DEFAULT_TIMEOUT) * 1000;
         enableStatusEvents = Context.getConfig().getBoolean("event.enable");
+        if (Context.getConfig().getBoolean("status.updateDeviceState")) {
+            tripsConfig = ReportUtils.initTripsConfig();
+        }
     }
 
     public void addActiveDevice(long deviceId, Protocol protocol, Channel channel, SocketAddress remoteAddress) {
@@ -80,21 +87,30 @@ public class ConnectionManager {
 
         if (enableStatusEvents && !status.equals(oldStatus)) {
             String eventType;
+            Event stateEvent = null;
             switch (status) {
                 case Device.STATUS_ONLINE:
                     eventType = Event.TYPE_DEVICE_ONLINE;
                     break;
                 case Device.STATUS_UNKNOWN:
                     eventType = Event.TYPE_DEVICE_UNKNOWN;
+                    if (tripsConfig != null) {
+                        stateEvent = updateDeviceState(deviceId);
+                    }
                     break;
                 default:
                     eventType = Event.TYPE_DEVICE_OFFLINE;
+                    if (tripsConfig != null) {
+                        stateEvent = updateDeviceState(deviceId);
+                    }
                     break;
             }
-            Event event = new Event(eventType, deviceId);
-            if (Context.getNotificationManager() != null) {
-                Context.getNotificationManager().updateEvent(event, null);
+            Set<Event> events = new HashSet<>();
+            if (stateEvent != null) {
+                events.add(stateEvent);
             }
+            events.add(new Event(eventType, deviceId));
+            Context.getNotificationManager().updateEvents(events, null);
         }
 
         Timeout timeout = timeouts.remove(deviceId);
@@ -124,6 +140,29 @@ public class ConnectionManager {
         }
 
         updateDevice(device);
+    }
+
+    public Event updateDeviceState(long deviceId) {
+        DeviceState deviceState = Context.getDeviceManager().getDeviceState(deviceId);
+        if (deviceState == null || deviceState.getMotionState() == null) {
+            return null;
+        }
+        Event result = null;
+        Boolean oldMotion = deviceState.getMotionState();
+        long currentTime = new Date().getTime();
+        boolean newMotion = !oldMotion;
+        Position potentialPosition = deviceState.getMotionPosition();
+        if (potentialPosition != null) {
+            long potentialTime = potentialPosition.getFixTime().getTime()
+                    + (newMotion ? tripsConfig.getMinimalTripDuration() : tripsConfig.getMinimalParkingDuration());
+            if (potentialTime <= currentTime) {
+                String eventType = newMotion ? Event.TYPE_DEVICE_MOVING : Event.TYPE_DEVICE_STOPPED;
+                result = new Event(eventType, potentialPosition.getDeviceId(), potentialPosition.getId());
+                deviceState.setMotionState(newMotion);
+                deviceState.setMotionPosition(null);
+            }
+        }
+        return result;
     }
 
     public synchronized void updateDevice(Device device) {
--- a/src/org/traccar/database/ConnectionManager.java
+++ b/src/org/traccar/database/ConnectionManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 - 2016 Anton Tananaev (anton@traccar.org)
+ * Copyright 2015 - 2017 Anton Tananaev (anton@traccar.org)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -149,15 +149,15 @@ public class ConnectionManager {
         }
         Event result = null;
         Boolean oldMotion = deviceState.getMotionState();
-        long currentTime = new Date().getTime();
+        long currentTime = System.currentTimeMillis();
         boolean newMotion = !oldMotion;
-        Position potentialPosition = deviceState.getMotionPosition();
-        if (potentialPosition != null) {
-            long potentialTime = potentialPosition.getFixTime().getTime()
+        Position motionPosition = deviceState.getMotionPosition();
+        if (motionPosition != null) {
+            long motionTime = motionPosition.getFixTime().getTime()
                     + (newMotion ? tripsConfig.getMinimalTripDuration() : tripsConfig.getMinimalParkingDuration());
-            if (potentialTime <= currentTime) {
+            if (motionTime <= currentTime) {
                 String eventType = newMotion ? Event.TYPE_DEVICE_MOVING : Event.TYPE_DEVICE_STOPPED;
-                result = new Event(eventType, potentialPosition.getDeviceId(), potentialPosition.getId());
+                result = new Event(eventType, motionPosition.getDeviceId(), motionPosition.getId());
                 deviceState.setMotionState(newMotion);
                 deviceState.setMotionPosition(null);
             }
--- a/src/org/traccar/database/DeviceManager.java
+++ b/src/org/traccar/database/DeviceManager.java
@@ -33,6 +33,7 @@ import org.traccar.helper.Log;
 import org.traccar.model.Command;
 import org.traccar.model.CommandType;
 import org.traccar.model.Device;
+import org.traccar.model.DeviceState;
 import org.traccar.model.DeviceTotalDistance;
 import org.traccar.model.Group;
 import org.traccar.model.Position;
@@ -51,6 +52,8 @@ public class DeviceManager extends BaseObjectManager<Device> implements Identity
     private AtomicLong devicesLastUpdate = new AtomicLong();
 
     private final Map<Long, Position> positions = new ConcurrentHashMap<>();
+
+    private final Map<Long, DeviceState> deviceStates = new ConcurrentHashMap<>();
 
     private boolean fallbackToText;
 
@@ -387,4 +390,13 @@ public class DeviceManager extends BaseObjectManager<Device> implements Identity
         }
         return result;
     }
+
+    public DeviceState getDeviceState(long deviceId) {
+        return deviceStates.get(deviceId);
+    }
+
+    public void setDeviceState(long deviceId, DeviceState deviceState) {
+        deviceStates.put(deviceId, deviceState);
+    }
+
 }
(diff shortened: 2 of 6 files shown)


src/org/traccar/database/ConnectionManager.java
```diff
@@ -86,15 +88,18 @@ public void updateDevice(final long deviceId, String status, Date time) {
...
                     break;
                 default:
                     eventType = Event.TYPE_DEVICE_OFFLINE;
+                    if (correctMotionState) {
+                        Context.getDeviceManager().updateMotionState(deviceId, null);
+                    }
                     break;
             }
-            Event event = new Event(eventType, deviceId);
-            if (Context.getNotificationManager() != null) {
```
Abyss777: Do not need this check because switch is common for every events

src/org/traccar/database/ConnectionManager.java
```diff
@@ -140,10 +145,14 @@ public synchronized void updatePosition(Position position) {
...
         for (long userId : Context.getPermissionsManager().getDeviceUsers(deviceId)) {
-            if (listeners.containsKey(userId)) {
-                for (UpdateListener listener : listeners.get(userId)) {
-                    listener.onUpdatePosition(position);
-                }
+            updatePosition(userId, position);
+        }
+    }
+
+    public synchronized void updatePosition(long userId, Position position) {
```
Abyss777: Implemented function to update position for single user

src/org/traccar/database/DeviceManager.java
```diff
@@ -66,6 +76,11 @@ public DeviceManager(DataManager dataManager) {
         dataRefreshDelay = config.getLong("database.refreshDelay", DEFAULT_REFRESH_DELAY) * 1000;
         lookupGroupsAttribute = config.getBoolean("deviceManager.lookupGroupsAttribute");
         fallbackToText = config.getBoolean("command.fallbackToSms");
+
+        minimalParkingDuration = config.getLong("report.trip.minimalParkingDuration", 300) * 1000;
```
Abyss777: Constants initialized with same values as for trips/stops

src/org/traccar/database/DeviceManager.java
```diff
@@ -387,4 +402,54 @@ public void sendCommand(Command command) throws Exception {
...
+            Position potentialPosition = motionStatePositions.get(deviceId);
+            if (potentialPosition != null) {
+                long potentialTime = potentialPosition.getFixTime().getTime();
+                Double distance = null;
+                if (position != null) {
+                    distance = ReportUtils.calculateDistance(potentialPosition, position, false);
+                }
+                if (newMotion) {
+                    if (potentialTime + minimalTripDuration <= currentTime
+                            || distance == null || distance >= minimalTripDistance) {
```
Abyss777: Detect motion by time and by distance

src/org/traccar/database/NotificationManager.java
```diff
@@ -62,6 +66,9 @@ public void updateEvent(Event event, Position position) {
                 if (notification != null) {
                     if (notification.getWeb()) {
                         Context.getConnectionManager().updateEvent(userId, event);
+                        if (withPosition) {
+                            Context.getConnectionManager().updatePosition(userId, position);
```
Abyss777: Because related position from past we need resent it after event to correctly link in web-client in "Events panel"

src/org/traccar/events/MotionEventHandler.java
```diff
@@ -37,19 +36,7 @@
...
-            oldMotion = lastPosition.getBoolean(Position.KEY_MOTION);
-        }
-        if (motion && !oldMotion) {
-            return Collections.singleton(
-                    new Event(Event.TYPE_DEVICE_MOVING, position.getDeviceId(), position.getId()));
-        } else if (!motion && oldMotion) {
-            return Collections.singleton(
-                    new Event(Event.TYPE_DEVICE_STOPPED, position.getDeviceId(), position.getId()));
-        }
+        Context.getDeviceManager().updateMotionState(position.getDeviceId(), position);
```
Abyss777: Return nothing because event never be related to current position.

src/org/traccar/database/ConnectionManager.java
```diff
@@ -41,6 +41,7 @@
 
     private final long deviceTimeout;
     private final boolean enableStatusEvents;
+    private final boolean correctMotionState;
```
tananaev: What does this parameter do? I think we need a better name because it's not very clear.

src/org/traccar/database/DeviceManager.java
```diff
@@ -52,6 +55,13 @@
 
     private final Map<Long, Position> positions = new ConcurrentHashMap<>();
 
+    private final Map<Long, Boolean> motionStates = new ConcurrentHashMap<>();
+    private final Map<Long, Position> motionStatePositions = new ConcurrentHashMap<>();
```
tananaev: Don't we already store positions above? or is it a first position? In that case do we really need to store the whole position or just the time would be enough?

src/org/traccar/database/DeviceManager.java
```diff
@@ -52,6 +55,13 @@
 
     private final Map<Long, Position> positions = new ConcurrentHashMap<>();
 
+    private final Map<Long, Boolean> motionStates = new ConcurrentHashMap<>();
```
tananaev: Do you think it's a good idea to create some "device state" class to store last position, motion status, and possibly some other information in future?

src/org/traccar/database/DeviceManager.java
```diff
@@ -387,4 +402,54 @@ public void sendCommand(Command command) throws Exception {
         }
         return result;
     }
+
+    public void updateMotionState(long deviceId, Position position) {
```
tananaev: It would be nice to extract logic back into motionhandler to avoid adding to much code device manager.

src/org/traccar/database/NotificationManager.java
```diff
@@ -48,6 +48,10 @@ public NotificationManager(DataManager dataManager) {
     }
 
     public void updateEvent(Event event, Position position) {
+        updateEvent(event, position, false);
```
tananaev: Is there any time when position is not null and we don't want to update it? Possibly not for this pull request, but it would help with handling events in the web app if event and position come in the same message.

src/org/traccar/database/NotificationManager.java
tananaev: Please elaborate. It doesn't seem right to send position twice. Am I understanding correctly that it's what you are doing?

src/org/traccar/events/MotionEventHandler.java
tananaev: I think we should move motion detection logic back. See comment above.

test/org/traccar/events/MotionEventHandlerTest.java
```diff
@@ -1,29 +0,0 @@
...
-import static org.junit.Assert.assertNotNull;
-
-import java.util.Collection;
-
-import org.junit.Test;
-import org.traccar.BaseTest;
-import org.traccar.model.Event;
-import org.traccar.model.Position;
-
-public class MotionEventHandlerTest extends BaseTest {
```
tananaev: It would be nice to implement some unit testing for the new logic.

src/org/traccar/database/ConnectionManager.java
Abyss777: It enables additional check motion state on offline/unknown status changes.

src/org/traccar/database/DeviceManager.java
Abyss777: Not sure, may be it will be more convenient.

src/org/traccar/database/DeviceManager.java
Abyss777: We need also distance for detect moving, link `positionId` to new event and pass appropriate `position` (whole) to notification.

src/org/traccar/database/DeviceManager.java
Abyss777: We can't `MotionHandler` are multiple for every pipeline, but we need some singleton to store per device motion states. Or am I mistaken?

src/org/traccar/database/NotificationManager.java
Abyss777: Positions come through pipeline and go to web-client at the end.

Abyss777:
I'm not sure how to implement such events without resend position. For "Events panel" it should come after Event.

Abyss777:
Drew flow scheme to represent it visually.
flows

I feel we need implement something essential (base) for such Events. I'm sure we want make some other events similar.

Abyss777:
One option is We might not resend past position but request it via API if user click it on Events Panel

src/org/traccar/database/ConnectionManager.java
tananaev: Can't think of any good name at the moment, we definitely need something better here.

src/org/traccar/database/DeviceManager.java
tananaev: It is faster than queries each parameter separate and it will enable us to easily add some extra information later.

src/org/traccar/database/DeviceManager.java
tananaev: Makes sense. Ignore my comment.

src/org/traccar/database/DeviceManager.java
tananaev: Yes, my idea was that you store any required state in device manager, but actual calculation code is here. This way we separate the code to make it easier to read.

tananaev:
Requesting position through API is also not ideal solution. We have the position. Why would we need to duplicate effort.

I think maybe we should modify pipeline to push events up the chain to the main event handler. This way we can keep position and event together. This will require quite a bit of refactoring, but seems like the right solution. It will also unify the way events and positions are pushed to the web app.

Abyss777:
But still, do not understand how it help. Event still be related to position from past (user can login in period between this position and related event). Also we have some event types not related to positions and still need a way to generate and transport events directly (without pipeline)
Do not think splitting a ways how we deliver events is good idea (through pipeline or directly)

tsmgeek:
Is this for devices that do not have movement detection built in or are you defining movement as a positional change?

tananaev:
I see the problem now. It's an old position, not current one. I missed that point. Then using API might be a good idea.

src/org/traccar/model/DeviceState.java
```diff
@@ -0,0 +1,41 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.model;
+
+public class DeviceState {
+
+    private Boolean motionState;
```
Abyss777: It is not primitive intentionally. To determine if it is not initialized.

src/org/traccar/database/ConnectionManager.java
```diff
@@ -49,6 +53,9 @@
     public ConnectionManager() {
         deviceTimeout = Context.getConfig().getLong("status.timeout", DEFAULT_TIMEOUT) * 1000;
         enableStatusEvents = Context.getConfig().getBoolean("event.enable");
+        if (Context.getConfig().getBoolean("status.updateDeviceState")) {
```
Abyss777: Is the name OK?

Abyss777:
Replaced the commit.
Implemented `DeviceState` class, in future we can use it for overspeeding state or other events.
Events now return back in flow, not send directly.
Implemented test.

I decided not over complicate `MotionEventHandler.updateMotionState` to not use it if we handle offline/unknown status. Just implemented simple variant in `ConnectionManager`

Also fixed issue in `ReportUtils`, parameters were initialized in wrong order.

src/org/traccar/model/DeviceState.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.model;
+
+public class DeviceState {
```
tananaev: Do you think it makes sense to add last position here as well?

Abyss777: Yes I think it is reasonable. Concentrate it in one place.
Also I was thinking that it is good place for device Status, but remember that we notify web client about it.

src/org/traccar/database/ConnectionManager.java
tananaev: I think it's OK. I'm assuming that the idea is that state will have more information than just motion state.

src/org/traccar/database/ConnectionManager.java
```diff
@@ -126,6 +142,29 @@ public void run(Timeout timeout) throws Exception {
...
     }
 
+    public Event updateDeviceState(long deviceId) {
+        DeviceState deviceState = Context.getDeviceManager().getDeviceState(deviceId);
+        if (deviceState == null || deviceState.getMotionState() == null) {
+            return null;
+        }
+        Event result = null;
+        Boolean oldMotion = deviceState.getMotionState();
+        long currentTime = new Date().getTime();
```
tananaev: Use `System.currentTimeMillis()` instead.

src/org/traccar/events/MotionEventHandler.java
```diff
@@ -21,11 +21,60 @@
...
 import org.traccar.model.Position;
+import org.traccar.reports.ReportUtils;
+import org.traccar.reports.model.TripsConfig;
 
 public class MotionEventHandler extends BaseEventHandler {
 
+    private TripsConfig tripsConfig;
+
+    public MotionEventHandler() {
+        if (Context.getConfig() != null) {
```
tananaev: Is there any case when we don't have config?

src/org/traccar/database/ConnectionManager.java
```diff
@@ -126,6 +142,29 @@ public void run(Timeout timeout) throws Exception {
...
+        boolean newMotion = !oldMotion;
+        Position potentialPosition = deviceState.getMotionPosition();
```
tananaev: I think we should just name this `motionPosition`.

src/org/traccar/database/ConnectionManager.java
```diff
@@ -126,6 +142,29 @@ public void run(Timeout timeout) throws Exception {
...
+        if (potentialPosition != null) {
+            long potentialTime = potentialPosition.getFixTime().getTime()
```
tananaev: I don't like `potential` prefix. Maybe `thresholdTime`?

src/org/traccar/model/DeviceState.java
tananaev: I guess we can leave it as it is for now...

Abyss777:
Renamed variables by next logic `potentialPosition` to `motionPosition`, `potentialTime` to `motionPositionTime`, but it is too long, leaved just `motionTime`

src/org/traccar/events/MotionEventHandler.java
Abyss777: Forgot to remove this check, when make next function `static` for tests.

Abyss777:
Sorry, occasionally closed.

tananaev:
Merged, thanks.

Abyss777:
I applied this and next changes to my server at Friday and tested them on real data during weekend.
It works correctly with some parameters tuning, but wants a bit more.
We have second parameter to detect motion - distance, but we detect stop only by time. I believe we should add second parameter to detect stop.
I'm suggesting - Ignition (of course if position contains it). If ignition is ON - it does not mean that car is moving, but usually car can't move with stopped engine. What do you think?

tananaev:
Makes sense to me.

PR #3431, opened 2017-08-10
Implement delayed overspeed events

Abyss777:
Implementation is simpler than motion because there is no "not-overspeed" event.
Logic stayed the same, but with delay. For example:
`minimalDuration = 15 seconds`, but device "overspeeding"  for a minute, if `notRepeat = false`  about 4 events will be generated (if reported with enough frequency) if `notRepeat = true` - only one event.

Also made a few small optimizations with previous code.
- Initialize `DeviceState` in get method
- Unwrapped some not obvious nested checks

### DIFF
--- a/src/org/traccar/database/ConnectionManager.java
+++ b/src/org/traccar/database/ConnectionManager.java
@@ -21,15 +21,12 @@ import org.jboss.netty.util.TimerTask;
 import org.traccar.Context;
 import org.traccar.GlobalTimer;
 import org.traccar.Protocol;
-import org.traccar.events.MotionEventHandler;
 import org.traccar.events.OverspeedEventHandler;
 import org.traccar.helper.Log;
 import org.traccar.model.Device;
 import org.traccar.model.DeviceState;
 import org.traccar.model.Event;
 import org.traccar.model.Position;
-import org.traccar.reports.ReportUtils;
-import org.traccar.reports.model.TripsConfig;
 
 import java.net.SocketAddress;
 import java.sql.SQLException;
@@ -47,9 +44,6 @@ public class ConnectionManager {
     private final long deviceTimeout;
     private final boolean enableStatusEvents;
     private final boolean updateDeviceState;
-    private TripsConfig tripsConfig = null;
-    private long minimalOverspeedDuration;
-    private boolean overspeedNotRepeat;
 
     private final Map<Long, ActiveDevice> activeDevices = new ConcurrentHashMap<>();
     private final Map<Long, Set<UpdateListener>> listeners = new ConcurrentHashMap<>();
@@ -59,11 +53,6 @@ public class ConnectionManager {
         deviceTimeout = Context.getConfig().getLong("status.timeout", DEFAULT_TIMEOUT) * 1000;
         enableStatusEvents = Context.getConfig().getBoolean("event.enable");
         updateDeviceState = Context.getConfig().getBoolean("status.updateDeviceState");
-        if (updateDeviceState) {
-            tripsConfig = ReportUtils.initTripsConfig();
-            minimalOverspeedDuration = Context.getConfig().getLong("event.overspeed.minimalDuration") * 1000;
-            overspeedNotRepeat = Context.getConfig().getBoolean("event.overspeed.notRepeat");
-        }
     }
 
     public void addActiveDevice(long deviceId, Protocol protocol, Channel channel, SocketAddress remoteAddress) {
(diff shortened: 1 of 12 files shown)


tananaev:
Do we really need this for overspeed? Are there any complaints about current implementation?

Abyss777:
Basically it is protection from unwanted events during short overtaking another car.

I have found only one https://www.traccar.org/forums/topic/speed-limits-configuration/#post-10686
but I believe there was others. Also count me who voted for this feature :)

tananaev:
OK, but how does it work by default? Will it work same as before or does user need to add some extra configuration?

Abyss777:
By default it works absolutely same as before.(ran code is longer of course, but result is the same) If `minimalDuration` is 0 or omitted, events are generated instantly.

tananaev:
Everything looks good except for one thing. I don't like that we have part of the logic in `ConnectionManager` class. Is there a way we can move it to corresponding handler? Possibly as a simple static method? Same applies for motion as well.

Abyss777:
Frankly, I also thought about it. I'll move it.

Abyss777:
Done

src/org/traccar/events/OverspeedEventHandler.java
```diff
@@ -29,40 +30,95 @@
...
+        } else if (oldOverspeed && !newOverspeed) {
+            deviceState.setOverspeedState(false);
+            deviceState.setOverspeedPosition(null);
+        } else {
+            deviceState.setOverspeedPosition(null);
+        }
+        Position overspeedPosition = deviceState.getOverspeedPosition();
+        if (overspeedPosition != null) {
+            long overspeedTime = overspeedPosition.getFixTime().getTime();
+            if (newOverspeed && overspeedTime + minimalOverspeedDuration <= currentTime) {
```
tananaev: Now it seems like there is some duplication between two methods. Is it possible to share the code? Does it make sense?

src/org/traccar/events/MotionEventHandler.java
```diff
@@ -36,6 +36,24 @@ public MotionEventHandler() {
...
 
+    public static Event updateMotionState(DeviceState deviceState, TripsConfig tripsConfig) {
+        Event result = null;
+        if (deviceState.getMotionState() != null && deviceState.getMotionPosition() != null) {
+            boolean newMotion = !deviceState.getMotionState();
+            Position motionPosition = deviceState.getMotionPosition();
+            long currentTime = System.currentTimeMillis();
+            long motionTime = motionPosition.getFixTime().getTime()
+                    + (newMotion ? tripsConfig.getMinimalTripDuration() : tripsConfig.getMinimalParkingDuration());
+            if (motionTime <= currentTime) {
```
tananaev: Same question here.

src/org/traccar/events/OverspeedEventHandler.java
Abyss777: It looks like, but one method become looks monstrous.

tananaev: It doesn't have to be one. Shared code can be split into a separate method.

src/org/traccar/database/ConnectionManager.java
```diff
@@ -142,26 +146,22 @@ public void run(Timeout timeout) throws Exception {
...
-                    + (newMotion ? tripsConfig.getMinimalTripDuration() : tripsConfig.getMinimalParkingDuration());
-            if (motionTime <= currentTime) {
-                String eventType = newMotion ? Event.TYPE_DEVICE_MOVING : Event.TYPE_DEVICE_STOPPED;
-                result = new Event(eventType, motionPosition.getDeviceId(), motionPosition.getId());
-                deviceState.setMotionState(newMotion);
-                deviceState.setMotionPosition(null);
-            }
+
+        event = OverspeedEventHandler.updateOverspeedState(deviceState, Context.getDeviceManager().
+                lookupAttributeDouble(deviceId, OverspeedEventHandler.ATTRIBUTE_SPEED_LIMIT, 0, false),
```
tananaev: Another thing I noticed here is that there is a mess with attributes. They are all over the place. Some are here, and some are in the event handler. Maybe we should just make `OverspeedEventHandler` a singleton and do everything there, instead of having duplication everywhere? That's just one solution. I think we need to take time and make it right.

tananaev: Kind of a similar story with `TripsConfig`. The more worrying is the fact that we solve same problems in slightly different ways. Focus should be on code quality, not a number of features.

Abyss777: What if we make `updateXxx` functions part of `DeviceState` class?
Agree, initializing parameters looks mess, because it is shared, may be we place them in `Context` ?

tananaev: Why can't we just make `OverspeedEventHandler` a singleton in `Context`? This way all the logic and all parameters will be in a single place.

Abyss777:
I've made changes.
There are only 2 parameters related to overspeed, do not think we need separate class like `TripsConfig`

tananaev:
Merged, looks good now.

PR #3443, opened 2017-08-14
Correct stop state by ignition.

Abyss777:
Here is the implementation of additional stop detection by ignition.

In case devices always report `ignition = false` made this function switchable.

### DIFF
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -257,7 +257,7 @@ public final class Context {
                 config.getLong("report.trip.minimalParkingDuration", 300) * 1000,
                 config.getBoolean("report.trip.greedyParking"),
                 config.getLong("report.trip.minimalNoDataDuration", 3600) * 1000,
-                config.getBoolean("report.trip.stopOnIgnitionOff"));
+                config.getBoolean("report.trip.useIgnition"));
     }
 
     public static void init(String[] arguments) throws Exception {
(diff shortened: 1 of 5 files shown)


src/org/traccar/Context.java
```diff
@@ -256,7 +256,8 @@ public static TripsConfig initTripsConfig() {
                 config.getLong("report.trip.minimalTripDuration", 300) * 1000,
                 config.getLong("report.trip.minimalParkingDuration", 300) * 1000,
                 config.getBoolean("report.trip.greedyParking"),
-                config.getLong("report.trip.minimalNoDataDuration", 3600) * 1000);
+                config.getLong("report.trip.minimalNoDataDuration", 3600) * 1000,
+                config.getBoolean("report.trip.stopOnIgnitionOff"));
```
Abyss777: Parameter name is controversial.

tsmgeek:
Does stop detection also take note of the movement flag that some devices can report?

src/org/traccar/Context.java
tananaev: Can we just name it something like `report.trip.useIgnition`?

Abyss777:
Yes, it uses `KEY_MOTION` attribute that reported by device or calculated automatically by speed if not.

Abyss777:
Renamed parameter.

tananaev:
Merged, thanks.

PR #3444, opened 2017-08-14
Update to EElink Decoder/Encoder

tsmgeek:
Supersedes #3420 due to correction of branches


tananaev:
Please let me know when all the comments from the previous pull request are fixed, so I can review it.

tsmgeek:
Remaining comments I am leaving as they are helpful.

tananaev:
You are free to leave it of course, but in that case unfortunately it's not going to be merged. Let me know if you decide to change your mind.

tsmgeek:
Why do you have such a big problem with comments?
It seems a form of protectionism to keep things obscure.

tananaev:
I don't like unnecessary code and duplication. It always results in inconsistency and maintenance headache in the long term.

tsmgeek:
That is an excuse, never heard anyone saying something is a maintenance headache when there are some comments, bad code yes. I understand if someone did a massive commit with silly things like `woops why are we here` etc but this is not the case. Its a protocol decoder not voodoo.

PR #3445, opened 2017-08-14
Trim email field on auth

tsmgeek:
Trim email field on user create/update.
On login trim input and force to lowercase as is standard with most sites.

### DIFF
--- a/src/org/traccar/database/DataManager.java
+++ b/src/org/traccar/database/DataManager.java
@@ -297,7 +297,7 @@ public class DataManager {
 
     public User login(String email, String password) throws SQLException {
         User user = QueryBuilder.create(dataSource, getQuery("database.loginUser"))
-                .setString("email", email.trim().toLowerCase())
+                .setString("email", email.trim())
                 .executeQuerySingle(User.class);
         if (user != null && user.isPasswordValid(password)) {
             return user;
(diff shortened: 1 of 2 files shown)


src/org/traccar/database/DataManager.java
```diff
@@ -297,7 +297,7 @@ private void initDatabaseSchema() throws SQLException, LiquibaseException {
 
     public User login(String email, String password) throws SQLException {
         User user = QueryBuilder.create(dataSource, getQuery("database.loginUser"))
-                .setString("email", email)
+                .setString("email", email.trim().toLowerCase())
```
tananaev: I agree with trimming, but I don't understand why we need to lower-case it.

tsmgeek: 1. database would allow not allow two users to be created with same address in different cases as it is using `utf8_general_ci` collation.
2. many users mess up with capslock on/off or capping certain letters in their email
3. this is standard practice with most sites to prevent user errors due to 2

tananaev: Sorry, but neither of your points answer my question. Field is case-insensitive in the database, so why do we care what user enters and why do we need to lower-case it?

tsmgeek: Query cache optimisation. If two queries came in for the same row it would always look the same to the DB and therefor cached.

tananaev:
Merged, thanks.

PR #3447, opened 2017-08-15
Reimplement trips detector using MotionEventHandler

Abyss777:
I hope it is final changes related to motion detection. Your decision about way of motion detection was more than wise. We can apply it to Trips and Stops detection.
Old monstrous, unreadable function was transformed in two linear and much more clear (I hope so)

Key difference from previous algorithm:
- removed `greedyParking` parameter, it is not needed because `MotionEventHandler` is "greedy both" - only continuous trips or stops can switch the state.
- `DeviceState` is initialized by first position in report, therefore initial trips or stops also detected (even if it is shorter than needed)
- To switch the state we need at least two positions, therefore we are assuming that previous position before gap and next after the gap is stopped position. It does not influence on the result, just notice.
- To not rewrite `MotionEventHandler` I just replace `KEY_MOTION` attribute in retrieved from DB positions.

Two functions look similar but inverse, I think combining them in one is not good, function become monstrous again, also we will avoid unchecked type casting.


src/org/traccar/reports/ReportUtils.java
```diff
@@ -237,8 +239,11 @@ private static StopReport calculateStop(ArrayList<Position> positions, int start
 
     private static boolean isMoving(ArrayList<Position> positions, int index,
             TripsConfig tripsConfig, double speedThreshold) {
-        if (tripsConfig.getMinimalNoDataDuration() > 0 && index < positions.size() - 1
+        if (tripsConfig.getMinimalNoDataDuration() > 0 && (index < positions.size() - 1
                 && positions.get(index + 1).getFixTime().getTime() - positions.get(index).getFixTime().getTime()
+                >= tripsConfig.getMinimalNoDataDuration())
+                || index > 0
+                && positions.get(index).getFixTime().getTime() - positions.get(index - 1).getFixTime().getTime()
```
Abyss777: Next after the gap also mark as stopped to make possible switch the state.

src/org/traccar/reports/ReportUtils.java
```diff
@@ -250,100 +255,79 @@ private static boolean isMoving(ArrayList<Position> positions, int index,
...
+    public static Collection<TripReport> detectTrips(Collection<Position> positionCollection, TripsConfig tripsConfig,
+            boolean ignoreOdometer, double speedThreshold) {
+        Collection<TripReport> result = new ArrayList<>();
 
         ArrayList<Position> positions = new ArrayList<>(positionCollection);
         if (positions != null && !positions.isEmpty()) {
-            int previousStartParkingIndex = 0;
+            MotionEventHandler  motionHandler = new MotionEventHandler(tripsConfig);
+            DeviceState deviceState = new DeviceState();
+            deviceState.setMotionState(isMoving(positions, 0, tripsConfig, speedThreshold));
```
Abyss777: Correct motion attribute according trips detector rules (for gaps and backward compatibility)

src/org/traccar/reports/ReportUtils.java
```diff
@@ -250,100 +255,79 @@ private static boolean isMoving(ArrayList<Position> positions, int index,
...
             for (int i = 0; i < positions.size(); i++) {
-                isMoving = isMoving(positions, i, tripsConfig, speedThreshold);
-                isLast = i == positions.size() - 1;
-
-                if ((isMoving || isLast) && startParkingIndex != -1) {
-                    if (!skipped || previousEndParkingIndex == 0) {
-                        previousEndParkingIndex = endParkingIndex;
-                    }
-                    endParkingIndex = i;
+                positions.get(i).set(Position.KEY_MOTION, isMoving(positions, i, tripsConfig, speedThreshold));
```
Abyss777: Correct motion attribute according trips detector rules (for gaps and backward compatibility)

src/org/traccar/reports/ReportUtils.java
Abyss777: Initialize `deviceState` by first position

src/org/traccar/reports/ReportUtils.java
```diff
@@ -250,100 +255,79 @@ private static boolean isMoving(ArrayList<Position> positions, int index,
...
-                        skipped = false;
-                    } else {
-                        skipped = true;
+                if (startTripIndex != -1 && startParkingIndex != -1 && event != null && !deviceState.getMotionState()) {
+                    result.add(calculateTrip(positions, startTripIndex, startParkingIndex, ignoreOdometer));
+                    startTripIndex = -1;
+                }
+            }
+            if (startTripIndex != -1 && startParkingIndex != -1) {
+                result.add(calculateTrip(positions, startTripIndex, startParkingIndex, ignoreOdometer));
```
Abyss777: If period ended too short parking still add detected trip

src/org/traccar/reports/ReportUtils.java
```diff
@@ -250,100 +255,79 @@ private static boolean isMoving(ArrayList<Position> positions, int index,
...
-                int end = isMoving && !tripsConfig.getGreedyParking()
-                        ? Math.max(endParkingIndex, previousEndParkingIndex) : positions.size() - 1;
-                long parkingDuration = positions.get(end).getFixTime().getTime()
-                        - positions.get(previousStartParkingIndex).getFixTime().getTime();
-                if (parkingDuration >= tripsConfig.getMinimalParkingDuration()) {
-                    result.add(calculateStop(positions, previousStartParkingIndex, end));
-                }
+            if (startParkingIndex != -1) {
+                result.add(calculateStop(positions, startParkingIndex,
+                        startTripIndex != -1 ? startTripIndex : positions.size() - 1));
```
Abyss777: If we detected too short trip add parking, or assume parking till the end of period.

test/org/traccar/reports/ReportUtilsTest.java
```diff
@@ -95,12 +95,20 @@ public void testDetectTripsSimple() throws ParseException {
...
-        assertFalse(result.isEmpty());
+        assertNotNull(stops);
+        assertFalse(stops.isEmpty());
 
-        StopReport itemStop = (StopReport) result.iterator().next();
+        Iterator<StopReport> iterator = stops.iterator();
+
+        StopReport itemStop = iterator.next();
+
+        assertEquals(date("2016-01-01 00:00:00.000"), itemStop.getStartTime());
```
Abyss777: It is the first difference in behavior: initial parking/trip will detected even if it is too short.

test/org/traccar/reports/ReportUtilsTest.java
```diff
@@ -184,18 +179,18 @@ public void testDetectStopsStartedFromTrip() throws ParseException {
...
+        Collection<StopReport> result = ReportUtils.detectStops(data, tripsConfig, false, 0.01);
 
         assertNotNull(result);
         assertFalse(result.isEmpty());
 
-        StopReport itemStop = (StopReport) result.iterator().next();
+        StopReport itemStop = result.iterator().next();
 
-        assertEquals(date("2016-01-01 00:00:00.000"), itemStop.getStartTime());
+        assertEquals(date("2016-01-01 00:02:00.000"), itemStop.getStartTime());
```
Abyss777: The second difference: If period started from moving it will not be added to detected parking (more logically I think)

src/org/traccar/reports/ReportUtils.java
tananaev: Can you please split parts of this expression into variables. It's very hard to understand what's going on.

src/org/traccar/reports/ReportUtils.java
tananaev: Maybe we can check if the attribute is already there and only calculate it if it's not?

src/org/traccar/reports/ReportUtils.java
```diff
@@ -250,100 +255,79 @@ private static boolean isMoving(ArrayList<Position> positions, int index,
...
+
+        ArrayList<Position> positions = new ArrayList<>(positionCollection);
+        if (positions != null && !positions.isEmpty()) {
+            MotionEventHandler  motionHandler = new MotionEventHandler(tripsConfig);
+            DeviceState deviceState = new DeviceState();
+            deviceState.setMotionState(isMoving(positions, 0, tripsConfig, speedThreshold));
+            int startTripIndex = -1;
+            int startParkingIndex = deviceState.getMotionState() ? -1 : 0;
+            for (int i = 0; i < positions.size(); i++) {
+                positions.get(i).set(Position.KEY_MOTION, isMoving(positions, i, tripsConfig, speedThreshold));
```
tananaev: Same question here.

src/org/traccar/reports/ReportUtils.java
```diff
@@ -250,100 +255,79 @@ private static boolean isMoving(ArrayList<Position> positions, int index,
...
+                }
+            }
+            if (startTripIndex != -1 && startParkingIndex != -1) {
+                result.add(calculateTrip(positions, startTripIndex, startParkingIndex, ignoreOdometer));
+            }
+        }
+        return result;
+    }
+
+    public static Collection<StopReport> detectStops(Collection<Position> positionCollection, TripsConfig tripsConfig,
```
tananaev: Detect trips and detect stops logic is very similar, but basically inverse. Is there a way we can combine two?

src/org/traccar/reports/ReportUtils.java
Abyss777: Check for gap has priority, because motion might be always true, but we still need override the value.

tananaev: OK, but then I think it's better to use some variable to store value instead of updating positions.

Abyss777: `MotionEventHandler` switch `deviceState` based on `KEY_MOTION` we need change its behavior then.

tananaev: The point is that we shouldn't alter positions after they are stored. There must be a better solution.

Abyss777: Got it. Will try find something.

Abyss777:
- Combined methods in one with help of generics.
- Made `MotionEventHandler` accept external `newMotion`
- Split gap detection on variables

src/org/traccar/reports/ReportUtils.java
```diff
@@ -237,10 +240,16 @@ private static StopReport calculateStop(ArrayList<Position> positions, int start
 
     private static boolean isMoving(ArrayList<Position> positions, int index,
             TripsConfig tripsConfig, double speedThreshold) {
-        if (tripsConfig.getMinimalNoDataDuration() > 0 && index < positions.size() - 1
-                && positions.get(index + 1).getFixTime().getTime() - positions.get(index).getFixTime().getTime()
-                >= tripsConfig.getMinimalNoDataDuration()) {
-            return false;
+        if (tripsConfig.getMinimalNoDataDuration() > 0) {
+            boolean beforeGap = index < positions.size() - 1
+                    && positions.get(index + 1).getFixTime().getTime() - positions.get(index).getFixTime().getTime()
```
tananaev: I think it can be split even more. Maybe extract time difference?

src/org/traccar/reports/ReportUtils.java
```diff
@@ -250,100 +259,65 @@ private static boolean isMoving(ArrayList<Position> positions, int index,
...
+                    } else if (!trips && deviceState.getMotionState()) {
+                        result.add((T) calculateStop(positions, startParkingIndex, startTripIndex));
+                        startParkingIndex = -1;
+                    }
                 }
             }
+            if (trips && startTripIndex != -1 && startParkingIndex != -1) {
+                result.add((T) calculateTrip(positions, startTripIndex, startParkingIndex, ignoreOdometer));
+            }
+            if (!trips && startParkingIndex != -1) {
```
tananaev: Else?

src/org/traccar/reports/ReportUtils.java
```diff
@@ -250,100 +259,65 @@ private static boolean isMoving(ArrayList<Position> positions, int index,
...
-            boolean isMoving = false;
-            boolean isLast = false;
-            boolean skipped = false;
-            boolean tripFiltered = false;
-
+            boolean trips = reportClass.equals(TripReport.class);
+            MotionEventHandler  motionHandler = new MotionEventHandler(tripsConfig);
+            DeviceState deviceState = new DeviceState();
+            deviceState.setMotionState(isMoving(positions, 0, tripsConfig, speedThreshold));
+            int startTripIndex = trips && deviceState.getMotionState() ? 0 : -1;
```
tananaev: You have combined two methods without actually combining logic. If you pick some more generic names, you won't need to check "trips" variable everywhere and ideally you will only need half lines of code.

You can use something like `startEventIndex` and `startNoEventIndex`. In case of trips "event" means trip and "no event" means parking, and vice versa in case of parking.

Abyss777:
You are right, combined the logic, sorry that take your time.
Also unwrapped check for gap, I think it became less clear than previous version.

tananaev:
Yes, I think we should probably revert the check back to previous version. Sorry about that.

Everything else looks good now.

tananaev:
Thanks. I've merged it.

PR #3449, opened 2017-08-16
Map correct position to delayed event notification

Abyss777:
Totally slip out of mind, we also have to load old position to map it to notification like we do on web side.

It might be looking like quickfix, if you are totally against requesting DB here I have a couple of ideas:
- Simply cache such delayed positions (map `positionId`->`Position` but I do not know when cleanup them, some positions might be related more than one event)
- Rethink how we call event handlers, now, they basically inside separate pipeline handlers. We can implement common pipeline handler to call all event handlers, collect results and call notification once. Because `DeviceState` has cached related position we can pass them to notification and avoid additional DB requests.

Also fixed nested condition

### DIFF
--- a/src/org/traccar/BaseEventHandler.java
+++ b/src/org/traccar/BaseEventHandler.java
@@ -15,7 +15,7 @@
  */
 package org.traccar;
 
-import java.util.Collection;
+import java.util.Map;
 
 import org.traccar.model.Event;
 import org.traccar.model.Position;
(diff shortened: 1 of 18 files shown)


tananaev:
Don't we pass position along with event? Why do we need to request it. Maybe I misunderstand how it works.

Abyss777:
Here, in this function we pass `position` that triggered event but it can be not related, for delayed events it is not related position it is usually after 5 minutes or 500 meters distance or other. Moreover `position` might be `null` if delayed event was triggered in `ConnectionManager` by timeout.

tananaev:
Sorry, but I don't understand the logic. Why would we pass unrelated position to this function?

Abyss777:
We call it from https://github.com/tananaev/traccar/blob/master/src/org/traccar/BaseEventHandler.java

it is one of pipeline handlers we handle one current position by every event handler one by one

tananaev:
I guess we need to change it. We have that position, so I don't see any good reason to request it again. Maybe we can have a map (Event -> Position) instead of simple collection of events.

Abyss777:
OK. I'll investigate it.

tananaev:
Merging, looks much better now. Thanks.

Abyss777:
Good, I also like how it looks.

PR #3450, opened 2017-08-17
Fix trips and stops detector

Abyss777:
Testing trips detector on real data noticed a couple of problems:
- In case `deviceState` switched instantly in one position (stopped and switched ignition off) we missed it
- Fluctuations of opposite state incorrectly handled (did not reset `startEventIndex`)

Added appropriate tests


tananaev:
Merged.

PR #3473, opened 2017-08-25
GT06 alarm packet issue

AnshulJain1985:


tananaev:
There are conflicts. Also, you shouldn't change submodule.

PR #3474, opened 2017-08-25
Gt06 alarm packet

AnshulJain1985:
Skipbyte 8 is correct for location data. But for alarm packet it is 9. Also alarm data is after LBS, due to which it is giving wrong alarms and it needs to be changed to skipbyte 9 as it was previously. 

Also, some gt06 devices sending GPS information length 13 instead of 12, so I have added length to maximum 12 incase it is greater than 12.
- Status packet is sending 13 length for GPS information. But as per protocol document, it cannot be greater than 12.

Example Login Packet
```
78780d01035873507021706101314f600d0a
```
Status Packet
```
78782516110819110208d402f264dc08289a4058d4c70901944600d300c0954606040600014057e90d0a
```
Due to above issue, it is sending alarm power cut as LBS data is skipping one extra value due to which it is taking language which is 02 and 02 is also for power cut.

Please check and let me know if you think there is any issue with this change.


tananaev:
Please provide documentation and fix build issues.

PR #3481, opened 2017-08-28
GT06 alarm packet issue

AnshulJain1985:
Skipbyte 8 is correct for location data. But for alarm packet it is 9. Also alarm data is after LBS, due to which it is giving wrong alarms and it needs to be changed to skipbyte 9 as it was previously.

Also, some gt06 devices sending GPS information length 13 instead of 12, so I have added length to maximum 12 incase it is greater than 12.

Status packet is sending 13 length for GPS information. But as per protocol document, it cannot be greater than 12.
Example Login Packet
```
78780d01035873507021706101314f600d0a
```
Status Packet
```
78782516110819110208d402f264dc08289a4058d4c70901944600d300c0954606040600014057e90d0a
```
Due to above issue, it is sending alarm power cut as LBS data is skipping one extra value due to which it is taking language which is 02 and 02 is also for power cut.

Location data :- Please check table 5.2.1. page no 12 for lbs data with 8 bytes.
Alarm Data :- Please check table 5.3.1. page no 16 for lbs data with 9 bytes.
PFA document for the same.
[new protocol gt06n.pdf](https://github.com/tananaev/traccar/files/1255225/new.protocol.gt06n.pdf)

Please check and let me know if you think there is any issue with this change.

### DIFF
--- a/src/org/traccar/protocol/Gt06ProtocolDecoder.java
+++ b/src/org/traccar/protocol/Gt06ProtocolDecoder.java
@@ -191,7 +191,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
         return true;
     }
 
-    private boolean decodeLbs(Position position, ChannelBuffer buf, boolean hasLength, boolean isExtended) {
+    private boolean decodeLbs(Position position, ChannelBuffer buf, boolean hasLength) {
 
         int length = 0;
         if (hasLength) {
@@ -205,11 +205,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
                 buf.readUnsignedShort(), buf.readUnsignedByte(), buf.readUnsignedShort(), buf.readUnsignedMedium())));
 
         if (length > 0) {
-            if (hasLength && !isExtended) {
-                buf.skipBytes(length - 9);
-            } else {
-                buf.skipBytes(length - 8);
-            }
+            buf.skipBytes(length - (hasLength ? 9 : 8));
         }
 
         return true;
@@ -468,7 +464,8 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
             }
 
             if (hasLbs(type)) {
-                decodeLbs(position, buf, hasStatus(type), false);
+                decodeLbs(position, buf, hasStatus(type));
+                buf.skipBytes(-1);
             }
 
             if (hasStatus(type)) {
@@ -580,7 +577,9 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
                 getLastLocation(position, position.getDeviceTime());
             }
 
-            decodeLbs(position, buf, true, true);
+            if (decodeLbs(position, buf, true)) {
+                buf.skipBytes(1); //Extended data
+            }
 
             buf.skipBytes(buf.readUnsignedByte()); // additional cell towers
             buf.skipBytes(buf.readUnsignedByte()); // wifi access point
--- a/src/org/traccar/protocol/Gt06ProtocolDecoder.java
+++ b/src/org/traccar/protocol/Gt06ProtocolDecoder.java
@@ -182,6 +182,9 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
         }
 
         if (length > 0) {
+            if (length > 12) {
+                length = 12;
+            }
             buf.skipBytes(length - 12); // skip reserved
         }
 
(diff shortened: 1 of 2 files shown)


tananaev:
Please fix the built issue.

AnshulJain1985:
Build is fixed. Sorry about that.
I have added additional parameter to determine extended decode type.

tananaev:
Please provide a test case from real device.

AnshulJain1985:
Hi Anton

Below are hex from real device. I have update the test cases.
Login Packet
78780d01035873507021706101314f600d0a
Location Packet
78781f12110819110216d402f250340828924055d4c801944600d300c09501429c830d0a
Alarm packet
78782516110819110208d402f264dc08289a4058d4c70901944600d300c0954606040600014057e90d0a

Please check.

src/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -202,7 +205,11 @@ private boolean decodeLbs(Position position, ChannelBuffer buf, boolean hasLengt
                 buf.readUnsignedShort(), buf.readUnsignedByte(), buf.readUnsignedShort(), buf.readUnsignedMedium())));
 
         if (length > 0) {
-            buf.skipBytes(length - 8);
+            if (hasLength && !isExtended) {
```
tananaev: Why do we need `isExtended`? Should this just be:
```java
buf.skipBytes(length - (hasLength ? 9 : 8));
```

src/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -182,13 +182,16 @@ private boolean decodeGps(Position position, ChannelBuffer buf, boolean hasLengt
         }
 
         if (length > 0) {
+            if (length > 12) {
```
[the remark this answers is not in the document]
AnshulJain1985: GPS tracker device is sending wrong length for GPS in alarm data. As GPS length cannont be greater than 12, I have added this condition. It is true for Position data also.
78782516110819110208d402f264dc08289a4058d4c70901944600d300c0954606040600014057e90d0a
Quality Of GPS
d4 |  where first bit is length of GPS. d => 13

src/org/traccar/protocol/Gt06ProtocolDecoder.java
AnshulJain1985: I added this because of the test case is failing for Extended decode for 7979.
If I wont add this then one of the test case is failing. As this is true for 7878 as per the documentation I have for GT06.

tananaev: Extended option has extra RSSI field after standard LBS data. I propose we add it directly into `decodeExtended` code.

src/org/traccar/protocol/Gt06ProtocolDecoder.java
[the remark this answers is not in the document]
tananaev: First of all, "GPS length cannont be greater than 12" statement is incorrect according to the documentation. Secondly, this code doesn't make any sense:
```java
if (length > 12) {
    length = 12;
}
buf.skipBytes(length - 12); // skip reserved
```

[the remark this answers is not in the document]
AnshulJain1985: Yes Anton, I totally agree but this is happening from the device. As per documentation it should not happen. The hex string I provided is from actual device of gt06 wetrack 2.

src/org/traccar/protocol/Gt06ProtocolDecoder.java
AnshulJain1985: Sure thats like a good suggestion. I will change this.

AnshulJain1985: I have made the changes. Please check.

tananaev:
Merged, thanks.

AnshulJain1985:
Thanks Anton.

SiddiqueAhmad:
Issue solved? 
We applied same change but still have issues
We have Jimmy gt06

tananaev:
As you can see, the code is merged, so the issue should be fixed.

SantiagoFD:
Hi,
We are still having problems with this using a gt800. If we comment the following code it works correctly.
`if (hasLbs(type)) {`
    `decodeLbs(position, buf, hasStatus(type));`
    `//buf.skipBytes(-1);`
`}`

tananaev:
Not sure how it slipped through the code review. Removed it now.

SantiagoFD:
Thanks

PR #3484, opened 2017-08-28
Do not filter alarms

Abyss777:
Some protocols report alarms without coordinates, we use last position to construct new one. Such position very likely will be filtered (if filter enabled)

Implemented configuration parameter to skip such positions (not filter)

Parameter name is controversial.

Also fixed http://findbugs.sourceforge.net/bugDescriptions.html#ICAST_IDIV_CAST_TO_DOUBLE

### DIFF
--- a/src/org/traccar/FilterHandler.java
+++ b/src/org/traccar/FilterHandler.java
@@ -147,10 +147,6 @@ public class FilterHandler extends BaseDataHandler {
 
     private boolean filter(Position position) {
 
-        if (keepAlarms && position.getAttributes().containsKey(Position.KEY_ALARM)) {
-            return false;
-        }
-
         StringBuilder filterType = new StringBuilder();
 
         Position last = null;
@@ -158,9 +154,10 @@ public class FilterHandler extends BaseDataHandler {
             last = Context.getIdentityManager().getLastPosition(position.getDeviceId());
         }
 
-        if (filterLimit(position, last)) {
-            filterType.append("Limit ");
+        if (filterLimit(position, last) || keepAlarms && position.getAttributes().containsKey(Position.KEY_ALARM)) {
+            return false;
         }
+
         if (filterInvalid(position)) {
             filterType.append("Invalid ");
         }
--- a/src/org/traccar/FilterHandler.java
+++ b/src/org/traccar/FilterHandler.java
@@ -138,11 +138,15 @@ public class FilterHandler extends BaseDataHandler {
         return false;
     }
 
-    private boolean filterLimit(Position position, Position last) {
+    private boolean skipLimit(Position position, Position last) {
         if (filterLimit != 0 && last != null) {
             return (position.getFixTime().getTime() - last.getFixTime().getTime()) > filterLimit;
         }
         return false;
+    }
+
+    private boolean skipAlarms(Position position) {
+        return keepAlarms && position.getAttributes().containsKey(Position.KEY_ALARM);
     }
 
     private boolean filter(Position position) {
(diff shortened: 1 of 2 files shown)


src/org/traccar/FilterHandler.java
```diff
@@ -127,7 +133,7 @@ private boolean filterMaxSpeed(Position position, Position last) {
         if (filterMaxSpeed != 0 && last != null) {
             double distance = position.getDouble(Position.KEY_DISTANCE);
             long time = position.getFixTime().getTime() - last.getFixTime().getTime();
-            return UnitsConverter.knotsFromMps(distance / (time / 1000)) > filterMaxSpeed;
+            return UnitsConverter.knotsFromMps(distance / (time / 1000.0)) > filterMaxSpeed;
```
tananaev: Why do we need decimal point here?

src/org/traccar/FilterHandler.java
```diff
@@ -30,6 +30,7 @@
     private int filterDistance;
     private int filterMaxSpeed;
     private long filterLimit;
+    private boolean allowAlarms;
```
tananaev: I would rename it to `keepAlarms` to make it clearer.

src/org/traccar/FilterHandler.java
```diff
@@ -146,6 +152,10 @@ private boolean filterLimit(Position position, Position last) {
 
     private boolean filter(Position position) {
 
+        if (allowAlarms && position.getAttributes().containsKey(Position.KEY_ALARM)) {
```
tananaev: Can you please move `filterLimit` check to the beginning as well.

src/org/traccar/FilterHandler.java
Abyss777: Here is FindBug complain http://findbugs.sourceforge.net/bugDescriptions.html#ICAST_IDIV_CAST_TO_DOUBLE
And I agree with it, we are loosing some precision.

tananaev: In most cases devices report only seconds, so we don't lose anything. Also, usually max speed is very high value, so 0.1% doesn't make much difference.

Anyway, I just don't like how it looks. Maybe we can change variable from `long` to `double`?

Abyss777:
Renamed parameter, changed variable type, moved Limit filter to the beginning

src/org/traccar/FilterHandler.java
```diff
@@ -126,33 +132,35 @@ private boolean filterDistance(Position position, Position last) {
...
+        }
+
         StringBuilder filterType = new StringBuilder();
 
         Position last = null;
         if (Context.getIdentityManager() != null) {
             last = Context.getIdentityManager().getLastPosition(position.getDeviceId());
         }
 
+        if (filterLimit(position, last)) {
```
tananaev: I think you misunderstood how limit works. It's works similar to "keepAlarms". Basically, if time between last position is more than X, then don't filter. Maybe we should rename it to avoid confusion.

Abyss777: I have inverted the comparison, now it is like other functions

tananaev: I see that, but limit is not for filtering, it's for preventing any other filtering.

Abyss777: Got it

Abyss777:
Replaced last commit

Abyss777:
I just thought may be rename `filterLimit` to `skipLimit` (or something) and move next code to function. Result will look very clear, like this
````java
        if (skipLimit(position, last) || skipAlarms(position)) {
            return false;
        }
````

**Comment by tananaev** [comment]
Yes, I was going to ask you if you can think of a better name. I like `skip`.

File: src/org/traccar/FilterHandler.java [hunk 5]
```diff
@@ -30,6 +30,7 @@
     private int filterDistance;
     private int filterMaxSpeed;
     private long filterLimit;
```
tananaev: Please rename this variable and config option as well.

tananaev:
Merged, thanks.

Abyss777:
I've renamed both variables and parameters to make them consistence

PR #3491, opened 2017-08-29
Make OsmAnd speed units configurable

### LINKED ISSUE #3487: Wrong speed with OsmAnd?
If I configure OsmAnd~ 2.7.5 online logging as follows:

`http://example.com:5055/?lat={0}&lon={1}&timestamp={2}&hdop={3}&altitude={4}&speed={5}& bearing={6}&id=666&trip=Test+2`

The speeds reported in Traccar are incorrect: it shows half the actual speed in km/h. For example, driving at 50 km/h results in values around 25 being displayed (see also #3486 concerning units). Driving at 70 km/h shows values around 35.

Car speeds were stable due to cruise control. Odometer speeds are unusually accurate (1 km/h slow). Speeds on OsmAnd are shown correctly as 50 km/h and 70 km/h (plus/minus GPS jitter).

Abyss777:
It is not obvious but looks like OsmAnd application post speed in m/s units http://osmand.net/features?id=trip-recording-plugin#Online_tracking , but Traccar Client uses knots.

I think it is good idea to make it configurable, at least server wide.

Also fix #3487

### DIFF
--- a/src/org/traccar/protocol/OsmAndProtocolDecoder.java
+++ b/src/org/traccar/protocol/OsmAndProtocolDecoder.java
@@ -25,9 +25,7 @@ import org.jboss.netty.handler.codec.http.HttpVersion;
 import org.jboss.netty.handler.codec.http.QueryStringDecoder;
 import org.joda.time.format.ISODateTimeFormat;
 import org.traccar.BaseProtocolDecoder;
-import org.traccar.Context;
 import org.traccar.DeviceSession;
-import org.traccar.helper.UnitsConverter;
 import org.traccar.model.Position;
 
 import java.net.SocketAddress;
@@ -40,11 +38,8 @@ import java.util.Map;
 
 public class OsmAndProtocolDecoder extends BaseProtocolDecoder {
 
-    private String speedUnits;
-
     public OsmAndProtocolDecoder(OsmAndProtocol protocol) {
         super(protocol);
-        speedUnits = Context.getConfig().getString("osmand.speedUnits", "kn");
     }
 
     private void sendResponse(Channel channel, HttpResponseStatus status) {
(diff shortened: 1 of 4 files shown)


src/org/traccar/protocol/OsmAndProtocolDecoder.java
```diff
@@ -38,8 +40,11 @@
 
 public class OsmAndProtocolDecoder extends BaseProtocolDecoder {
 
+    private String speedUnits;
+
     public OsmAndProtocolDecoder(OsmAndProtocol protocol) {
         super(protocol);
+        speedUnits = Context.getConfig().getString("osmand.speedUnits", "kn");
```
tananaev: Please make it consistent with other places:

https://github.com/tananaev/traccar/blob/master/src/org/traccar/protocol/Tk103ProtocolDecoder.java#L256

Possibly updating TK103 decoder would be even better.

Abyss777:
May be we should implement some task tracker or exchange plans somehow.

src/org/traccar/protocol/OsmAndProtocolDecoder.java
Abyss777: Don't you mind if I move common code to `BaseProtocolDecoder` ?

tananaev: I guess it makes sense.

tananaev:
I think it works fine without any tracking. If you start working on some bigger change, just let me know. Email or comment on github issue would be fine.

tananaev:
Merged, thanks.

9662:
Marginal comment:

> May be we should implement some task tracker or exchange plans somehow.

In my projects over at [GitLab](https://gitlab.com/), whoever starts working on a task [assigns it to himself](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/), usually followed by creation of a merge request (pull request in GitHub speak?)

tananaev:
That's also an option.

Issue #3496, opened 2017-09-01
Commands improvement


Abyss777:
I'm developing this https://github.com/tananaev/traccar-web/issues/568 and want to discuss some things related to server side.

Some devices have `devicePassword` to send command, we can set additional password parameter in command attributes every time or set it in device attributes ones. https://github.com/tananaev/traccar/blob/master/src/org/traccar/BaseProtocolEncoder.java#L31

Some other protocols require send IMEI like password https://github.com/tananaev/traccar/blob/master/src/org/traccar/protocol/Tk103ProtocolEncoder.java

- Can we implement function `initUniqueID` that will take `deviceUniqueId` if it is omitted in command attributes?

-  Some protocols need password, some imei, some nothing. To dynamically add password or imei field to command dialog we somehow need to notify web-part about it. I can't find simple way to do that. I see two options: always add password and uniqueId fields to send command dialog. Or opposite do not add this fields and take them only from device fields/attributes. What do you think?

tananaev:
We never send `uniqueId` in attributes because it's already available on the server. Or am I missing something?

Password is more complicated, but I think we can just ask user to provide it in device attributes. It's possible to get it from command attributes as well, but this option is mostly for API use.

Abyss777:
Oh, sorry, It is me who missed something. We already initialize `uniqueId` from device.

Got it for password, I incline to the same.

Abyss777:
Another suggestion... now we are setting timezone (for command) in hours and pass in command attributes in seconds. But we already have list of all timezones names on web side and can make selection convenient for user.
We can pass timezone name from web side and convert it to necessary offset on server side. Just can't decide: introduce new attribute like `Command.KEY_TIMEZONE_NAME` or automatically detect `Command.KEY_TIMEZONE` string value. Or maybe even accept only timezone names.

tananaev:
Agree in general. As for approach, whatever is easier and nicer is probably the right way.

Abyss777:
All necessary server side changes for https://github.com/tananaev/traccar-web/pull/571 was implemented. Closing this.

PR #3502, opened 2017-09-04
Accept timezone id instead of offset in commands

Abyss777:
In continuous #3496  
Decided to replace how we treat `Command.KEY_TIMEZONE`. Now it is timezone id (like user/server timezone)
As about backward compatibility, i think it will be easy convert seconds to "GMT+XX:XX" if somebody use this API.

Unfortunately `TimeZone.getTimeZone()` return "GMT" not exception if passed unknown/incorrect id but I thing it is OK.

PS: Pt502 does not have any Timezone formatting, and worked incorrectly.

### DIFF
--- a/src/org/traccar/protocol/MiniFinderProtocolEncoder.java
+++ b/src/org/traccar/protocol/MiniFinderProtocolEncoder.java
@@ -29,7 +29,7 @@ public class MiniFinderProtocolEncoder extends StringProtocolEncoder implements 
         if (key.equals(Command.KEY_ENABLE)) {
             return (Boolean) value ? "1" : "0";
         } else if (key.equals(Command.KEY_TIMEZONE)) {
-            return String.format("%+03d", TimeZone.getTimeZone(value.toString()).getRawOffset() / 3600000);
+            return String.format("%+03d", TimeZone.getTimeZone((String) value).getRawOffset() / 3600000);
         } else if (key.equals(Command.KEY_INDEX)) {
             switch (((Number) value).intValue()) {
                 case 0:
--- a/src/org/traccar/protocol/MiniFinderProtocolEncoder.java
+++ b/src/org/traccar/protocol/MiniFinderProtocolEncoder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Anton Tananaev (anton@traccar.org)
+ * Copyright 2016 - 2017 Anton Tananaev (anton@traccar.org)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


src/org/traccar/protocol/MiniFinderProtocolEncoder.java
```diff
@@ -27,7 +29,7 @@ public String formatValue(String key, Object value) {
         if (key.equals(Command.KEY_ENABLE)) {
             return (Boolean) value ? "1" : "0";
         } else if (key.equals(Command.KEY_TIMEZONE)) {
-            return String.format("%+03d", ((Number) value).longValue() / 3600);
+            return String.format("%+03d", TimeZone.getTimeZone(value.toString()).getRawOffset() / 3600000);
```
tananaev: Why do we need to call `toString()`? Same for other places as well.

src/org/traccar/protocol/Pt502ProtocolEncoder.java
```diff
@@ -15,11 +15,27 @@
...
+        if (key.equals(Command.KEY_TIMEZONE)) {
+            return String.valueOf(TimeZone.getTimeZone(value.toString()).getRawOffset() / 3600000);
+        }
+
+        return null;
+    }
+
+    @Override
+    protected String formatCommand(Command command, String format, String... keys) {
+        return super.formatCommand(command, format, this, keys);
```
tananaev: Why `super`?

src/org/traccar/protocol/MiniFinderProtocolEncoder.java
Abyss777: Because `value` is `Object`. Better use casting?

src/org/traccar/protocol/Pt502ProtocolEncoder.java
Abyss777: Actually, I've copied from here https://github.com/tananaev/traccar/blob/master/src/org/traccar/protocol/WatchProtocolEncoder.java#L45
Should I remove both?

src/org/traccar/protocol/MiniFinderProtocolEncoder.java
tananaev: Yes, casting is better if we know it's a String.

src/org/traccar/protocol/Pt502ProtocolEncoder.java
tananaev: I think originally it was same signature in watch encoder, so super was required. Now it looks like it's not required. Please use it as well.

tananaev:
Merged, thanks.

PR #3504, opened 2017-09-05
Remove Attribute Aliases

Abyss777:
Removed Attribute Aliases

Also added missed `api/report/stops` to `swagger.json`


tananaev:
Merged, thanks.

PR #3507, opened 2017-09-06
Move units to attributes and add volume units

Abyss777:
- Removed speed, distance units and timezone from user/server model
- Get speed, distance, volume units and timezone from user/server attributes for reports and notifications
- Map other units to notifications in case somebody want use them.
- Adjusted templates

According to Wikipedia there are two wide spread types of gallons for fuel and they have the same common used abbreviation - "gal".  I'm not sure if we have to invent different abbreviation. I used "gal" for both. But they coded as `impGal` and `usGal`, liters as `ltr` . Here is changes in Excel templates:
````
${summary.spentFuel}
````
changed to
````
${volumeUnit.equals("impGal") ? "".format("%.1f gal", summary.spentFuel / 4.546) : volumeUnit.equals("usGal") ? "".format("%.1f gal", summary.spentFuel / 3.785) : "".format("%.1f l", summary.spentFuel)}
````

````
${trip.spentFuel}
````
changed to
````
${volumeUnit.equals("impGal") ? "".format("%.1f gal", trip.spentFuel / 4.546) : volumeUnit.equals("usGal") ? "".format("%.1f gal", trip.spentFuel / 3.785) : "".format("%.1f l", trip.spentFuel)}
````

````
${stop.spentFuel}
````
changed to
````
${volumeUnit.equals("impGal") ? "".format("%.1f gal", stop.spentFuel / 4.546) : volumeUnit.equals("usGal") ? "".format("%.1f gal", stop.spentFuel / 3.785) : "".format("%.1f l", stop.spentFuel)}
````


tananaev:
Merged, thanks.

PR #3511, opened 2017-09-07
Handle correct and incorrect attribute types

Abyss777:
- Moved Boolean workaround outside the `ExtendedModel`. I think it should be clean.
- Adapted `DeviceManager.lookupAttributeXxx` to handle correct and incorrect types (temporary solution, for year or so)
- Handle boolean attributes in `OsmAndProtocol`
- Handle `mail.smtp.port` as integer

Some comments in code

### DIFF
--- a/src/org/traccar/database/DeviceManager.java
+++ b/src/org/traccar/database/DeviceManager.java
@@ -262,59 +262,44 @@ public class DeviceManager extends BaseObjectManager<Device> implements Identity
 
     public boolean lookupAttributeBoolean(
             long deviceId, String attributeName, boolean defaultValue, boolean lookupConfig) {
-        String result = lookupAttribute(deviceId, attributeName, lookupConfig);
-        if (result != null) {
-            return Boolean.parseBoolean(result);
-        }
-        return defaultValue;
+        Object result = lookupAttribute(deviceId, attributeName, lookupConfig);
+        return result != null ? Boolean.parseBoolean(result.toString()) : defaultValue;
     }
 
     public String lookupAttributeString(
             long deviceId, String attributeName, String defaultValue, boolean lookupConfig) {
-        String result = lookupAttribute(deviceId, attributeName, lookupConfig);
-        if (result != null) {
-            return result;
-        }
-        return defaultValue;
+        Object result = lookupAttribute(deviceId, attributeName, lookupConfig);
+        return result != null ? result.toString() : defaultValue;
     }
 
     public int lookupAttributeInteger(long deviceId, String attributeName, int defaultValue, boolean lookupConfig) {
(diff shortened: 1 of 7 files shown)


src/org/traccar/notification/NotificationMail.java
```diff
@@ -43,7 +43,7 @@ private static Properties getProperties(PropertiesProvider provider) {
         if (host != null) {
             properties.put("mail.transport.protocol", provider.getString("mail.transport.protocol", "smtp"));
             properties.put("mail.smtp.host", host);
-            properties.put("mail.smtp.port", provider.getString("mail.smtp.port", "25"));
+            properties.put("mail.smtp.port", String.valueOf(provider.getInteger("mail.smtp.port", 25)));
```
Abyss777: According to documentation it should be integer but stored as string https://javaee.github.io/javamail/docs/api/com/sun/mail/smtp/package-summary.html

> The SMTP protocol provider supports the following properties, which may be set in the JavaMail Session object. The properties are always set as strings; the Type column describes how the string is interpreted. For example, use
        props.put("mail.smtp.port", "888");
to set the mail.smtp.port property, which is of type int.

src/org/traccar/database/DeviceManager.java
```diff
@@ -262,59 +262,44 @@ public Position getLastPosition(long deviceId) {
 
     public boolean lookupAttributeBoolean(
             long deviceId, String attributeName, boolean defaultValue, boolean lookupConfig) {
-        String result = lookupAttribute(deviceId, attributeName, lookupConfig);
-        if (result != null) {
-            return Boolean.parseBoolean(result);
-        }
-        return defaultValue;
+        Object result = lookupAttribute(deviceId, attributeName, lookupConfig);
+        return result != null ? Boolean.parseBoolean(result.toString()) : defaultValue;
```
tananaev: I don't think it's the right way. It's better to check object type, and only parse if it's a string. Applies to all other methods.

Abyss777: Something like this?
````java
        Object result = lookupAttribute(deviceId, attributeName, lookupConfig);
        if (result != null) {
            if (result instanceof String) {
                return Boolean.parseBoolean((String) result);
            } else {
                return (Boolean) result;
            }
        } else {
            return defaultValue;
        }
    }

````

Comment (tananaev) [reply src/org/traccar/database/DeviceManager.java]: Yes, maybe something like this to make it shorter:
```java
return result instanceof String ? Boolean.parseBoolean((String) result) : (Boolean) result;
```

Abyss777:
Done

tananaev:
Merged, thanks.

PR #3514, opened 2017-09-08
Implement skipping filtering by list of attributes

### LINKED ISSUE #3513: FR: do not filter command results
Command result usually comes without real position, we construct it from last one. It is very likely it will be filtered by duplicate or distance.
What do you think about adding new switch to exclude from filtering positions with command results? Like we do for alarms #3484

Abyss777:
Main switch is `filter.skipAttributes.enable` because lookup might costs a lot.
Attributes list should be in `filter.skipAttributes`

fix #3513 

More comments in code

### DIFF
--- a/src/org/traccar/FilterHandler.java
+++ b/src/org/traccar/FilterHandler.java
@@ -30,7 +30,7 @@ public class FilterHandler extends BaseDataHandler {
     private int filterDistance;
     private int filterMaxSpeed;
     private long skipLimit;
-    private boolean skipAlarms;
+    private boolean skipAttributes;
 
     public void setFilterInvalid(boolean filterInvalid) {
         this.filterInvalid = filterInvalid;
@@ -68,8 +68,8 @@ public class FilterHandler extends BaseDataHandler {
         this.skipLimit = skipLimit;
     }
 
-    public void setSkipAlarms(boolean skipAlarms) {
-        this.skipAlarms = skipAlarms;
+    public void setSkipAttributes(boolean skipAttributes) {
+        this.skipAttributes = skipAttributes;
     }
 
     public FilterHandler() {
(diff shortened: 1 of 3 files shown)


test/org/traccar/BaseTest.java
```diff
@@ -40,25 +40,25 @@ public boolean isLatestPosition(Position position) {
...
                     long deviceId, String attributeName, boolean defaultValue, boolean lookupConfig) {
-                return false;
+                return defaultValue;
             }
 
             @Override
             public String lookupAttributeString(
                     long deviceId, String attributeName, String defaultValue, boolean lookupConfig) {
-                return null;
+                return defaultValue;
```
Abyss777: Some pieces of code might rely on `defaultValue` and not waited `null`

test/org/traccar/FilterHandlerTest.java
```diff
@@ -3,14 +3,74 @@
...
 import java.util.Date;
 
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 
-public class FilterHandlerTest extends BaseTest {
+public class FilterHandlerTest {
+
+    static {
+        Context.init(new IdentityManager() {
```
Abyss777: Have not found better way to override `lookupAttribute` returned value than totally reinitialize `IdentityManager`

src/org/traccar/FilterHandler.java
```diff
@@ -145,8 +145,17 @@ private boolean skipLimit(Position position, Position last) {
         return false;
     }
 
-    private boolean skipAlarms(Position position) {
-        return skipAlarms && position.getAttributes().containsKey(Position.KEY_ALARM);
+    private boolean skipAttributes(Position position) {
+        if (skipAttributes) {
+            String attributesString = Context.getIdentityManager().lookupAttributeString(
+                    position.getDeviceId(), "filter.skipAttributes", "", true);
+            for (String attribute : attributesString.split("[ ,]")) {
```
tananaev: I think we need to make it more efficient than parsing string every time. Maybe parse it into a list once when we create handler.

Abyss777: We are looking up in devices tree, not only in config. It is impossible to "pre-parse"

tananaev: Good point. I forgot about it.

tananaev:
Merged, thanks.

PR #3517, opened 2017-09-09
Improve ADM protocol

naihil:
Improve ADM protocol decoder;
Implement ADM protocol commands.


src/org/traccar/protocol/AdmProtocolDecoder.java
```diff
@@ -33,76 +33,80 @@ public AdmProtocolDecoder(AdmProtocol protocol) {
...
-        int type = buf.readUnsignedByte();
-
-        DeviceSession deviceSession;
-        if (type == MSG_IMEI) {
-            deviceSession = getDeviceSession(
-                    channel, remoteAddress, buf.readBytes(15).toString(StandardCharsets.US_ASCII));
-        } else {
-            deviceSession = getDeviceSession(channel, remoteAddress);
-        }
+    private DeviceSession parseIdentification(Channel channel, SocketAddress remoteAddress, ChannelBuffer buf) {
```
tananaev: I don't see any good reason for separating this method.

src/org/traccar/protocol/AdmProtocolDecoder.java
```diff
@@ -33,76 +33,80 @@ public AdmProtocolDecoder(AdmProtocol protocol) {
...
 
             position.set(Position.KEY_VERSION_FW, buf.readUnsignedByte());
-            buf.readUnsignedShort(); // index
+            position.set(Position.KEY_INDEX, buf.readUnsignedShort());
+
+            int status = buf.readUnsignedShort();
+            position.set(Position.KEY_STATUS, status);
 
-            position.set(Position.KEY_STATUS, buf.readUnsignedShort());
+            boolean isValid = !BitUtil.check(status, 5);
```
tananaev: Why do you need a variable here?

src/org/traccar/protocol/AdmProtocolDecoder.java
```diff
@@ -119,4 +123,41 @@ protected Object decode(
...
+    }
+
+    @Override
+    protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
+        ChannelBuffer buf = (ChannelBuffer) msg;
+
+        buf.readUnsignedShort(); // device id
+
+        int size = buf.readUnsignedByte();
+        if (size != CMD_RESPONSE_SIZE) {
```
tananaev: Is there a better way to identify command response?

src/org/traccar/protocol/AdmProtocolEncoder.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.jboss.netty.buffer.ChannelBuffer;
+import org.jboss.netty.buffer.ChannelBuffers;
+import org.traccar.BaseProtocolEncoder;
+import org.traccar.helper.Log;
+import org.traccar.model.Command;
+
+import java.nio.charset.StandardCharsets;
+
+public class AdmProtocolEncoder extends BaseProtocolEncoder {
```
tananaev: Looks like device is using string format for commands. Use `StringEncoder` instead of converting it into buffer here.

src/org/traccar/protocol/AdmProtocolEncoder.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private ChannelBuffer encodeContent(String data) {
```
tananaev: I don't think this method would be required. You just need to add `\r\n` to the string in `encodeCommand`.

src/org/traccar/protocol/AdmProtocolDecoder.java
naihil: No. Command response identified only by packet size

naihil:
I`ll fix issues and look at StringEncoder, documentation attached
[protocol_ADM.pdf](https://github.com/tananaev/traccar/files/1290112/protocol_ADM.pdf)

src/org/traccar/protocol/AdmProtocolDecoder.java
```diff
@@ -119,4 +115,41 @@ protected Object decode(
...
+            return null;
+        }
+
+        Position position = new Position();
+        position.setProtocol(getProtocolName());
+        position.setDeviceId(deviceSession.getDeviceId());
+
+        getLastLocation(position, null);
+
+        position.set(Position.KEY_RESULT, buf.readBytes(129).toString(StandardCharsets.UTF_8));
```
tananaev: We probably need to trim this string. Also, can we re-use `CMD_RESPONSE_SIZE` constant here?

src/org/traccar/protocol/AdmProtocolDecoder.java
```diff
@@ -33,76 +33,72 @@ public AdmProtocolDecoder(AdmProtocol protocol) {
...
-
-        DeviceSession deviceSession;
-        if (type == MSG_IMEI) {
-            deviceSession = getDeviceSession(
-                    channel, remoteAddress, buf.readBytes(15).toString(StandardCharsets.US_ASCII));
-        } else {
-            deviceSession = getDeviceSession(channel, remoteAddress);
-        }
-
+    private Position parseData(Channel channel, SocketAddress remoteAddress, ChannelBuffer buf, int type) {
```
tananaev: Please rename to `decodeData`. Parsing usually implies string content.

src/org/traccar/protocol/AdmProtocolDecoder.java
naihil: 129 bytes is fixed length for response text. We can use (CMD_RESPONSE_SIZE - 3) if its important.

tananaev: Yes, that's what I meant. Can you please also add a unit test for it.

naihil: I checked some commands and i think that simple trim is not enough. Device dont truncate previous command result,
Example:
Exec STATUS: 
"result":"ID=1 Soft=0x51 GPS=919 Time=22:36:21 09.09.17 Val=0 Lat=61.6778 Lon=50.8242 V=0 SatCnt=3+8 Stat=0x0010 In_alarm=0x00\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000"
Then exec INPUT 0: 
"result":"INPUT0 = 0\u0000x51 GPS=919 Time=22:36:21 09.09.17 Val=0 Lat=61.6778 Lon=50.8242 V=0 SatCnt=3+8 Stat=0x0010 In_alarm=0x00\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000"
Input command result rewrote beginning of previous command with trailing \u0000 and not truncate remaining buffer. 
Maybe better not trim but get string before first \u0000 symbol?

tananaev: Yes. Looks like it's a zero-terminated string, so we need to properly cut it.

tananaev:
Merged, thanks.

PR #3534, opened 2017-09-19
Implement per device Notifications

### LINKED ISSUE #2893: Notification per device
Hi,
Can I set individual notification for each device, instead per account?

Regards,

Abyss777:
- Made `Notification` standard objects, managed by `ExtendedObjectManager`
- Add `allways` field to Notification model, this flag means that user wants to get notification of desired type from all his devices.
- Made `Typed` class to use when we return Command Types and Notification Types

More comments in code

fix #2893

### DIFF
--- a/schema/changelog-3.15.xml
+++ b/schema/changelog-3.15.xml
@@ -83,5 +83,55 @@
       <column name="limitcommands" type="BOOLEAN" defaultValueBoolean="false" />
     </addColumn>
 
+    <addColumn tableName="notifications">
+      <column name="always" type="BOOLEAN" defaultValueBoolean="false" valueBoolean="true">
+        <constraints nullable="false" />
+      </column>
+    </addColumn>
+
+    <createTable tableName="user_notification">
+      <column name="userid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="notificationid" type="INT">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <addForeignKeyConstraint baseTableName="user_notification" baseColumnNames="userid" constraintName="fk_user_notification_userid" referencedTableName="users" referencedColumnNames="id" onDelete="CASCADE" />
+
+    <sql>
+      INSERT INTO user_notification (notificationid, userid) SELECT id AS notificationid, userid FROM notifications;
+    </sql>
+
+    <dropForeignKeyConstraint baseTableName="notifications" constraintName="fk_notifications_userid"/>
+    <dropColumn tableName="notifications" columnName="userid" />
+
+    <addForeignKeyConstraint baseTableName="user_notification" baseColumnNames="notificationid" constraintName="fk_user_notification_notificationid" referencedTableName="notifications" referencedColumnNames="id" onDelete="CASCADE" />
+
+    <createTable tableName="group_notification">
+      <column name="groupid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="notificationid" type="INT">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <addForeignKeyConstraint baseTableName="group_notification" baseColumnNames="groupid" constraintName="fk_group_notification_groupid" referencedTableName="groups" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="group_notification" baseColumnNames="notificationid" constraintName="fk_group_notification_notificationid" referencedTableName="notifications" referencedColumnNames="id" onDelete="CASCADE" />
+
+    <createTable tableName="device_notification">
+      <column name="deviceid" type="INT">
+        <constraints nullable="false" />
+      </column>
+      <column name="notificationid" type="INT">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <addForeignKeyConstraint baseTableName="device_notification" baseColumnNames="deviceid" constraintName="fk_device_notification_deviceid" referencedTableName="devices" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="device_notification" baseColumnNames="notificationid" constraintName="fk_device_notification_notificationid" referencedTableName="notifications" referencedColumnNames="id" onDelete="CASCADE" />
+
   </changeSet>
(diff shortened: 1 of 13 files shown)


src/org/traccar/database/NotificationManager.java
```diff
@@ -18,57 +19,70 @@
...
-    private final ReadWriteLock notificationsLock = new ReentrantReadWriteLock();
+public class NotificationManager extends ExtendedObjectManager<Notification> {
 
     public NotificationManager(DataManager dataManager) {
-        this.dataManager = dataManager;
-        refresh();
+        super(dataManager, Notification.class);
+    }
+
+    private Set<Long> getEffectiveNotifications(long userId, long deviceId) {
```
Abyss777: This function return ids of notifications that linked to user and either "always" or linked to device.
I decided that maps for every used objects are not necessary. Here is only one iteration for pair user-device.

src/org/traccar/database/NotificationManager.java
```diff
@@ -18,57 +19,70 @@
...
             if (event.getGeofenceId() == 0 || Context.getGeofenceManager() != null
                     && Context.getGeofenceManager().checkItemPermission(userId, event.getGeofenceId())) {
-                Notification notification = getUserNotificationByType(userId, event.getType());
-                if (notification != null) {
-                    if (notification.getWeb()) {
-                        Context.getConnectionManager().updateEvent(userId, event);
-                    }
-                    if (notification.getMail()) {
-                        NotificationMail.sendMailAsync(userId, event, position);
+                boolean webSent = false;
```
Abyss777: This flags help to not do extra iterations and not duplicate notifications

schema/changelog-3.15.xml
```diff
@@ -83,5 +83,54 @@
       <column name="limitcommands" type="BOOLEAN" defaultValueBoolean="false" />
     </addColumn>
 
+    <addColumn tableName="notifications">
+      <column name="allways" type="BOOLEAN" defaultValueBoolean="false" valueBoolean="true">
```
Abyss777: Make old notifications "always" to make migration transparent to users.

schema/changelog-3.15.xml
```diff
@@ -83,5 +83,54 @@
...
+      <column name="notificationid" type="INT">
+        <constraints nullable="false" />
+      </column>
+    </createTable>
+
+    <addForeignKeyConstraint baseTableName="user_notification" baseColumnNames="userid" constraintName="fk_user_notification_userid" referencedTableName="users" referencedColumnNames="id" onDelete="CASCADE" />
+    <addForeignKeyConstraint baseTableName="user_notification" baseColumnNames="notificationid" constraintName="fk_user_notification_notificationid" referencedTableName="notifications" referencedColumnNames="id" onDelete="CASCADE" />
+
+    <sql>
+      INSERT INTO user_notification (notificationid, userid) SELECT id AS notificationid, userid FROM notifications;
```
Abyss777: Migrate old notifications to new scheme transparent.

schema/changelog-3.15.xml
tananaev: There is typo in `always`. There should be only one `L`.

schema/changelog-3.15.xml
tananaev: By the way, have you tested this on different databases? I think it should work, but I'm not completely sure. I think we should test at least H2, MySQL and MS SQL.

Abyss777:
Shame on me for such typo. You were right, there was issue in MSSQL, but not related to that sql request.
H2 does not have issues.

Replaced commit.

schema/changelog-3.15.xml
```diff
@@ -83,5 +83,55 @@
...
+    <addForeignKeyConstraint baseTableName="user_notification" baseColumnNames="userid" constraintName="fk_user_notification_userid" referencedTableName="users" referencedColumnNames="id" onDelete="CASCADE" />
+
+    <sql>
+      INSERT INTO user_notification (notificationid, userid) SELECT id AS notificationid, userid FROM notifications;
+    </sql>
+
+    <dropForeignKeyConstraint baseTableName="notifications" constraintName="fk_notifications_userid"/>
+    <dropColumn tableName="notifications" columnName="userid" />
+
+    <addForeignKeyConstraint baseTableName="user_notification" baseColumnNames="notificationid" constraintName="fk_user_notification_notificationid" referencedTableName="notifications" referencedColumnNames="id" onDelete="CASCADE" />
```
Abyss777: This constraint must go after drop `userId` column for MSSQL, otherwise it compliance about possible loop.

src/org/traccar/database/NotificationManager.java
tananaev: One small comment here. I think if we move "sent" to be a prefix, it will be easier to read the code.

Abyss777:
Done

tananaev:
Merged, thanks.

Abyss777:
Agree that selecting the notifications is slower than previous implementation. But I think not dramatic, even not remarkable. I'll try to benchmark it somehow. But I think any complexity in implementation will make process slower.

tananaev:
I was thinking about it because now we have two nested loops. One for all users for a device, and then for notifications, but I guess in most cases there will be only one user per device, so it shouldn't be a big problem.

Abyss777:
I've tested it with simple benchmark
at the begining of `updateEvent` function
````java
long start = System.currentTimeMillis();
````
and at the end
````
System.out.println(System.currentTimeMillis() - start);
````

I enabled all web notifications and alarms and speed as mail also in 3.14

Run test-generator and get following numbers
````
12
12
1
1
3
2
6
13
4
3
3
5
3
2
2
3
1
2
3
3
5
10
6
11
5
9
1
````

after that I migrated to current and make following changes:
leave "always" only for web, create "alarm" and link to device, create "speed" and link to device group.
and got the following numbers
````
5
5
3
3
1
2
1
24
1
1
1
3
2
2
1
3
2
7
````

Funny, but I think there is no difference moreover new scheme gives more stable numbers.

hairy-dog:
I'm really sorry to dig up such an old thread, but I am trying to work out how to set notifications for individual devices inn the GUI and I can't find it. I'm sure there must be a way to do it.

tananaev:
For questions please use the forum, not GitHub.

PR #3538, opened 2017-09-20
Move Commandtypes API to Commands

Abyss777:
We have not discussed it but I think it is reasonable
- It is more logically to have the same first level API point to commands related requests
- It is consistent with Notifications https://github.com/tananaev/traccar/blob/master/src/org/traccar/api/resource/NotificationResource.java#L52
- We removing some code duplication

PS: Do not know how to treat with Gabor Somogyi copyright in this case


tananaev:
You can probably add copyright line to the CommandResource class. Otherwise looks great.

Abyss777:
Added copyright to CommandResource

tananaev:
Merged, thanks.

PR #3540, opened 2017-09-21
Make check for valid in MotionEventHandler optional

### LINKED ISSUE #3539: Checks for Valid in event handlers
I'm testing teltonika FMB920 and faced with problem that traccar does not detect motion/stop. The problem is that stopped positions device reports as not valid.

We have checks for Valid in some events handlers. I think we should remove them from handlers not related to coordinates:
- motion
- fuel drop

But about motion I have some doubts... On the one hand, if motion automatically calculated from speed it might be incorrect for invalid position, on the other hand if motion reported from device it is correct even if position is invalid.

@tananaev what do you think?

Abyss777:
fix #3539 

Also removed check for valid from `FuelDropEventHandler`

### DIFF
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -258,7 +258,8 @@ public final class Context {
                 config.getLong("report.trip.minimalTripDuration", 300) * 1000,
                 config.getLong("report.trip.minimalParkingDuration", 300) * 1000,
                 config.getLong("report.trip.minimalNoDataDuration", 3600) * 1000,
-                config.getBoolean("report.trip.useIgnition"));
+                config.getBoolean("report.trip.useIgnition"),
+                config.getBoolean("event.motion.processInvalidPositions"));
     }
 
     public static void init(String[] arguments) throws Exception {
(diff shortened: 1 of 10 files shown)


src/org/traccar/Context.java
```diff
@@ -258,7 +258,8 @@ public static TripsConfig initTripsConfig() {
                 config.getLong("report.trip.minimalTripDuration", 300) * 1000,
                 config.getLong("report.trip.minimalParkingDuration", 300) * 1000,
                 config.getLong("report.trip.minimalNoDataDuration", 3600) * 1000,
-                config.getBoolean("report.trip.useIgnition"));
+                config.getBoolean("report.trip.useIgnition"),
+                config.getBoolean("event.motion.processInvalidPositions"));
```
tananaev: Maybe we should rename it to `report.trip.` to be consistent with other parameters?

Abyss777: I'm not sure. Firstly we have another parameter started from "event.motion" - `"event.motion.speedThreshold`, secondly frankly it is not working with trips, just with positions received online.

tananaev: Can we move `event.motion.speedThreshold` to the `TripsConfig` then?

Abyss777: Sure. It will take a time.

Abyss777:
Moved `speedThreshold` to `TripsConfig`, it is good decision, we removed one parameter from some functions in `ReportUtils`

tananaev:
Merged, thanks.

PR #3541, opened 2017-09-21
Teltonika: store command response as KEY_RESULT

Abyss777:
Looks like it was implemented before we started handle command responses in modern way
https://github.com/tananaev/traccar/commit/17744264e2938cf8c87346c175e458cd89a1f3c6#diff-1fcc7f951825555b40d62ec37802158f

I've checked on real device, send command "getparam 800" and got this:

image


tananaev:
Merged, thanks.

tananaev:
By the way, should we just remove `KEY_COMMAND`?

Abyss777:
It is used in 4 other protocols, and I'm not sure if it has the same meaning there.

tananaev:
I suspect it is, but we can leave it for now and remove eventually when we revise those protocols.

Issue #3544, opened 2017-09-22
Teltonika IO decoding


Abyss777:
I think it is more convenient to discuss and collect it here
I've collected IO meanings from some documentation, last column from official for FMBXXX devices http://teltonika.lt/?media_dl=6079 only permanent and eventual IOs.

[Teltonika Summary IO Table.xlsx](https://github.com/tananaev/traccar/files/1323815/Teltonika.Summary.IO.Table.xlsx)

Marked green that totally compatible, marked red that is not compatible from device to device.

tananaev:
Uploaded to Google Docs:

https://docs.google.com/a/traccar.org/spreadsheets/d/1BTbIQ9nlkwqDO1s9AnoighdxyiwCSCCVVEBdi0dOjTU

tananaev:
Sounds good to me.

PR #3547, opened 2017-09-26
Improve Teltonika IO decoding

Abyss777:
Improved IO values decoding according to #3544 

Added a couple of tests from FMB920

More comments in code

### DIFF
--- a/src/org/traccar/model/CellTower.java
+++ b/src/org/traccar/model/CellTower.java
@@ -106,4 +106,10 @@ public class CellTower {
         this.signalStrength = signalStrength;
     }
 
+    public void setOperator(long operator) {
+        String operatorString = String.valueOf(operator);
+        mobileCountryCode = Integer.parseInt(operatorString.substring(0, 3));
+        mobileNetworkCode = Integer.parseInt(operatorString.substring(3));
+    }
+
 }
--- a/src/org/traccar/protocol/TeltonikaProtocolDecoder.java
+++ b/src/org/traccar/protocol/TeltonikaProtocolDecoder.java
@@ -93,7 +93,7 @@ public class TeltonikaProtocolDecoder extends BaseProtocolDecoder {
         }
     }
 
-    private void decodeParameter(Position position, int id, ChannelBuffer buf, int length) {
+    private void decodeOtherParameter(Position position, int id, ChannelBuffer buf, int length) {
         switch (id) {
             case 1:
             case 2:
@@ -199,11 +199,19 @@ public class TeltonikaProtocolDecoder extends BaseProtocolDecoder {
                 position.set(Position.KEY_MOTION, readValue(buf, length, false) == 1);
                 break;
             case 244:
-                position.set("roaming", readValue(buf, length, false) == 1);
+                position.set(Position.KEY_ROAMING, readValue(buf, length, false) == 1);
                 break;
             default:
                 position.set(Position.PREFIX_IO + id, readValue(buf, length, false));
                 break;
+        }
+    }
+
+    private void decodeParameter(Position position, int id, ChannelBuffer buf, int length, int codec) {
+        if (codec == CODEC_GH3000) {
+            decodeGh3000Parameter(position, id, buf, length);
+        } else {
+            decodeOtherParameter(position, id, buf, length);
         }
     }
 
@@ -315,11 +323,7 @@ public class TeltonikaProtocolDecoder extends BaseProtocolDecoder {
         if (BitUtil.check(globalMask, 1)) {
             int cnt = buf.readUnsignedByte();
             for (int j = 0; j < cnt; j++) {
-                if (codec == CODEC_GH3000) {
-                    decodeGh3000Parameter(position, buf.readUnsignedByte(), buf, 1);
-                } else {
-                    decodeParameter(position, buf.readUnsignedByte(), buf, 1);
-                }
+                decodeParameter(position, buf.readUnsignedByte(), buf, 1, codec);
             }
         }
 
@@ -327,11 +331,7 @@ public class TeltonikaProtocolDecoder extends BaseProtocolDecoder {
         if (BitUtil.check(globalMask, 2)) {
             int cnt = buf.readUnsignedByte();
             for (int j = 0; j < cnt; j++) {
-                if (codec == CODEC_GH3000) {
-                    decodeGh3000Parameter(position, buf.readUnsignedByte(), buf, 2);
-                } else {
-                    decodeParameter(position, buf.readUnsignedByte(), buf, 2);
-                }
+                decodeParameter(position, buf.readUnsignedByte(), buf, 2, codec);
             }
         }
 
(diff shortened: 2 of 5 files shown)


src/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -113,15 +113,15 @@ private void decodeParameter(Position position, int id, ChannelBuffer buf, int l
             case 19:
                 position.set("axisZ", readValue(buf, length, true));
                 break;
+            case 21:
+                position.set("gsmLevel", readValue(buf, length, false));
```
tananaev: We usually use `RSSI` key. Is it something different?

src/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -134,15 +134,92 @@ private void decodeParameter(Position position, int id, ChannelBuffer buf, int l
...
+            case 222:
+                if (readValue(buf, length, false) == 1) {
+                    position.set(Position.KEY_ALARM, Position.ALARM_SOS);
+                }
+                break;
+            case 240:
+                position.set(Position.KEY_MOTION, readValue(buf, length, false) == 1);
+                break;
+            case 244:
+                position.set("roaming", readValue(buf, length, false) == 1);
```
tananaev: I believe we can have a dedicated key for this. I recall there are some other protocol that report this information.

src/org/traccar/model/CellTower.java
```diff
@@ -106,4 +106,10 @@ public void setSignalStrength(Integer signalStrength) {
         this.signalStrength = signalStrength;
     }
 
+    public void setOperator(long operator) {
```
Abyss777: Function in case we have operator code (MCC+MNC)

src/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -113,15 +113,15 @@ private void decodeParameter(Position position, int id, ChannelBuffer buf, int l
             case 19:
                 position.set("axisZ", readValue(buf, length, true));
                 break;
+            case 21:
```
Abyss777: It is 1-5 range, not RSSI 0-31. As an option we can multiply it to fit RSSI range and store as RSSI

src/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -113,15 +113,15 @@ private void decodeParameter(Position position, int id, ChannelBuffer buf, int l
...
+                position.set("gsmLevel", readValue(buf, length, false));
+                break;
             case 66:
                 position.set(Position.KEY_POWER, readValue(buf, length, false) * 0.001);
                 break;
             case 67:
                 position.set(Position.KEY_BATTERY, readValue(buf, length, false) * 0.001);
                 break;
-            case 70:
```
Abyss777: It has other meaning for FMB devices

src/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -134,15 +134,92 @@ private void decodeParameter(Position position, int id, ChannelBuffer buf, int l
...
+                position.set(Position.KEY_PDOP, readValue(buf, length, false) * 0.1);
+                break;
+            case 67:
+                position.set(Position.KEY_BATTERY, readValue(buf, length, false) * 0.001);
+                break;
+            case 221:
+                position.set("button", readValue(buf, length, false));
+                break;
+            case 222:
+                if (readValue(buf, length, false) == 1) {
```
Abyss777: Only one alarm reason I decided to use. It is pressing big red button.

src/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -134,15 +134,92 @@ private void decodeParameter(Position position, int id, ChannelBuffer buf, int l
...
                 break;
         }
     }
 
+    private void decodeNetwork(Position position) {
+        long cid = position.getLong(Position.PREFIX_IO + 205);
+        int lac = position.getInteger(Position.PREFIX_IO + 206);
+        if (cid != 0 && lac != 0) {
+            CellTower cellTower = CellTower.fromLacCid(lac, cid);
+            long operator = position.getInteger(Position.KEY_OPERATOR);
```
Abyss777: Should we remove 205, 206 and operator attributes if successfully decoded network?

src/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -269,6 +363,8 @@ private void decodeLocation(Position position, ChannelBuffer buf, int codec) {
             }
         }
 
+        decodeNetwork(position);
```
Abyss777: We can try decode Network only after decoding all IOs. Have to moved code to function, because method to long.

test/org/traccar/ProtocolTest.java
```diff
@@ -257,6 +257,10 @@ private void verifyDecodedPosition(Object decodedObject, boolean checkLocation,
             Assert.assertTrue(attributes.get(Position.KEY_DRIVER_UNIQUE_ID) instanceof String);
         }
 
+        if (attributes.containsKey(Position.KEY_STEPS)) {
```
Abyss777: Added type test for recently added known attribute

src/org/traccar/model/CellTower.java
tananaev: Is it a common way of representing the data? If it's specific to Teltonika, it should be in the decoder.

src/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -233,23 +315,35 @@ private void decodeLocation(Position position, ChannelBuffer buf, int codec) {
...
+                }
             }
         }
 
         // Read 4 byte data
         if (BitUtil.check(globalMask, 3)) {
             int cnt = buf.readUnsignedByte();
             for (int j = 0; j < cnt; j++) {
-                decodeParameter(position, buf.readUnsignedByte(), buf, 4);
+                if (codec == CODEC_GH3000) {
```
tananaev: There is a bit of duplication. I suggest we have 3 methods `decodeGh3000Parameter`, `decodeOtherParameter` and `decodeParameter`. Last one will take `codec` parameter and based on it decide which of the first two methods to call.

src/org/traccar/protocol/TeltonikaProtocolDecoder.java
Abyss777: I noticed that in other protocols RSSI usually value from 0 to 31 even in this GH3000, but here it is value in range 1 - 5.
We can multiply it on `6,2` to fit the range 0-31

src/org/traccar/protocol/TeltonikaProtocolDecoder.java
Abyss777: Yes, it worth to be dedicated key. But only one place I've found with "roam/roaming" words is https://github.com/tananaev/traccar/blob/master/src/org/traccar/protocol/Gl200BinaryProtocolDecoder.java#L267

src/org/traccar/model/CellTower.java
Abyss777: I'm sure it is common way. From Wikipedia:
> A mobile country code (MCC) is used in combination with a mobile network code (MNC) (a combination known as an "MCC/MNC tuple") to uniquely identify a mobile network operator (carrier) using the GSM (including GSM-R), UMTS, and LTE public land mobile networks.

src/org/traccar/protocol/TeltonikaProtocolDecoder.java
tananaev: I don't think we need to match ranges. RSSI is not really standardised for other protocols.

src/org/traccar/protocol/TeltonikaProtocolDecoder.java
tananaev: No I think it's fine to leave them.

tananaev:
Merged, thanks.

PR #3553, opened 2017-09-28
Make user SMTP config more priority than server

### LINKED ISSUE #3552: Email Server-wide Vs Per-user configuration
I've noticed that when server-wide email configuration is done then Per-user don't work. Traccar always sends email using the default, not the user configured SMTP server. Is that the desired behavor?

A user might want to use their own SMTP server, not the default server-wide. Is that possible to change this behavior

Abyss777:
Made user SMTP config more priority and added switch to ignore user config

fix #3552

### DIFF
--- a/src/org/traccar/notification/NotificationMail.java
+++ b/src/org/traccar/notification/NotificationMail.java
@@ -87,13 +87,16 @@ public final class NotificationMail {
     public static void sendMailSync(long userId, Event event, Position position) throws MessagingException {
         User user = Context.getPermissionsManager().getUser(userId);
 
-        Properties properties = getProperties(new PropertiesProvider(Context.getConfig()));
-        if (!properties.containsKey("mail.smtp.host")) {
+        Properties properties = null;
+        if (!Context.getConfig().getBoolean("mail.smtp.ignoreUserConfig")) {
             properties = getProperties(new PropertiesProvider(user));
-            if (!properties.containsKey("mail.smtp.host")) {
-                Log.warning("No SMTP configuration found");
-                return;
-            }
+        }
+        if (properties == null || !properties.containsKey("mail.smtp.host")) {
+            properties = getProperties(new PropertiesProvider(Context.getConfig()));
+        }
+        if (!properties.containsKey("mail.smtp.host")) {
+            Log.warning("No SMTP configuration found");
+            return;
         }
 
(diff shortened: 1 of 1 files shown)


src/org/traccar/notification/NotificationMail.java
```diff
@@ -87,9 +87,9 @@ private static Properties getProperties(PropertiesProvider provider) {
     public static void sendMailSync(long userId, Event event, Position position) throws MessagingException {
         User user = Context.getPermissionsManager().getUser(userId);
 
-        Properties properties = getProperties(new PropertiesProvider(Context.getConfig()));
-        if (!properties.containsKey("mail.smtp.host")) {
-            properties = getProperties(new PropertiesProvider(user));
+        Properties properties = getProperties(new PropertiesProvider(user));
+        if (Context.getConfig().getBoolean("mail.smtp.ignoreUserConfig") || !properties.containsKey("mail.smtp.host")) {
```
tananaev: Seems a bit inefficient to initialise user properties if `ignoreUserConfig` is set to `true`.

Abyss777: I also noticed it, but it become look more complex
```java
        Properties properties;
        if (!Context.getConfig().getBoolean("mail.smtp.ignoreUserConfig")) {
            properties = getProperties(new PropertiesProvider(user));
            if (!properties.containsKey("mail.smtp.host")) {
                properties = getProperties(new PropertiesProvider(Context.getConfig()));
            }
        } else {
            properties = getProperties(new PropertiesProvider(Context.getConfig()));
        }
        if (!properties.containsKey("mail.smtp.host")) {
            Log.warning("No SMTP configuration found");
            return;
        }
```
is it OK? or you see how make simpler?

tananaev: How about something like this:
```java
Properties properties = null;
if (!Context.getConfig().getBoolean("mail.smtp.ignoreUserConfig")) {
    properties = getProperties(new PropertiesProvider(user));
}
if (properties == null || !properties.containsKey("mail.smtp.host")) {
    properties = getProperties(new PropertiesProvider(Context.getConfig()));
}
```

Abyss777:
Of course! Thanks for advice. Replaced the commit.

tananaev:
Merged, thanks.

PR #3556, opened 2017-09-28
Socket servlet Authentication with Authorization Basic

diefferson:
Basic authorization in socket is interesting to consume in another app for example


src/org/traccar/api/AsyncSocketServlet.java
```diff
@@ -22,25 +22,113 @@
...
-                }
+
+                auth(req);
+                return new AsyncSocket(getUserId());
+
             }
         });
     }
 
+    private void auth(ServletUpgradeRequest req) {
```
tananaev: It looks very similar to the code that we have in `SecurityRequestFilter`. Is there a way to avoid duplication?

diefferson: I do not have much experience with Servlet Filters, so I could not add SecurityRequestFilter to this servlet.

tananaev:
Closing as it seems outdated.

PR #3562, opened 2017-10-02
Update GL200 IGN Pattern

fheisenberg:
Fixed sequence odometer, hour meter in IGN pattern

### DIFF
--- a/src/org/traccar/protocol/Gl200TextProtocolDecoder.java
+++ b/src/org/traccar/protocol/Gl200TextProtocolDecoder.java
@@ -234,8 +234,8 @@ public class Gl200TextProtocolDecoder extends BaseProtocolDecoder {
             .expression("[^,]*,")                // device name
             .number("d+,")                       // ignition off duration
             .expression(PATTERN_LOCATION.pattern())
-            .number("(d{5}:dd:dd)?,")            // hour meter
             .number("(d{1,7}.d)?,")              // odometer
+            .number("(d{5}:dd:dd)?,")            // hour meter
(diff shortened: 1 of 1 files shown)


tananaev:
I think the issue is that we already have odometer in the `PATTERN_LOCATION`, so we need to update other patterns to not have it.

Do you have protocol documentation? I have one that doesn't have hour meter at all:

We should probably support that as well.

fheisenberg:
Yes, i have the documentation of the protocol , how can i send you?

tananaev:
You can send to my email. Address is in profile.

fheisenberg:
Done!, i sent you the protocol documentation, i wait for your comments

tananaev:
What about my other comment?

fheisenberg:
I am little confuse, when i said "i wait for your comments" i meant the comments to the email that i send you, because in the email  it was included a excel file where i compare the location patterns in response to the odometer issue

betonme:
Any news here, which implementatiuon is the correct one?

christiankyte:
@tananaev what was blocking this from being merged? I have the formatting documentation and have noticed some changes that are not implemented in the pattern for IGN|F and OBD
https://cloud.ckyte.com/s/j2rSbam9j4GY9NC

christiankyte:
No worries, I've already started a deep dive on GV500MAP variations of the gl200 protocol. It's shocking they would implement a slightly different version with each device. What I have noticed so far is that the selection of the Objects to be sent, via their tool, allows options that break the decoding. Anyways- ill submit a pull request when I finish it up.

tananaev:
Closing as no updates.

PR #3577, opened 2017-10-10
Update case statement Xexun Protocol

kirkryan:
A proposed change to allow for the correct use of value.toLowerCase() handling (case-insensitive switch statement functionality.


tananaev:
Merged, thanks.

Issue #3575, opened 2017-10-09
flespi integration


janbartnitsky:
Hello
Let's discuss the possibility of integrating [flespi](https://flespi.com) platform into traccar.
There are useful features that can be integrated to traccar via flespi [REST API](https://flespi.io/docs).

tananaev:
I would recommend to discuss features separately. Do you have a list of features that you want to integrate? Any technical details on you want to integrate them?

ashchurko:
The simpliest way is to link one flespi channel (by specifying token and channel id) into traccar protocol instance and communicate with it via HTTP REST API. 

But give me few days to look into the architecture and source code of traccar and I will be back with the proposition.

ashchurko:
Anton, please review pull request above by @nadezhdamosunova as a proposed integration between traccar and flespi. The poll method we implemented is different to current tcp/udp connections listeners protocols, so we are open for any feedback.

tananaev:
Flespi has been support for a while.

PR #3582, opened 2017-10-13
initial flespi integtation #3575

nadezhdamosunova:
Here is a flespi integration pull-request: issue #3575
    
To start pulling messages from flespi:
1. Get authorization token: login at flespi panel https://flespi.io and copy token in "Tokens" folder
2. Create one or more channels (with any protocol) at "Channels" tab. Point some devices or simulators to send messages in created flespi channels
3. Put token and channels ids in config file default.xml (keys "flespi.token" & "flespi.channel.ids"). Channels ids can be specified as comma-separated list, e.g. 14,15,78 
    
When application starts it begins pulling messages from specified channels.
Position is parsed  in received messages, and if it's valid and device id is known, then devices' locations are updated.

This is initial basic integration of flespi, conserning only devices' locations. It gives the idea of flespi: to convert different protocols into unified format.
We can parse all messages parameters, implement commands etc as further integration.
  
Anton, please take a look at the code. You probably will suggest some improvements. Looking forward to proceed with the integration.


tananaev:
Before implementing anything I would like to discuss why we need polling API in the first. I was told that flespi supports pushing data as well. That option should work much better with Traccar. I can be implemented as a separate protocol.

ashchurko:
Polling API is better to use to move the processing load of devices communications outside of the traccar. In that case you have all connections processing in the flespi channel and just fetch messages from channel buffer for all devices.

Current implementation of push method - we call it [device streams](https://flespi.com/blog/flespi-streams-overview-telematics-data-and-commands-sent-without-a-hitch) - require device and stream to be registered on the flespi platform - it is additional actions by the user.

Push method of all channel messages from all devices can be implementing by using addiotional adapter - [flespi_receiver](https://github.com/flespi-software/flespi_receiver/) - it poll channel data and can push it via HTTP REST request to separate web-server. Do you think such method is preffered in traccar operation? 

Anyway commands to devices should be sent via outgoing HTTP request from traccar. Will it be possible in latest scenario?

tananaev:
Polling data every 5 seconds seems inefficient. What was the reason for implementing it this way? Is there a plan to implement better API for real time tracking?

Because your code is not part of the standard pipeline, it basically bypasses most of the Traccar functionality. Things like notifications won't work.

ashchurko:
Do you have any suggestions on better API for real time tracking?

It will be possible to bind via MQTT to the flespi platform and receive incoming messages in push-like mode, but MQTT protocol, especially async, is even more complicated on the java then HTTP.

tananaev:
What we have in Traccar is WebSocket connection to push updates live. I'm sure there are many other options as well.

ashchurko:
WebSockets - you mean Traccar to be used initially as client or as server?

I still do not understand the what is the problem with HTTP connections? We talking about server vs server communication. Both are located in datacenters, we do not have any feasible restrictions on the traffic or latency. In normal workflow we have messages from multiple or device buffered on the flespi and need just normal type of moving them to Traccar with acknowledge of each transferred block by Traccar. Two common used transfer protocol are proposed => HTTP request/response sequence and MQTT PUB/SUB. 

So what's wrong with HTTP request/response? Traccar can be used as HTTP server or HTTP client, there is no difference from our side, except less or more to configure. 

We can even send data from flespi to Traccar HTTP server separately per device if you need this. But this will be less optimal of course.

tananaev:
Using Traccar as HTTP server would work.

ashchurko:
Just to be sure: Traccar as HTTP server => can we post messages from multiple devices(tracking units) in one request or it should be separated request per device?
And outgoing commands => will it be possible to use HTTP client on Traccar to POST commands to the flespi?

tananaev:
I think it's possible to post data from several devices in one request, but I would avoid it.

I guess there is no other option for sending commands, so yes.

nadezhdamosunova:
Here is an implementation where Traccar act as a server: it listens http pachets on specified port.
https://github.com/tananaev/traccar/pull/3582/commits/fb82a8167b99f706cb25d828bf1d72424ba98a75

It requires some transmitter from flespi channel, and flespi_receiver can be used as such transmitter.
https://github.com/flespi-software/flespi_receiver

src/org/traccar/ExtendedObjectDecoder.java
```diff
@@ -69,6 +72,16 @@ public void handleUpstream(
                         saveOriginal(o, originalMessage);
                         Channels.fireMessageReceived(ctx, o, e.getRemoteAddress());
                     }
+                } else if (decodedMessage instanceof Map) {
```
tananaev: Why do we need a map?

nadezhdamosunova:
Yes, you are right.
It can be simplified: https://github.com/nadezhdamosunova/traccar/commit/563776604e5acfcde8f83ab779406caa8a069ef4

src/org/traccar/protocol/FlespiProtocol.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2015 Anton Tananaev (anton@traccar.org)
```
tananaev: Please update dates.

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    protected Object decode(
+            Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
+
+        HttpRequest request = (HttpRequest) msg;
+        JsonArray result = Json.createReader(new StringReader(request.getContent().toString(StandardCharsets.UTF_8)))
+                .readArray();
+        List<Position> positions = new LinkedList<>();
+        Log.debug(String.format("messages received msgs_count=%d", result.size()));
```
tananaev: Remove extra logging.

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,116 @@
...
+        if (channel != null) {
+            HttpResponse response = new DefaultHttpResponse(HttpVersion.HTTP_1_1, status);
+            response.headers().add(HttpHeaders.Names.CONTENT_LENGTH, 10);
+            response.setContent(ChannelBuffers.copiedBuffer("Hello namo", CharsetUtil.US_ASCII));
+            channel.write(response);
+        }
+    }
+
+    private void decodePosition(JsonObject msg, Position position) {
+        position.setProtocol("flespi");
```
tananaev: Use protocol name from protocol object. Don't hardcode it here.

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,116 @@
...
+    private void sendResponse(Channel channel, HttpResponseStatus status) {
+        if (channel != null) {
+            HttpResponse response = new DefaultHttpResponse(HttpVersion.HTTP_1_1, status);
+            response.headers().add(HttpHeaders.Names.CONTENT_LENGTH, 10);
+            response.setContent(ChannelBuffers.copiedBuffer("Hello namo", CharsetUtil.US_ASCII));
+            channel.write(response);
+        }
+    }
+
+    private void decodePosition(JsonObject msg, Position position) {
```
tananaev: I would recommend to rename `msg` to `object` or `json` to avoid confusion with original `msg`.

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,116 @@
...
+        position.setProtocol("flespi");
+
+        position.setTime(new Date((long) msg.getJsonNumber("timestamp").doubleValue() * 1000));
+        JsonNumber lat = msg.getJsonNumber("position.latitude");
+        JsonNumber lon = msg.getJsonNumber("position.longitude");
+        position.setLatitude((lat != null && lon != null) ? lat.doubleValue() : 0);
```
tananaev: Why do we need to check `lon` here?

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,116 @@
...
+        position.setLongitude((lat != null && lon != null) ? lon.doubleValue() : 0);
```
tananaev: Same question.

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,116 @@
...
+
+        JsonNumber speed = msg.getJsonNumber("position.speed");
+        position.setSpeed(speed != null ? speed.doubleValue() : 0);
+        if (position.getSpeed() == 111) {
```
tananaev: What does `111` mean? Why is it an alarm?

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        JsonNumber course = msg.getJsonNumber("position.direction");
+        position.setCourse(course != null ? course.doubleValue() : 0);
+
+        JsonNumber altitude = msg.getJsonNumber("position.altitude");
+        position.setAltitude(altitude != null ? altitude.doubleValue() : 0);
+
+        int satellites = msg.getInt("position.satellites", 0);
+        position.setValid(position.getLatitude() != 0 && position.getLongitude() != 0 && satellites >= 3);
```
tananaev: This doesn't look like a correct check for validity.

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,116 @@
...
+        position.set(Position.KEY_SATELLITES, satellites);
+
+        if (msg.getBoolean("alarm.event.trigger", false)) {
```
tananaev: What does this alarm mean?

src/org/traccar/protocol/FlespiProtocol.java
nadezhdamosunova: corrected

src/org/traccar/protocol/FlespiProtocolDecoder.java
nadezhdamosunova: corrected

src/org/traccar/protocol/FlespiProtocolDecoder.java
nadezhdamosunova: renamed to `object`

src/org/traccar/protocol/FlespiProtocolDecoder.java
nadezhdamosunova: corrected

src/org/traccar/protocol/FlespiProtocolDecoder.java
nadezhdamosunova: It was done to ensure both coordinates are present in position.
But it's minor case and can be neglected.
Retained only `lat` check here.

src/org/traccar/protocol/FlespiProtocolDecoder.java
nadezhdamosunova: Retained only `lon` check here.

src/org/traccar/protocol/FlespiProtocolDecoder.java
nadezhdamosunova: Removed (tested notifications).

src/org/traccar/protocol/FlespiProtocolDecoder.java
nadezhdamosunova: Corrected: position validity is taken from "position.valid" field of message json-object.
(true by default)

src/org/traccar/protocol/FlespiProtocolDecoder.java
nadezhdamosunova: "alarm.event.trigger":true field is added in message json if flespi gateway receives "alarm" status from device, e.g.:
https://flespi.com/protocols/galileosky
https://flespi.com/protocols/teltonika

nadezhdamosunova:
Anton! Corrections are made according to the comments. f20b926497a4d61b6fefcdba57c0ebfb23ef5fc2

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,111 @@
...
+        JsonNumber speed = object.getJsonNumber("position.speed");
+        position.setSpeed(speed != null ? speed.doubleValue() : 0);
+
+        JsonNumber course = object.getJsonNumber("position.direction");
+        position.setCourse(course != null ? course.doubleValue() : 0);
+
+        JsonNumber altitude = object.getJsonNumber("position.altitude");
+        position.setAltitude(altitude != null ? altitude.doubleValue() : 0);
+
+        int satellites = object.getInt("position.satellites", 0);
```
tananaev: Why do we need a separate variable here?

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    private void decodePosition(JsonObject object, Position position) {
+        position.setProtocol(protocolName);
+
+        position.setTime(new Date((long) object.getJsonNumber("timestamp").doubleValue() * 1000));
+        JsonNumber lat = object.getJsonNumber("position.latitude");
+        JsonNumber lon = object.getJsonNumber("position.longitude");
+        position.setLatitude(lat != null ? lat.doubleValue() : 0);
```
tananaev: I think it's better to just have `if (lat != null && lon != null)` block for both.

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,111 @@
...
+        position.setLongitude(lon != null ? lon.doubleValue() : 0);
+
+        JsonNumber speed = object.getJsonNumber("position.speed");
+        position.setSpeed(speed != null ? speed.doubleValue() : 0);
```
tananaev: I would recommend to change this and other places to `if` code blocks like this:
```java
if (speed != null) {
    position.setSpeed(speed.doubleValue());
}
```

It's easier to read and no need for fallback zero value.

nadezhdamosunova:
Anton, corresponding corrections are made.

Concerning the question about alarm.
Devices can send alarm state to server: 
Galileosky device sends alarm state in 15th bit of tag 0x40.
Teltonika device - sends priority value (1 byte) in AVL Data; priority values >= 2 mean alarm state of device.
etc.
In all such cases flespi gateway sets "alarm.event.trigger":true field in message json, which means that device reported about alarm state. It can bee used in notifications triggers.

tananaev:
OK, the code looks good now, but can you please add a couple of unit test cases, ideally with data from various different devices and case (e.g. alarm / no alarm).

tananaev:
I don't see any commits or unit tests. Have you pushed?

nadezhdamosunova:
Yeah, I didn't push the commit.
Here it is: ad8ee1702e02e02533e9f20e06af960b2cf3c647

test/org/traccar/ProtocolTest.java
```diff
@@ -98,6 +98,10 @@ protected void verifyAttributes(BaseProtocolDecoder decoder, Object object) thro
         verifyDecodedPosition(decoder.decode(null, null, object), false, true, null);
     }
 
+    protected void verifyAttributesList(BaseProtocolDecoder decoder, Object object) throws Exception {
```
tananaev: Why do you need a new method?

nadezhdamosunova: This is to verify attributes in the case when decoder.decode returns List<Positions>.
I don't persist it to be a separate method. 
I'm ready to reimplement it if you could advise where it's better to place is.

tananaev: Let's try without changing this base class for now. I believe you can use `verifyPositions` method.

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+    private void decodePosition(JsonObject object, Position position) {
+        position.setProtocol(protocolName);
+
+        position.setTime(new Date((long) object.getJsonNumber("timestamp").doubleValue() * 1000));
+        JsonNumber lat = object.getJsonNumber("position.latitude");
+        JsonNumber lon = object.getJsonNumber("position.longitude");
+        if (lat != null && lon != null) {
```
tananaev: By the way, I noticed another thing here. If location is not provided, you should call `getLastLocation`.

nadezhdamosunova:
Unit tests fixed; `getLastLocation` call is added if location not provided
2ef987e1352c9d9483e062ff82fa83d0511a0972

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -0,0 +1,119 @@
...
+import java.io.StringReader;
+import java.net.SocketAddress;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.LinkedList;
+import java.util.Date;
+
+public class FlespiProtocolDecoder extends BaseProtocolDecoder {
+
+    private final String protocolName;
```
tananaev: Why do you need this variable? Just use `getProtocolName()` method.

nadezhdamosunova:
Protocol name fixed: 8e14763c45f2a10ac198d311e770282387cd9070
Thank you!

tananaev:
Merged, thanks.

ashchurko:
Anton, we have implemented direct streaming from flespi platform, so it is possible to interconnect flespi and traccar directly (with HTTP push-like streaming from flespi). Here is some more information: https://flespi.com/blog/streams-now-pushing-as-well-as-pulling

We will integrate more parameters from flespi messages into traccar soon.

tananaev:
Great, thanks.

PR #3589, opened 2017-10-19
Implement buffered/delayed commands

Abyss777:
- If device is not online commands put in by device queues
- When device becomes online again traccar send all queued commands.

Queues are in memory only, lost if server restarted.

More comments in code

### DIFF
--- a/src/org/traccar/api/resource/CommandResource.java
+++ b/src/org/traccar/api/resource/CommandResource.java
@@ -62,15 +62,19 @@ public class CommandResource extends ExtendedObjectResource<Command> {
         Context.getPermissionsManager().checkReadonly(getUserId());
         long deviceId = entity.getDeviceId();
         long id = entity.getId();
+        boolean sent;
         if (deviceId != 0 && id != 0) {
             Context.getPermissionsManager().checkPermission(Command.class, getUserId(), id);
             Context.getPermissionsManager().checkDevice(getUserId(), deviceId);
             Context.getPermissionsManager().checkUserDeviceCommand(getUserId(), deviceId, id);
-            Context.getCommandsManager().sendCommand(id, deviceId);
+            sent = Context.getCommandsManager().sendCommand(id, deviceId);
         } else {
             Context.getPermissionsManager().checkLimitCommands(getUserId());
             Context.getPermissionsManager().checkDevice(getUserId(), deviceId);
-            Context.getCommandsManager().sendCommand(entity);
+            sent = Context.getCommandsManager().sendCommand(entity, deviceId);
+        }
+        if (!sent) {
+            return Response.accepted(entity).build();
         }
         return Response.ok(entity).build();
     }
--- a/src/org/traccar/database/CommandsManager.java
+++ b/src/org/traccar/database/CommandsManager.java
@@ -22,9 +22,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
-import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentLinkedQueue;
 
 import org.traccar.BaseProtocol;
 import org.traccar.Context;
@@ -35,7 +33,7 @@ import org.traccar.model.Position;
 
 public class CommandsManager  extends ExtendedObjectManager<Command> {
 
-    private final Map<Long, Queue<Command>> deviceQueues = new ConcurrentHashMap<>();
+    private final Map<Long, List<Command>> deviceQueues = new ConcurrentHashMap<>();
 
     public CommandsManager(DataManager dataManager) {
         super(dataManager, Command.class);
@@ -127,20 +125,18 @@ public class CommandsManager  extends ExtendedObjectManager<Command> {
         return result;
     }
 
-    private Queue<Command> getDeviceQueue(long deviceId) {
+    private List<Command> getDeviceQueue(long deviceId) {
         if (!deviceQueues.containsKey(deviceId)) {
-            deviceQueues.put(deviceId, new ConcurrentLinkedQueue<Command>());
+            deviceQueues.put(deviceId, new ArrayList<Command>());
         }
         return deviceQueues.get(deviceId);
     }
 
     public void sendQueuedCommands(ActiveDevice activeDevice) {
-        Queue<Command> deviceQueue = deviceQueues.get(activeDevice.getDeviceId());
+        List<Command> deviceQueue = deviceQueues.get(activeDevice.getDeviceId());
         if (deviceQueue != null) {
-            Command command = deviceQueue.poll();
-            while (command != null) {
+            for (Command command : deviceQueue) {
                 activeDevice.sendCommand(command);
-                command = deviceQueue.poll();
             }
         }
     }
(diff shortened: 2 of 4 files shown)


src/org/traccar/api/resource/CommandResource.java
```diff
@@ -62,15 +62,19 @@ public Response send(Command entity) throws Exception {
...
-            Context.getCommandsManager().sendCommand(id, deviceId);
+            sent = Context.getCommandsManager().sendCommand(id, deviceId);
         } else {
             Context.getPermissionsManager().checkLimitCommands(getUserId());
             Context.getPermissionsManager().checkDevice(getUserId(), deviceId);
-            Context.getCommandsManager().sendCommand(entity);
+            sent = Context.getCommandsManager().sendCommand(entity, deviceId);
+        }
+        if (!sent) {
+            return Response.accepted(entity).build();
```
Abyss777: I decided to notify API client that command was queued. I think code "202" - "Accepted" fits.

src/org/traccar/database/CommandsManager.java
```diff
@@ -31,26 +35,22 @@
...
+    public boolean sendCommand(long commandId, long deviceId) throws Exception {
+        return sendCommand(getById(commandId), deviceId);
     }
 
-    public void sendCommand(long commandId, long deviceId) throws Exception {
-        sendCommand(getById(commandId), deviceId, false);
-    }
-
-    public void sendCommand(Command command, long deviceId, boolean fallbackToText) throws Exception {
+    public boolean sendCommand(Command command, long deviceId) throws Exception {
```
Abyss777: return `true` if command sent, `false` if queued

src/org/traccar/database/CommandsManager.java
```diff
@@ -71,32 +71,26 @@ public void sendCommand(Command command, long deviceId, boolean fallbackToText)
             if (activeDevice != null) {
                 activeDevice.sendCommand(command);
             } else {
-                if (fallbackToText) {
-                    command.setTextChannel(true);
-                    sendCommand(command, deviceId, false);
-                } else {
-                    throw new RuntimeException("Device is not online");
-                }
+                sent = !getDeviceQueue(deviceId).add(command);
```
Abyss777: `add` always return true if not exception

src/org/traccar/database/CommandsManager.java
```diff
@@ -71,32 +71,26 @@ public void sendCommand(Command command, long deviceId, boolean fallbackToText)
...
-                if (lastPosition != null) {
-                    BaseProtocol protocol = Context.getServerManager().getProtocol(lastPosition.getProtocol());
-                    if (protocol.getSupportedTextCommands().contains(command.getType())
-                            || online && protocol.getSupportedDataCommands().contains(command.getType())) {
-                        result.add(commandId);
-                    }
-                } else if (command.getType().equals(Command.TYPE_CUSTOM)) {
+            if (lastPosition != null) {
+                BaseProtocol protocol = Context.getServerManager().getProtocol(lastPosition.getProtocol());
+                if ((command.getTextChannel()
```
Abyss777: This construction might looks unclear, can unwrap ternary operator.

src/org/traccar/database/CommandsManager.java
```diff
@@ -133,4 +127,22 @@ public void sendCommand(Command command, long deviceId, boolean fallbackToText)
...
+        }
+        return deviceQueues.get(deviceId);
+    }
+
+    public void sendQueuedCommands(ActiveDevice activeDevice) {
+        Queue<Command> deviceQueue = deviceQueues.get(activeDevice.getDeviceId());
+        if (deviceQueue != null) {
+            Command command = deviceQueue.poll();
+            while (command != null) {
+                activeDevice.sendCommand(command);
```
Abyss777: If exception happen here, command will be lost and rest of queue will be sent when device goes online again.

src/org/traccar/database/ConnectionManager.java
```diff
@@ -122,6 +124,7 @@ public void updateDevice(final long deviceId, String status, Date time) {
                 public void run(Timeout timeout) throws Exception {
                     if (!timeout.isCancelled()) {
                         updateDevice(deviceId, Device.STATUS_UNKNOWN, null);
+                        activeDevices.remove(deviceId);
```
Abyss777: Assuming device is not active when its state unknown, to not send commands in emptiness.

tsmgeek:
Is there to be any consideration if commands could have a time limit or if a user was to send the same command multiple times.

It may be an idea to put these commands in a DB to persist a restart, there are asset trackers which may have 24h intervals, a restart will loose any pending commands for these.

src/org/traccar/database/CommandsManager.java
```diff
@@ -31,26 +35,22 @@
 
 public class CommandsManager  extends ExtendedObjectManager<Command> {
 
-    private boolean fallbackToText;
+    private final Map<Long, Queue<Command>> deviceQueues = new ConcurrentHashMap<>();
```
tananaev: I know queue sounds like a right choice here, but really we can as well use regular list here. As soon as device is online we send all commands anyway.

src/org/traccar/database/CommandsManager.java
```diff
@@ -133,4 +127,22 @@ public void sendCommand(Command command, long deviceId, boolean fallbackToText)
...
+            deviceQueues.put(deviceId, new ConcurrentLinkedQueue<Command>());
+        }
+        return deviceQueues.get(deviceId);
+    }
+
+    public void sendQueuedCommands(ActiveDevice activeDevice) {
+        Queue<Command> deviceQueue = deviceQueues.get(activeDevice.getDeviceId());
+        if (deviceQueue != null) {
+            Command command = deviceQueue.poll();
+            while (command != null) {
```
tananaev: Regular for each loop should work here as well. And is simpler.

tananaev:
@tsmgeek, it has been considered and for now we decided to go with simpler solution and in-memory buffer.

Abyss777:
You are right. `ArrayList` is enough.

src/org/traccar/database/CommandsManager.java
tananaev: Just noticed this one. Please unwrap it.

Abyss777:
Done

tananaev:
I think I just found a bug. You have replaced queue poll with a for each loop, but now you don't clear the buffer.

Abyss777:
Oh, sorry.
But now I see the problem, if we clear list after loop, it might cause infinite resend whole list in case exception. If we are going to remove entities in loop, we have to do it with help of `iterator` it will not look simple as we expected.

tananaev:
You are right. Let's revert back to the queue.

Abyss777:
Reverted.

tsmgeek:
I'm just going to put a side case in, device is listed as connected but has just disappeared ie network just failed but socket has yet to close.
Would the message be queued or lost, I guess there is no confirmation of reception by device, or will it work on if an ACK was received from TCP packet, UDP another issue.

tananaev:
Merged, thanks.

tsmgeek:
Will the API return a flag to say the command was queued?

Abyss777:
@tsmgeek 
It will return 202 code

PR #3597, opened 2017-10-21
Added new Parser.DateTimeFormat, HMSno0

jon-stumpf:
I added a new date/time format to deal with the fact that the Cradlepoint protocol provides time with no leading zeros.  I didn't like having the special case in the protocol decoder.

Even though there is only one case that does this, no doubt another protocol may do the same thing in the future and it is best to have all the date/time logic in one place so the protocol decoders can look similar.


tananaev:
I don't think we need this. It's basically a workaround for poorly designed protocol.

PR #3599, opened 2017-10-21
Updated eelink protocol

tsmgeek:
Updated encoder/decoder and tests to parse V2 protocol.
Multiple BSS & BSSID, device status with report packets and many other fixes.


tsmgeek:
I have had this running now for several weeks without issues.

Issue #3591, opened 2017-10-19
Computed Attributes - unable remove attribute


tsmgeek:
This in relation to the JEXL in #3550

I have looked at the code and from my understanding there is no way to unset an attribute, I may be mistaken though.
If you return null it leaves it as is, if you return false and type is String then it just becomes a string, for other types it gets cast.
I propose a way to return a value that means to unset the attr.
Obviously this presents an issue with Boolean type as how do you differentiate between true/false/null while also allowing unset.
What about using the return string "unset" this would remove the attribute completely and would work across all datatypes.

tananaev:
I believe `null` should mean unset. I don't see any problem with boolean.

tananaev:
@Abyss777, what do you think?

tsmgeek:
https://github.com/tananaev/traccar/blob/master/src/org/traccar/processing/ComputedAttributesHandler.java#L99-L114
Null is just ignored and left as is.

Boolean return value would be fine for a string/number, but for a Boolean data type it would not work as false may be valid vs unset.

I need to specifically unset the 'alarm' attr.

tananaev:
I still don't understand your point about boolean. False is false, null is null. I said null should unset the value.

tsmgeek:
If you return null from jexl it does not change anything, there is a test to see if result is null and just returns position as it went in.

Look at L99. It seems if your expression returns null then the result is a literal null.

tananaev:
Yes, that's why I'm proposing to unset it if the result is null. What I don't understand is what the problem with boolean.

tsmgeek:
While it may be a good idea to use null, it would break any expressions that people have already written returning null when they do not want to change anything.
If you did use null to unset then how would someone make sure the value was not unset but no nothing to it, just pass the param back i assume, main point is backwards compatibility with existing installed base.
My problem with Boolean is based on the assumption that null is not changed to mean unset, leaving just True/False, if that is the case what do you use to unset.

tananaev:
Yes, if you want to keep parameter you can use existing value.

Abyss777:
Hello
The idea of computed attributes is not correct existed reported by device, but create new one based on reported. As I understand philosophy of traccar we decoding "known" attributes in common way and they are standardized, it is not needed correct them.

I believe raw data should be leaved as is. Why it is needed to delete any attribute? Let them lay in DB.

tananaev:
[...]

PR #3615, opened 2017-10-30
Alter ComputedAttributes to remove attribs

tsmgeek:
#3591 

- change ComputedAttributes to handle null as unset of attribute
- add ability to set CopyAttributes to all to copy all

### DIFF
--- a/src/org/traccar/processing/ComputedAttributesHandler.java
+++ b/src/org/traccar/processing/ComputedAttributesHandler.java
@@ -103,7 +103,7 @@ public class ComputedAttributesHandler extends BaseDataHandler {
                             case "number":
                                 if (attribute.getAttribute().equals(false)) {
                                     position.getAttributes().remove(attribute.getAttribute());
-                                }else{
+                                } else {
                                     position.getAttributes().put(attribute.getAttribute(), (Number) result);
                                 }
                                 break;
@@ -113,14 +113,14 @@ public class ComputedAttributesHandler extends BaseDataHandler {
                             default:
                                 if (attribute.getAttribute().equals(false)) {
                                     position.getAttributes().remove(attribute.getAttribute());
-                                }else{
+                                } else {
                                     position.getAttributes().put(attribute.getAttribute(), result.toString());
                                 }
                         }
                     } catch (ClassCastException error) {
                         Log.warning(error);
                     }
-                }else{
+                } else {
                     position.getAttributes().remove(attribute.getAttribute());
                 }
             }
--- a/src/org/traccar/processing/ComputedAttributesHandler.java
+++ b/src/org/traccar/processing/ComputedAttributesHandler.java
@@ -96,21 +96,32 @@ public class ComputedAttributesHandler extends BaseDataHandler {
                 } catch (JexlException error) {
                     Log.warning(error);
                 }
+
                 if (result != null) {
                     try {
                         switch (attribute.getType()) {
                             case "number":
-                                position.getAttributes().put(attribute.getAttribute(), (Number) result);
+                                if (attribute.getAttribute().equals(false)) {
+                                    position.getAttributes().remove(attribute.getAttribute());
+                                }else{
+                                    position.getAttributes().put(attribute.getAttribute(), (Number) result);
+                                }
                                 break;
                             case "boolean":
                                 position.getAttributes().put(attribute.getAttribute(), (Boolean) result);
                                 break;
                             default:
-                                position.getAttributes().put(attribute.getAttribute(), result.toString());
+                                if (attribute.getAttribute().equals(false)) {
+                                    position.getAttributes().remove(attribute.getAttribute());
+                                }else{
+                                    position.getAttributes().put(attribute.getAttribute(), result.toString());
+                                }
                         }
                     } catch (ClassCastException error) {
                         Log.warning(error);
                     }
+                }else{
+                    position.getAttributes().remove(attribute.getAttribute());
                 }
             }
         }
--- a/src/org/traccar/processing/ComputedAttributesHandler.java
+++ b/src/org/traccar/processing/ComputedAttributesHandler.java
@@ -95,14 +95,15 @@ public class ComputedAttributesHandler extends BaseDataHandler {
                     result = computeAttribute(attribute, position);
                 } catch (JexlException error) {
                     Log.warning(error);
+                    return position;
                 }
 
                 if (result != null) {
                     try {
                         switch (attribute.getType()) {
                             case "number":
-                                if (attribute.getAttribute().equals(false)) {
-                                    position.getAttributes().remove(attribute.getAttribute());
+                                if (attribute.getAttribute().equals(false) || attribute.getAttribute().equals(true)) {
+                                    throw new IllegalArgumentException("Number Attribute Type was returned Boolean instead of Number");
                                 } else {
                                     position.getAttributes().put(attribute.getAttribute(), (Number) result);
                                 }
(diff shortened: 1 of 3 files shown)


src/org/traccar/processing/ComputedAttributesHandler.java
```diff
@@ -96,21 +96,32 @@ protected Position handlePosition(Position position) {
                 } catch (JexlException error) {
                     Log.warning(error);
                 }
+
                 if (result != null) {
                     try {
                         switch (attribute.getType()) {
                             case "number":
-                                position.getAttributes().put(attribute.getAttribute(), (Number) result);
+                                if (attribute.getAttribute().equals(false)) {
```
tananaev: How can it be equal to `false` if it's a number? Am I missing something? Same question for default case as well.

src/org/traccar/processing/CopyAttributesHandler.java
```diff
@@ -37,7 +37,12 @@ protected Position handlePosition(Position position) {
         if (attributesString.isEmpty()) {
             attributesString = Position.KEY_DRIVER_UNIQUE_ID;
         } else {
-            attributesString += "," + Position.KEY_DRIVER_UNIQUE_ID;
+            if (attributesString.equals("all")) {
```
tananaev: This is not related to computed attributes in any way. Also, setting attributes would clear existing attributes, so I don't think it's correct. Also, I don't think using special keyword is a good idea. We need some different way of copying all attributes. For example, if `copyAttributes` is not specified it would mean copy app.

tsmgeek: No its not related. Just happened I made the changes on the same branch by accident.

Line 36, 'if (attributesString.isEmpty()) {'
My understanding of 'lookupAttributeString' is that the default is "" empty. and would match line 36. Is that incorrect?

src/org/traccar/processing/ComputedAttributesHandler.java
tsmgeek: Its a catch, in case user returns false then it handles it like null else there may be a cast error trying to convert a boolean to Number.

src/org/traccar/processing/CopyAttributesHandler.java
tananaev: I'm proposing to use absence of the parameter, not empty value.

src/org/traccar/processing/ComputedAttributesHandler.java
tananaev: Why would it be boolean value?

tsmgeek: Im not sure JEXL cares, it will return false/null as the expression is written. Its just a failsafe check in-case the expression is written in that way. More belts and braces approach rather than allow exceptions to happen. For string type if expression returns literal false it would be cast to string, here it tests the literal and handles it while if the user wants text false they should write literal string "false".

src/org/traccar/processing/CopyAttributesHandler.java
tsmgeek: More like this, if null then do as previous isEmpty, if specified but empty then its all else its limited?
```
        String attributesString = Context.getDeviceManager().lookupAttributeString(
                position.getDeviceId(), "processing.copyAttributes", null, true);
        if (attributesString == null) {
            attributesString = Position.KEY_DRIVER_UNIQUE_ID;
        } else {
            if (attributesString.isEmpty()) {
                position.setAttributes(last.getAttributes());
                return position;
            } else {
                attributesString += "," + Position.KEY_DRIVER_UNIQUE_ID;
            }
        }
```

tananaev: Something like it.

src/org/traccar/processing/ComputedAttributesHandler.java
tananaev: I don't agree with this approach. If expression returns wrong type it should fail with exception.

tsmgeek: Ok ile put in exceptions instead.

Abyss777:
Want to point two things...
1. There is a different `null`: intentional and because of exception. We should delete attribute only if it is intentional I believe
2. We are copying attributes only if they are not exist in current position. I think `"all"` logic should have the same behavior.

tsmgeek:
@Abyss777 
1. good catch, ile amend to handle the exception properly
2. I see, yup, it could cause error currently if last is null, ile correct

tsmgeek:
Are there anything remaining issues on this PR that I need to correct?

src/org/traccar/processing/ComputedAttributesHandler.java
```diff
@@ -95,22 +95,36 @@ protected Position handlePosition(Position position) {
...
                     Log.warning(error);
+                    return position;
                 }
+
                 if (result != null) {
                     try {
                         switch (attribute.getType()) {
                             case "number":
-                                position.getAttributes().put(attribute.getAttribute(), (Number) result);
+                                if (attribute.getAttribute().equals(false) || attribute.getAttribute().equals(true)) {
```
tananaev: Can't we just check type?

src/org/traccar/processing/ComputedAttributesHandler.java
```diff
@@ -95,22 +95,36 @@ protected Position handlePosition(Position position) {
...
+                                } else {
+                                    position.getAttributes().put(attribute.getAttribute(), (Number) result);
+                                }
                                 break;
                             case "boolean":
                                 position.getAttributes().put(attribute.getAttribute(), (Boolean) result);
                                 break;
                             default:
-                                position.getAttributes().put(attribute.getAttribute(), result.toString());
+                                if (attribute.getAttribute().equals(false) || attribute.getAttribute().equals(true)) {
```
tananaev: Same here. Also, why do we need to check this? What's the use case?

src/org/traccar/processing/CopyAttributesHandler.java
```diff
@@ -19,6 +19,7 @@
 import org.traccar.BaseDataHandler;
 import org.traccar.Context;
 import org.traccar.model.Position;
+import org.apache.commons.lang3.StringUtils;
```
tananaev: Please don't use this library. We don't have explicit dependency on it.

src/org/traccar/processing/CopyAttributesHandler.java
```diff
@@ -31,21 +32,29 @@ private Position getLastPosition(long deviceId) {
...
-        }
+
         if (last != null) {
+            String attributesString = Context.getDeviceManager().lookupAttributeString(
+                    position.getDeviceId(), "processing.copyAttributes", null, true);
+
+            if (attributesString == null) {
+                attributesString = Position.KEY_DRIVER_UNIQUE_ID;
+            } else {
+                if (attributesString.isEmpty()) {
```
tananaev: Are we copying all attributes when the parameter is empty? Doesn't make much sense to me at the moment. Did we decide to do it like this for some reason?

tananaev: Also, we don't need attribute string if we want to copy everything.

tsmgeek:
Not sure why CI is failing on not being able to load StringJoiner, I am unsure another way to do this.

`[ERROR] /home/travis/build/tananaev/traccar/src/org/traccar/processing/CopyAttributesHandler.java:[19,17] cannot find symbol
  symbol:   class StringJoiner
  location: package java.util
[ERROR] /home/travis/build/tananaev/traccar/src/org/traccar/processing/CopyAttributesHandler.java:[40,21] cannot find symbol
  symbol:   class StringJoiner
  location: class org.traccar.processing.CopyAttributesHandler
[ERROR] /home/travis/build/tananaev/traccar/src/org/traccar/processing/CopyAttributesHandler.java:[40,48] cannot find symbol
  symbol:   class StringJoiner
  location: class org.traccar.processing.CopyAttributesHandler`

src/org/traccar/processing/ComputedAttributesHandler.java
tsmgeek: This was to make sure we are not casting a literal false/true into a number/string

src/org/traccar/processing/CopyAttributesHandler.java
tananaev: I would make sense to use "null" value to indicate copying of all attributes.

src/org/traccar/processing/ComputedAttributesHandler.java
tananaev: It doesn't answer my question. Why don't you just check with `instaceof`?

tananaev:
Don't use `StringJoiner`. Just write regular Java code to join string.

tananaev:
This was implemented recently.

PR #3617, opened 2017-10-30
Map "Math" to Computed Attributes context

Abyss777:
It would be useful to have `Math` functions in Computed Attributes. Class registered as function namespaces http://commons.apache.org/proper/commons-jexl/reference/syntax.html#Functions

https://www.traccar.org/forums/topic/power-math-operation-in-computed-attributes/


tananaev:
Merged, thanks.

PR #3624, opened 2017-11-04
Support for custom params on EventForwarder (support for telegram, pushover, etc)

Shinryuken:
Hi,

i implemented a new modality when sending event messages inside EventForwarder, now you can send requests as a "form-urlencoded", this means that the actual payload will be sent as a form parameter (the parameter name is configurable), this strategy will let you to add another parameters inside the form body message too.

You can add new custom form params almost on the same way as the custom headers are inserted (event.forward.header), the only difference is that you will use the equals "=" char instead of ":" as key/value separator. 

You can enable this mode using the option: "event.forward.payloadAsParamMode", when using this mode you can use the other 2 configuration elements, "event.forward.paramMode.payloadParamName" if you want to change the form name param of the payload (default is "payload") and "event.forward.paramMode.additionalParams" to specify additional key/value form params.

I added an option to prettyPrint the payload messages too, you can enable it using the global option 'event.forward.prettyPrintedPayload'.

This is an example (using all the new params) that lets you integrate traccar server with the pushover platform.

**Update: options updated**
```
	<entry key='status.timeout'>60</entry>
	<entry key='event.statusHandler'>true</entry>
	<entry key="event.forward.enable">true</entry>
	<entry key='event.forward.url'>https://api.pushover.net/1/messages.json</entry>

	<entry key='mapper.prettyPrintedJson>true</entry>
	<entry key='event.forward.payloadAsParamMode'>true</entry>
	<entry key='event.forward.paramMode.payloadParamName'>message</entry>
	<entry key='event.forward.paramMode.additionalParams'>
		token=YOUR_TOKEN_ID
		user=YOUR_USER_ID
	</entry>
```

I tested the new options and they work pretty well.

Tried to make the change as clean as possible and avoiding code duplications. This mode does not change or affects the actual behaviour of the module.

Let me know what do you think.


tananaev:
Can you please some examples of what payload looks like.

Also, I think if we include pretty formatter option, it should be global.

Shinryuken:
Hi tananaev,

yes sure, this is how an http request looks like:

```
Client 127.0.0.1:51317 accepted at 12:16:12 AM
12:16:12 AM received [POST / HTTP/1.1
Content-Type: application/x-www-form-urlencoded; charset=UTF-8
Content-Length: 987
Connection: keep-alive
Host: localhost:9000
Accept: */*
User-Agent: AHC/1.0

user=342323424&token=kfysngfy53dg&message=%7B%0D%0A++%22event%22+%3A+%7B%0D%0A++++%22id%22+%3A+233%2C%0D%0A++++%22attributes%22+%3A+%7B+%7D%2C%0D%0A++++%22deviceId%22+%3A+2%2C%0D%0A++++%22type%22+%3A+%22deviceOnline%22%2C%0D%0A++++%22serverTime%22+%3A+%222017-11-05T23%3A16%3A11.993%2B0000%22%2C%0D%0A++++%22positionId%22+%3A+0%2C%0D%0A++++%22geofenceId%22+%3A+0%0D%0A++%7D%2C%0D%0A++%22device%22+%3A+%7B%0D%0A++++%22id%22+%3A+2%2C%0D%0A++++%22attributes%22+%3A+%7B+%7D%2C%0D%0A++++%22name%22+%3A+%22tk%22%2C%0D%0A++++%22uniqueId%22+%3A+%2286%22%2C%0D%0A++++%22status%22+%3A+%22online%22%2C%0D%0A++++%22lastUpdate%22+%3A+%222017-11-04T17%3A34%3A36.429%2B0000%22%2C%0D%0A++++%22positionId%22+%3A+185%2C%0D%0A++++%22groupId%22+%3A+0%2C%0D%0A++++%22geofenceIds%22+%3A+%5B+%5D%2C%0D%0A++++%22phone%22+%3A+%22%22%2C%0D%0A++++%22model%22+%3A+%22%22%2C%0D%0A++++%22contact%22+%3A+%22%22%2C%0D%0A++++%22category%22+%3A+null%0D%0A++%7D%0D%0A%7D]
```

You could use this online tool to decode it:

h**ps://meyerweb.com/eric/tools/dencoder/

Regarding the pretty formatter, yes i agree, i did not made a global change because i was not sure if it could break something. We could keep 2 mappers globally (normal/prettyPrinted), use the pretty one when it could improve data readability (using atomic configuration options) and keep the other one for internal transformations. Or we could just modify directly the behaviour of the single global mapper. I don't have enough knowledge of the server atm to see which one could be a better solution.

tananaev:
So you are actually sending JSON as a message. It's not really very user-friendly.

Shinryuken:
By the way, i was checking new messages on my phone and i thought " why not just use **telegram** to receive messages?"

Wow, it took me 5 minutes to create a bot, it's ridiculously simple, after that you could use a configuration like this one:

```
	<entry key='event.statusHandler'>true</entry>
	<entry key="event.forward.enable">true</entry>
	<entry key='event.forward.url'>https://api.telegram.org/bot123:456789/sendMessage</entry>

	<entry key='event.forward.prettyPrintedPayload'>true</entry>
	<entry key='event.forward.payloadAsParamMode'>true</entry>
	<entry key='event.forward.paramMode.payloadParamName'>text</entry>
	<entry key='event.forward.paramMode.additionalParams'>
		chat_id=123456
	</entry>
```

It works, maybe we could use another format for messages

Updating the title.

tananaev:
I think the biggest problem is the format. I don't think end users would be interested in JSON messages.

Shinryuken:
Yes, i agree, is there something already implemented on server that we could use to format messages?, if not, we could start to define a new message format, maybe configurable using verbosity levels. We could start from the sms message provided by some trackers:

```
Power alarm!
lat:77.739288
long:12.977558
speed:0.00
T:17/09/23 16:09
http://maps.google.com/maps?f=q&q=12.977558,77.739288&z=16
```

tananaev:
Yes, there is already a system for email and SMS notifications that uses templates.

Shinryuken:
NotificationFormatter, i will give a look on it

Shinryuken:
- Added a new dedicated method on NotificationFormatter for EventForwarder messages
- Added new 'forwarder' templates (copy from the sms ones)
- Some refactoring
- New instance for pretty printer on configuration (used only by json payloads)

Made some tests and i'm receiving messages, but there is a problem, there is no User on eventForwarded messages so all the fields that are crafted using User configuration or data could be not available, i put for testing the first one available for the device but this **should be fixed somehow**, do you have any suggestions?

P.s. I prepared the EventForwarder class to be splitted, this could be the next step, leaving common logic on this class and creating one new class for each sending mode.

Shinryuken:
> but there is a problem, there is no User on eventForwarded messages so all the fields that are crafted using User configuration or data could not be available, i put for testing the first one available for the device but this should be fixed somehow

I noticed now that there are already default values for those params (distanceUnit, SpeedUnit, etc), maybe with minor changes it could work, i will give it a look.

tananaev:
I noticed that you have duplicated templates from SMS. I think that instead of duplicating we should make it use SMS templates by default and provide an option for user to customise it.

Shinryuken:
- EventForwarder refactoring: noew thare are 2 event forwarders, one for the actual json payload and another one for the new method using form params, it should be possible to add more in the future.

- User default settings are now accessible when no User is available.

Shinryuken:
@tananaev

It's not clear to me if you want to keep separate templates for the EventForwarder, the server could already provide 2 different types for text messages, "sms" and "forwarder", forwarder templates could be more prolix and sms ones to be more concise, then let the user choose by configuration.

They are both plaintext so you could have "basic" and "verbosed" templates instead of "sms" and "forwarder" and then provide configuration options to choose the verbosity level for each message system (sms, eventForwarder, etc).

The new form params modality for EventForwarder should be ok now.

tananaev:
I have thought about this functionality a bit more and it doesn't really make sense to have it global like this. It would make sense to have push notifications per user like other types of notifications.

Shinryuken:
Taking into account the services that i tested, the params that you insert are linked to an account and are specific, so, if you would like to send notifications by user, that means that you need to keep those specific params by user. For sms messages you have phones, for emails you have mail addresses, for form notifications you would have something more dynamic and not so homogeneous. 

Once you send a message on an external platform you could decide the audience inside the platform itself (broadcast group, a group of accounts, a single person, etc), almost like the service that was already available using json payloads.  

Personally i'm not sure that switching to user notifications could be suitable for this implementation.

tananaev:
I just don't see much use case for this feature then. Basically it's only useful for someone who hosts a server just for himself. Traccar is multi-user platform, so it's not really a good fit, unless we can find some way to make it per-user configuration.

Shinryuken:
I'm still seeing this implementation as a variant of the module that was already present, events forwarded to an external service, i just added the possibility to call these services more dynamically, incrementing the supported platforms/services list.

It could be useful for anyone that is using or wants to use the "event forward" functionality and wanted more features/flexibility, so they will be able to reach services that are not supported by configuration atm.

I don't undestand your point, do you want to migrate the existing EventForwarder to the user configuration too?

tananaev:
Well, your original use case mentions push notifications and telegram which are clearly for end-user notifications. Forwarding is intended for some third party service, not for end user. Those are completely different use cases.

Shinryuken:
I can see your point, well, it's not necessary to be used only for push notifications, i see it on a more generic and abstract way, but yes, you could use it for that purpose.

I think that this implementation could be eventually a good start point for what you have on mind.

Shinryuken:
maybe the "forwarder" templates must be enriched a bit

tananaev:
I'm not sure templates are the right approach here. Templates are designed to display something for end users. I would like to make forwarding more customisable, but don't know if this is a good way.

tsmgeek:
This may be a useful function but slightly limited without the multi-user ability.
Some of the functions added are nice to have in the EventForwarder for more general purposes but also building it as NotificationPusher or NotificationTelegram may be better. Of course the notifications system would need to have ability to handle more than just App/SMS/Mail. If that was plugable so when additional notification options exist it allows them to be selected ie App/SMS/Mail/Telegram/Pusher etc.
This may beyond the scope of this initial commit though but the general idea has promise.

Shinryuken:
@tananaev 
that's a good point, json data format should be available, we could let the user decide, adding an option to choose between json formatted data (default) or use the dedicated templates (seen as plaintext), 
i think that leaving the template option could be a welcome feature, i will make the changes.

@tsmgeek
Your proposal could be very useful to someone that intends to use those specific services but i agree with you that this is not the scope of this addition but as i said before it could be a good starting point for those ideas.

Shinryuken:
- New option available

`<entry key='event.forward.paramMode.useTemplatesForPayload'>true</entry>`

tsmgeek:
I still think that the initial point of EventForwarder was more to pass everything onto another system in a standard format rather than act as a pseudo notification system, there are specific modules for that, you should be able to convert this over into a new notification type easy, ie NotificationWebhooks as this essentially is that. IMO.

Shinryuken:
I think that there was some confusion regarding the pull request, as said before, push notifications were just an use case (i think a very good one). I saw this as a simple change that could improve a lot eventForwarder; now it's able to adapt better when trying to communicate with an external service, rather that making or letting an external service to adapt to traccar, sometimes you don't have the control of the external system, telegram http requests could be an example.

I would like to make a bigger contribution to this project but i just don't have the free time atm to make a deep analysis and/or to go further to implement another kind of notification (I'm pretty sure that that would require at least web interface configuration and database management) and personally i don't like to make poor contributions.

I think that even if traccar system is going to have an explicit Telegram integration (i'm mentioning only this one because it's free) or some kind of "web notifications", this pull could be added, because it makes eventForwarder more versatile.

After this explanation, i would like to say that i removed the possibility to use NotificationFormatter inside EventForwarder (only json will be used), but i left the re-factoring changes, this because I'm looking forward to what you suggested, i think that with some time and with the help of someone else this could be implemented, so if you agree, i could modify this pull until it's ok and create a new one to start discussing on what should be done to implement this new notification system.

Let me know.

src/org/traccar/Context.java
```diff
@@ -413,9 +423,17 @@ public static void init(String[] arguments) throws Exception {
 
     }
 
+    private static EventForwarder getConfiguredEventForwarder() {
```
tananaev: I don't think we need a separate method for this.

src/org/traccar/Context.java
```diff
@@ -413,9 +423,17 @@ public static void init(String[] arguments) throws Exception {
...
+        if (Context.getConfig().getBoolean("event.forward.payloadAsParamMode")) {
+            return new FormTypeEventForwarder();
+        }
+        return new JsonTypeEventForwarder();
+    }
+
     public static void init(IdentityManager testIdentityManager) {
         config = new Config();
         objectMapper = new ObjectMapper();
+        objectWriterPrettyPrinter = objectMapper.writerWithDefaultPrettyPrinter();
```
tananaev: I don't think we need both options at the same time.

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -68,31 +73,52 @@ public static VelocityContext prepareContext(long userId, Event event, Position
...
     }
 
     public static String formatSmsMessage(long userId, Event event, Position position) {
-        VelocityContext velocityContext = prepareContext(userId, event, position);
+        String templatePath = Context.getConfig().getString("sms.templatesPath", "sms");
+
+        return formatterLogic(null, userId, event, position, templatePath);
+    }
+
+    public static String formatForwarderMessage(Event event, Position position) {
```
tananaev: Why do we still have this and template files?

src/org/traccar/reports/ReportUtils.java
```diff
@@ -61,19 +61,19 @@ public static void checkPeriodLimit(Date from, Date to) {
         }
     }
 
-    public static String getDistanceUnit(long userId) {
+    public static String getDistanceUnit(Long userId) {
```
tananaev: Why are we changing everything to `Long` class?

src/org/traccar/Context.java
Shinryuken: Method length would be 154 lines (max allowed is 150)

Shinryuken:
Hi,

i made some changes as referenced on the code comments (and commented the remaining points).

Regarding the "good use case" question, i already explained the intentions and purpose of the changes (and eventually some future changes), i don't have any other explicit use case on my hands.

src/org/traccar/Context.java
Shinryuken: This was already explained:

- I like the idea to keep 2 separate serializers, one for internal use and another one eventually for external communication, new modules could decide to use one or another.

- I prefered this approach because they are 2 different objects, _ObjectWriter_ and  _ObjectMapper_, see EventForwarder.java at line 115

- I did not found a method that could let you to keep an instance of ObjectMapper having a default writer with prettyPrinter.

- The main purpose of having a specific instance of the writer is because .writerWithDefaultPrettyPrinter() will leak a new writer instance each time that is called so i preferred to do it once and keep it.

tananaev: I don't see any good reason to have two. The only reason I can think of to use "pretty" writer is some kind of debugging. In that case we can enable it globally.

src/org/traccar/Context.java
tananaev: You need to find a better way to fix it and split original method into meaningful chunks.

Shinryuken:
- Made the prettyPrinted option global as asked
- Made the refactoring as asked

Changed the pretty printed option name from:

`event.forward.prettyPrintedPayload`

to

`config.mapper.prettyPrintedJson`

src/org/traccar/Context.java
Shinryuken: Done

src/org/traccar/Context.java
Shinryuken: Done

src/org/traccar/notification/NotificationFormatter.java
Shinryuken: Removed, done

src/org/traccar/reports/ReportUtils.java
Shinryuken: Rollbacked, done.

src/org/traccar/Context.java
```diff
@@ -323,6 +325,9 @@ public static void init(String[] arguments) throws Exception {
         objectMapper = new ObjectMapper();
         objectMapper.setConfig(
                 objectMapper.getSerializationConfig().without(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS));
+        if (Context.getConfig().getBoolean("config.mapper.prettyPrintedJson")) {
```
tananaev: Please remove `config.` prefix from. It doesn't really make sense because all parameters are in "config".

src/org/traccar/Context.java
```diff
@@ -349,24 +354,7 @@ public static void init(String[] arguments) throws Exception {
...
-                    geolocationProvider = new OpenCellIdGeolocationProvider(key);
-                    break;
-                case "unwired":
-                    geolocationProvider = new UnwiredGeolocationProvider(url, key);
-                    break;
-                default:
-                    geolocationProvider = new MozillaGeolocationProvider(key);
-                    break;
-            }
+            initLocationResolutionSystem();
```
tananaev: Please rename to `initGeolocationSystem` or even better `initGeolocationModule`.

src/org/traccar/Context.java
```diff
@@ -429,6 +395,59 @@ public static void init(String[] arguments) throws Exception {
...
+            case "unwired":
+                geolocationProvider = new UnwiredGeolocationProvider(url, key);
+                break;
+            default:
+                geolocationProvider = new MozillaGeolocationProvider(key);
+                break;
+        }
+    }
+
+    private static void initEventsSubsystem() {
```
tananaev: Rename to `initEventsSystem` or `initEventsModule` to be consistent with other method.

src/org/traccar/notification/JsonTypeEventForwarder.java
```diff
@@ -0,0 +1,21 @@
...
+import org.traccar.model.Event;
+import org.traccar.model.Position;
+
+import com.ning.http.client.AsyncHttpClient.BoundRequestBuilder;
+
+public class JsonTypeEventForwarder extends EventForwarder {
+
+    @Override
+    protected String getContentType() {
+
```
tananaev: Please remove this line break.

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -68,31 +70,46 @@ public static VelocityContext prepareContext(long userId, Event event, Position
...
     }
 
     public static String formatSmsMessage(long userId, Event event, Position position) {
-        VelocityContext velocityContext = prepareContext(userId, event, position);
+        String templatePath = Context.getConfig().getString("sms.templatesPath", "sms");
+
+        return formatterLogic(null, userId, event, position, templatePath);
+    }
+
+    private static String formatterLogic(VelocityContext vc, Long userId, Event event, Position position,
```
tananaev: I think `formatMessage` would be more appropriate. "Logic" doesn't really explain what the method does.

src/org/traccar/notification/EventForwarder.java
```diff
@@ -15,54 +15,75 @@
...
 import org.traccar.helper.Log;
 import org.traccar.model.Device;
 import org.traccar.model.Event;
 import org.traccar.model.Geofence;
 import org.traccar.model.Position;
 
-import java.nio.charset.StandardCharsets;
-import java.util.HashMap;
-import java.util.Map;
+import com.fasterxml.jackson.core.JsonProcessingException;
```
tananaev: Can you please not re-format the imports to avoid potential merge conflicts.

src/org/traccar/notification/EventForwarder.java
```diff
@@ -15,54 +15,75 @@
...
 
     private String url;
     private String header;
+    private final String payloadParamName;
+    private final String additionalParams;
 
     public EventForwarder() {
         url = Context.getConfig().getString("event.forward.url", "http://localhost/");
         header = Context.getConfig().getString("event.forward.header", "");
+        payloadParamName = Context.getConfig().getString("event.forward.paramMode.payloadParamName", "payload");
```
tananaev: Please shorten these two config keys.

src/org/traccar/notification/EventForwarder.java
Shinryuken: i try to avoid these issues but my ide does it automatically

src/org/traccar/Context.java
Shinryuken: done

src/org/traccar/Context.java
Shinryuken: done

src/org/traccar/Context.java
Shinryuken: done

src/org/traccar/notification/JsonTypeEventForwarder.java
Shinryuken: done

src/org/traccar/notification/NotificationFormatter.java
Shinryuken: done

src/org/traccar/notification/EventForwarder.java
Shinryuken: i thought that could be a good idea to use that name because they indicate that are exlusive params for the 'paramMode' and they will have no effect using the existing 'json' mode (or another one) of 'event.forward', so leaving something like 'event.forward.payloadParamName' could lead to confusion imho.

tananaev: OK, if it's only for paramMode what is it doing in the base class then?

src/org/traccar/notification/FormTypeEventForwarder.java
```diff
@@ -0,0 +1,24 @@
...
+public class FormTypeEventForwarder extends EventForwarder {
+
+    @Override
+    protected String getContentType() {
+        return "application/x-www-form-urlencoded; charset=UTF-8";
+    }
+
+    @Override
+    protected void setContent(Event event, Position position, BoundRequestBuilder requestBuilder) {
+        if (!getAdditionalParams().equals("")) {
```
tananaev: Use `isEmpty` instead.

src/org/traccar/notification/EventForwarder.java
```diff
@@ -27,42 +30,60 @@
...
+        for (String paramLine: paramsLines) {
+            splitedLine = paramLine.split(separator, 2);
+            if (splitedLine.length == 2) {
+                paramsMap.put(splitedLine[0].trim(), Arrays.asList(splitedLine[1].trim()));
+            }
+        }
+        return paramsMap;
+    }
+
+    protected String prepareJsonPayload(Event event, Position position) {
```
tananaev: Why is this in the base class as well?

Shinryuken: it converts event data to json data, so it's used to create all the body content for "JsonMode" and the payload parameter for "paramMode"

src/org/traccar/notification/FormTypeEventForwarder.java
Shinryuken: I just followed the existing convention

header = Context.getConfig().getString("event.forward.header", "");
...
 if (!header.equals("")) {

tananaev: You followed bad example. That place should be fixed as well.

src/org/traccar/notification/EventForwarder.java
tananaev: You reminded me my problem with this feature. Can we just have a pull request for improvements to formatter and context classes? I don't really like sending JSON as encoded form parameter. I don't see any good use case for it.

src/org/traccar/notification/EventForwarder.java
Shinryuken: yes you are right, this no longer makes sense to be on the base class, fixed.

src/org/traccar/notification/EventForwarder.java
Shinryuken: Even spending considerable efforts on discussions and explanations we still have trivial misunderstandings, i changed the course of this implementation about 3 times. 

Anyway, i see your point, i made some changes, please see the main thread.

Shinryuken:
I changed the form type to send the data, now it's using a multipart form, json payload will be sent with the content type "application/json", Tested.

It reflects the standard, it's already supported by the traccar http client library and the most important thing is that if you implement a service and map directly the payload as a complex entity you should be able to decode it automatically using conventional APIs because its content type is explicitly specified and matched.

This does not change the original behaviour, only "low" level data communication has changed.

I removed the class to send urlencoded data because it's not used anymore.

tananaev:
What does the request look like now?

Shinryuken:
This is an example with json prettyPrint enabled, don't consider the "Content-Length" http header because sensitive data was edited.

```
POST / HTTP/1.1
Content-Length: 932
Content-Type: multipart/form-data; boundary=SIC9-pndhELxu_DsAqpExjxgOVkdEn
Connection: keep-alive
Host: localhost:55001
Accept: */*
User-Agent: AHC/1.0

--SIC9-pndhELxu_DsAqpExjxgOVkdEn
Content-Disposition: form-data; name="some_param"
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

33333333
--SIC9-pndhELxu_DsAqpExjxgOVkdEn
Content-Disposition: form-data; name="payload_param"
Content-Type: application/json; charset=UTF-8
Content-Transfer-Encoding: 8bit

{
  "event" : {
    "id" : 111,
    "attributes" : { },
    "deviceId" : 1,
    "type" : "deviceOnline",
    "serverTime" : "2017-11-11T00:00:00.000+0000",
    "positionId" : 0,
    "geofenceId" : 0
  },
  "device" : {
    "id" : 1,
    "attributes" : { },
    "name" : "tt",
    "uniqueId" : "55555555",
    "status" : "online",
    "lastUpdate" : "2017-11-11T00:00:00.000+0000",
    "positionId" : 111,
    "groupId" : 0,
    "geofenceIds" : [ ],
    "phone" : "",
    "model" : "",
    "contact" : "",
    "category" : null
  }
}
--SIC9-pndhELxu_DsAqpExjxgOVkdEn--
```

src/org/traccar/notification/MultiPartEventForwarder.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    protected String getContentType() {
+        return "multipart/form-data";
+    }
+
+    @Override
+    protected void setContent(Event event, Position position, BoundRequestBuilder requestBuilder) {
+
+        if (!additionalParams.equals("")) {
```
tananaev: Please use `isEmpty` and fix any other place you can find that uses anything like this.

src/org/traccar/notification/MultiPartEventForwarder.java
```diff
@@ -0,0 +1,47 @@
...
+import com.ning.http.client.multipart.StringPart;
+
+public class MultiPartEventForwarder extends EventForwarder {
+
+    private final String payloadParamName;
+    private final String additionalParams;
+
+    public MultiPartEventForwarder() {
+        payloadParamName = Context.getConfig().getString("event.forward.paramMode.payloadParamName", "payload");
+        additionalParams = Context.getConfig().getString("event.forward.paramMode.additionalParams", "");
```
tananaev: Why empty string and not null?

src/org/traccar/notification/MultiPartEventForwarder.java
```diff
@@ -0,0 +1,47 @@
...
+            for (Entry<String, List<String>> en : paramsToAdd.entrySet()) {
+                for (String value : en.getValue()) {
+                    requestBuilder.addBodyPart(new StringPart(en.getKey(), value, null, StandardCharsets.UTF_8));
+                }
+            }
+        }
+        requestBuilder.addBodyPart(new StringPart(payloadParamName,
+                                                  prepareJsonPayload(event, position),
+                                                  "application/json",
+                                                  StandardCharsets.UTF_8));
```
tananaev: Please fix indentation here. We never use formatting like this in the code.

src/org/traccar/notification/MultiPartEventForwarder.java
```diff
@@ -0,0 +1,47 @@
...
+        return "multipart/form-data";
+    }
+
+    @Override
+    protected void setContent(Event event, Position position, BoundRequestBuilder requestBuilder) {
+
+        if (!additionalParams.equals("")) {
+            Map<String, List<String>> paramsToAdd = splitParams(additionalParams, "=");
+
+            for (Entry<String, List<String>> en : paramsToAdd.entrySet()) {
```
tananaev: Please use proper name instead of `en`.

src/org/traccar/notification/EventForwarder.java
```diff
@@ -43,26 +47,39 @@ public EventForwarder() {
...
+            requestBuilder.setHeaders(params);
         }
 
-        requestBuilder.setBody(preparePayload(event, position));
+        setContent(event, position, requestBuilder);
         requestBuilder.execute();
     }
 
-    private byte[] preparePayload(Event event, Position position) {
+    protected Map<String, List<String>> splitParams(String params, String separator) {
```
tananaev: Why is this in the base class?

Shinryuken: method name was not so accurate, it splits http headers and request params, renamed

src/org/traccar/notification/MultiPartEventForwarder.java
Shinryuken: fixed

src/org/traccar/notification/MultiPartEventForwarder.java
Shinryuken: done

src/org/traccar/notification/MultiPartEventForwarder.java
Shinryuken: done

src/org/traccar/notification/MultiPartEventForwarder.java
Shinryuken: done

Shinryuken:
Changes made

src/org/traccar/notification/EventForwarder.java
```diff
@@ -20,6 +20,7 @@
 import java.util.Arrays;
 import java.util.List;
 
+import org.apache.commons.lang3.StringUtils;
```
tananaev: Please don't use this. There is no explicit dependency for this library and I don't see any reason for adding dependency just to check for empty string.

src/org/traccar/notification/MultiPartEventForwarder.java
```diff
@@ -1,10 +1,11 @@
 package org.traccar.notification;
 
-import java.nio.charset.StandardCharsets;
+import static java.nio.charset.StandardCharsets.UTF_8;
```
tananaev: Remove static import.

Shinryuken: Done.

src/org/traccar/notification/EventForwarder.java
Shinryuken: Done.

I would like to add my opinion on this,

you need to do 2 checks, null and isEmpty, if you search on the project you will find several occurrences of this kind of checks, it could be simpler using apis.

You could consider to re-add those library apis in the future, they are almost used everywhere, so you will end having them inside your project frequently, personally i don't see why this should be avoided, if you don't start to use them you will never use them and that could lead to repeatedly reinvent the wheel on your project or to start an implementation of your own set of utilities for simple operations; both solutions could be highly ineffective. Specially when you have around 40 Mb of "ready to use" code on your project.

Shinryuken:
Code updated

Shinryuken:
thank you

Here is a recap of the new configuration options:

**mapper.prettyPrintedJson**
_boolean (default: false)_, global option to enable pretty printing on json serializations
	
**event.forward.payloadAsParamMode**
_boolean (default: false)_, enable the new sending mode

**event.forward.paramMode.payloadParamName**
_String (default: "payload")_ lets you specify the name of the payload form param

**event.forward.paramMode.additionalParams**
_String (default: null)_, can be used to send additional form parameters

baerni:
How can I configure a pushover push notification now without event.forward.paramMode.additionalParams and using only event.forward.header?

Fmstrat:
Just adding another question to this. I'm trying to get Pushover to work as a notification service, but I'm not sure what to do given all the conversation in this ticket. Can anyone help out?

PR #3632, opened 2017-11-08
Retry geocoding for trips/stops

Abyss777:
In some cases it is not necessary to revers geocoding every position. It might be enough find out addresses of trips endpoints for reports and allow user check address on demand from web client.

- Moved `AddressFormat` inside `Geocoder`
- Implement synchronous geocoding
- Implement retry geocoding for trips/stops reports
- Implement API for revers geocoding

if `geocoder.enable` - `true` - it will try for all positions, if `report.retryGeocoding` - `true` it will retry find out addresses of trips and stops. If both are false, geocoder is disabled.

I have an initial implementation for web client 
image

image

more comments in code...

### DIFF
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -333,7 +333,7 @@ public final class Context {
 
         identityManager = deviceManager;
 
-        if (config.getBoolean("geocoder.enable") || config.getBoolean("report.retryGeocoding")) {
+        if (config.getBoolean("geocoder.enable")) {
             geocoder = initGeocoder();
         }
 
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -43,6 +43,7 @@ import org.traccar.database.StatisticsManager;
 import org.traccar.database.UsersManager;
 import org.traccar.events.MotionEventHandler;
 import org.traccar.events.OverspeedEventHandler;
+import org.traccar.geocoder.AddressFormat;
 import org.traccar.geocoder.BingMapsGeocoder;
 import org.traccar.geocoder.FactualGeocoder;
 import org.traccar.geocoder.GeocodeFarmGeocoder;
@@ -263,6 +264,41 @@ public final class Context {
                 config.getDouble("event.motion.speedThreshold", 0.01));
     }
 
+    public static Geocoder initGeocoder() {
+        String type = config.getString("geocoder.type", "google");
+        String url = config.getString("geocoder.url");
+        String key = config.getString("geocoder.key");
+        String language = config.getString("geocoder.language");
+
+        String formatString = config.getString("geocoder.format");
+        AddressFormat addressFormat;
+        if (formatString != null) {
+            addressFormat = new AddressFormat(formatString);
+        } else {
+            addressFormat = new AddressFormat();
+        }
+
+        int cacheSize = config.getInteger("geocoder.cacheSize");
+        switch (type) {
+            case "nominatim":
+                return new NominatimGeocoder(url, key, language, cacheSize, addressFormat);
+            case "gisgraphy":
+                return new GisgraphyGeocoder(url, cacheSize, addressFormat);
+            case "mapquest":
+                return new MapQuestGeocoder(url, key, cacheSize, addressFormat);
+            case "opencage":
+                return new OpenCageGeocoder(url, key, cacheSize, addressFormat);
+            case "bingmaps":
+                return new BingMapsGeocoder(url, key, cacheSize, addressFormat);
+            case "factual":
+                return new FactualGeocoder(url, key, cacheSize, addressFormat);
+            case "geocodefarm":
+                return new GeocodeFarmGeocoder(key, language, cacheSize, addressFormat);
+            default:
+                return new GoogleGeocoder(key, language, cacheSize, addressFormat);
+        }
+    }
+
     public static void init(String[] arguments) throws Exception {
 
         config = new Config();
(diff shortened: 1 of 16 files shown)


src/org/traccar/Context.java
```diff
@@ -263,6 +264,41 @@ public static TripsConfig initTripsConfig() {
                 config.getDouble("event.motion.speedThreshold", 0.01));
     }
 
+    public static Geocoder initGeocoder() {
```
Abyss777: Moved in separate function because `init` too long

src/org/traccar/api/resource/ServerResource.java
```diff
@@ -47,4 +48,10 @@ public Response update(Server entity) throws SQLException {
         return Response.ok(entity).build();
     }
 
+    @Path("geocoding")
```
Abyss777: Have not found better place for now.
May be some day we will implement some "Utility" endpoint

src/org/traccar/geocoder/JsonGeocoder.java
```diff
@@ -58,14 +64,14 @@ public void getAddress(
             }
         }
 
-        Context.getAsyncHttpClient().prepareGet(String.format(url, latitude, longitude))
+        Context.getAsyncHttpClient().prepareGet(String.format(Locale.US, url, latitude, longitude))
```
Abyss777: I believe it is a bug. Because if default locale is not standard (like Russian) it formats numbers with comma instead of point and it always fail.

src/org/traccar/api/resource/ServerResource.java
tananaev: I think we can rename path to `geocode`.

src/org/traccar/geocoder/JsonGeocoder.java
tananaev: We actually set locale in `Main` class. I guess it's not set for unit tests.

src/org/traccar/geocoder/JsonGeocoder.java
```diff
@@ -84,6 +90,36 @@ public void onThrowable(Throwable t) {
         });
     }
 
+    @Override
+    public String getAddress(final double latitude, final double longitude) {
```
tananaev: There seems to be quite a bit of duplication. Is there way we can avoid it?

src/org/traccar/Context.java
```diff
@@ -297,39 +333,8 @@ public static void init(String[] arguments) throws Exception {
...
-                    geocoder = new FactualGeocoder(url, key, cacheSize);
-                    break;
-                case "geocodefarm":
-                    geocoder = new GeocodeFarmGeocoder(key, language, cacheSize);
-                    break;
-                default:
-                    geocoder = new GoogleGeocoder(key, language, cacheSize);
-                    break;
-            }
+        if (config.getBoolean("geocoder.enable") || config.getBoolean("report.retryGeocoding")) {
```
tananaev: I think `retryGeocoding` is not a very good name. Also, there is no way to disable geocoding for each location. Maybe instead of `retryGeocoding` we can add a parameter `geocoder.lazy` or something like that?

Abyss777: `geocoder.enable` = `false` disables geocoder for each position.
May be `geocoder.onDemand` ?

tananaev: It's a bit confusing if with `geocoder.enable` you can still get geocoding results. Maybe `geocoder.onRequest`?

Abyss777: Agree, if we were not concern of backward compatibility we could make like this
`geocoder.enable` - main switch
`geocoder.everyPosition` - enables Geocoder pipeline handler
`geocoder.onRequest` - enables geocoding for trips/stops

src/org/traccar/geocoder/JsonGeocoder.java
Abyss777: It looks like that, I can try, but the problem is that some pieces of code inside callbacks and call other callbacks but similar are linear.

src/org/traccar/Context.java
tananaev: I think we don't need 3 options. We just need two:
`geocoder.enable` - main switch
`geocoder.onRequest` - disable geocoding for every position and enable for reports

src/org/traccar/geocoder/JsonGeocoder.java
tananaev: The part that needs to be async can be separated into a method or something.

src/org/traccar/Context.java
Abyss777: I want to cover a case when geocoding in every position was enabled, but address of endpoints is empty by some reason (network or invalid position, or limit reached)

tananaev: Then i suggest:
`geocoder.enable` - main switch
`geocoder.ignorePositions` - disable automated geocoding for every position; disabled by default because `false` is default value
`geocoder.onRequest` - enable geocoding by request

Abyss777:
I've experimented with locales in `GeocoderTest`
It fails with `Locale.ENGLISH` but pass with `Locale.UK` and `Locale.US`
My developer machine is:
```bash
$ cat /etc/lsb-release 
DISTRIB_ID=neon
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="KDE neon User Edition 5.11"
$ java -version
openjdk version "1.8.0_151"
OpenJDK Runtime Environment (build 1.8.0_151-8u151-b12-0ubuntu0.16.04.2-b12)
OpenJDK 64-Bit Server VM (build 25.151-b12, mixed mode)
```

tananaev:
I think we should use `Locale.US`.

Abyss777:
I've combined `getAddress` logic in one function, if `callback` is null we try to do synchronous request.
Changes switches logic as we discussed

tananaev:
Merged, thanks.

Abyss777:
Thanks.
I'm also thinking of implementation retry geocoding every position related to events.
It will allow switch off each position geocoding, but still get addresses in notifications.
What do you think?

tananaev:
Sounds good to me. What about the web part?

Abyss777:
I'll send PR on Monday

renaudallard:
From a end user perspective, I think it would be better if it queries the address directly when the user focus on the device (or point in route/trips) instead of putting a link requesting the address. That way, you still save most of the requests, but you can have addresses in follow mode and you are not requested to click anywhere to show the address. So the end user doesn't see any difference in the web interface, but you still save queries.

tananaev:
@renaudallard, please create a new thread instead of commenting on closed pull request.

PR #3638, opened 2017-11-14
Improve LDAP

Abyss777:
Made it much more flexible.
- Implemented the way with service login and pre lookup user in catalog.
- Fill user fields from catalog
- Added switch to disable internal login (users are still in DB)
- Allow to make user admin if he is in predefined group (on creation stage)

More comments in code

### DIFF
--- a/src/org/traccar/database/LdapProvider.java
+++ b/src/org/traccar/database/LdapProvider.java
@@ -16,36 +16,161 @@
 package org.traccar.database;
 
 import javax.naming.Context;
+import javax.naming.NamingEnumeration;
 import javax.naming.NamingException;
+import javax.naming.directory.Attribute;
 import javax.naming.directory.InitialDirContext;
+import javax.naming.directory.SearchControls;
+import javax.naming.directory.SearchResult;
+
+import org.traccar.Config;
+import org.traccar.helper.Log;
+import org.traccar.model.User;
+
 import java.util.Hashtable;
 
 public class LdapProvider {
 
     private String url;
-    private String context;
+    private String searchBase;
+    private String idAttribute;
+    private String nameAttribute;
+    private String mailAttribute;
+    private String searchFilter;
+    private String adminFilter;
+    private String serviceUser;
+    private String servicePassword;
 
-    public LdapProvider(String url, String context) {
-        this.url = url;
-        this.context = context;
+    public LdapProvider(Config config) {
+        String url = config.getString("ldap.url");
+        if (url != null) {
+            this.url = url;
+        } else {
+            this.url = "ldap://" + config.getString("ldap.server") + ":" + config.getInteger("ldap.port", 389);
+        }
+        this.searchBase = config.getString("ldap.base");
+        this.idAttribute = config.getString("ldap.idAttribute", "uid");
+        this.nameAttribute = config.getString("ldap.nameAttribute", "cn");
+        this.mailAttribute = config.getString("ldap.mailAttribute", "mail");
+        this.searchFilter = config.getString("ldap.searchFilter", "(" + idAttribute + "=:login)");
+        String adminGroup = config.getString("ldap.adminGroup");
+        this.adminFilter = config.getString("ldap.adminFilter");
+        if (this.adminFilter == null && adminGroup != null) {
+            this.adminFilter = "(&(" + idAttribute + "=:login)(memberOf=" + adminGroup + "))";
+        }
+        this.serviceUser = config.getString("ldap.user");
+        this.servicePassword = config.getString("ldap.password");
     }
 
-    public boolean login(String username, String password) {
-
+    private InitialDirContext auth(String accountName, String password) throws NamingException {
         Hashtable<String, String> env = new Hashtable<>();
         env.put(Context.INITIAL_CONTEXT_FACTORY, "com.sun.jndi.ldap.LdapCtxFactory");
         env.put(Context.PROVIDER_URL, url);
 
         env.put(Context.SECURITY_AUTHENTICATION, "simple");
-        env.put(Context.SECURITY_PRINCIPAL, "uid=" + username + "," + context);
+        env.put(Context.SECURITY_PRINCIPAL, accountName);
         env.put(Context.SECURITY_CREDENTIALS, password);
 
+        return new InitialDirContext(env);
+    }
+
+    private boolean isAdmin(String accountName) {
+        if (this.adminFilter != null) {
+            try {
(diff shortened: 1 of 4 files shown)


tananaev:
Merged, thanks. Have you tested it?

src/org/traccar/database/LdapProvider.java
```diff
@@ -16,36 +16,161 @@
...
 
-    public LdapProvider(String url, String context) {
-        this.url = url;
-        this.context = context;
+    public LdapProvider(Config config) {
+        String url = config.getString("ldap.url");
+        if (url != null) {
+            this.url = url;
+        } else {
+            this.url = "ldap://" + config.getString("ldap.server") + ":" + config.getInteger("ldap.port", 389);
```
Abyss777: In most cases it enough to set `ldap.server` but still there is a way to set custom `url`

src/org/traccar/database/LdapProvider.java
```diff
@@ -16,36 +16,161 @@
...
+        }
+        this.searchBase = config.getString("ldap.base");
```
Abyss777: Search base is the way to limit search scope, to not lookup full domain, but some OU only for example
`ou=Users,ou=Branch,dc=domain,dc=local`

src/org/traccar/database/LdapProvider.java
```diff
@@ -16,36 +16,161 @@
...
+        this.idAttribute = config.getString("ldap.idAttribute", "uid");
```
Abyss777: This attribute is used as `login`. For OpenLDAP it is `uid`, for Active Directory it is usually `sAMAccountName`

src/org/traccar/database/LdapProvider.java
```diff
@@ -16,36 +16,161 @@
...
+        this.nameAttribute = config.getString("ldap.nameAttribute", "cn");
```
Abyss777: It is Common Name attribute, usually `cn` for both implementations.

src/org/traccar/database/LdapProvider.java
```diff
@@ -16,36 +16,161 @@
...
+        this.mailAttribute = config.getString("ldap.mailAttribute", "mail");
```
Abyss777: It is email attribute, usually `mail` for both implementations.

src/org/traccar/database/LdapProvider.java
```diff
@@ -16,36 +16,161 @@
...
+        this.searchFilter = config.getString("ldap.searchFilter", "(" + idAttribute + "=:login)");
```
Abyss777: It is customized search filter to pre lookup user, `:login` replaced by searching login, by default it is `(uid=:login)` or `(sAMAccountName=:login)`
Here is couple of examples:
- `(|(sAMAccountName=:login)(mail=:login))` login by account name or email address
- `(&(sAMAccountName=:login)(memberOf:1.2.840.113556.1.4.1941:=CN=Traccar users,OU=Traccar,OU=Security,DC=domain,DC=local)(!(UserAccountControl:1.2.840.113556.1.4.803:=2)))` user must be member of `Traccar users` group or its nested groups (!) and not disabled. Valid for Active Directory

src/org/traccar/database/LdapProvider.java
```diff
@@ -16,36 +16,161 @@
...
+        String adminGroup = config.getString("ldap.adminGroup");
+        this.adminFilter = config.getString("ldap.adminFilter");
```
Abyss777: Filter do determine if user is admin

src/org/traccar/database/LdapProvider.java
```diff
@@ -16,36 +16,161 @@
...
+        if (this.adminFilter == null && adminGroup != null) {
+            this.adminFilter = "(&(" + idAttribute + "=:login)(memberOf=" + adminGroup + "))";
+        }
+        this.serviceUser = config.getString("ldap.user");
```
Abyss777: It is recommended to use full Distinguish Name here (like `cn=traccar,ou=service,dc=domain,dc=local`). But frankly Active Directory accept any like `user`, `DOMAIN\user`, `user@DOMAIN.LOCAL` etc.

src/org/traccar/database/LdapProvider.java
```diff
@@ -16,36 +16,161 @@
...
+            try {
+                InitialDirContext context = initContext();
+                String searchString = new String(adminFilter).replace(":login", accountName);
+                SearchControls searchControls = new SearchControls();
+                searchControls.setSearchScope(SearchControls.SUBTREE_SCOPE);
+                NamingEnumeration<SearchResult> results = context.search(searchBase, searchString, searchControls);
+                if (results.hasMoreElements()) {
+                    results.nextElement();
+                    if (results.hasMoreElements()) {
+                        Log.warning("Matched multiple users for the accountName: " + accountName);
```
Abyss777: This situation usually indicate something misconfigured (filter for example)

src/org/traccar/database/LdapProvider.java
```diff
@@ -16,36 +16,161 @@
...
+                attribute = ldapUser.getAttributes().get(mailAttribute);
+                if (attribute != null) {
+                    user.setEmail((String) attribute.get());
+                } else {
+                    user.setEmail(accountName);
+                }
+            }
+            user.setAdmin(isAdmin(accountName));
+        } catch (NamingException e) {
+            user.setLogin(accountName);
```
Abyss777: Very unlikely situation, because user was just authenticated but we should implement it.

src/org/traccar/database/LdapProvider.java
```diff
@@ -16,36 +16,161 @@
...
+        return user;
+    }
+
+    public boolean login(String username, String password) {
         try {
-            new InitialDirContext(env).close();
-            return true;
+            SearchResult ldapUser = lookupUser(username);
+            if (ldapUser != null) {
+                auth(ldapUser.getNameInNamespace(), password).close();
```
Abyss777: We are try authenticate found user with full Distinguish Name

tananaev:
I have already merged, but those details might be useful in future. By the way it would be nice to write some documentation on how to configure it and how it works.

Abyss777:
Sure. I've tested it with Active Directory (2008 R2 level)

Abyss777:
It was mostly notes for you, for myself and for everebody who wants to understand how code works. There is some amount of documentation to be written, but as you know it is not the most interesting activity :) I'll try to do my best.

PR #3639, opened 2017-11-14
Do geocoding for every events related position

Abyss777:
If `geocoder.onRequest` enabled retry to geocode event related position before sending notification.

Also fixed unnecessary String creation in `LdapProvider`


tananaev:
Merged, thanks.

PR #3643, opened 2017-11-18
T580w

ValeraVi:
Support for [Vjoycar T580W](http://www.vjoycar.com/Mini-GPS-Tracker/Wifi-mini-GPS-tracker-T580w.html) device with GPS, LPS and WiFi positioning in existing tk103 protocol.

We have some discussion about this device some time ago: https://github.com/tananaev/traccar/pull/2837

This is an implementation of protocol for devices of http://www.vjoycar.com/
Some description of protocol are here: http://www.vjoycar.com/File.asp?Id=119
also here: https://goo.gl/8DKadX

According to documentation supported devices are:
T16 T18 T18H TK05 TK10 Tk15 TK20 T12 T4400 T8800 T15400
with WiFi:
T510W, T580W, TK101W, T531W,TK05SDW,TK10SDW,TK15SDW,TK20SDW
List of commands (traccar -> Device) are from user guide from T580W: https://goo.gl/M4ancp
I suspect, that at least latest wifi devices also using same commands.


src/org/traccar/protocol/Tk103Protocol.java
```diff
@@ -40,7 +40,22 @@ public Tk103Protocol() {
                 Command.TYPE_REBOOT_DEVICE,
                 Command.TYPE_SET_ODOMETER,
                 Command.TYPE_ENGINE_STOP,
-                Command.TYPE_ENGINE_RESUME);
+                Command.TYPE_ENGINE_RESUME,
+                //T580W commands:
+                "T580W_positionSingle",
```
tananaev: You need to map commands to standard ones. Applies to all.

src/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -63,6 +66,10 @@ public Tk103ProtocolDecoder(Tk103Protocol protocol) {
             .number("(?:L(x+))?")                // odometer
             .any()
             .number("([+-]ddd.d)?")              // temperature
+            .groupBegin()
+            .number("([+-]?d+.d{1,2}),").optional() // altitude
+            .number("(d+)$").optional()         // number of visible satellites
+            .groupEnd("?")
```
tananaev: Why do you have this group when both items in it are already optional?

src/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -87,6 +94,20 @@ public Tk103ProtocolDecoder(Tk103Protocol protocol) {
...
 
+    private static final Pattern PATTERN_LBSWIFI = new PatternBuilder()
+            .number("(d+),")                     // device id
+            .expression("(.{4}),")               // command
+            .number("(d+),")                     // mcc
+            .number("(d+),")                     // mnc
+            .number("(d+),")                     // lac
+            .number("(d+),")                     // cid
+            .number("(d+),")                     // number of wifi macs
+            .expression("((?:(?:[0-9A-Fa-f]{2}:){5}(?:[0-9A-Fa-f]{2})\\*[-+]?\\d{1,2}\\*\\d{1,2},)*)")
```
tananaev: This should be a number. Also please provide details on what this is. It seems like it can be split into separate items.

src/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -111,38 +132,140 @@ private void decodeType(Position position, String type, String data) {
             case "BO01":
                 position.set(Position.KEY_ALARM, decodeAlarm(data.charAt(0) - '0'));
                 break;
+            // Motion Alert with Location
```
tananaev: Remove all unnecessary comments like this.

src/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -111,38 +132,140 @@ private void decodeType(Position position, String type, String data) {
...
+            case "ZC28":
+            case "DW41":
+            case "DW61":
+                position.set(Position.KEY_ALARM, "badBattery");
+                break;
+            // Fuel Cut On with Location
+            case "ZC2A":
+            case "DW43":
+            case "DW63":
+                position.set(Position.KEY_FUEL_LEVEL, 0.0);
```
tananaev: This and next one doesn't make sense.

src/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -111,38 +132,140 @@ private void decodeType(Position position, String type, String data) {
...
+            case "ZC2B":
+            case "DW44":
+            case "DW64":
+                position.set(Position.KEY_FUEL_LEVEL, 1.0);
+                break;
+            // Fuel Cut Alert with Location:
+            case "ZC2C":
+            case "DW45":
+            case "DW65":
+                position.set(Position.KEY_ALARM, "fuelCut");
```
tananaev: What does this alert mean? How is it different from "Fuel Cut On" above?

src/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -111,38 +132,140 @@ private void decodeType(Position position, String type, String data) {
...
+    private static Map<Long, Position> batteryInfo = new HashMap<Long, Position>();
+
+    public void setLastBatteryPower(Position positionWithCurrentBatteryInfo) {
+        batteryInfo.put(positionWithCurrentBatteryInfo.getDeviceId(), positionWithCurrentBatteryInfo);
+    }
+
+    // It is possible to use "processing.copyAttributes" configuration to have battery level in all events,
+    //  but this will not work if battery events will be filtered somehow and dropped from DB, so it will be no
+    //  event in DB with latest battery level. So, ensure all our events have latest battery level.
+    public void getLastBatteryPower(Position position) {
```
tananaev: This has to be removed. Copy attributes functionality covers it.

src/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -190,11 +320,73 @@ private Position decodeNetwork(Channel channel, SocketAddress remoteAddress, Str
...
+
+        decodeType(position, parser.next(), "0");
+
+        getLastLocation(position, null);
+
+        getLastBatteryPower(position);
+
+        Network network = new Network();
+
+        // Parse LBS
```
tananaev: Please remove all unnecessary comments like this.

src/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -190,11 +320,73 @@ private Position decodeNetwork(Channel channel, SocketAddress remoteAddress, Str
...
+                }
+            }
+        }
+
+        if (network.getCellTowers() != null || network.getWifiAccessPoints() != null) {
+            position.setNetwork(network);
+        }
+
+        // Parse date-time
+        DateBuilder dateBuilder = new DateBuilder()
```
tananaev: Why are you not using `parser.nextDateTime`?

src/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -223,6 +415,11 @@ protected Object decode(
             return decodeNetwork(channel, remoteAddress, sentence);
         }
 
+        Position positionWifi = decodeLbsWifi(channel, remoteAddress, sentence);
+        if (positionWifi != null) {
```
tananaev: Can this be identified by message type instead?

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: According to documentation, there is a list of wifi MAC addresses of variable size. Any link to some samples how such list is defined in regex?

info from: https://goo.gl/8DKadX
```
1.6.1 Cell Tower & WIFI format: (DW50)
(013632651491,DW50,460,0,0,6,2,aa*bb*cc*dd*ee*ff:-
8:0,aa*bb*cc*dd*ee*ff:-8:0,040613,040137)
"(" stands for packet start
"013632651491" stands for tracker ID
"DW50" stands for data command
"460" mcc
"1" mnc
"0" lac
"6" cid
“2” wifi number
“aa*bb*cc*dd*ee*ff:-8:0” wifi string(mac:rssi:channel_number)
“aa*bb*cc*dd*ee*ff:-8:0” wifi string(mac:rssi:channel_number)
"040613" stands for data, 04/June/2013
"040137" stands for time, 04:01:37, UTC time
")" stands for packet end
```
in actual device the symbols "*" replaced by ":" and vice versa

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: done

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: Those alarms from documentation: https://goo.gl/8DKadX
I do not have any clue what they mean and does not have actual device to check. I will remove this completely..

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: same as above, found in doc, i will remove...

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: even if battery events will be filtered? (they have timestamp (fixTime) same as previous event (because of getLastLocation()), so alway filtered by "filter.duplicate")

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: done

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: done

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: intention was that some other tk103-like devices will have the same format for wifi-positioning, but if not needed i can check for specific T580W commands. The other problem is that command depends from alarm code, but i can check at least known commands if needed...

src/org/traccar/protocol/Tk103ProtocolDecoder.java
tananaev: There is an option to not filter messages with certain attributes. Either way it's not a good idea to implement some workaround for a single protocol.

src/org/traccar/protocol/Tk103ProtocolDecoder.java
tananaev: Well, at least it should be something like this:
```java
number("((?:(?:xx:){5}(?:xx)\\*[-+]?d+\\*d+,)*)");
```
or if you know it's always just MAC addresses:
```java
expression("(.*)");
```
and parse it in the code. This is what we do in some other decoders.

src/org/traccar/protocol/Tk103ProtocolDecoder.java
tananaev: If they are present together, then you should remove `optional` from parameters themselves.

src/org/traccar/protocol/Tk103Protocol.java
tananaev: 1. Decide which command to use based on some configuration attribute. See GT06 encoder for example.
2. Either remove or add it to the command types if you think it's generic enough that would be useful for many protocols. What does `multiControllerOn` do?

ValeraVi: from user guide from T580W: https://goo.gl/M4ancp: 
`12. Multi Controller
Allows any cellphone number to check the device location and status besides the
master number.`

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: done

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: ok, i will try to fix this...

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: ok, i will remove...

src/org/traccar/protocol/Tk103Protocol.java
tananaev: This might be too specific. I would recommend to remove it for now and leave only generic enough commands.

ValeraVi: done

ValeraVi: > ...See GT06 encoder for example....
How to set this configuration attribute from test case code, so i can test my implementation of the command? Thanks!

tananaev: Add a setter.

src/org/traccar/protocol/Tk103ProtocolEncoder.java
```diff
@@ -16,36 +16,89 @@
...
-            case Command.TYPE_POSITION_STOP:
-                return formatCommand(command, "({%s}AR0000000000)", Command.KEY_UNIQUE_ID);
-            case Command.TYPE_ENGINE_STOP:
-                return formatCommand(command, "({%s}AV011)", Command.KEY_UNIQUE_ID);
-            case Command.TYPE_ENGINE_RESUME:
-                return formatCommand(command, "({%s}AV010)", Command.KEY_UNIQUE_ID);
-            default:
-                Log.warning(new UnsupportedOperationException(command.getType()));
-                break;
+        boolean deviceT580W = Context.getIdentityManager().lookupAttributeBoolean(
```
tananaev: I would recommend to use something more generic instead of model name. I think `alternative` as we use in GT06 would be better and more consistent.

src/org/traccar/protocol/Tk103ProtocolEncoder.java
```diff
@@ -16,36 +16,89 @@
...
 package org.traccar.protocol;
 
+import org.traccar.Context;
 import org.traccar.StringProtocolEncoder;
 import org.traccar.helper.Log;
 import org.traccar.model.Command;
 
 public class Tk103ProtocolEncoder extends StringProtocolEncoder {
 
+    protected String t580WFormatCommand(Command command, String format, String... keys) {
```
tananaev: Why do we need this method? I suggest we remove it.

src/org/traccar/protocol/Tk103ProtocolEncoder.java
```diff
@@ -16,36 +16,89 @@
...
+        String content = super.formatCommand(command, format, keys);
+        return String.format("[begin]sms2,%s,[end]", content);
+    }
+
+    protected String checkEn(Command command, String ifEnabled, String ifDisabled) {
```
tananaev: Method and parameter names are not per guidelines. Also, I think it should be inlined.

test/org/traccar/protocol/Tk103ProtocolDecoderTest.java
```diff
@@ -159,6 +159,48 @@ public void testDecode() throws Exception {
...
+
+        verifyNotNull(decoder, text(
+                "(013632651491,DW50,460,0,0,6,0,040613,040137"));
+
+        verifyNotNull(decoder, text(
+                "(864555555555555,ZC03,191117,234207,$Notice: Device version: 1.0$"));
+
+        verifyNotNull(decoder, text(
+                "(864555555555555,ZC03,191117,234207,$1 .Sensor sensitivity: 1\r\n" +
+                        "2 .Alert status: Off\r\n" +
```
tananaev: Remove extra indentation.

test/org/traccar/protocol/Tk103ProtocolEncoderTest.java
```diff
@@ -2,11 +2,71 @@
...
+import org.traccar.Context;
 import org.traccar.ProtocolTest;
+import org.traccar.database.IdentityManager;
 import org.traccar.model.Command;
+import org.traccar.model.Device;
+import org.traccar.model.Position;
 
 public class Tk103ProtocolEncoderTest extends ProtocolTest {
 
+    private static IdentityManager t580wManager = new IdentityManager() {
```
tananaev: I don't think this is necessary. There must be a better way.

test/org/traccar/protocol/Tk103ProtocolEncoderTest.java
```diff
@@ -31,6 +91,15 @@ public void testEncodePositionSingle() throws Exception {
 
         Assert.assertEquals("(123456789012345AP00)", encoder.encodeCommand(command));
 
+        IdentityManager old = Context.getIdentityManager();
+        Context.init(t580wManager);
+
+        try {
+            Assert.assertEquals("[begin]sms2,*getposl*,[end]", encoder.encodeCommand(command));
+        } finally {
```
tananaev: This is just wrong.

src/org/traccar/protocol/Tk103ProtocolEncoder.java
```diff
@@ -16,36 +16,89 @@
...
+                    return t580WFormatCommand(command, "*routetrackoff*");
+                case Command.TYPE_MODE_DEEP_SLEEP:
+                    return t580WFormatCommand(command, checkEn(command, "*sleep*2*", "*sleepoff*"));
+                case Command.TYPE_MODE_POWER_SAVING:
+                    return t580WFormatCommand(command, checkEn(command, "*sleepv*", "*sleepoff*"));
+                case Command.TYPE_ALARM_SOS:
+                    return t580WFormatCommand(command, checkEn(command, "*soson*", "*sosoff*"));
+                case Command.TYPE_SET_CONNECTION:
+                    String ipAddress = command.getString(Command.KEY_SERVER);
+                    ipAddress = ipAddress.replace(".", "*");
```
tananaev: This can be inlined as well.

src/org/traccar/protocol/Tk103ProtocolEncoder.java
ValeraVi: done

src/org/traccar/protocol/Tk103ProtocolEncoder.java
ValeraVi: done

src/org/traccar/protocol/Tk103ProtocolEncoder.java
ValeraVi: done

src/org/traccar/protocol/Tk103ProtocolEncoder.java
ValeraVi: Thanks for the tip! I will implement exactly the same approach as it was done in GT06.

test/org/traccar/protocol/Tk103ProtocolDecoderTest.java
ValeraVi: done

test/org/traccar/protocol/Tk103ProtocolEncoderTest.java
ValeraVi: as you suggested above, i will implement exactly the same approach as it was done in GT06.

test/org/traccar/protocol/Tk103ProtocolEncoderTest.java
ValeraVi: as you suggested above, i will implement exactly the same approach as it was done in GT06.

test/org/traccar/BaseTest.java
```diff
@@ -40,7 +40,7 @@ public boolean isLatestPosition(Position position) {
             @Override
             public boolean lookupAttributeBoolean(
                     long deviceId, String attributeName, boolean defaultValue, boolean lookupConfig) {
-                return defaultValue;
+                return (deviceId == 2 && attributeName == "tk103.alternative" && lookupConfig) ? true : defaultValue;
```
ValeraVi: Just FYI.

During debugging testcases, i found, that Context can have different IdentityManager, depending of in what order tests are run. This is because Context is a static stuff, so it will just contain IdentityManager from last test case (it can be based on FilterHandlerTest (below) or BaseTest (here) ).

So i had to modify both for now...

src/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -206,7 +353,7 @@ protected Object decode(
             sentence = sentence.substring(beginIndex + 1);
         }
 
-        if (channel != null) {
+        if (channel != null && sentence.length() >= 16) {
```
tananaev: Why are you adding this?

test/org/traccar/BaseTest.java
tananaev: We need to find a better way. I think you should just use setter like we do for decoder configuration parameters.

src/org/traccar/protocol/Tk103ProtocolDecoder.java
ValeraVi: to prevent this error/exception:

`2017-11-20 00:58:10 DEBUG: [BE5D7A93: 5002 < x.x.x.xx] STR: (864555555555,ZC03,191117,235808,$1 .Sensor sensitivity: 1
2 .Alert status: Off
3 .Check interval is set to 240 minute(s).
4 .Checkgsm interval is set to 60 minute(s).
5 .SOS SMS Alert: On
6 .SOS Call Alert: On
7 . Power: 94%$) HEX: xxxxxxx

2017-11-20 00:59:54  WARN: [BE5D7A93] error - String index out of range: 12 - StringIndexOutOfBoundsException (... < Tk103ProtocolDecoder:356 < ExtendedObjectDecoder:58 < ...)

2017-11-20 00:59:54  INFO: [BE5D7A93] disconnected
`
Some "Command Results" contains very "free/random" text, which can include frame end symbol ")", which result in several frames, even with very small sizes. They are ignored everywhere, except in this code.

test/org/traccar/BaseTest.java
ValeraVi: Context is static. So, even if i implement some other IdentityManager in BaseTest/FilterHandlerTest, which will have some setter for Attributes, i had to set some attribute before test and unset it after test (using try-finally), so it will not change behavior for other tests. But, as you mentioned previously you does not like try-finally. I am understanding you correctly about "use setter like we do for decoder configuration parameters"? If not - can you please share a link to code snippet with setter for decoder configuration parameters.

Anyway. If this problem is not already solved in any other xxxProtocolEncoderTest (and there are nothig related to this in GT06 protocol, which you mentioned before), and as this is not very related to T580w GPS/LBS/WIFI positioning actually. Do you agree to remove testcases for commands, which shares same name? There are still a lot of tests for other commands. And then try to solve this problem separately?

src/org/traccar/protocol/Tk103ProtocolDecoder.java
tananaev: Can I see proper original HEX? I don't see why it would not work.

ValeraVi: I have modified the traccar to log plain text, not hex, to easy debug...

Device sends to traccar:
(864555555555,ZC03,191117,235808,$1 .Sensor sensitivity: 1
2 .Alert status: Off
3 .Check interval is set to 240 minute(s).
4 .Checkgsm interval is set to 60 minute(s).
5 .SOS SMS Alert: On
6 .SOS Call Alert: On
7 . Power: 94%$)
(with \r\n as "new lines" but this does not matter for this case).

As tk103 protocol has ")" symbol as frame end symbol, this one message from device will be actually separated to 3 messages and "decode" will be called 3 times like this:

a) decode("(864555555555,ZC03,191117,235808,$1 .Sensor sensitivity: 1
2 .Alert status: Off
3 .Check interval is set to 240 minute(s");

b) decode(".
4 .Checkgsm interval is set to 60 minute(s");  <----- here will be exception,  because sentence = "s" - only one symbol. It is like that, because inside decode() sentence.indexOf('(') will remove everything until and including "(".

c) decode(".
5 .SOS SMS Alert: On
6 .SOS Call Alert: On
7 . Power: 94%$").

tananaev: Then the problem is with frame decoder. It needs to be fixed instead of implementing some workaround in the protocol decoder.

ValeraVi: Yes, i am agree that one frame decoder can't support all devices... For this device, seems like proper decoder should not try to search frame end symbol between "$" symbols...

tananaev: No, that's not what I said. In this case it's definitely possible to have single frame decoder.

ValeraVi: yes it is possible, but for this device the only solution to properly decode all data as single frame is to not try to search frame end symbol ")" between "$" symbols. Here is a format or this command from documentation:
"return content" as we can see also can contain frame end symbol ")", but it should be ignored. Hope that they does not send "$" symbol inside return content :)

1.4 Return SMS (ZC03)
(013632651491,ZC03, 040613, 040137,$...$)
"(" stands for packet start
"013632651491" stands for tracker ID
" ZC03" stands for data command
"040613" stands for date, 04/June/2013
"040137" stands for time, 04:01:37, UTC time
“$...$” stands for return content
")" stands for packet end

tananaev: I was thinking about counting opening and closing brackets, but ignoring content between `$` should also work.

tananaev:
I would recommend to use `indexOf` method in the frame decode instead of iterating through the whole buffer yourself.

src/org/traccar/protocol/Tk103FrameDecoder.java
```diff
@@ -0,0 +1,87 @@
...
+    protected Object decode(
+            ChannelHandlerContext ctx,
+            Channel channel,
+            ChannelBuffer buf) throws Exception {
+
+        if (buf.readableBytes() < FRAME_MIN_SIZE) {
+            return null;
+        }
+
+        int indexStart = buf.indexOf(buf.readerIndex(), buf.writerIndex(), FRAME_START_SYMBOL);
```
tananaev: I believe this is not necessary. It's always the first character.

src/org/traccar/protocol/Tk103FrameDecoder.java
```diff
@@ -0,0 +1,87 @@
...
+import org.jboss.netty.channel.ChannelHandlerContext;
+import org.jboss.netty.handler.codec.frame.FrameDecoder;
+
+public class Tk103FrameDecoder extends FrameDecoder {
+
+    private static final int FRAME_MIN_SIZE = 2;
+    private static final int FRAME_MAX_SIZE = 1024;
+    private static final byte FRAME_START_SYMBOL = (byte) '(';
+    private static final byte FRAME_END_SYMBOL = (byte) ')';
+    private static final byte FRAME_FREE_TEXT_BLOCK_SYMBOL = (byte) '$';
```
tananaev: Not sure why you decided to make all these constants. Most of them are used one-time only and I think code would be much more readable without them.

src/org/traccar/protocol/Tk103FrameDecoder.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        int indexStart = buf.indexOf(buf.readerIndex(), buf.writerIndex(), FRAME_START_SYMBOL);
+        if (indexStart == -1) {
+            buf.clear();
+            return null;
+        }
+        indexStart++;
+
+        int index, cnt;
```
tananaev: Please use proper names for variables. In this case probably `count` instead of `cnt`.

src/org/traccar/protocol/Tk103FrameDecoder.java
```diff
@@ -0,0 +1,87 @@
...
+        for (index = indexStart, cnt = 0;; index++) {
+            int i = index;
+            index = buf.indexOf(index, buf.writerIndex(), FRAME_END_SYMBOL);
+            if (index == -1) {
```
tananaev: I don't think the logic is correct. Please also add unit tests for frame decoder.

test/org/traccar/TestIdentityManager.java
```diff
@@ -0,0 +1,122 @@
+package org.traccar;
+
+import org.traccar.database.*;
+import org.traccar.model.*;
+import java.util.concurrent.ConcurrentHashMap;
+
+import java.util.Map;
+
+public final class TestIdentityManager implements IdentityManager {
```
tananaev: This class seems to be overcomplicated. I prefer unit tests to be as simple as possible.

test/org/traccar/TestIdentityManager.java
```diff
@@ -0,0 +1,122 @@
+package org.traccar;
+
+import org.traccar.database.*;
```
tananaev: We never use star-imports.

src/org/traccar/protocol/Tk103FrameDecoder.java
ValeraVi: done

src/org/traccar/protocol/Tk103FrameDecoder.java
ValeraVi: As i understand this is not always the case, that why this was implemented in decoder: https://github.com/tsmgeek/traccar/blob/65fcb5b18289b645f0b26138b35af8eb224b7ce1/src/org/traccar/protocol/Tk103ProtocolDecoder.java#L204
but i can remove indexStart if you do not want it.

src/org/traccar/protocol/Tk103FrameDecoder.java
ValeraVi: Tested by debugging and also code works on my server several days without problems. Please share your concerns.
Can you please suggest what existing protocol frame decoder i can use as a sample for unit tests?

test/org/traccar/TestIdentityManager.java
ValeraVi: done

test/org/traccar/TestIdentityManager.java
ValeraVi: Should i remove unused code and simplify the code, or you does not accept the whole approach?

src/org/traccar/protocol/Tk103FrameDecoder.java
tananaev: Looks like you are right, but then it needs to be in a single place, not both. Frame decoder sounds like a better place.

src/org/traccar/protocol/Tk103FrameDecoder.java
tananaev: Most custom frame decoders have tests. They all look the same.

test/org/traccar/TestIdentityManager.java
tananaev: I think it's a bad idea for unit tests to have some internal state, but I like the idea of having a separate class for it. I guess to simplify the class to something close to the original one, possibly with some small additions/modifications.

src/org/traccar/protocol/Tk103FrameDecoder.java
ValeraVi: done

src/org/traccar/protocol/Tk103FrameDecoder.java
ValeraVi: done

src/org/traccar/protocol/Tk103FrameDecoder.java
ValeraVi: done

test/org/traccar/TestIdentityManager.java
ValeraVi: done

ValeraVi:
Frame decoder unit tests coming soon..

ValeraVi:
Some more fixes for this PR are needed?

tananaev:
I don't know. Is it ready to be reviewed? You haven't said anything.

ValeraVi:
Yes, sorry for delay, but i uploaded all what i have for now, so it is ready for review.

tananaev:
I had a brief look and noticed that you are removing brackets from the message in your frame decoder. Please don't do it. Keep message in original form.

ValeraVi:
Old frame decoder also removed closing bracket ")", but ok, i will keep both...

ValeraVi:
I have modified frame decoder and related code to keep brackets in messages. Please, review. Thanks!

test/org/traccar/TestIdentityManager.java
tananaev: You still internal state in this class.

ValeraVi:
Another implementation of the unit tests for "alternative" tk103 commands in protocol encoder uploaded. Internal state in TestIdentityManager class was removed. Please, review. Thanks.

tananaev:
I think moving it into a separate class is overkill. Also you create and destroy an instance every time. We can probably just have a parameter in constructor to force alternative. This way we can easily test it without a need for creating a separate class.

ValeraVi:
Change for force alternative parameter in constructor and modified tests are uploaded. Please, review. Thanks!

tananaev:
Merged, thanks. I will do some minor cleanup, but mostly it looks good.

loiclecapitaine:
dear all
I have followed your discussion about T580W but i am a newbie so it's difficult for me to all understand. I try to set my T580W tracker on my traccar server but unsuccesfully . I use 5002 port. 
so according to your message what i have to do to configure the server to use fine the T580 WIFI tracker from VJOY

Many thanks for your precious help.

tananaev:
@loiclecapitaine, this is not the right place to ask questions. Use forum and ideally create your own thread. Also, make sure you follow troubleshooting guide BEFORE posting anything.

loiclecapitaine:
Sorry i will post it on traccar forum
regards

PR #3658, opened 2017-11-27
- Fix local login if ldap enabled & Limit Command permission

Abyss777:
Fixed couple of issues:
- Fix local login if ldap enabled
- Do not allow user edit own "Limit Command" permission. Missed this check on backend side.


tananaev:
Merged, thanks.

PR #3663, opened 2017-11-29
handle cut power alarm for pt502

categulario:
Although this alarm is not sent always in this way, I have received some
messages with this format


tananaev:
Merged, thanks.

PR #3667, opened 2017-12-04
flespi protocol remastered to receive data from flespi stream

nadezhdamosunova:
Hi Anton,
This is flespi protocol remastered to receive data from [flespi stream](https://flespi.io/docs/#/registry/!/streams).
Also more parameters are converted into internal representation.
Could you please review the commit to merge it into the Traccar project.


src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -69,36 +85,246 @@ protected Object decode(
...
+        }
+        if (name.equals("vehicle.vin")) {
+            position.set(Position.KEY_VIN, ((JsonString) value).getString());
+            return 1;
+        }
+        return 0;
+    }
+
+    private int saveAlarm(String name, JsonValue value, Position position) {
+        if ("alarm.event.trigger".equals(name)) {
```
tananaev: Please use `switch` everywhere instead of `if`s.

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -69,36 +85,246 @@ protected Object decode(
...
     private void decodePosition(JsonObject object, Position position) {
-        position.setProtocol(getProtocolName());
+        // store all params in position
+        Set<Map.Entry<String, JsonValue>> params = object.entrySet();
+        Iterator<Map.Entry<String, JsonValue>> it = params.iterator();
+        while (it.hasNext()) {
+            Map.Entry<String, JsonValue> param = it.next();
+            String paramName = param.getKey();
+            JsonValue paramValue = param.getValue();
+            if ("timestamp".equals(paramName)) {
```
tananaev: Why do we need split it into several methods? Why can't we just have a big switch for everything?

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -69,36 +85,246 @@ protected Object decode(
     }
 
     private void decodePosition(JsonObject object, Position position) {
-        position.setProtocol(getProtocolName());
+        // store all params in position
```
tananaev: I think this comment doesn't add any value and should be removed.

src/org/traccar/protocol/FlespiProtocolDecoder.java
nadezhdamosunova: Having one big switch will violate MethodLength checkstyle rule. Is it possible to discard this check?

tananaev: Are you sure you have tried switch and not ifs?

nadezhdamosunova:
Code refactored to use switch instead of ifs: 445b1e1

src/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -69,36 +85,185 @@ protected Object decode(
...
+    }
 
-        position.setValid(object.getBoolean("position.valid", true));
-        position.set(Position.KEY_SATELLITES, object.getInt("position.satellites", 0));
+    private int saveParam(String name, int index, JsonValue value, Position position) {
+        switch (name) {
+            case "timestamp":
+                Date deviceTime = new Date((long) (((JsonNumber) value).doubleValue() * 1000));
+                position.setTime(deviceTime);
+                return 1;
```
tananaev: Why is there `return 1` everywhere? I think it makes more sense to use boolean.

nadezhdamosunova:
Return value of saveParam method is changed to boolean: 4d188ce

tananaev:
Merged, thanks.

PR #3680, opened 2017-12-15
Handle boolean parameters in mail configuration attributes

### LINKED ISSUE #3679: Some types incompatibility in user attributes
https://www.traccar.org/forums/topic/oshibka-v-pochtovom-otpravlenii-cherez-atributy-javalangboolean/

Mail attributes must be strings in backend https://github.com/tananaev/traccar/blob/master/src/org/traccar/notification/NotificationMail.java

But in the web client they have different types https://github.com/tananaev/traccar-web/blob/master/web/app/store/UserAttributes.js
boolean is boolean

What will we do? I think making them all strings on web side will decrease users ease of use.

Abyss777:
Handle "boolean" and "string as boolean" attributes in mail configuration

fix #3679

### DIFF
--- a/src/org/traccar/notification/NotificationMail.java
+++ b/src/org/traccar/notification/NotificationMail.java
@@ -47,16 +47,16 @@ public final class NotificationMail {
 
             Boolean starttlsEnable = provider.getBoolean("mail.smtp.starttls.enable");
             if (starttlsEnable != null) {
-                properties.put("mail.smtp.starttls.enable", starttlsEnable);
+                properties.put("mail.smtp.starttls.enable", String.valueOf(starttlsEnable));
             }
             Boolean starttlsRequired = provider.getBoolean("mail.smtp.starttls.required");
             if (starttlsRequired != null) {
-                properties.put("mail.smtp.starttls.required", starttlsRequired);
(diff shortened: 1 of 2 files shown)


src/org/traccar/notification/NotificationMail.java
```diff
@@ -45,18 +45,18 @@ private static Properties getProperties(PropertiesProvider provider) {
             properties.put("mail.smtp.host", host);
             properties.put("mail.smtp.port", String.valueOf(provider.getInteger("mail.smtp.port", 25)));
 
-            String starttlsEnable = provider.getString("mail.smtp.starttls.enable");
+            Boolean starttlsEnable = provider.getBoolean("mail.smtp.starttls.enable");
             if (starttlsEnable != null) {
-                properties.put("mail.smtp.starttls.enable", Boolean.parseBoolean(starttlsEnable));
+                properties.put("mail.smtp.starttls.enable", starttlsEnable);
```
tananaev: I thought `String` values are required here and in other places.

Abyss777:
It is working (looks like internally all converted to strings).
But you are right, better be consistent.

tananaev:
Merged, thanks.

PR #3690, opened 2017-12-20
din and dout parameters remastered

nadezhdamosunova:
Hi Anton,
We changed registration of "din" and "dout" parameters in flespi platform.
Now the parameters contain a bit-mask of digital inputs or outputs on/off values.
I've changed FlespiProtocolDecoder accordingly.
Could you please consider the pull request.


tananaev:
Merged, thanks.

PR #3695, opened 2017-12-23
FCM Implementation

### LINKED ISSUE #3551: Feature Request: GCM & APNS support
Ability for notifications for user to be forwarded to GCM/APNS.
Would need webapp to have ability to set these up.

parveenkumardeeva:
Hi Tananaev,

Implemented the FCM feature compatible with Java 1.7 (traccar's current version) but didn't tested with java 1.7.
I have successfully tested with Java 1.8 and notifications are coming to Android app successfully.

It's the backport version of the following library which is written in Java 1.8 and all the features are listed there.

[https://github.com/bytefish/FcmJava](url)

Here are the config attributes:

web.templatesPath = sms  (Using SMS templates as fcm templates)
fcm.ignoreUserConfig =true
fcm.api.key= {fcm server key}
fcm.api.url = https://fcm.googleapis.com/fcm/send ( By default ping to https://fcm.googleapis.com/fcm/send )

And in User attributes:
fcm.token = fcm registration token key, currently in the user attributes. I didn't implemented it in the User Model directly because didn't find very much use case except that we can update the User's fcm token which we have received from the FCM Server in response to the old fcm token while sending Notification. Here is the Link:
[https://github.com/bytefish/FcmJava/issues/30](url)

fix #3551


tananaev:
I don't think copying the whole library is an acceptable solution.

parveenkumardeeva:
Yeah, It's not. We can use library directly via maven but only in the case of Java 1.8.

So, meanwhile we can use the library as a module. After the Java version upgrade we can use only the implementation classes.

atiqsamtia:
I have implemented the FCM without an external library, simply using HTTP calls. check my Pull request https://github.com/traccar/traccar/pull/4002

PR #3697, opened 2017-12-26
Device disable

### LINKED ISSUE #3683: Feature Request - Device Disable
Good Day

Would be nice to add Device Enable / Disable feature:

Device Status: Disabled. (Still get location data and save to DB)
Can not select in reporting until enabled. (Accounting Platforms Integration)
Rest-Api to be able to modify between enabled / disabled. (Accounting Platforms Integration)

Could you add something like this?

Thank You

Albertus Geyser

Abyss777:
Here is implementation discussed in #3683

Disabled device excluded from reports of ordinary users and managers and from events notifications (via all transports).
Current device position is not updated via websocket for ordinary user, but device object is still updated.
Theoretically user can understand where device is placed now by `device.geofenceIds[]` but I do not think it is big leakage.

Only admin can enable device back.

There is switch `database.storeDisabled` to store or not store data from disabled device.

Also It is easy to add device expiration in future.

Frankly I'm not very satisfied with changes, some constructions/checks become rather complex. I have some thoughts about permission optimizations, but it is need to be discussed.

fix #3683

### DIFF
--- a/schema/changelog-3.16.xml
+++ b/schema/changelog-3.16.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd"
+  logicalFilePath="changelog-3.16">
+
+  <changeSet author="author" id="changelog-3.16">
+
+    <addColumn tableName="devices">
+      <column name="disabled" type="BOOLEAN" defaultValueBoolean="false" />
+    </addColumn>
+
+  </changeSet>
+</databaseChangeLog>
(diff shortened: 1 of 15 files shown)


tananaev:
I agree that it seems like a lot of extra complexity. That's why I suggested to basically exclude devices from user permissions if it's disabled. That would have been much simpler. Maybe we should still do it, but have some extra API to return disabled devices? To display it only in "devices" panel?

Abyss777:
I'm not sure that it will be less complex, just in other places of code. We still need exclude devices from reports and positions.

My thoughts is about improve logic of `checkDevice` function. Basically we have next permissions:
- read device object - all users who have access to device including device readonly users
- edit/remove device object - not readonly  and device readonly users who have access to device
- send command - all who can read ( if not disabled? )
- reset total distance - admins only (?)
- retrieve current position - all who can read if device not disabled
- be notified about events related to device - same as previous
- get reports - same as previous
- ...

What is complex is that in some cases we should rise exception, but in other we should skip device silently.

tananaev:
Even if we implement that, we would need to basically double that logic on the web side as well. I feel like my option is much much simpler.

Abyss777:
I'll try, how do you see admin interface to enable/disable devices?

tananaev:
I guess just a checkbox on the device model.

Abyss777:
But disabled devices are excluded from user lists, how to get to device model?

tananaev:
No, obviously they are not excluded from admin.

Abyss777:
Implemented simple disable in other PR, want to leave this variant as is just in case.

PR #3698, opened 2017-12-26
flespi protocol: content-type=application/json

nadezhdamosunova:
Hi Anton,
We changed flespi stream to send json data in request body instead of in url-encoded data  (Content-Type:application/json).
Please, consider the corresponding changes in FlespiProtocolDecoder.


tananaev:
Merged, thanks.

PR #3700, opened 2017-12-27
Implement device disable (simple)

Abyss777:
Here is simple implementation of device disabling. In continue of https://github.com/tananaev/traccar/pull/3697

Device is basically hidden from all users except admins.

Incoming positions might be stored or not in DB. (`database.storeDisabled`)

### DIFF
--- a/src/org/traccar/database/DeviceManager.java
+++ b/src/org/traccar/database/DeviceManager.java
@@ -104,13 +104,33 @@ public class DeviceManager extends BaseObjectManager<Device> implements Identity
         return getItems(getAllItems());
     }
 
+    public Set<Long> getAllUserItems(long userId) {
+        return Context.getPermissionsManager().getDevicePermissions(userId);
+    }
+
     @Override
     public Set<Long> getUserItems(long userId) {
         if (Context.getPermissionsManager() != null) {
-            return Context.getPermissionsManager().getDevicePermissions(userId);
+            Set<Long> result = new HashSet<>();
+            for (long deviceId : Context.getPermissionsManager().getDevicePermissions(userId)) {
+                Device device = Context.getIdentityManager().getById(deviceId);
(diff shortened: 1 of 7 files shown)


src/org/traccar/database/PermissionsManager.java
```diff
@@ -179,9 +194,9 @@ public void checkDeviceLimit(long userId) throws SecurityException, SQLException
         if (deviceLimit != -1) {
             int deviceCount = 0;
             if (getUserManager(userId)) {
-                deviceCount = Context.getDeviceManager().getManagedItems(userId).size();
+                deviceCount = Context.getDeviceManager().getAllManagedItems(userId).size();
             } else {
-                deviceCount = Context.getDeviceManager().getUserItems(userId).size();
+                deviceCount = Context.getDeviceManager().getAllUserItems(userId).size();
```
Abyss777: Calculate limit using all devices including disabled to avoid users tricks.

src/org/traccar/database/DeviceManager.java
```diff
@@ -104,15 +104,35 @@ public Device getDeviceByPhone(String phone) {
...
+        return Context.getPermissionsManager().getDevicePermissions(userId);
+    }
+
     @Override
     public Set<Long> getUserItems(long userId) {
         if (Context.getPermissionsManager() != null) {
-            return Context.getPermissionsManager().getDevicePermissions(userId);
+            Set<Long> result = new HashSet<>();
+            for (long deviceId : Context.getPermissionsManager().getDevicePermissions(userId)) {
+                Device device = Context.getIdentityManager().getById(deviceId);
```
tananaev: Do we really need `getIdentityManager` here? You are already in `DeviceManager`.

Abyss777:
Oh, sorry, removed. Initially it was in other class.

tananaev:
Merged, thanks.

PR #3709, opened 2018-01-10
Odometer permission to both Admin or Manager

parveenkumardeeva:


tananaev:
This doesn't seem correct. It will allow any user to do it.

parveenkumardeeva:
Oh Sorry, Got the error.

src/org/traccar/database/PermissionsManager.java
```diff
@@ -300,7 +300,13 @@ public void checkGroup(long userId, long groupId) throws SecurityException {
     }
 
     public void checkDevice(long userId, long deviceId) throws SecurityException {
-        if (!Context.getDeviceManager().getUserItems(userId).contains(deviceId) && !getUserAdmin(userId)) {
+        if (!Context.getDeviceManager().getUserItems(userId).contains(deviceId)) {
+            checkDeviceManagerOrAdmin(userId, deviceId);
+        }
+    }
+
+    public void checkDeviceManagerOrAdmin(long userId, long deviceId) throws SecurityException {
```
tananaev: I don't think we need a separate method. Please implement it the same way it's implement in other places.

parveenkumardeeva: Then We have to implement whole logic in the Device Resource's updateTotalDistance method or change the checkDeviceManagerOrAdmin method name?

tananaev: You don't need another method and you don't need to move the logic anywhere. Just look at the `BaseObjectResource` example.

parveenkumardeeva:
Dear Anton,

I don't find any better way to do this. Please add this feature yourself if you still don't found the OK code.

tananaev:
Looks good now. Merged, thanks.

PR #3710, opened 2018-01-11
Update group and device cache on group delete.

Abyss777:
Deleting group may cause cascade clear `groupId` fields in `groups` and `devices` DB tables. We need refresh cached objects.

https://www.traccar.org/forums/topic/group-delete-issue/


tananaev:
Merged, thanks.

parveenkumardeeva:
Thank you both for quick response.

PR #3719, opened 2018-01-17
Add poiLayer field to User/Server model

### LINKED ISSUE #513: Few suggestions
Hi, Anton!
I have a few questions for Traccar.
It would be really great if we could add their own POI and pics of the cars to the map. To facilitate the localization of the interface, must bind variables in the .war file to an external lang files. For example make a /lang directory and place langs files as a txt, xml, or some other files that could interact with variables from .war file. Also, would be nice if there was a possibility create your own themes using css and html markup. This step would help increase the amount of third-party themes for interface. Actually I want to say that Traccar wonderful product that has great potential and I want to say thank you for the development of this amazing tracking system!
I very much hope that my suggestions will be heard.
Thanks.

### LINKED ISSUE #3714: Adding custom POI
Sorry, but i can write this only bu Russian now.

Предлагаем ввести функцию добавления точек ПОИ с выпадающим описанием при наведении, чтобы была возможность вставлять фото, телефон и ссылку. Желательно чтобы ПОИ можно было группировать и управлять их отображением для различных групп пользователей. Мы используем эту систему в автоклубе УАЗ Патриот Санкт-Петербург и предполагаем введение всех магазинов запчастей для наших машин, СТО партнеров. При проведении клубных мероприятий будем обозначать точки сбора, место расположения базового лагеря и т.п... Сейчас пока реализовали просто добавление таких точек как объектов мониторинга, что конечно же не удобно, и не дает возможность получить информацию по точке (телефон, адрес сайта...)...

Abyss777:
Added field to store POI layer URL.

Made `VARCHAR` length as for `mapUrl`

fix #3714 
fix #513

### DIFF
--- a/schema/changelog-3.16.xml
+++ b/schema/changelog-3.16.xml
@@ -12,5 +12,13 @@
       <column name="disabled" type="BOOLEAN" defaultValueBoolean="false" />
     </addColumn>
 
+    <addColumn tableName="users">
+        <column name="poilayer" type="VARCHAR(512)" />
+    </addColumn>
+
+    <addColumn tableName="servers">
+        <column name="poilayer" type="VARCHAR(512)" />
+    </addColumn>
+
   </changeSet>
 </databaseChangeLog>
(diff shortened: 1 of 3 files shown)


tananaev:
Merged, thanks. It's much easier to modify models now. So much less code :)

PR #3721, opened 2018-01-19
Adding filter.skipLimit functionality based on Server Time Only

parveenkumardeeva:
https://www.traccar.org/forums/topic/does-filterduplicate-and-filterskiplimit-works-well-for-gt06-protocol/


tananaev:
Merged, thanks.

tananaev:
You are right, they will be filtered, but I think the main idea behind filter limit was to allow periodic live update of the device. For example, device doesn't move for a day or even longer with distance filtering, but you still want a message from it once in a while (e.g. once an hour) to make sure device is still working.

Abyss777:
Hm... you are right `skipLimit` should not influence on such "positions bunch" theoretically.

PR #3723, opened 2018-01-19
org.traccar.ServerManager: ignore stray files in the jar archive

matjon:
Somehow, non-class files may find their way into the jar archive.
For example, I had Tk103ProtocolDecoder.java open in Vim. Vim created a
temporary file called .Tk103ProtocolDecoder.java.swp which was imported
into the .jar archive.  ServerManager then tried to load a class named
.Tk103ProtocolDecoder.java and failed.
This patch prevents ServerManager from trying to load such bogus files.

This work was funded by Partner Security, http://www.partnersecurity.pl/ but code is owned by me.


tananaev:
I think you need to investigate why it's happening for you instead of implementing this workaround. It shouldn't happen.

tananaev:
Closing as no updates.

PR #3724, opened 2018-01-19
[WIP] org.traccar.MainEventHandler: ignore rapid position changes

matjon:
Sometimes, the GPS trackers seem to send erroneous data that
teleport the tracked unit to Africa or somewhere similar.
Detect and ignore such reported positions.

This work was funded by Partner Security,
http://www.partnersecurity.pl . but code is owned by me.

This is not the final code to be merged, but rather I am seeking some comments on the approach.


Abyss777:
This (and much more) already implemented in filter https://github.com/tananaev/traccar/blob/master/src/org/traccar/FilterHandler.java

matjon:
Thanks, Abyss777

PR #3731, opened 2018-01-20
NPE bug fix for the Position's serverTime

parveenkumardeeva:
I was not aware about the initialization of serverTime and didn't found any better way than this.

Pull Request:
https://github.com/tananaev/traccar/pull/3721

### DIFF
--- a/src/org/traccar/model/Position.java
+++ b/src/org/traccar/model/Position.java
@@ -131,6 +131,10 @@ public class Position extends Message {
     public static final String ALARM_TAMPERING = "tampering";
     public static final String ALARM_REMOVING = "removing";
 
+    public Position() {
+        this.serverTime = new Date();
+    }
+
     private String protocol;
 
     public String getProtocol() {


src/org/traccar/model/Position.java
```diff
@@ -131,6 +131,10 @@
     public static final String ALARM_TAMPERING = "tampering";
     public static final String ALARM_REMOVING = "removing";
 
+    public Position() {
+        this.serverTime = new Date();
```
tananaev: You don't need `this`.

parveenkumardeeva:
Thanks,
Sure, In future I'll take care of it definitely.

Abyss777:
Don't this contradict with https://github.com/tananaev/traccar/blob/master/src/org/traccar/database/DataManager.java#L341
I mean internal `serverTime` will be a bit different with that goes in DB.

tananaev:
Not a big issue, but yes, we should probably remove that from data manager.

tananaev:
Fixed in 68b2d3ab8897114dac8cd92618cf0777f630750a. @Abyss777, do you see any issues with the change?

Abyss777:
No, I don't see. Good code cleanup.

But I see another problem with initial code, `Date` object creation happen twice when we loading positions from DB. First one in constructor, second in `setServerTime`. It increases amount "garbage" objects and amount used memory (not much but still) that is already problem with reports.

tananaev:
Any ideas how to fix it? Set it every time in decoder?

Abyss777:
May be like this?
```java
    public void setServerTime(Date serverTime) {
        if (serverTime != null) {
            if (this.serverTime != null) {
                this.serverTime.setTime(serverTime.getTime());
            } else {
                this.serverTime = new Date(serverTime.getTime());
            }
        } else {
            this.serverTime = null;
        }
    }

```

tananaev:
How about set it in getter if it's not already set?

Abyss777:
getter called only from `filter.skipLimit`. twice for `current` and `last` at the same time.
If we return `serverTime` initialization as was (when write in DB). `Last` do not get applied `serverTime` from DB and values in `skipLimit` will be the same.

parveenkumardeeva:
I don't know is it right or wrong but using for other framework.
And also for the Traccar and working well:

    private Date serverTime;

    public Date getServerTime() {
        return serverTime;
    }

    public void setServerTime(Date serverTime) {
        this.serverTime = serverTime;
    }

tananaev:
@Abyss777, main place where getter is called is `QueryBuilder`. I'll try to add new constructor.

tananaev:
I have added a new constructor and removed new `Date` object creation every time. It was originally done to fix FindBugs issues, but I think faster code is better than some warnings. Ignored those warnings. @Abyss777, please check if you see any issues.

Abyss777:
Good. I do not see any issues.
Just to fix is in mind, `Date` is not primitive and we should handle it more carefully as pointer.

PR #3736, opened 2018-01-23
Implement scheduled notifications

Abyss777:
A calendar can be selected for a notification now. If it is selected, notification will be sent only in calendar period.

- Introduced `ScheduledModel` for objects with `calendarId` field
- Filter "effective" notifications by calendar
- Added permission check when add or update `ScheduledModel` objects (User should not be able select calendar he do not have access)
- Removing calendar can cause cascade clear of `calendarId` fields we should refresh geofences and notifications.

### DIFF
--- a/schema/changelog-3.16.xml
+++ b/schema/changelog-3.16.xml
@@ -24,7 +24,7 @@
       <column name="calendarid" type="INT" />
     </addColumn>
 
-    <addForeignKeyConstraint baseColumnNames="calendarid" baseTableName="notifications" constraintName="fk_notification_calendar_calendarid" onDelete="SET NULL" onUpdate="RESTRICT" referencedColumnNames="id" referencedTableName="calendars"/>
+    <addForeignKeyConstraint baseColumnNames="calendarid" baseTableName="notifications" constraintName="fk_notification_calendar_calendarid" onDelete="SET NULL" onUpdate="RESTRICT" referencedColumnNames="id" referencedTableName="calendars" />
 
   </changeSet>
 </databaseChangeLog>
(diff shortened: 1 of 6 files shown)


schema/changelog-3.16.xml
```diff
@@ -20,5 +20,11 @@
         <column name="poilayer" type="VARCHAR(512)" />
     </addColumn>
 
+    <addColumn tableName="notifications">
+      <column name="calendarid" type="INT" />
+    </addColumn>
+
+    <addForeignKeyConstraint baseColumnNames="calendarid" baseTableName="notifications" constraintName="fk_notification_calendar_calendarid" onDelete="SET NULL" onUpdate="RESTRICT" referencedColumnNames="id" referencedTableName="calendars"/>
```
tananaev: Please add space before the closing bracket.

src/org/traccar/database/NotificationManager.java
```diff
@@ -42,12 +44,16 @@ public NotificationManager(DataManager dataManager) {
         geocodeOnRequest = Context.getConfig().getBoolean("geocoder.onRequest");
     }
 
-    private Set<Long> getEffectiveNotifications(long userId, long deviceId) {
+    private Set<Long> getEffectiveNotifications(long userId, long deviceId, Date date) {
```
tananaev: I would rename it to `time` to make it clear that it's a full time, not just date.

Abyss777:
Done

tananaev:
Merged, thanks.

parveenkumardeeva:
Dear @Abyss777 ,

User can't edit or create new Notification or Geofence unless he's an Admin.
It's because they have by default calenderId set to 0. I think this should not happen according to design or Am I understanding something wrong?

@tananaev , It's not so old request so I have decided to write here.

PR #3741, opened 2018-01-29
Fix ScheduledModel objects permission check

Abyss777:
Fix https://github.com/tananaev/traccar/pull/3736#issuecomment-360967138

And implemented check for `groupId` for Devices and Groups in the same way.

### DIFF
--- a/src/org/traccar/api/BaseObjectResource.java
+++ b/src/org/traccar/api/BaseObjectResource.java
@@ -37,6 +37,7 @@ import org.traccar.model.Calendar;
 import org.traccar.model.Command;
 import org.traccar.model.Device;
 import org.traccar.model.Group;
+import org.traccar.model.GroupedModel;
 import org.traccar.model.ScheduledModel;
 import org.traccar.model.User;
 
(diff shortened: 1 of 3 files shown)


src/org/traccar/api/BaseObjectResource.java
```diff
@@ -79,7 +80,10 @@ public Response add(T entity) throws SQLException {
             Context.getPermissionsManager().checkDeviceLimit(getUserId());
         } else if (baseClass.equals(Command.class)) {
             Context.getPermissionsManager().checkLimitCommands(getUserId());
-        } else if (entity instanceof ScheduledModel) {
+        } else if (entity instanceof GroupedModel && ((GroupedModel) entity).getGroupId() != 0) {
+            Context.getPermissionsManager().checkPermission(Group.class, getUserId(),
```
tananaev: It's a bit of a formatting preference, but I think if parameters don't fit it's better to new-line all parameters like this:
```java
Context.getPermissionsManager().checkPermission(
        Group.class, getUserId(), ((GroupedModel) entity).getGroupId());
```
This way you need method name on one line and all parameters on the other line. Easier to read. Also, some code editors handle it better.

Abyss777: I like it in general with reserve. In case all parameters does not fit new line, may be better leave part on first line?

tananaev: In some rare case, but most of the time I would prefer two lines with just parameters and keep method on the first line.

Abyss777:
Updated formatting

tananaev:
Merged, thanks.

parveenkumardeeva:
Thank you both.

PR #3749, opened 2018-02-01
Provide web access to media files received from devices

### LINKED ISSUE #3688: Media access
Hi Anton,

Where to view media files in the application ? 
is their any way to access media files on another server?

Abyss777:
There is a switch `media.dirAllowed` to allow access directories. Admin can enumerate root directory, users only device directories they have access.

image

- Added `DefaultServlet` to serve folder with media files as static content
- Added filter to check access to device

fix #3688 

More comments in code...

### DIFF
--- a/src/org/traccar/api/MediaFilter.java
+++ b/src/org/traccar/api/MediaFilter.java
@@ -17,6 +17,7 @@
 package org.traccar.api;
 
 import java.io.IOException;
+import java.sql.SQLException;
 
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
@@ -27,7 +28,6 @@ import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
-import javax.ws.rs.NotAuthorizedException;
 
 import org.traccar.Context;
 import org.traccar.api.resource.SessionResource;
@@ -54,7 +54,9 @@ public class MediaFilter implements Filter {
                 }
             }
             if (userId == null) {
-                throw new NotAuthorizedException("Not authorized");
+                ((HttpServletResponse) response).setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+                response.getWriter().println("Not authorized");
+                return;
             }
 
             String path = ((HttpServletRequest) request).getPathInfo();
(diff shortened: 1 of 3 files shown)


src/org/traccar/api/MediaFilter.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        try {
+            HttpSession session = ((HttpServletRequest) request).getSession(false);
+            Long userId = null;
+            if (session != null) {
+                userId = (Long) session.getAttribute(SessionResource.USER_ID_KEY);
```
Abyss777: Authorization by session only is supported

src/org/traccar/web/WebServer.java
```diff
@@ -156,6 +161,19 @@ private void initOldWebApp() {
         handlers.addHandler(app);
     }
 
+    private void initMedia() {
+        ServletContextHandler servletHandler = new ServletContextHandler(ServletContextHandler.SESSIONS);
+        servletHandler.setContextPath("/api/media");
```
Abyss777: We have to use sub path to make SessionManager use same sessions as for `/api`

tananaev: Can we add it to the API ServletContextHandler?

src/org/traccar/api/MediaFilter.java
```diff
@@ -0,0 +1,104 @@
...
+import org.traccar.helper.Log;
+import org.traccar.model.Device;
+
+public class MediaFilter implements Filter {
+
+    private boolean dirAllowed;
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+        dirAllowed = Context.getConfig().getBoolean("media.dirAllowed");
```
tananaev: I would prefer to use full words. For example `directoryAllow`.

src/org/traccar/api/MediaFilter.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private static void formatError(HttpServletResponse response, Exception e) throws IOException {
```
tananaev: Can we set status and return directly in the `doFilter` method?

src/org/traccar/api/MediaFilter.java
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            String[] parts = ((HttpServletRequest) request).getPathInfo().split("/");
+            if (parts.length < 2) {
+                if (dirAllowed) {
+                    Context.getPermissionsManager().checkAdmin(userId);
+                } else {
+                    throw new SecurityException("Wrong path");
+                }
+            } else if (parts.length == 2 && !dirAllowed) {
```
tananaev: What is this check for?

Abyss777: It is to prevent list directory for ordinary users, but just thought may be it is not needed, because it is forbidden on servlet level anyway.

src/org/traccar/api/MediaFilter.java
Abyss777: Sure

Abyss777:
- Move media servlet to api context (it is much better, thanks for the tip)
- Removed any directory check from filter, let servlet handle it
- Renamed switch to `media.directoryAllow`
- Moved error formatting back to `doFilter`
- Some other cleanup

src/org/traccar/api/MediaFilter.java
```diff
@@ -93,7 +72,17 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 
             chain.doFilter(request, response);
         } catch (Exception e) {
-            formatError((HttpServletResponse) response, e);
+            HttpServletResponse httpResponse = (HttpServletResponse) response;
+            if (e instanceof SecurityException) {
```
tananaev: Actually what I meant is that instead of throwing exceptions and catching them you can just set status directly in most cases.

Abyss777: We still need catch Security exceptions when checking user and device. Just made it in common way.

tananaev: Yes, but do we need to catch anything else other than security? It seems like a lot of overhead. And even if we do, `instanceof` is not the best option. You can just have separate catch clauses. But I think if we don't need exception we shouldn't use it.

Abyss777:
You are right. Removed throwing of exceptions. Used `DeviceManager` instead identity to `getByUniqueId` to catch only `SQLException`

src/org/traccar/api/MediaFilter.java
```diff
@@ -54,34 +54,32 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
                 }
             }
             if (userId == null) {
-                throw new NotAuthorizedException("Not authorized");
+                ((HttpServletResponse) response).setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+                response.getWriter().println("Not authorized");
```
tananaev: Do we really need to write some text in the output? I think it makes sense for exception only. Hardcoded text doesn't provide any value over status code.

src/org/traccar/api/MediaFilter.java
```diff
@@ -54,34 +54,32 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
                 }
             }
             if (userId == null) {
-                throw new NotAuthorizedException("Not authorized");
+                ((HttpServletResponse) response).setStatus(HttpServletResponse.SC_UNAUTHORIZED);
```
tananaev: You can cast it once at the beginning of the method. It will probably make code easier to read.

tananaev:
Merged, thanks.

Abyss777:
Thanks. Sorry for so much corrections in this PR

tananaev:
One more thing I though of. What would happen if media folder is not specified?

Abyss777:
Good question. It returns 404 for me, but I think it tries open some folder. But do not know which one, neither `./api/media` /  `./media` or `/api/media` /  `/media`

Abyss777:
What about choose some default folder? like `./media` ?

tananaev:
I think it would be better to allow disabling of this functionality.

PR #3752, opened 2018-02-02
Add switch to enable media web access

Abyss777:
Added switch `media.web.enable` and renamed `media.web.directoryAllow` for consistence

### DIFF
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -337,9 +337,7 @@ public final class Context {
             ldapProvider = new LdapProvider(config);
         }
 
-        if (config.hasKey("media.path")) {
-            mediaManager = new MediaManager(config);
-        }
+        mediaManager = new MediaManager(config.getString("media.path"));
 
         if (dataManager != null) {
             usersManager = new UsersManager(dataManager);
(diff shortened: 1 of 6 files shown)


tananaev:
Can you please also implement it for `MediaManager`.

Abyss777:
Sure. Which values we will place in `/setup/default.xml`

tananaev:
Enable by default. Maybe we should just enable it if path is provided?

Abyss777:
Agree, extra parameters are useless.

Abyss777:
Now protocols do not handle case when MediaManager is not defined. How does traccar should behave in case `MediaManager` disabled but got photo? Still create position from last and put something in attributes to inform what photo was skipped or silently skip them?

tananaev:
Just return null and attribute will be zero. I don't think there is any problem with it.

Abyss777:
Used `media.path` as switch and added handles when `MediaManager` is undefined in protocols.

tananaev:
I was thinking of returning null from `writeFile` to avoid null checks.

Abyss777:
But MediaManager might be not initialized https://github.com/tananaev/traccar/blob/master/src/org/traccar/Context.java#L340

tananaev:
Yes, but maybe it's better to initialize it and then just return `null` if path is not specified. This way we won't need all those null checks.

Abyss777:
Done

tananaev:
Merged, thanks.

tananaev:
I actually decided to revert back parameter name to the original one because now we use it in a single place and it makes much more sense: https://github.com/tananaev/traccar/commit/654116f06161a6f73d23a6bb610f5420cebfdfc9.

Sorry about that.

Abyss777:
Thank you. It's OK.

PR #3754, opened 2018-02-04
Added a post option to forward url

NateZ7:
Hi Anton,

As we discussed, forward.json will determine if the forward url would be POST or GET.

### DIFF
--- a/src/org/traccar/WebDataHandler.java
+++ b/src/org/traccar/WebDataHandler.java
@@ -144,7 +144,7 @@ public class WebDataHandler extends BaseDataHandler {
     }
 
     protected String prepareJsonPayload(Position position) {
-        
+
         Map<String, Object> data = new HashMap<>();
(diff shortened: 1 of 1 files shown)


tananaev:
Why are you manually create a hash map for JSON object? You need to use existing models.

Also, build failed, so you would need to fix the error before we can proceed.

NateZ7:
Hi Anton,

You asked me to follow the event forwarding logic, so I did.
As I saw in EventForwarder, you created hash map manually.
Let me know if i'm mistaken about that.
Plus, I had some issues with the fork, so please address only the second
fork, which has no build errors (fixed it)

Let me know for any fix needed in order to merge my changes.
Thanks,
Nativ.

tananaev:
You need to use full objects same way as in event forwarding, but not individual primitive types. Please close second pull request.

NateZ7:
Deleted the wrong pull request, and fixed what you asked.

PR #3755, opened 2018-02-04
added POST option to forward url

NateZ7:
Hi Anton,

As we discussed, I added the forward.json to determine POST or GET in forward url.

### DIFF
--- a/src/org/traccar/BasePipelineFactory.java
+++ b/src/org/traccar/BasePipelineFactory.java
@@ -238,9 +238,8 @@ public abstract class BasePipelineFactory implements ChannelPipelineFactory {
         }
 
         if (Context.getConfig().getBoolean("forward.enable")) {
-            final Boolean json = Context.getConfig().getBoolean("forward.json");
-            final String url = Context.getConfig().getString("forward.url");
-            pipeline.addLast("webHandler", new WebDataHandler(url, json));
+            pipeline.addLast("webHandler", new WebDataHandler(Context.getConfig().getString("forward.url"),
+                    Context.getConfig().getBoolean("forward.json")));
         }
 
         if (commandResultEventHandler != null) {
--- a/src/org/traccar/BasePipelineFactory.java
+++ b/src/org/traccar/BasePipelineFactory.java
@@ -238,7 +238,7 @@ public abstract class BasePipelineFactory implements ChannelPipelineFactory {
         }
 
         if (Context.getConfig().getBoolean("forward.enable")) {
-            pipeline.addLast("webHandler", new WebDataHandler(Context.getConfig().getString("forward.url")));
+            pipeline.addLast("webHandler", new WebDataHandler(Context.getConfig().getString("forward.url"), Context.getConfig().getBoolean("forward.json")));
         }
 
         if (commandResultEventHandler != null) {
--- a/src/org/traccar/BasePipelineFactory.java
+++ b/src/org/traccar/BasePipelineFactory.java
@@ -238,7 +238,9 @@ public abstract class BasePipelineFactory implements ChannelPipelineFactory {
         }
 
         if (Context.getConfig().getBoolean("forward.enable")) {
-            pipeline.addLast("webHandler", new WebDataHandler(Context.getConfig().getString("forward.url"), Context.getConfig().getBoolean("forward.json")));
+            final Boolean json = Context.getConfig().getBoolean("forward.json");
+            final String url = Context.getConfig().getString("forward.url");
+            pipeline.addLast("webHandler", new WebDataHandler(url, json));
         }
 
         if (commandResultEventHandler != null) {
--- a/src/org/traccar/WebDataHandler.java
+++ b/src/org/traccar/WebDataHandler.java
@@ -146,17 +146,10 @@ public class WebDataHandler extends BaseDataHandler {
     protected String prepareJsonPayload(Position position) {
 
         Map<String, Object> data = new HashMap<>();
+        Device device = Context.getIdentityManager().getById(position.getDeviceId());
 
-        if (position != null) {
-            data.put(KEY_POSITION, position);
-        }
-
-        if (position.getDeviceId() != 0) {
-            Device device = Context.getIdentityManager().getById(position.getDeviceId());
-            if (device != null) {
-                data.put(KEY_DEVICE, device);
-            }
-        }
+        data.put(KEY_POSITION, position);
+        data.put(KEY_DEVICE, device);
 
         try {
             return Context.getObjectMapper().writeValueAsString(data);
--- a/src/org/traccar/WebDataHandler.java
+++ b/src/org/traccar/WebDataHandler.java
@@ -148,8 +148,13 @@ public class WebDataHandler extends BaseDataHandler {
         Map<String, Object> data = new HashMap<>();
         Device device = Context.getIdentityManager().getById(position.getDeviceId());
 
-        data.put(KEY_POSITION, position);
-        data.put(KEY_DEVICE, device);
+        if (position != null) {
+            data.put(KEY_POSITION, position);
+        }
+
+        if (device != null) {
+            data.put(KEY_DEVICE, device);
+        }
 
         try {
             return Context.getObjectMapper().writeValueAsString(data);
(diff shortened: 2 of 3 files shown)


tananaev:
Why are there two pull requests? Also, please address comment from the second thread.

NateZ7:
This one is the relevant PR. deleted the other one and fixed what you said, thanks.

src/org/traccar/BasePipelineFactory.java
```diff
@@ -238,7 +238,9 @@ public ChannelPipeline getPipeline() {
         }
 
         if (Context.getConfig().getBoolean("forward.enable")) {
-            pipeline.addLast("webHandler", new WebDataHandler(Context.getConfig().getString("forward.url")));
+            final Boolean json = Context.getConfig().getBoolean("forward.json");
```
tananaev: I think we can safely inline the variables.

src/org/traccar/WebDataHandler.java
```diff
@@ -29,12 +32,17 @@
 import java.util.Locale;
 import java.util.TimeZone;
 
+
```
tananaev: Remove extra spaces like this.

src/org/traccar/WebDataHandler.java
```diff
@@ -29,12 +32,17 @@
...
 public class WebDataHandler extends BaseDataHandler {
 
     private final String url;
+    private final Boolean json;
```
tananaev: Why is it a boolean class and not a primitive type?

src/org/traccar/WebDataHandler.java
```diff
@@ -29,12 +32,17 @@
...
+    private static final String KEY_POSITION = "position";
+    private static final String KEY_DEVICE = "device";
 
-    public WebDataHandler(String url) {
+    public WebDataHandler(String url, Boolean json) {
```
tananaev: Same here.

test/org/traccar/WebDataHandlerTest.java
```diff
@@ -12,7 +12,7 @@ public void testFormatRequest() throws Exception {
 
         Position p = position("2016-01-01 01:02:03.000", true, 20, 30);
 
-        WebDataHandler handler = new WebDataHandler("http://localhost/?fixTime={fixTime}&gprmc={gprmc}&name={name}");
+        WebDataHandler handler = new WebDataHandler("http://localhost/?fixTime={fixTime}&gprmc={gprmc}&name={name}", false);
```
tananaev: Please add unit test for JSON as well.

src/org/traccar/WebDataHandler.java
```diff
@@ -121,10 +129,50 @@ public String formatRequest(Position position) {
...
+            setContent(position, requestBuilder);
+            requestBuilder.execute();
 
+        } else {
+            Context.getAsyncHttpClient().prepareGet(formatRequest(position)).execute();
+        }
         return position;
     }
 
+    protected String getContentType() {
```
tananaev: Why do we need a separate method?

src/org/traccar/WebDataHandler.java
```diff
@@ -121,10 +129,50 @@ public String formatRequest(Position position) {
...
+        return "application/json; charset=utf-8";
+    }
+
+    protected void setContent(Position position, AsyncHttpClient.BoundRequestBuilder requestBuilder) {
```
tananaev: Same question.

src/org/traccar/BasePipelineFactory.java
NateZ7: Its longer than 150 lines, Travis CI didn't accept it

src/org/traccar/WebDataHandler.java
NateZ7: Here, we need a separate method because I don't have Event object in WebDataHandler, so that method is edited accordingly

src/org/traccar/WebDataHandler.java
NateZ7: How would you want me to get the content type from 'JsonTypeEventForwarder' ??

src/org/traccar/BasePipelineFactory.java
tananaev: You can separate it into several lines.

src/org/traccar/WebDataHandler.java
tananaev: What does `JsonTypeEventForwarder` have to do with this? It has nothing to do with events.

src/org/traccar/WebDataHandler.java
tananaev: I don't understand what events have to do with this functionality.

NateZ7: What did you mean by separate method ? maybe I dont get it.

src/org/traccar/WebDataHandler.java
NateZ7: What did you mean by separate method ? maybe I dont get it.

src/org/traccar/BasePipelineFactory.java
NateZ7: That's what I did

src/org/traccar/WebDataHandler.java
tananaev: Why do you have a method with one line of code? It doesn't make sense to me.

NateZ7: understood. Fixed and pushed

tananaev:
You haven't fixed one of the comments (about removing variables). Also, why is test case called `testBuilderRequest`?

NateZ7:
Thanks for the comment, Fixed the name to match previous syntax, plus fixed the inlined variables.

tananaev:
@Abyss777, can you please take a quick look and see if you can think of any issues or problems.

Abyss777:
What is the reason of sending Device object with every position?

NateZ7:
@Abyss777 We got a request from a client that matches the data we get from the position, and maybe it will be helpful for others. Nevertheless, there is a flag to activate it or not.

tananaev:
I think it makes sense so that third party app has access to device model as well, if it needs. Also, I imagine most of the time this will be used on a same machine or at least local network, so some extra traffic shouldn't be a problem.

Also, it's consistent with event forwarding.

src/org/traccar/WebDataHandler.java
```diff
@@ -121,10 +128,41 @@ public String formatRequest(Position position) {
...
 
+    protected String prepareJsonPayload(Position position) {
+
+        Map<String, Object> data = new HashMap<>();
+
+        if (position != null) {
+            data.put(KEY_POSITION, position);
+        }
+
+        if (position.getDeviceId() != 0) {
```
Abyss777: I think this check if not necessary and next check device for null be enough .

Abyss777:
OK. Then one small comment.

src/org/traccar/WebDataHandler.java
tananaev: @Abyss777, should we remove same checks from `EventForwarder`?

NateZ7: @Abyss777  Again, the check of the position is not necessary ? just leave the check device for null ?

tananaev: I think it's safe to assume that there is a device.

NateZ7: So remove both checks ?

Abyss777: Hm...I believe yes we can remove.
Map.get(0) will return `null`  - "the value to which the specified key is mapped, or null if this map contains no mapping for the key"

NateZ7:
@Abyss777 remove the checks.

tananaev:
What about device? Why are we removing it?

Abyss777:
@NateZ7 
You need remove checks id for `0`, but leave checks objects for `null`

NateZ7:
@Abyss777  Got ya ! did the changes

tananaev:
Merged it, thanks.

slobglob:
@tananaev Please add it the `forward.json` option to https://www.traccar.org/configuration-file/
Thank you.

tananaev:
Updated.

PR #3765, opened 2018-02-09
Unified Device timezone handling

### LINKED ISSUE #3572: Time offset for each device
Please implement configurable time offset for each device. There are plenty of cheap Chinese GPS trackers with poor documentation where correct time zone is impossible to setup. Even when using multiple GT06 devices, it is not possible to configure this "globally" as each device sends different timezone timestamp.

So I think the best is to add device attribute with Time Offset, so I can specify different offset for particular device.

Abyss777:
Added `timeZone` field to `DeviceSession`
- It might be initialized once on login message and used in all next packets
- It might be initialized once on device session creation to not lookup all tree on every position. But to apply timezone changes device has to reconnect.

Also removed unnecessary checks in forwarders

fix #3572

### DIFF
--- a/src/org/traccar/WebDataHandler.java
+++ b/src/org/traccar/WebDataHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 - 2016 Anton Tananaev (anton@traccar.org)
+ * Copyright 2015 - 2018 Anton Tananaev (anton@traccar.org)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


src/org/traccar/WebDataHandler.java
```diff
@@ -148,9 +148,7 @@ protected String prepareJsonPayload(Position position) {
         Map<String, Object> data = new HashMap<>();
         Device device = Context.getIdentityManager().getById(position.getDeviceId());
 
-        if (position != null) {
-            data.put(KEY_POSITION, position);
-        }
+        data.put(KEY_POSITION, position);
```
Abyss777: Here was FindBug warning

> Nullcheck of value previously dereferenced (RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE)

tananaev:
Merged, thanks.

PR #3782, opened 2018-02-22
Add support for OpenGTS compatible clients (GPSLogger for Android)

vblack:
Signed-off-by: Kostiantyn Luzan <vblack2006@gmail.com>


tananaev:
Please rename classes from "OpenGTS" to "OpenGts". That's the coding style that we use.

Looks like it's an HTTP-based protocol. Why are you decoding it as a raw TCP protocol?

Do you have proper documentation for the format?

vblack:
Ok, I will rename - not a problem.
Yes it is actually HTTP GET requests, but I am no familiar with your system, I just see that Sanav protocol did similar things, that needed for this.
Unfortunately I do not have documentation, however OpenGTS is opensource, so there should be something at least in the code.

tananaev:
You need to take OsmAnd protocol as an example. It's also http-based.

vblack:
Ok. I'll rework it based on OsmAnd

PR #3784, opened 2018-02-25
Add support for OpenGTS compatible clients (GPSLogger for Android)

vblack:
Signed-off-by: Kostiantyn Luzan <vblack2006@gmail.com>

### DIFF
--- a/src/org/traccar/protocol/OpenGtsProtocolDecoder.java
+++ b/src/org/traccar/protocol/OpenGtsProtocolDecoder.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2013 - 2017 Anton Tananaev (anton@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.protocol;
+
+import org.jboss.netty.channel.Channel;
+import org.jboss.netty.handler.codec.http.HttpRequest;
+import org.jboss.netty.handler.codec.http.HttpResponseStatus;
+import org.jboss.netty.handler.codec.http.QueryStringDecoder;
+import org.traccar.BaseHttpProtocolDecoder;
+import org.traccar.DeviceSession;
+import org.traccar.helper.DateBuilder;
+import org.traccar.helper.Parser;
+import org.traccar.helper.PatternBuilder;
+import org.traccar.model.Position;
+
+import java.net.SocketAddress;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+public class OpenGtsProtocolDecoder extends BaseHttpProtocolDecoder {
+
+    private static final Pattern PATTERN = new PatternBuilder()
+            .text("$GPRMC,")
+            .number("(dd)(dd)(dd),")          // time (hhmmss)
+            .expression("([AV]),")               // validity
+            .number("(d+)(dd.d+),")              // latitude
+            .expression("([NS]),")
+            .number("(d+)(dd.d+),")              // longitude
+            .expression("([EW]),")
+            .number("(d+.d+),")                  // speed
+            .number("(d+.d+),")                 // course
+            .number("(dd)(dd)(dd),")             // date (ddmmyy)
+            .any()
+            .compile();
+
+    public OpenGtsProtocolDecoder(OpenGtsProtocol protocol) {
+        super(protocol);
+    }
+
+    @Override
+    protected Object decode(
+            Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
+
+        HttpRequest request = (HttpRequest) msg;
+        QueryStringDecoder decoder = new QueryStringDecoder(request.getUri());
+        Map<String, List<String>> params = decoder.getParameters();
+
+        Position position = new Position();
+        position.setProtocol(getProtocolName());
+
+        for (Map.Entry<String, List<String>> entry : params.entrySet()) {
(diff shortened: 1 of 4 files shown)


test/org/traccar/protocol/OpenGtsProtocolDecoderTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class OpenGtsProtocolDecoderTest extends ProtocolTest {
+
+    @Test
+    public void testDecode() throws Exception {
+
+        OpenGtsProtocolDecoder decoder = new OpenGtsProtocolDecoder(new OpenGtsProtocol());
+
+        verifyNull(decoder, request(
+                "/?bat=0"));
```
tananaev: Is this a real test case?

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
```diff
@@ -0,0 +1,141 @@
...
+                    position.setSpeed(parser.nextDouble(0));
+                    position.setCourse(parser.nextDouble(0));
+
+                    dateBuilder.setDateReverse(parser.nextInt(0), parser.nextInt(0), parser.nextInt(0));
+                    position.setTime(dateBuilder.getDate());
+                    break;
+                case "alt":
+                    position.setAltitude(Double.parseDouble(value));
+                    break;
+                case "batt":
```
tananaev: Are all parameters real? Please include only ones that OpenGTS protocol actually has.

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
```diff
@@ -0,0 +1,141 @@
...
+                    position.set(Position.KEY_BATTERY_LEVEL, Double.parseDouble(value));
+                    break;
+                default:
```
tananaev: Same question here.

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Override
+    protected Object decode(
+            Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
+
+        HttpRequest request = (HttpRequest) msg;
+        QueryStringDecoder decoder = new QueryStringDecoder(request.getUri());
+        Map<String, List<String>> params = decoder.getParameters();
+        if (params.isEmpty()) {
+            decoder = new QueryStringDecoder(request.getContent().toString(StandardCharsets.US_ASCII), false);
```
tananaev: Why do you need this?

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
```diff
@@ -0,0 +1,141 @@
...
+                                position.set(entry.getKey(), value);
+                                break;
+                        }
+                    }
+                    break;
+            }
+        }
+
+        if (position.getFixTime() == null) {
+            position.setTime(new Date());
```
tananaev: Why do we need this?

test/org/traccar/protocol/OpenGtsProtocolDecoderTest.java
vblack: Well, I don't see situation when id actually not coming, so we do not need to cover such possibility? Cause I don't have protocol documentation, I couldn't say for sure.

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
vblack: Yes, as you could see from test string there is actual altitude and battery. Depending on device battery reported zero or real value. Altitude seems real on all devices I have.

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
vblack: I'm not sure, but why OsmAnd need this? To make sure that HTTP request contains only ASCII ?

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
vblack: Yes, looks like this is irrelevant. Removed.

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
tananaev: OK.

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
tananaev: What about this one?

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
tananaev: No, OsmAnd protocol supports data in the body content, no just query string.

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
vblack: On some of devices it is always reporting zero (mostly Car Head units and some mobiles), on some - it is real battery percentage.

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
vblack: Ok, I got it.

src/org/traccar/protocol/OpenGtsProtocolDecoder.java
tananaev: The question is about default case.

vblack: I think OsmAnd use this for all non-handled fields (like future update or something), so do I, because I don't have a protocol documentation. If you think this is bad idea - I could remove it, no problem.

tananaev: OsmAnd handles additional fields because there are some clients that send additional data. We shouldn't add it unless it's also the case for OpenGTS protocol.

vblack: Ok, I got your point

vblack:
Hi, Is there any more issues before merging this?

tananaev:
I don't know. Have you fixed all the issues? You have to notify me if that's the case.

vblack:
Yes, I fixed all the issues.

tananaev:

PR #3788, opened 2018-02-28
Cycle in group hierarchy bug fix while editing Groups

parveenkumardeeva:
Cycle in group hierarchy check only works during cache update but not checking at the time of entry updation into the DB.

### DIFF
--- a/src/org/traccar/database/GroupsManager.java
+++ b/src/org/traccar/database/GroupsManager.java
@@ -78,7 +78,8 @@ public class GroupsManager extends BaseObjectManager<Group> implements Managable
     @Override
     public void updateItem(Group group) throws SQLException {
         checkGroupCycles(group);
-        super.updateItem(group);
+        getDataManager().updateObject(group);
+        super.updateCachedItem(group);
     }
 
     @Override
--- a/src/org/traccar/database/GroupsManager.java
+++ b/src/org/traccar/database/GroupsManager.java
@@ -78,14 +78,7 @@ public class GroupsManager extends BaseObjectManager<Group> implements Managable
     @Override
     public void updateItem(Group group) throws SQLException {
         checkGroupCycles(group);
-        getDataManager().updateObject(group);
-        super.updateCachedItem(group);
-    }
-
-    @Override
-    protected void updateCachedItem(Group group) {
-        checkGroupCycles(group);
-        super.updateCachedItem(group);
+        super.updateItem(group);
     }
 
     @Override
--- a/src/org/traccar/database/GroupsManager.java
+++ b/src/org/traccar/database/GroupsManager.java
@@ -76,6 +76,12 @@ public class GroupsManager extends BaseObjectManager<Group> implements Managable
     }
 
     @Override
+    public void updateItem(Group group) throws SQLException {
+        checkGroupCycles(group);
+        super.updateItem(group);
+    }
+
+    @Override
     protected void updateCachedItem(Group group) {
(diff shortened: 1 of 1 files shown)


Abyss777:
I'm not sure it is needed. Cycle checked here https://github.com/tananaev/traccar/blob/master/src/org/traccar/database/GroupsManager.java#L80

It is called from parent object here https://github.com/tananaev/traccar/blob/master/src/org/traccar/database/BaseObjectManager.java#L97

parveenkumardeeva:
It's needed because it updates the entry into the DB but throws error at the time of cache update.
This cause the server crash at the time of server restart because that time it checks for Cycle in group hierarchy.

Abyss777:
Got it now.
But we got double check for cycle. It is need to find how to remove it.

parveenkumardeeva:
We can remove it from the updateCachedItem method but currently as per my honest opinion it won't check for circular dependency during server restart and the cache can still have wrong data into it.
Any another view in your mind?

parveenkumardeeva:
Yes, sure @Abyss777 

I think it can be only removed after clearing the db for such type of entries.

tananaev:
Just want to add a note here that this is a very common problem. I have seen it on demo servers multiple times and it would be great if we can fix it before next release.

Abyss777:
@parveenkumardeeva, @tananaev 
How about this?
```java
    @Override
    public void updateItem(Group group) throws SQLException {
         checkGroupCycles(group);
         getDataManager().updateObject(group);
         super.updateCachedItem(group);
    }
```

parveenkumardeeva:
I think the main problem is not resolved until the DB gets cleared for such type of entries.
Then this workaround should not be needed because circular dependency will not get into the DB because of the pre-checks.
So, we can simply remove the overrided updateCachedItem method or Is it neccessary to check into the refreshItems for the circular dependency even after pre-checks? Any specific reason?

Abyss777:
We have some logic what allow work direct updates Devices and Groups in DB. We can't absolutely rely that DB is always correct.

parveenkumardeeva:
OK, I see. Then we can skip the double check by your code.

parveenkumardeeva:
Also then we can write some logic while updateCachedItem for the removal of the circular dependecies that exists into the DB.

Abyss777:
I'm not sure we need some automatic correction. After this fix we will not get new cycles hopefully. But old should be fixed manually.

tananaev:
I have just checked the code. Why do we need to override `updateCachedItem`? Can't we just do something like following and remove `updateCachedItem` override?
```java
    @Override
    public void updateItem(Group group) throws SQLException {
        checkGroupCycles(group);
        super.updateItem(group);
    }
```

Abyss777:
> Why do we need to override updateCachedItem?
We are refreshing items (groups) periodically from DB (in case it is updated directly).
https://github.com/tananaev/traccar/blob/master/src/org/traccar/database/BaseObjectManager.java#L66

I believe we need this protection to have cache without cycles even if they are in DB. Do we?

tananaev:
I think as long as we ensure that it's impossible to create cycle through API, we shouldn't check data when we load database. If user manually messed up something in the database then it's his problem.

Abyss777:
OK. Then we can remove `updateCachedItem` override.

tananaev:
@parveenkumardeeva, can you please update pull request.

tananaev:
Merged, thanks.

PR #3802, opened 2018-03-10
store valid position as latest

alisol911:


tananaev:
This will not be merged if you don't provide any explanation.

alisol911:
Invalid positions are used as event log only, we lost current position when latest is invalid

tananaev:
I don't agree. Someone might be interested to see additional sensor information even if position is invalid.

alisol911:
Ok, I handle this in my own branch, thanks

PR #3810, opened 2018-03-16
Added LaipacSFKamel protocol

kevingoos:
I added the LaipacSFKamel protocol, this protocol is used be Laipac devices like Lola, Bracelet and PL-V4.
We already use it in our environment and it can maybe be useful for some other Traccar users.


src/org/traccar/helper/DataConverter.java
```diff
@@ -26,7 +26,7 @@ private DataConverter() {
 
     public static byte[] parseHex(String string) {
         try {
-            return Hex.decodeHex(string);
+            return Hex.decodeHex(string.toCharArray());
```
tananaev: Why do we need this change?

tools/translate.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/python
```
tananaev: Why do we need this file here?

src/org/traccar/protocol/OsmAndProtocolDecoder.java
```diff
@@ -42,8 +42,7 @@ public OsmAndProtocolDecoder(OsmAndProtocol protocol) {
     }
 
     @Override
-    protected Object decode(
-            Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
+    protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
```
tananaev: Why are you changing OsmAnd protocol?

src/org/traccar/model/Position.java
```diff
@@ -52,6 +52,8 @@
     public static final String KEY_BATTERY_LEVEL = "batteryLevel"; // percentage
     public static final String KEY_FUEL_LEVEL = "fuel"; // liters
     public static final String KEY_FUEL_CONSUMPTION = "fuelConsumption"; // liters/hour
+    public static final String KEY_ANALOG_1 = "analog 1"; // volts
```
tananaev: For ADC we use `PREFIX_ADC`.

src/org/traccar/model/Position.java
```diff
@@ -88,6 +90,10 @@
 
     public static final String KEY_DRIVER_UNIQUE_ID = "driverUniqueId";
 
+    public static final String KEY_CELL_NET_CODE = "cellNetCode";
```
tananaev: Please explain new attributes.

src/org/traccar/helper/DataConverter.java
kevingoos: I had a build error on this line. It could be because I am using maybe an older version of java?
I am running JDK 1.8.0_161 on my computer. The toCharArray() function fixed the build error for me.

tools/translate.py
kevingoos: Can be removed, my mistake.

src/org/traccar/protocol/OsmAndProtocolDecoder.java
kevingoos: I looked into this file for some how to. But it looks like my environment tried to change the formatting, can be undone.

src/org/traccar/model/Position.java
kevingoos: I will adapt this.

src/org/traccar/model/Position.java
kevingoos: Inside the Laipac AVRMC message, there are two part included called Cell ID 1 and Cell ID 2. These to part contain info about the simcard and network inside the device. Cell ID 1 is split up in two parts, the first 4 bytes telling about the Cell Network Code (this is a code which is dependent on the cell network you are connected to). And the last 4 bytes which are the Cell ID Code (which are the codes connected to the simcard itself)

kevingoos:
1. The protocol inside traccar for Laipac wasn't working correctly with the new devices that we have currently. Apparently they changed there messages for the new devices, and added some extra info. With the old regex, the new message doesn't get parsed correctly and this create a mayor bug in the system because it will not send the $EAVACK message correctly. This means that the messages get stacked up in the device.

2. I have the protocol documentation from Laipac themselves, but for legal reasons I cannot share this documentation.

tananaev:
If seems like the format is almost the same as before. If the difference is just a few fields, then probably the best solution is to modify existing decoder.

kevingoos:
@tananaev Can you do a none regression test on an older liapac device with the new code, because we don't have one of the old devices, only the new once.

tananaev:
There are unit tests for that.

kevingoos:
@tananaev I adapted the Liapac protocol to work with the new version and info. I also added some extra unit tests for special senarios.

src/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -35,25 +35,34 @@ public LaipacProtocolDecoder(LaipacProtocol protocol) {
 
     private static final Pattern PATTERN = new PatternBuilder()
             .text("$AVRMC,")
-            .expression("([^,]+),")              // identifier
```
tananaev: You changed spacing everywhere which messed up the diff. Please revert comments on any unchanged code.

src/org/traccar/model/Position.java
tananaev: Cell information has to be decoded into network object.

setup/default.xml
```diff
@@ -231,5 +231,4 @@
     <entry key='opengts.port'>5159</entry>
     <entry key='cautela.port'>5160</entry>
     <entry key='continental'>5161</entry>
-
```
tananaev: Please revert this change.

test/org/traccar/protocol/LaipacProtocolDecoderTest.java
```diff
@@ -28,8 +28,7 @@ public void testDecode() throws Exception {
         verifyNull(decoder, text(
                 "$ECHK,MSG00002,0*5E"));
 
-        verifyPosition(decoder, text(
-                "$AVRMC,99999999,164339,A,4351.0542,N,07923.5445,W,0.29,78.66,180703,0,3.727,17,1,0,0*37"),
+        verifyPosition(decoder, text("$AVRMC,99999999,164339,A,4351.0542,N,07923.5445,W,0.29,78.66,180703,0,3.727,17,1,0,0*37"),
```
tananaev: Please keep original formatting.

src/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -35,25 +35,34 @@ public LaipacProtocolDecoder(LaipacProtocol protocol) {
...
+            .number("(d{3})")                   // Cell 2 - Operator Code
+            .optional(4)
+            .text("*")
+            .number("(xx)")                     // checksum
             .compile();
 
     @Override
     protected Object decode(
             Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
-
```
tananaev: Please don't make unnecessary formatting changes.

kevingoos: Fixed

test/org/traccar/protocol/LaipacProtocolDecoderTest.java
kevingoos: Fixed

src/org/traccar/protocol/LaipacProtocolDecoder.java
kevingoos: Fixed

setup/default.xml
kevingoos: Fixed

kevingoos:
@tananaev now the network is also fixed :)

setup/default.xml
```diff
@@ -33,7 +33,7 @@
     <entry key='database.generateQueries'>true</entry>
 
     <entry key='database.changelog'>./schema/changelog-master.xml</entry>
-
```
tananaev: This file is still in the diff. Please check diff yourself before saying that it's fixed.

src/org/traccar/helper/DataConverter.java
tananaev: This is a part of third party library and has nothing to do with version of Java that you are using. If it doesn't work, it means there is some issue with your SDK or development tools. This needs to be reverted.

src/org/traccar/model/Position.java
```diff
@@ -291,5 +291,4 @@ public void setNetwork(Network network) {
     public String getType() {
         return super.getType();
     }
-
```
tananaev: Again, revert any unnecessary changes to the file. Please check diff on GitHub yourself before asking to review.

src/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -88,28 +101,86 @@ protected Object decode(
         dateBuilder.setDateReverse(parser.nextInt(0), parser.nextInt(0), parser.nextInt(0));
         position.setTime(dateBuilder.getDate());
 
-        String type = parser.next();
+        String eventCode = parser.next();
+        String decodedAlarm = decodeAlarm(eventCode);
+        if (decodedAlarm != null) {
+            position.set(Position.KEY_ALARM, decodeAlarm(eventCode));
```
tananaev: Why do you call `decodeAlarm` twice? You don't need null check here and probably don't need variable as well.

src/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -45,8 +47,18 @@ public LaipacProtocolDecoder(LaipacProtocol protocol) {
             .number("(d+.d+),")                  // speed
             .number("(d+.d+),")                  // course
             .number("(dd)(dd)(dd),")             // date (ddmmyy)
-            .expression("(.),")                  // type
-            .expression("[^*]+").text("*")
+            .expression("([abZXTSMHFE86430]),")  // event code
+            .number("(d+)").expression("(\\.?)").number("(d*),")    // battery voltage
```
tananaev: Can you please explain logic here.

src/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -88,28 +101,86 @@ protected Object decode(
...
-        String type = parser.next();
+        String eventCode = parser.next();
+        String decodedAlarm = decodeAlarm(eventCode);
+        if (decodedAlarm != null) {
+            position.set(Position.KEY_ALARM, decodeAlarm(eventCode));
+        }
+        position.set(Position.KEY_EVENT, eventCode);
+
+        double batteryVoltage = parser.nextDouble();
+        if (parser.next().isEmpty()) {
```
tananaev: Logic here looks overcomplicated.

src/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -88,28 +101,86 @@ protected Object decode(
...
+            parser.next();
+            batteryVoltage *= 0.001;
+        } else {
+            batteryVoltage += parser.nextDouble() * 0.001;
+        }
+        position.set(Position.KEY_BATTERY, batteryVoltage);
+
+        position.set(Position.KEY_TOTAL_DISTANCE, parser.nextDouble());
```
tananaev: Do not use `KEY_TOTAL_DISTANCE`. It's reserved for Traccar internal use. You should be using `KEY_ODOMETER`.

src/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -88,28 +101,86 @@ protected Object decode(
...
+        position.set(Position.KEY_GPS, parser.nextInt());
+        position.set(Position.PREFIX_ADC + 1, parser.nextDouble() * 0.001);
+        position.set(Position.PREFIX_ADC + 2, parser.nextDouble() * 0.001);
+
+        String cellNetCodeString = parser.next();
```
tananaev: Use standard naming for these (`cid`, `mcc`, `mnc` and `lac`).

src/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -88,28 +101,86 @@ protected Object decode(
...
+        String cellIdString = parser.next();
+        String countryCodeString = parser.next();
+        String operatorCodeString = parser.next();
+        if (cellNetCodeString != null
+                && cellIdString != null
+                && countryCodeString != null
+                && operatorCodeString != null) {
+            int cellNetCode = Integer.parseInt(cellNetCodeString, 16);
```
tananaev: Avoid creating second set of variables. Convert directly in the parameters.

src/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -88,28 +101,86 @@ protected Object decode(
...
+            Long cellId = Long.parseLong(cellIdString, 16);
+            int countryCode = Integer.parseInt(countryCodeString);
+            int operatorCode = Integer.parseInt(operatorCodeString);
+            position.setNetwork(new Network(CellTower.from(countryCode, operatorCode, cellNetCode, cellId)));
+        }
+
         String checksum = parser.next();
+        String result = sentence.replaceAll("^\\$(.*)\\*[0-9a-fA-F]{2}$", "$1");
```
tananaev: What is this for? I suggest we remove checksum checking as we don't use it in other places. TCP has own checksum which is more than enough.

src/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -88,28 +101,86 @@ protected Object decode(
...
+            return Position.ALARM_GEOFENCE_ENTER;
+        } else if (event.equals('T')) {
+            return Position.ALARM_TAMPERING;
+        } else if (event.equals("H")) {
+            return Position.ALARM_POWER_OFF;
+        } else if (event.equals('X')) {
+            return Position.ALARM_GEOFENCE_ENTER;
+        } else if (event.equals('8')) {
+            return Position.ALARM_SHOCK;
+        } else if (event.equals('7') && event.equals('4')) {
```
tananaev: This will always be false? What did you mean?

src/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -88,28 +101,86 @@ protected Object decode(
...
-                channel.write("$AVCFG,00000000,x*2D");
-            }
-
         }
 
         return position;
     }
 
+    private String decodeAlarm(String event) {
+        if (event.equals('Z')) {
```
tananaev: I think it's better to use switch here.

src/org/traccar/protocol/LaipacProtocolDecoder.java
kevingoos: This is the battery voltage. There are two formats for this, you have the old format "3.785" and the new format "3785". They changed the format to not contain the point in the new version. So to support the old an new version, I added an optional point inside the regex.

tananaev: The can be implemented easier. Just remove dots in string and then divide by 1000.

kevingoos:
I think I fixed everything you ask for. And I checked the file changes.

tananaev:
Merged, thanks.

src/org/traccar/protocol/LaipacProtocolDecoder.java
kevingoos: @tananaev I removed it but me and my engineering team are not sure why you wouldn't check it. Is there any special reason why you don't check this?

tananaev: It's about consistency and performance.

If you want to verify checksum, we should add it to all other protocol, not just this one.

Second is performance. It just waste of CPU time. I don't see any value in checksum because TCP and UDP protocol already have a better checksum. For example, you don't see checksum in HTTP protocol because it's an application level protocol and integrity of the data should be checked on the transport layer (which is TCP).

tananaev:
I have done some further clean up:

https://github.com/tananaev/traccar/commit/d2a130b71e9d8064515b16dff865f476a1aadf15

PR #3815, opened 2018-03-20
Add user objects to Event forward payload

### LINKED ISSUE #3814: [Feature request] Per user event forwarding / simplify PUSH implementation
Looks like this feature rather demanded
https://github.com/tananaev/traccar/issues/3164
https://github.com/tananaev/traccar/issues/3757
https://github.com/tananaev/traccar/issues/3707
and others

I see it as fourth transport (called "Push" may be) in user interface and simple POSTing json to configured URL on backend. I believe it is easy to implement some PHP/perl/ruby/whatever page to proxy it to desired push service or FCM or even SMS via http
Initially I thought about implementing it as fallback for SMS but I believe we need separate set of templates for formatting: most push services support "tittle" and we should not economize letters as with SMS, but still not so long as emails. And some services can handle links differently.

Also I think it would be useful to post some user details with such object, here is draft of json object to POST:
This part is extracted from user object:
- `login` (why not to integrate with LDAP)
- `email`
- `phone`
- `serviceId` (value of configured user attribute, I think some services distinguish users by token, jid, internal id or something)

This part is constructed with template:
- `tittle`
- `body`
- `link`

Abyss777:
This change will allow to make traccar integration a bit more simpler 
The set of users is the same as traccar uses to notify about forwarded Event.

fix #3814

### DIFF
--- a/src/org/traccar/database/NotificationManager.java
+++ b/src/org/traccar/database/NotificationManager.java
@@ -73,9 +73,16 @@ public class NotificationManager extends ExtendedObjectManager<Notification> {
 
         long deviceId = event.getDeviceId();
         Set<Long> users = Context.getPermissionsManager().getDeviceUsers(deviceId);
+        Set<Long> usersToForward = null;
+        if (Context.getEventForwarder() != null) {
+            usersToForward = new HashSet<>();
+        }
         for (long userId : users) {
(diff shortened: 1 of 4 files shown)


tananaev:
Merged, thanks.

PR #3822, opened 2018-03-25
method to calculate distance from center of circle

ivanfmartinez:
This method is used in my mqtt integration with traccar which is being refactored to be published


tananaev:
Merged.

Issue #3111, opened 2017-04-21
SMS HTTP protocol support


himsashu:
Hi Tananaev
Love to know that traccar now support SMPP protocol for sending SMS notification in latest version. I have two question regarding SMS notification:

1. Can be be possible to support HTTP protocol for sending SMS. As it is more easy to implement and easy to customize. In my case, I have a SMS gateway which has very easy API. The sample of http gateway is as below:
**http://GATEWAY_IP_ADDRESS/unified.php?usr=USERNAME&pwd=PASSWORD&sndr=SENDERID**&ph=MOBILE_NUMBER&text=SMS_MESSAGE.

In above, the BOLD link is fixed which can be stored in each account and MOBILE_NUMBER and SMS_MESSAGE can be generated by Traccar. So we may add this link to each account which need SMS alert.

2. Do traccar save the SMS notifications or Email notification in database so that it can be tracker that how much email or sms is consumed by each account. A report can be generated.

Hopefully my above suggestions may be useful for other users as well.

tananaev:
1. The URL seems to be specific for some provider that you want to use. It's not a universal solution. Also there is already a thread with exactly the same request.

2. Events are saved in the database, so you can calculate how many emails and SMS messages were sent knowing user settings.

tananaev:
We already support HTTP SMS. In fact it's the only option we support now.

Issue #3636, opened 2017-11-13
Feature Requests HTTP SMS


kjetilpp:
Hello! 
I miss a function in traccar.

In gps-server.net you can use HTTP Api from "any" bulk sms provider. 
See: ![skjermbilde](https://user-images.githubusercontent.com/5620578/32735910-61751cce-c896-11e7-9789-9f8da3a0aaaa.PNG)

Perhaps this could be implemented together with SMPP, that already exists?

tsmgeek:
SMPP already exists, look at the configuration file https://www.traccar.org/configuration-file/

tananaev:
@tsmgeek, I think you have misread something.

tsmgeek:
Yup I did, woops.
#3624 seems to be part way to this but he will not implement this as a Notification handler instead.

Shinryuken:
It could be done eventually but GET requests are not supported on the actual implementation.

(the example shown is sending passwords using a GET?, really?)

pc500:
This actually appears to be good idea.  Ideally, we'd all use SMPP, but the commercial availability of this product is mostly inter-carrier in nature and not carrier-to-enterprise. 

The APIs are fairly simplistic.  Almost all offer http get and post for simplification, and most offer JSON if one wants to do it correctly.    Since many non-coders are writing scripts to use there things, and you are sending a 160 character message, there is nothing that horrible about encoding it in the HTTP request:

EG:
https://api.onesimcard.com/corp.php?user=XXXX&key=XXXX&command=sms&simcard=372

Support for a couple of the basic large providers  (IE: Twileo) for the proper JSON APIs would be great, but if you offer a way to simply to a http get/post and perform variable substitution based on attributes in the DB already for the user, it shouldn't be too bad.

Passwords over get are OK as the url is encoded in SSL.  However, keep in mind, this is only SMS.

kjetilpp:
Yes, that's one of the reasons I want this feature. It is not easy to get an SMPP account if you do not send over 20000 SMS in a month.

tsmgeek:
To throw a spanner in the works #3651 seems to want the ability (per user) basis to forward SMS messages to an email address for Email2SMS gateways. This is not the same as sending emails as they use different templates.

mauriceoduor:
[...]

Issue #3613, opened 2017-10-27
SMS without SMPP


veniciosn:
Our service to send SMS use the following:

http://54.173.24.177/painel/api.ashx?action=sendsms&lgn=LOGIN&pwd=PASS
&msg=Message%20ToSend&numbers=3199999999

They do not have smpp support. Can I use this URL to send SMS on events?

Thanks for the great work.

tananaev:
Unfortunately not.

tsmgeek:
Could the SMS provider be altered through config so a choice could be made between SMPP/HTTP POST?
What I mean is would you object if a change was made to allow for this if the module was created?

tananaev:
No, I don't have any objections to that.

filosganga:
Would be good to support AWS SNS as well

joseph-santos:
hello someone can help me to install a q24 modems to the traccar platform

tananaev:
We support HTTP-based SMS providers and AWS SNS.

PR #3823, opened 2018-03-26
refactor to make possible to use other sms provider

ivanfmartinez:


ivanfmartinez:
this will make easier to implement a class to close  #3111 #3613 and #3636

tananaev:
I think it would be better to generalise not just SMS manager, but all notification options.

ivanfmartinez:
I agree that the full notifications can be refactored, but in this case the SMS is an notification type and my refactor is just to select the provider for send the SMS.

This simple refactor can make possible to possible to attach new SMS providers.

To make all notifications attachable we must make structural changes to database and web UI also.

tananaev:
I still think we should implement universal solution for all notifications, not just SMS.

ivanfmartinez:
probably you should create a table for the notification methods, and this table can have the definition including the class to be used to process the notification.

the tables which currently have email, sms and web setting should be changed to use this notification 
methods reference and also the destination for the notification method.

At this moment I can see extra notification types :
 MQTT 
 URL call with GET or POST

But the SMS are also used for sending configuration to the devices, and in this case the SMS should still have the getSmsManager method in the Context class to send this commands to the devices.

This PR is simple and make possible to use other SMS providers by just defining one new class and keeping full compatibility with current versions. The refactor for notifications is a good thing to be done (I like this idea), but requires more planing because will have DB and UI changes .

tananaev:
I don't think we need to store classes in the database. Configuration file should be enough.

ivanfmartinez:
The current config only read by key, in this case we need an notification definition and then the class like this ?

    <entry key='notification.types'>sms,mqtt,get</entry>
    <entry key='notification.sms.class'>org.traccar.sms.SMSNotitificatior</entry>
    <entry key='notification.mqtt.class'>org.traccar.mqtt.MQTTNotitificatior</entry>
    <entry key='notification.get.class'>package.HttpGetNotitificatior</entry>

In this case we cannot ensure the integrity of the definitions in the database if someone remove the "mqtt"  notification type and there still an configured notification using this type.

The definitions like description will be at database or in the config xml ?

tananaev:
Why do we need anything in the database? I would imagine that set of available notification providers would be fixed for a server, so we don't need to keep anything in the database. Definition can also be in the config.

ivanfmartinez:
The user has to decide where the notifications will go. 

The notifications table has columns web, mail, sms that are direclty linked to the code in traccar. 

Creating generic notifications this has to be changed to relate the notifications record to the notifications implementations.

tananaev:
Of course we would need user notifications config in the database, but configuration on notification providers themselves can be static in the config.

ivanfmartinez:
traccar 4.3 has support for the configurable notitfication providers and SMS http providers

PR #3830, opened 2018-04-01
notifications refactor, generic internal representation

ivanfmartinez:
@tananaev  what you think about this instead of #3823 

It's just a first step without making changes in DB and UI.

### DIFF
--- a/schema/changelog-3.18.xml
+++ b/schema/changelog-3.18.xml
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd"
+  logicalFilePath="changelog-3.18">
+
+  <changeSet author="author" id="changelog-3.18">
+
+    <addColumn tableName="notifications">
+      <column name="transports" type="VARCHAR(128)" />
+    </addColumn>
+
+    <update tableName="notifications">
+        <column name="transports" value="web, mail, sms" />
+        <where>web = 1 AND mail = 1 AND sms = 1</where>
+    </update>
+    
+    <update tableName="notifications">
+        <column name="transports" value="web, mail" />
+        <where>web = 1 AND mail = 1 AND sms = 0</where>
+    </update>
+    
+    <update tableName="notifications">
+        <column name="transports" value="web" />
+        <where>web = 1 AND mail = 0 AND sms = 0</where>
+    </update>
+
+    <update tableName="notifications">
+        <column name="transports" value="web, sms" />
+        <where>web = 1 AND mail = 0 AND sms = 1</where>
+    </update>
+
+    <update tableName="notifications">
+        <column name="transports" value="mail, sms" />
+        <where>web = 0 AND mail = 1 AND sms = 1</where>
+    </update>
+
+    <update tableName="notifications">
+        <column name="transports" value="mail" />
+        <where>web = 0 AND mail = 1 AND sms = 0</where>
+    </update>
+
+    <update tableName="notifications">
+        <column name="transports" value="sms" />
+        <where>web = 0 AND mail = 0 AND sms = 1</where>
+    </update>
+    
+    <!-- 
+    <dropColumn tableName="notifications" columnName="web" />
+    <dropColumn tableName="notifications" columnName="mail" />
+    <dropColumn tableName="notifications" columnName="sms" />
+     -->
+
+  </changeSet>
+</databaseChangeLog>
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -18,6 +18,9 @@
     <entry key='logger.level'>all</entry>
     <entry key='logger.file'>./logs/tracker-server.log</entry>
 
+    <entry key='filter.enable'>true</entry>
+    <entry key='filter.future'>3600</entry>
+
     <entry key='event.enable'>true</entry>
     <entry key='event.ignoreDuplicateAlerts'>true</entry>
     <entry key='processing.computedAttributes.enable'>true</entry>
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -27,6 +27,11 @@
 
     <entry key='media.path'>./media</entry>
 
+    <entry key='notificator.types'>web,mail</entry>
+    <entry key='notificator.web.class'>org.traccar.notification.NotificationWeb</entry>
+    <entry key='notificator.sms.class'>org.traccar.notification.NotificationSms</entry>
+    <entry key='notificator.mail.class'>org.traccar.notification.NotificationMail</entry>
+
     <entry key='server.statistics'>https://www.traccar.org/analytics/</entry>
 
     <!-- DATABASE CONFIG -->
--- a/src/org/traccar/BaseProtocol.java
+++ b/src/org/traccar/BaseProtocol.java
@@ -92,18 +92,18 @@ public abstract class BaseProtocol implements Protocol {
 
     @Override
     public void sendTextCommand(String destAddress, Command command) throws Exception {
-        if (Context.getSmppManager() != null) {
+        if (Context.getSmsManager() != null) {
             if (command.getType().equals(Command.TYPE_CUSTOM)) {
-                Context.getSmppManager().sendMessageSync(destAddress, command.getString(Command.KEY_DATA), true);
+                Context.getSmsManager().sendMessageSync(destAddress, command.getString(Command.KEY_DATA), true);
             } else if (supportedTextCommands.contains(command.getType()) && textCommandEncoder != null) {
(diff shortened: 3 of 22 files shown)


tananaev:
@Abyss777, can you please review the code? I had a brief look and it mostly looks fine except for some minor issues, but you have more experience with notifications.

Abyss777:
I do not see an issues in logic. May be some minor issues with naming and style.

Do we have any reason to make other two types of notifications "replaceable" ?

ivanfmartinez:
One replacement for standard javamail is the use of SNS/SES : https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notifications-via-sns.html its very easy to use and can be defined just by the permissions for the EC2 machine, without manual configuration for user/password.

The only one that I cannot see a replacement is the web notification, but its easier to make all using same logic and in the future if the UI is changed will be easy to replace the logic.

ivanfmartinez:
How do you think the configurations can be done to make possible to use other notifications ? 

I want to move my mqtt processing logic to use this notificators because will make possible to work with all traccar events.

tananaev:
We don't really need to come up with examples on replacements. I think it's just a good idea to have flexible options. Also, ideally we need to think of a flexible way to add new notification types (e.g. chat or push notifications).

I don't really see any benefit of MQTT. People keep talking about it, but no-one can properly explain why we need it. I see it only as an extra complexity. Will a person with limited technical knowledge be able to set it up? I doubt it.

ivanfmartinez:
MQTT in an internal network is very easy. Just put the server on and point your publishers and subscribers.  No username/password, you decide your topics, I think is very easy for any user.

Using this integration I can make my openhab know when I'm at home without making traccar talk to my openhab, traccar just publish the states and openhab read the states in a unified bus.

Probably like you I don't like the idea of MQTT as public and available on the internet (which many people want to do), this will make more complicated, needing passwords, with many headaches for security.

Apps like mqttwarn https://github.com/jpmens/mqttwarn can translate data published in the topics to many other services without need to code inside.

Abyss777:
I would leave everything as is, just make web/email/sms classes replaceable, like `sms.manager.class`.
Renaming it is not a big deal.

Now we have three essential classes of notification:
- Web - online, for most of all notifications, just online
- Email - not urgent information, mostly for archive
- Sms - urgent type of notifications

Do we need an another class of notifications? Not transport but class.

There are a lot of wanted transports (push, telegram, jabber, may be some ECM) but I believe in most cases 3 classes enough.

Just thoughts, may be in future make it possible to select transports for every class in user/server attributes.

ivanfmartinez:
I think the user should decide which notification types it can use.

As an example here in Brazil the SMS providers has two options : promotional (less reliable) and transactional, the cost as different, delivery time and other things.

As a traccar user I want to define some notifications to be as sms_promotional and some as sms_transactional dependending of device, for some devices I want full reliablility for others it does not matter and I can use the cheaper.

The best provider for transactional is not the same as the transactional, if I can create as many notification types I want will be easy to implement the changes to support this .

Also I can consider telegram as reliable for urgent for some users/devices but not for others.

Please let the user decide which notification methods are better for each case.

Abyss777:
We can remove web/email/sms from Notification object and add some string field like `transport`. One Notification will be responsible for one event and one transport.

I believe initial configuration will be a pain, user need to create lots of Notification objects to enable all.

ivanfmartinez:
will be a lot of notification definitions but will make possible to a more granular configuration and expand to any number of transports.

Thinking about future I would suggest something like have two fields :

transport and options

mail    "EMPTY  OPTIONS"  - use default user email
mail    "something@domain" - use this specific email for this notification
sms   "EMPTY OPTIONS"  - use default user email
sms   "+111111" - use specific phone number.

Will be better if the transport class provides access to an json schema and the options can be a json based on the schema. 

We can have configuration like :
sms  { "type" : "transactional", "destination" : "+111111" }

The json appears to be difficult for users, but we can hide the json using this library : https://github.com/json-editor/json-editor  I have used it in some projects and works fine, I have the schema available on the webserver and the fields and created dynamically based on the schema file.

https://rawgit.com/jdorn/json-editor/master/examples/css_integration.html

tananaev:
I completely agree with @Abyss777. It makes sense to have a flexible option to select transport.

I don't agree with last comment here. We already have use email and phone configuration. I believe that's enough and we don't need another option to set it. We just need to provide user model to specific notificator and it can extract whatever it needs to deliver notification (e.g. email or phone).

Abyss777:
I just thought we can make this `transport` field comma-separated list, to not make users add a lot of Notification objects.

Also we need some API to enumerate currently available transports.

Have no idea for now how to configure unknown number of "notificators" in config file, may be also with friendly names.

ivanfmartinez:
What you think about

    <entry key='notificator.types'>web,sms,mail,mqtt</entry>
    <entry key='notificator.sms.class'>className</entry>
    <entry key='notificator.mail.class'>className</entry>
    <entry key='notificator.mqtt.class'>className</entry>

I sugest also to pass the "type" for the class constructor, this will make possible to have also generic configuration for each notificator as 
    <entry key='notificator.mqtt.url'>tcp://....</entry>

Abyss777:
Any news here?
There are so much requests of implementing different sms providers...

ivanfmartinez:
I'm using this implementation on my server, the internal things are working fine. 
But I don't know how  the web interfaces interact with the model to make the UI changes.

This format permits to have multiple dynamic notificators. The SMS default notificator implementation also uses a generic SMSManager. And will be very simple to implement different SMS options.

Abyss777:
UI later.

We need convert 3 fields web/mail/sms to one comma separated string like `transport`
And adjust your code to use list of transports from `notificator.types`

ivanfmartinez:
If you make the changes in the database the changes in my code is just in the method that returns a set of notification transports : https://github.com/traccar/traccar/pull/3830/files#diff-8c3515db6a18762b9cee7586050fb5bcR73

and remove the old methods for SMS, MAIL and WEB.

parveenkumardeeva:
Any update here?

ljporter1723:
+1 vote for built in traccar notifications for push and especially mqtt.  @ivanfmartinez do you have step by step guide to implement your mqtt traccar code for a non coder? SMS and mail is next to useless.

Abyss777:
Looks like there is no movement here.
I'm going to implement general approach discussed here if nobody is working on it.

ivanfmartinez:
@ljporter1723  in the repo : https://github.com/ivanfmartinez/ivanfm-traccar-mqtt  you can just compile the code and add configurations from the README and you can send data to mqtt. I use it in my system, I have an script which get messages from mqtt and send to telegram. You can use https://github.com/jpmens/mqttwarn for this gateway

I have made all the changes I know how to do in the traccar code in the backend. As  I don't have the experience or time learn to make the UI or database changes.

I have made some changes in Apr/09, and no one commented about these. As the PR for traccar need to be exactly how @tananaev  wants, I will not make more changes without specific directions.

In my mind this part can be merged as it currently works without changes in current code / db. This make possible to people start testing and creating their own code for notifications, and more people help.

ivanfmartinez:
@tananaev  the Null notification make easy for any place use notificators without checking if they exist or not. 

If  someone configure an notificator and it does not exist it will not give a null object, and the server will stay running fine without a lot of null checkings.

Abyss777:
- I think it is better initialize `NotificatorManager` in `Context` like we do with all other managers.
- I think we do not need initialize two `SmppClient`. I believe we can use something like `Context.getNotificatorManager.getNotificator("sms") instead of `Context.getSmsManager()`

Abyss777:
Or you have a reason for separate `SmsManager`?

ivanfmartinez:
@Abyss777 where did you find two initiatilization for SmppClient ?

It is just instantiated in the Context.

The SMS is an special case where some devices receives commands by SMS. The SMSManager is the Interface that can be implemented by any SMS implemenation, and the SMSNotification uses this SMSManager implementation for sending SMS messsages.

The Notificator send notifications of events/positions to users, and the SMSManager sends messages to phone numbers.

To replace the SMS implementation the user should define sms.manager.class parameter with the class name for an SMSManager implementation.

Abyss777:
@ivanfmartinez Sorry, you are right. `NotificationSms` initialized secondly and uses `SmsManager`.

Thanks for explanation. That's more clear now for me.

Do you use `NotificatorManager.getMail()` and `getSms()` only for testing? I believe we should test all configured notificators by list. (there might be other types, not only "mail" and "sms")

Abyss777:
Please add the `changelog-3.18` to `changelog-master` and comment out `dropColumn`s from changelog-3.18 if schema migration goes OK. 

General logic is fine to me.

parveenkumardeeva:
@tananaev Would you also consider this pull request with Netty 4?

tananaev:
I don't think this pull request has anything to do with Netty.

parveenkumardeeva:
No, I'm asking for Traccar v4.0 as I read your comments during Netty 4 development that you are going to release next version as Traccar 4.0

tananaev:
That's not what you asked.

parveenkumardeeva:
Sorry my bad, for not describing the question completely.

tananaev:
If it's ready by the time we release then we'll include it. There is no specific plan for it.

parveenkumardeeva:
Ok, Thanks.

PR #3835, opened 2018-04-06
Fix the asynchronous position issue

ssianky:
The asyncronous upstreams are executed last and position is updated in the database afterward.

### DIFF
--- a/src/org/traccar/AsyncDataHandler.java
+++ b/src/org/traccar/AsyncDataHandler.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2018 Anton Tananaev (anton@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


tananaev:
This is incorrect solution. Position has to be set as last only when it's fully processed.

ssianky:
Then you shouldn't write positions into database if any asynchronous precessing fails, but you do that.

ssianky:
Anyway, I don't think there's a better solution for the issue.

tananaev:
We shouldn't we write position if async processing failed? Are you saying losing data is better?

ssianky:
> Position has to be set as last only when it's fully processed.

You said that. So if the asynchronous precessing fails - it's not fully processed and the last position shouldn't be set.

tananaev:
Failed asynchronous request doesn't mean it hasn't been processed. It just means we didn't get information, so whatever data we were requesting will be empty in the position.

ssianky:
Yeah, and in my case it will work the same - if the asynchronous processing fails, it will remain null. The difference is that it won't destroy all other processings.

tananaev:
Failed asynchronous call is a corner case. Successful one is not. If we do geocoding after everything else it's probably not a big deal, but GSM-based location can affect other handlers and needs to be done first.

ssianky:
So I agree that the geolocation was intended as an additional way to find the track's location  so it's not good to move it to the end. I guess I will just remove it altogether since it's useless how it's made.

But the geocode is fine to work at the end. It's not a problem not having the address.

Anyway, the fact that asynchronous precessing fucks all other processings is worse. I don't know if that can be somehow fixed in a better way.

tananaev:
There is a better way. We need to delay processing of next position until first one is fully processed. It's not easy, but it's possible to implement.

tananaev:
Using synchronous network calls is not a good practice.

PR #3837, opened 2018-04-06
add position accuracy to log

ivanfmartinez:


tananaev:
I don't think it's a good idea to clutter log file with more data. Accuracy is not available for most of the data.

ivanfmartinez:
my devices have, my first idea was to check if accuracy > 0 but I followed the schema from other variables.

What you think If I check the speed, course, and accuracy and just print if different of zero ?

tananaev:
That sounds like a good idea.

ivanfmartinez:
As course is Angle and zero is valid, I will do check check for the others and leave course always printed.

PR #3838, opened 2018-04-08
use geofence attribute to permit per polyline distance configuration

ivanfmartinez:


src/org/traccar/model/Geofence.java
```diff
@@ -65,7 +65,9 @@ public void setArea(String area) throws ParseException {
         } else if (area.startsWith("POLYGON")) {
             geometry = new GeofencePolygon(area);
         } else if (area.startsWith("LINESTRING")) {
-            geometry = new GeofencePolyline(area, Context.getConfig().getDouble("geofence.polylineDistance", 25));
+            final double distance = getDouble("polylineDistance");
+            geometry = new GeofencePolyline(area, distance >= 1 ? distance
```
tananaev: Why `>= 1`?

ivanfmartinez:
it can be just > 0, but as I had before some precision errors during floating point comparision I considered this as a good value without precision errors, and non one will use 0.05 as precision for geofence....

What value do you want to be there ?

tananaev:
There is no precision error with zero value, which is probably what you want to check to make sure it's set.

tananaev:
Merged, thanks.

PR #3842, opened 2018-04-11
Add support of multiple Maintenances

### LINKED ISSUE #3548: Feature request: Multiple maintenance
Create different maintenance attributes. For example: oil change, change of tires, anual revision, etc.

Abyss777:
Implemented Maintenance objects. Logic is very similar to Geofence, but used to rise periodic (by distance) events. I think it is easy to extend it to handle engine hours in future.

Extended Event object with `maintenanceId` field to implement similar Event filtering in case user do not have access to particular Maintenance object. Everything is just like with Geofences.

Unfortunately `interval` is reserved word for MySQL, chose `lapse` as synonym.

Also I believe this feature can be used not only for car maintenance, but for example with fitness tracker to rise notification every tripped Km, or something other periodical.

fix #3548 

PS: good news, LibreOffice Calc (my 5.1.6.2) can be used to create xlsx templates now

fix https://github.com/traccar/traccar-web/issues/639


tananaev:
What do you think about making maintenance more flexible in terms of the attributes that it checks? I know many people are interested in engine hours maintenance as well as distance. I believe it can also be just fixed-time maintenance.

Abyss777:
I've added field `type` it is basically name of position attribute that will be analyzed.

There are some restrictions:
- There must be `lastPosition`
- Both old and new values must not be zero

Just to avoid unwanted events if attribute are missed in some positions.

schema/changelog-3.17.xml
```diff
@@ -0,0 +1,73 @@
...
+      <column name="name" type="VARCHAR(4000)">
+        <constraints nullable="false" />
+      </column>
+      <column name="type" type="VARCHAR(128)">
+        <constraints nullable="false" />
+      </column>
+      <column name="start" type="DOUBLE" defaultValueNumeric="0">
+        <constraints nullable="false" />
+      </column>
+      <column name="lapse" type="DOUBLE" defaultValueNumeric="0">
```
tananaev: I think it's better to rename this to interval.

src/org/traccar/database/NotificationManager.java
```diff
@@ -78,8 +78,10 @@ public void updateEvent(Event event, Position position) {
             usersToForward = new HashSet<>();
         }
         for (long userId : users) {
-            if (event.getGeofenceId() == 0 || Context.getGeofenceManager() != null
-                    && Context.getGeofenceManager().checkItemPermission(userId, event.getGeofenceId())) {
+            if ((event.getGeofenceId() == 0 || Context.getGeofenceManager() != null
+                    && Context.getGeofenceManager().checkItemPermission(userId, event.getGeofenceId()))
+                    && (event.getMaintenanceId() == 0
```
tananaev: Do we need to check if manager is null? If not, maybe we should remove geofence manager check?

src/org/traccar/events/MaintenanceEventHandler.java
```diff
@@ -16,54 +16,47 @@
...
 
-        if ((long) (oldTotalDistance / maintenanceInterval) < (long) (newTotalDistance / maintenanceInterval)) {
-            Event event = new Event(Event.TYPE_MAINTENANCE, position.getDeviceId(), position.getId());
-            event.set(Position.KEY_TOTAL_DISTANCE, newTotalDistance);
-            return Collections.singletonMap(event, position);
+        Map<Event, Position> events = new HashMap<>();
+        for (long maintenanceId : Context.getMaintenancesManager().getAllDeviceItems(position.getDeviceId())) {
+            Maintenance maintenance = Context.getMaintenancesManager().getById(maintenanceId);
+            if (maintenance.getLapse() != 0) {
+                double oldValue = lastPosition.getDouble(maintenance.getType());
```
tananaev: I wonder if there is an easy way to add simple time-based maintenance, but I think this pull request is already big enough and we can think about it later if someone requests.

schema/changelog-3.17.xml
tananaev: How about `period`?

src/org/traccar/events/MaintenanceEventHandler.java
tananaev: Agree, let's leave it like it is for now.

tananaev:
Merged, thanks.

PR #3847, opened 2018-04-14
Odometer in reports

ckrey:
Implements Odometer in Summary, Trips, and Stops Reports traccar/traccar-web#652 
relies on corresponding PR in traccar/traccar-web


ckrey:
Includes updates to OwnTracksProtocol

tananaev:
I think you need to split the two things.

PR #3848, opened 2018-04-15
Odometer in reports only

ckrey:
Implements Odometer in Summary, Trips, and Stops Reports traccar/traccar-web#652
relies on corresponding PR in traccar/traccar-web


tananaev:
In stops report shouldn't we just call it odometer?

ckrey:
we can do so but

a) Stop report shares startOdometer and endOdometer members from BaseReport
b) Technically a Stop is a set of positions with a start and end as summary or trip are although the odometer is not expected to change much between start and end

tananaev:
I'm talking about display part. I'm fine with storing it as it is internally.

ckrey:
ok done

tananaev:
Can you please clean up branch history. The list of commits is huge and some of them are completely irrelevant to this pull request.

PR #3859, opened 2018-04-27
org.traccar.DistanceHandler: refactor function roundDistance

matjon:
This was discussed in:
        https://www.traccar.org/forums/topic/strange-casting-and-scaling-of-distance-values-in-distancehandlerhandleposition/

I doubt that this rounding works in all cases as the values may get
converted to e.g. 12345.34999999999.

Full-disclosure: I have not run the code in its current form as I'm currently on 3.14.

### DIFF
--- a/src/org/traccar/DistanceHandler.java
+++ b/src/org/traccar/DistanceHandler.java
@@ -41,6 +41,18 @@ public class DistanceHandler extends BaseDataHandler {
         return null;
     }
 
+    /*
+     * HACK:
+     * Round distance and totalDistance to two
+     * fractional decimal digits to save space when storing these values
+     * in JSON.
+     */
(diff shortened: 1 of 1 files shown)


tananaev:
This doesn't address the issue discussed in the forum thread. It would be nice to round values before saving to the database.

matjon:
Yup, but at least the code is better than it was. At least the hack is explained and made clear.

matjon:
I'm sorry, but I'm not interested in fixing the problem properly now.
I have refactored this as I'm going to push a second PR that adds a new feature to DistanceHandler.

tananaev:
I don't see much value. You just moved rounding to the bottom and separated a method that fits on a single line anyway.

PR #3860, opened 2018-04-27
org.traccar.DistanceHandler: add parameter coordinates.minErrorStatic 

matjon:
This commit adds the following configuration parameter:

coordinates.minErrorStatic

        Distance in meters. Value of `coordinates.minErrorStatic` that
        applies when the speed *as reported by the device* is equal to
        zero (or, more exactly, less then 0.01 knots). By default this
        value is equal to `coordinates.minErrorStatic`.

Using this parameter works very well for filtering out erroneous position changes
when the device is static. As we all know, GPS is imprecise and using it
introduces position errors. When the device is static, every position
recorded by GPS is slightly different and position history - when displayed
on a map - looks like a distinctive "star".
In this situation, most device position updates feature a speed of (exactly) 0.0 knots.
By replacing all positions with a speed of 0.0 knots with last
position, we can make the device appear almost "static" on the map.

It is safe to use much higher values of `coordinates.minErrorStatic` than
values of `coordinates.minError`.

This filtering mechanism works very well in practice. It is not
necessary to make the threshold speed user-configurable.
It worked very well even if the comparison

    speed == 0.0

was used.

This work was funded by Partner Security (www.partnersecurity.pl) but
copyright remains with me (Mateusz Jończyk). Additionally, I have
obtained permission to release my work as open source.

### DIFF
--- a/src/org/traccar/BasePipelineFactory.java
+++ b/src/org/traccar/BasePipelineFactory.java
@@ -204,13 +204,9 @@ public abstract class BasePipelineFactory implements ChannelPipelineFactory {
             pipeline.addLast("hemisphere", hemisphereHandler);
         }
 
-        if (motionHandler != null) {
-            pipeline.addLast("motion", motionHandler);
-        }
+        pipeline.addLast("motion", motionHandler);
 
-        if (distanceHandler != null) {
-            pipeline.addLast("distance", distanceHandler);
-        }
+        pipeline.addLast("distance", distanceHandler);
 
         if (remoteAddressHandler != null) {
             pipeline.addLast("remoteAddress", remoteAddressHandler);
--- a/src/org/traccar/BasePipelineFactory.java
+++ b/src/org/traccar/BasePipelineFactory.java
@@ -131,7 +131,8 @@ public abstract class BasePipelineFactory implements ChannelPipelineFactory {
         distanceHandler = new DistanceHandler(
                 Context.getConfig().getBoolean("coordinates.filter"),
                 Context.getConfig().getInteger("coordinates.minError"),
-                Context.getConfig().getInteger("coordinates.maxError"));
+                Context.getConfig().getInteger("coordinates.maxError"),
+                Context.getConfig().getInteger("coordinates.minErrorStatic", -1));
 
         if (Context.getConfig().getBoolean("processing.remoteAddress.enable")) {
             remoteAddressHandler = new RemoteAddressHandler();
--- a/src/org/traccar/BasePipelineFactory.java
+++ b/src/org/traccar/BasePipelineFactory.java
@@ -204,6 +204,10 @@ public abstract class BasePipelineFactory implements ChannelPipelineFactory {
             pipeline.addLast("hemisphere", hemisphereHandler);
         }
 
+        if (motionHandler != null) {
+            pipeline.addLast("motion", motionHandler);
+        }
+
         if (distanceHandler != null) {
             pipeline.addLast("distance", distanceHandler);
         }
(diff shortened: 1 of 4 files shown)


tananaev:
The whole point of coordinate filtering is to do it when device is static. I don't see why we need second parameter.

matjon:
As I said:

> It is safe to use much higher values of `coordinates.minErrorStatic` than values of `coordinates.minError`.

Using `coordinates.minError` runs the risk of distorting device movements when the speed of the device is small. By taking into account the speed of the device (as reported by the GPS unit) it is possible to create a better classifier.

On our production server we use AFAIK an equivalent of `coordinates.minErrorStatic = 500 meters` and it's doing very well. Obviously we could not use `coordinates.minError == 500 meters`.

tananaev:
Then I think we should just add option to use speed for coordinates filtering.

matjon:
How it would work?

matjon:
> Even more generic than speed threshold would be to use motion attribute.

You mean a binary motion attribute: if it is running or not? Are there GPS units that report only if the device is running?

matjon:
OK, I have found `org/traccar/MotionHandler.java`.

Have a nice day, I'll go sleeping now.

matjon:
> Even more generic than speed threshold would be to use motion attribute.

I have pushed some commits to the branch.
Now I use

    boolean isInMotion = position.getBoolean(Position.KEY_MOTION);

to check whether the device is in motion.

tananaev:
OK, now I think we just need to remove `coordinatesMinErrorStatic`. I don't see any point in it.

matjon:
The value of `coordinates.minErrorStatic` is a safeguard, just like `filter.skipLimit`. It is to ensure that we don't filter out a position that has changed much.

After all, the device may have moved between position updates and still be static when reporting the final position. For example, if the GPS unit reports positions every 60 second, the device may have moved with high speed in seconds 10-50 and then be static in second 60. Both position updates would then specify speed = 0, but the position of the device has changed much and should not be filtered out.

tananaev:
Why can't we use `coordinatesMinError` for that?

matjon:
Because `coordinates.minError` is already used for filtering out events when the device is in motion. 

Setting high values of `coordinates.minError` risks movement distortion, but higher values of `coordinates.minErrorStatic` (e.g 500 meters) are safe.

On our production server we use AFAIK an equivalent of coordinates.minErrorStatic = 500 meters and it's doing very well. Obviously we could not use coordinates.minError == 500 meters.

matjon:
It would be possible to turn `coordinates.minErrorStatic` into a switch: if `coordinates.minErrorStatic == true`, then `coordinates.minError` is applied only to situations when the device is static. Such was a behaviour of a previous version of the patch (and what is currently running on our production server). I would claim that such a behaviour is less useful (as it is beneficial to also filter out events when the device reports to be in motion) and results in more complicated code.

It would be nice (i would say) to specify several values of `coordinates.minError` that apply for various speed thresholds. E.g.:

- if the device reports to be static, then apply filtering if position change is < 500 m,

- if the device's speed is <1 kn (which is rare when the device is in motion), then apply filtering if position change is < 50m,

- if the device's speed is higher, then apply filtering if position change is < 10m.

It would also be good to take into account history of the device's movements: if the device in past 10 minutes has been sitting mostly in a single place, we are able to apply higher filtering threshold.

I would claim that filtering out GPS inaccuracies is tricky and requires several approaches.

tananaev:
Nobody wants to filter moving device, so having two parameters for the same things makes absolutely zero sense to me.

matjon:
A device that is static may self-report as moving. It happens frequently. The speed is usually small, < 5 knots.

matjon:
Bump!
I have a mobile phone that is constantly sitting on my desk and reporting position events to a Traccar server on my laptop. Most of the position reports (like 80-90%) are with speed of 0.0 kn, but some of them are with a non-zero speed, like 0.5kn, 1.0kn or even higher (4.4 kn and 7.8kn).

PR #3862, opened 2018-04-28
Make Commands queueing switchable

### LINKED ISSUE #3861: Feature Request: Option to disable Delayed Commands functionality
According to what was discussed in the forum, in some cases it is important to disable the command queue functionality for the devices, also called Delayed Commands.

In some cases, putting the commands in a queue can cause a vehicle to turn off in the middle of a street

Forum Link: https://www.traccar.org/forums/topic/is-possible-disable-the-command-submission-feature-for-queue-delayed-commands/

Abyss777:
There is a reason to make it switchable.

Leaved it enabled in `default.xml`

fix #3861

### DIFF
--- a/src/org/traccar/database/CommandsManager.java
+++ b/src/org/traccar/database/CommandsManager.java
@@ -37,8 +37,11 @@ public class CommandsManager  extends ExtendedObjectManager<Command> {
 
     private final Map<Long, Queue<Command>> deviceQueues = new ConcurrentHashMap<>();
 
-    public CommandsManager(DataManager dataManager) {
+    private boolean queueing;
+
+    public CommandsManager(DataManager dataManager, boolean queueing) {
         super(dataManager, Command.class);
+        this.queueing = queueing;
     }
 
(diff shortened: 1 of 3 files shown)


src/org/traccar/database/CommandsManager.java
```diff
@@ -70,6 +73,8 @@ public boolean sendCommand(Command command) throws Exception {
             ActiveDevice activeDevice = Context.getConnectionManager().getActiveDevice(deviceId);
             if (activeDevice != null) {
                 activeDevice.sendCommand(command);
+            } else if (!queueing) {
+                throw new RuntimeException("Commands queueing is not enabled");
```
tananaev: Maybe we should just return something like "Device is offline"?

Abyss777: Yes, but a bit more correct "not online" it is unknown or offline. And like it was before:
```java
throw new RuntimeException("Device is not online");
```

tananaev:
Merged, thanks.

PR #3879, opened 2018-05-15
Actualize swagger.json

Abyss777:
Actualized `swagger.json`:
- Added Maintenances
- Renamed `admin` field
- Added `poiLayer` to User/Server
- Added `disabled` to Device
- Added `calendarId` to Notification


tananaev:
Merged, thanks.

Issue #2920, opened 2017-02-16
H02 vehicle status field


tananaev:
| Bit | | First | | Second | | Third |
|-|-|-|-|-|-|-|
| 0 | 0 | Illegal door open alarm | 0 | Thief alarm | 0 | Thief alarm |
| 1 | 0 | SOS | 0 | Rob alarm | 0 | Robbery alarm |
| 2 | 0 | Speed ​​alarm | 0 | Over speed alarm | 0 | overspeed speed |
| 3 | 0 | illegal ignition alarm | 0 | undefined | 0 | Illegal ignition alarm |
| 4 | 0 | Entering alarm | 0 | In geo-fence alarm | 0 | Entering alarm |
| 5 | 0 | GPS antenna disconnect alarm | 0 | undefined | 0 | GPS antenna disconnect alarm |
| 6 | 0 | GPS antenna short circuit alarm | 0 | undefined | 0 | GPS antenna short circuit alarm |
| 7 | 0 | Out alarm | 0 | Out geo-fence alarm | 0 | Out alarm |
| 8 | 0 | Door open | 0 | undefined | 0 | Door open |
| 9 | 0 | Armed | 0 | undefined | 0 | Vehicle fortification |
| 10 | 0 | ACC off | 0 | ACC OFF | 0 | ACC off |
| 11 | 0 | Crash alarm | 1 | undefined | 1 | reserves |
| 12 | 1 | keep | 1 | undefined | 1 | reserves |
| 13 | 0 | pump | 0 | undefined | 0 | engine |
| 14 | 0 | Custom alarm | 0 | undefined | 0 | Custom alarm |
| 15 | 0 | Over speed | 0 | undefined | 0 | overspeed |
| 16 | 0 | GPS mistake alarm | 0 | undefined | 0 | GPS receiver fault alarm |
| 17 | 0 | Shock alarm | 1 | Low battery alarm | 1 | reserves |
| 18 |0 | Tilt alarm | 1 | undefined | 1 | reserves |
| 19 | 0 | Use backup battery | 0 | GPS tracker is powered via built in battery | 0 | Terminal by backup battery power supply |
| 20 | 0 | Battery remove alarm | 0 | Main Power disconnect | 0 | Battery removed |
| 21 | 0 | GPS antenna disconnect | 0 | undefined | 0 | GPS antenna disconnect |
| 22 | 0 | GPS antenna short circuit | 0 | undefined | 0 | GPS antenna short circuit |
[...]

PR #3889, opened 2018-05-26
Fix parser for LKGPS (H02) protocol.

maximilianloy:
Properly parse alarms, LBS message and read battery level when available

Prevent status update messages from being filtered and use last position if current is invalid / zero.

[LKGPS protocol.pdf](https://github.com/traccar/traccar/files/2041404/LKGPS.protocol.pdf)

### DIFF
--- a/src/org/traccar/FilterHandler.java
+++ b/src/org/traccar/FilterHandler.java
@@ -227,6 +227,14 @@ public class FilterHandler extends BaseDataHandler {
         }
 
         if (filterType.length() > 0) {
+            if (last != null && !filterZero(last) && !filterInvalid(last)) {
+                if (position.getLong(Position.KEY_STATUS) != last.getLong(Position.KEY_STATUS)) {
+                    position.setLatitude(last.getLatitude());
+                    position.setLongitude(last.getLongitude());
+                    position.setFixTime(last.getFixTime());
+                    return false;
+                }
+            }
 
             StringBuilder message = new StringBuilder();
             message.append("Position filtered by ");
--- a/src/org/traccar/protocol/H02ProtocolDecoder.java
+++ b/src/org/traccar/protocol/H02ProtocolDecoder.java
@@ -25,6 +25,7 @@ import org.traccar.helper.BitUtil;
 import org.traccar.helper.DateBuilder;
 import org.traccar.helper.Parser;
 import org.traccar.helper.PatternBuilder;
+import org.traccar.helper.Log;
 import org.traccar.model.CellTower;
 import org.traccar.model.Network;
 import org.traccar.model.Position;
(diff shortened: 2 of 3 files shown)


src/org/traccar/FilterHandler.java
```diff
@@ -227,6 +227,14 @@ private boolean filter(Position position) {
         }
 
         if (filterType.length() > 0) {
+            if (last != null && !filterZero(last) && !filterInvalid(last)) {
+                if (position.getLong(Position.KEY_STATUS) != last.getLong(Position.KEY_STATUS)) {
```
tananaev: This seems to be duplication of existing functionality from `skipAttributes`.

tananaev:
Did you check it against other H02 protocols? I have spent quite a bit of time investigating it:

https://github.com/traccar/traccar/issues/2920

We can't just take a change that works for one device, but not others. Same applies to battery. Some other devices might report some other value in place of battery. How would we distinguish?

maximilianloy:
We only use devices from one manufacturer, so I could only verify by making sure all tests succeed. But I understand the problem. What about defining a new port and a separate H02 parser?

tananaev:
I don't think duplicating all the code just for that is a good option, if that's what you are suggesting.

src/org/traccar/FilterHandler.java
maximilianloy: As far as I remember I tried with skipAttributes but this resulted in invalid / zero GPS coordinates to be processed.
In our case the tracker goes to sleep and when it wakes up due to movement it sends an alarm message immediately, even if it does not have a fix yet. The alarm event needs to be forwarded to our push handler to inform the user, even if the fix is invalid / zero coordinates.

tananaev: Events work just fine with invalid and zero coordinates.

maximilianloy:
Than I guess the bits would have to be configured via configuration file.
Something like:

h02.alarmIndex = 20
h02....

src/org/traccar/FilterHandler.java
maximilianloy: But will it prevent the zero coordinate from being added to the route of the tracker?

maximilianloy: Now I did some further testing using various configuration options and 3.17 without the modifications in FilterHandler as in this PR.
The goal is to always update the status and attributes for the alarm, but use the last known position instead of an invalid / zero position:

I verified it using this LKGPS protocol heartbeat message with zero coordinates:
```
HEX: 2a48512c343731303030383838352c56312c3136333833312c562c303030302e303030302c4e2c30303030302e303030302c452c3030302e30302c3030302c3039303631382c46464535464246462c3233322c30312c666634312c302c3423
ASCII: *HQ,4710008885,V1,163831,V,0000.0000,N,00000.0000,E,000.00,000,090618,FFE5FBFF,232,01,ff41,0,4#
```

#### coordinates.maxError

I thought the `coordinates.maxError` setting should do the trick. Reading the documentation I understood that whenever the distance is greater than the value and position is invalid, the old position is used.

The relevant settings in the config file:
```xml
    <entry key='coordinates.filter'>true</entry>
    <entry key='coordinates.minError'>5</entry>
    <entry key='coordinates.maxError'>5</entry>
```

Still the message gets fully processed and the location jumps to zero. Either my understanding is wrong, or the message is not read as invalid.

#### filter.skipAttributes

Second attempt was to use `filter.zero` with `filter.skipAttributes`:
```xml
    <entry key='filter.zero'>true</entry>
    <entry key='filter.skipAttributes.enable'>true</entry>
    <entry key='filter.skipAttributes'>alarm,armed,battery,charge,result,status</entry>
```

Again location jumps to zero.
Using only `filter.zero` without `filter.skipAttributes` filters the message, but then also the alarm is not processed.

What else could I try to get the same behaviour as my patch?

tananaev: Skip attributes are applied before filtering, so if everything configured correctly positions with listed attributes won't be skipped.

maximilianloy: If I use skip attributes, it won't be skipped, that is correct. But if the message comes with zero coordinates they will be used too, that is the problem. The attributes should always be updated, but the coordinates only if they are non-zero, otherwise the previous coordinates should be used.
How to configure that?
Thanks!

Issue #3532, opened 2017-09-16
Feature Request: Total Hour for vehicle just like Total Distance 


darshan2210:
Hello Anton,

Request you for the feature for maintain the vehicle Total Hour just like as Total Distance option available right now in server.

This will help us for the vehicle that have the Total Hour instead of Odometer.

Thanks.

tsmgeek:
Working off the ACC signal? Does not guarantee that engine is running though. May need to be a configurable option as some trackers could use a GPIO port or a separate flag for if engine is running.

darshan2210:
Yes, that would be nice idea.

pc500:
Some trackers report "Ignition".  That could be used.  Yes, someone could have the key turned all the way without the engine running, but most won't as the battery would be depleted quick.  It's like GPS computed mileage... "close enough".

Abyss777:
Here is some thoughts about this feature.

Before implementing server side engine hours accumulator based on ignition we need to standardize a attribute.
I believe `KEY_HOURS` is for that.

From the server side it is simpler use milliseconds (just subtract unix time), but we have number of protocols already used this attribute.
- Aplicom
- Astra
- Continental
- Gl200Text
- GoSafe
- Gps103
- Mta6
- Mxt
- Suntech
- Vtfms
- Xirgo

Looks like most of them mean that as hours, I think it is not enough precise to accumulating. Also we already have engine hours in reports and it is in milliseconds. I think it is better make it consistent.

@tananaev what do you think? Or may be better add another attribute for that?

tananaev:
I think we should convert it to milliseconds, even if devices report in much lower precision. If device reports it, we shouldn't override it.

Abyss777:
Looks like here is small issue https://github.com/traccar/traccar/blob/master/src/org/traccar/protocol/GoSafeProtocolDecoder.java#L111

There must be `hours[1]` and `hours[2]`. It is OK, i'll fix it but...
there is a test 
https://github.com/traccar/traccar/blob/master/test/org/traccar/protocol/GoSafeProtocolDecoderTest.java#L14
with `COT:65;20` where `20` do not match documentation
> `Engine hour, time format: <Hour>-<Minute>-<Second>`
> `Range: 0-0-0~99999-0-0`

How will we treat it? Like hours with omitted minutes and seconds, or seconds without hours and minutes?

tananaev:
I think we can assume that it's hours and then fix if there is an issue.

[...]

PR #3894, opened 2018-05-28
Store engine hours in GL200 and GoSafe in milliseconds

Abyss777:
First step for #3532 

Implemented parsing engine hours in GoSafe and similar GL200.

Have not noticed any tests with missed minutes and seconds in GL200 but implemented like in GoSafe, just in case.


tananaev:
Merged, thanks.

PR #3896, opened 2018-05-29
Store engine hours in milliseconds for the rest of protocols

Abyss777:
Next step for #3532 

Assumed that most of protocol reports in hours.

### DIFF
--- a/src/org/traccar/protocol/AplicomProtocolDecoder.java
+++ b/src/org/traccar/protocol/AplicomProtocolDecoder.java
@@ -549,7 +549,7 @@ public class AplicomProtocolDecoder extends BaseProtocolDecoder {
         }
 
         if ((selector & 0x0020) != 0) {
-            position.set(Position.KEY_HOURS, UnitsConverter.millisecondsFromHours(buf.readUnsignedInt()));
+            position.set(Position.KEY_HOURS, UnitsConverter.msFromHours(buf.readUnsignedInt()));
             position.set("serviceDistance", buf.readInt());
             position.set("driverActivity", buf.readUnsignedByte());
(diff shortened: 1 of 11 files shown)


src/org/traccar/protocol/AplicomProtocolDecoder.java
```diff
@@ -549,7 +549,7 @@ private void decodeF(Position position, ChannelBuffer buf, int selector) {
         }
 
         if ((selector & 0x0020) != 0) {
-            position.set(Position.KEY_HOURS, buf.readUnsignedInt());
+            position.set(Position.KEY_HOURS, buf.readUnsignedInt() * 3600000);
```
tananaev: Do you think we should add this to unit converter class?

Abyss777:
Function name rather long, what about `msFromHours` ? Or it is OK?

tananaev:
Yeah, I like `msFromHours`.

Abyss777:
Renamed

src/org/traccar/protocol/MxtProtocolDecoder.java
```diff
@@ -145,7 +145,7 @@ protected Object decode(
             }
 
             if (BitUtil.check(infoGroups, 4)) {
-                position.set(Position.KEY_HOURS, buf.readUnsignedInt());
+                position.set(Position.KEY_HOURS, buf.readUnsignedInt() * 60000);
```
tananaev: Sorry, haven't noticed it before. I think all of the conversions should be using `UnitsConverter`.

tananaev:
Merged, thanks.

PR #3901, opened 2018-05-31
Implement engine hours accumulation

Abyss777:
I believe it is final server side step for #3532 

Main switch is `engineHours.enable` it controls two features:
- Enable `EngineHoursHandler` itself
- Enable engine hours calculation in reports (I think if switch is disabled user do not interested in engine hours at all, we just will not waste time on calculating it)

Handler designed as if device have never reported `hours` or `ignition` attribute will never appear.
Also it works as `copyAttributes` if device reports `hours` not in every position.

One side effect, if device reports `hours` periodically, traccar will still accumulate `hours` in between.

Some comments in code...

### DIFF
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -21,6 +21,8 @@
     <entry key='filter.enable'>true</entry>
     <entry key='filter.future'>3600</entry>
 
+    <entry key='engineHours.enable'>true</entry>
+
     <entry key='event.enable'>true</entry>
     <entry key='event.ignoreDuplicateAlerts'>true</entry>
     <entry key='processing.computedAttributes.enable'>true</entry>
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -21,11 +21,10 @@
     <entry key='filter.enable'>true</entry>
     <entry key='filter.future'>3600</entry>
 
-    <entry key='engineHours.enable'>true</entry>
-
     <entry key='event.enable'>true</entry>
     <entry key='event.ignoreDuplicateAlerts'>true</entry>
     <entry key='processing.computedAttributes.enable'>true</entry>
+    <entry key='processing.engineHours.enable'>true</entry>
 
     <entry key='media.path'>./media</entry>
 
--- a/src/org/traccar/EngineHoursHandler.java
+++ b/src/org/traccar/EngineHoursHandler.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2018 Anton Tananaev (anton@traccar.org)
+ * Copyright 2018 Andrey Kunitsyn (andrey@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar;
+
+import org.traccar.model.Position;
+
+public class EngineHoursHandler extends BaseDataHandler {
+
+    @Override
+    protected Position handlePosition(Position position) {
+        if (!position.getAttributes().containsKey(Position.KEY_HOURS)) {
+            Position last = Context.getIdentityManager().getLastPosition(position.getDeviceId());
+            if (last != null) {
+                long hours = last.getLong(Position.KEY_HOURS);
+                if (last.getBoolean(Position.KEY_IGNITION) && position.getBoolean(Position.KEY_IGNITION)) {
+                    hours += position.getFixTime().getTime() - last.getFixTime().getTime();
+                }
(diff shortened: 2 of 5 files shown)


src/org/traccar/reports/ReportUtils.java
```diff
@@ -263,10 +263,16 @@ private static StopReport calculateStop(
         stop.setSpentFuel(calculateFuel(startStop, endStop));
 
         long engineHours = 0;
-        for (int i = startIndex + 1; i <= endIndex; i++) {
-            if (positions.get(i).getBoolean(Position.KEY_IGNITION)
-                    && positions.get(i - 1).getBoolean(Position.KEY_IGNITION)) {
-                engineHours += positions.get(i).getFixTime().getTime() - positions.get(i - 1).getFixTime().getTime();
+        if (startStop.getAttributes().containsKey(Position.KEY_HOURS)
```
Abyss777: If first and last position contain `hours` we do not need recalculate it.

src/org/traccar/reports/Summary.java
```diff
@@ -44,11 +44,13 @@ private static SummaryReport calculateSummaryResult(long deviceId, Date from, Da
             Position firstPosition = null;
             Position previousPosition = null;
             double speedSum = 0;
+            boolean needCalculateEngineHours = Context.getConfig().getBoolean("engineHours.enable");
```
tananaev: Maybe rename to `engineHoursEnabled`?

src/org/traccar/reports/Summary.java
```diff
@@ -44,11 +44,13 @@ private static SummaryReport calculateSummaryResult(long deviceId, Date from, Da
...
             for (Position position : positions) {
                 if (firstPosition == null) {
                     firstPosition = position;
                 }
-                if (previousPosition != null && position.getBoolean(Position.KEY_IGNITION)
+                if (needCalculateEngineHours && previousPosition != null
```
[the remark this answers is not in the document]
tananaev: Why do we need to calculate hours if we already have it?

src/org/traccar/reports/ReportUtils.java
```diff
@@ -263,10 +263,16 @@ private static StopReport calculateStop(
...
+                && endStop.getAttributes().containsKey(Position.KEY_HOURS)) {
+            engineHours = endStop.getLong(Position.KEY_HOURS) - startStop.getLong(Position.KEY_HOURS);
+        } else if (Context.getConfig().getBoolean("engineHours.enable")) {
+            for (int i = startIndex + 1; i <= endIndex; i++) {
```
tananaev: Do we really need this fallback? Or just say that if you want engine hours enable it in the config?

src/org/traccar/EngineHoursHandler.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    protected Position handlePosition(Position position) {
+        if (!position.getAttributes().containsKey(Position.KEY_HOURS)) {
+            Position last = Context.getIdentityManager().getLastPosition(position.getDeviceId());
+            if (last != null) {
+                long hours = last.getLong(Position.KEY_HOURS);
+                if (last.getBoolean(Position.KEY_IGNITION) && position.getBoolean(Position.KEY_IGNITION)) {
+                    hours += position.getFixTime().getTime() - last.getFixTime().getTime();
+                }
+                if (hours != 0) {
```
tananaev: Do we need this check? Zero hours is also a valid value.

setup/default.xml
```diff
@@ -21,6 +21,8 @@
     <entry key='filter.enable'>true</entry>
     <entry key='filter.future'>3600</entry>
 
+    <entry key='engineHours.enable'>true</entry>
```
tananaev: Would it be better to rename it to something like `processing.engineHours.enable`?

src/org/traccar/reports/Summary.java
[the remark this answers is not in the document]
Abyss777: It is fallback for old data.

src/org/traccar/reports/ReportUtils.java
Abyss777: This fallback for old data when accumulation was not enabled. Like we did for `totalDistance` (in June 2018 we should remove it if I'm not mistaken)

src/org/traccar/EngineHoursHandler.java
Abyss777: It is in case server have vehicles and "not vehicles" devices. If device never reports `hours` and `ignition` the attribute will never be in State panel for such device.

src/org/traccar/reports/Summary.java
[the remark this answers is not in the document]
tananaev: So, it's also temporary, right?

src/org/traccar/EngineHoursHandler.java
tananaev: Makes sense.

src/org/traccar/reports/Summary.java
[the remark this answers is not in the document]
Abyss777: Yes, my bad, I did not mention it in description.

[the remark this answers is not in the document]
tananaev: We should probably have an issue to make sure we remove it at some point. So we don't forget.

tananaev:
Merged, thanks.

PR #3911, opened 2018-06-05
Migrate Wialon and Wondex

Abyss777:
Migrated Wialon and Wondex.

Also want to test with real wondex device tomorrow.

### DIFF
--- a/src/org/traccar/protocol/WondexFrameDecoder.java
+++ b/src/org/traccar/protocol/WondexFrameDecoder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 Anton Tananaev (anton@traccar.org)
+ * Copyright 2013 - 2018 Anton Tananaev (anton@traccar.org)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


src/org/traccar/protocol/WondexFrameDecoder.java
```diff
@@ -36,17 +38,17 @@ protected Object decode(
         if (buf.getUnsignedByte(buf.readerIndex()) == 0xD0) {
 
             // Send response
-            ChannelBuffer frame = buf.readBytes(KEEP_ALIVE_LENGTH);
-            if (channel != null) {
-                channel.write(frame);
+            ByteBuf frame = buf.readBytes(KEEP_ALIVE_LENGTH);
+            if (ctx != null && ctx.channel() != null) {
```
tananaev: You have channel here as a separate variable, so I don't think we need to use context.

src/org/traccar/protocol/WondexFrameDecoder.java
```diff
@@ -36,17 +38,17 @@ protected Object decode(
...
+                ctx.channel().writeAndFlush(new NetworkMessage(frame, ctx.channel().remoteAddress()));
             }
             return frame;
 
         } else {
 
-            int index = buf.indexOf(buf.readerIndex(), buf.writerIndex(), new StringFinder("\r\n"));
+            int index = new AsciiString(buf.array(), buf.readerIndex(), buf.writerIndex(), false).indexOf("\r\n");
```
tananaev: Just a note here. I like this option if buffer is a string, but in some cases it can be a mix of characters and binary data. In that case I like my solution better.

Abyss777:
I have replaced the commit (we will have a lot of commits, do not think extra is good)
Used your variant of CRLF searching.

tananaev:
Merged, thanks.

PR #3913, opened 2018-06-06
Migrate Y, X, W protocols

Abyss777:
Migrated protocols
Added `BufferUtil` helper class.

### DIFF
--- a/src/org/traccar/helper/BufferUtil.java
+++ b/src/org/traccar/helper/BufferUtil.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2018 Anton Tananaev (anton@traccar.org)
+ * Copyright 2018 Andrey Kunitsyn (andrey@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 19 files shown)


src/org/traccar/helper/BufferUtil.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.nio.charset.StandardCharsets;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufUtil;
+import io.netty.buffer.Unpooled;
+
+public class BufferUtil {
+
+    public static int indexOf(String needle, ByteBuf haystack) {
```
Abyss777: In most cases we search from the reader to writer index

src/org/traccar/helper/BufferUtil.java
```diff
@@ -0,0 +1,39 @@
...
+        return ByteBufUtil.indexOf(
+                Unpooled.wrappedBuffer(needle.getBytes(StandardCharsets.US_ASCII)), haystack);
+    }
+
+    public static int indexOf(String needle, ByteBuf haystack, int startIndex, int endIndex) {
```
Abyss777: But sometimes we need to search in limited piece of buffer

tananaev:
Perfect, thanks.

PR #3919, opened 2018-06-08
Rename all tables to avoid using reserved words

Abyss777:
I think it is right to detach renaming to separate changeset with precondition, can't say why but feel it is better.


Abyss777:
Sorry, forget that we have tests for query constructors.

tananaev:
Merged, thanks.

PR #3921, opened 2018-06-08
Migrate part of T protocols

Abyss777:
Here is migrated part of T protocols


tananaev:
Merged, thanks. I'll try to finish remaining tomorrow.

src/org/traccar/protocol/TmgFrameDecoder.java
```diff
@@ -15,34 +15,37 @@
...
 
     private boolean isLetter(byte c) {
         return c >= 'a' && c <= 'z';
     }
 
-    @Override
-    protected Object decode(
-            ChannelHandlerContext ctx, Channel channel, ChannelBuffer buf) throws Exception {
-
-        int beginIndex = buf.indexOf(buf.readerIndex(), buf.writerIndex(), new ChannelBufferIndexFinder() {
```
Abyss777: `ChannelBufferIndexFinder` is removed. There is nothing for replacement.

tananaev: Yeah, I had to replace it with a loop in one other decoder as well.

Issue #3666, opened 2017-12-01
ignition attribute for meitrack


slachtar:
Hello,
traccar is not setting the ignition attribute while meitrack protocol already supports it by checking event code.
if event code is 2 or 3 or 4 then ignition is on, else if event code is 10 or 11 or 12 then ignition is off.

http://www.meitrack.com/cd-download/Protocols/MEITRACK_GPRS_Protocol.pdf (page 9, 10)

Best regards,

tananaev:
Feel free to send a pull request.

PR #3943, opened 2018-06-22
Decode ignition states for meitrack

slachtar:
Hi Anton,
this pull request sets the ignition attribute for meitrack trackers.
It's related to https://github.com/traccar/traccar/issues/3666 and https://www.traccar.org/forums/topic/using-event-number-as-trigger-for-ignition-on-and-off/ 
I'm using it with the latest traccar version and it works great when combined with the attribute copying in the conf file.
Could you please review it?

### DIFF
--- a/src/org/traccar/protocol/MeitrackProtocolDecoder.java
+++ b/src/org/traccar/protocol/MeitrackProtocolDecoder.java
@@ -142,18 +142,20 @@ public class MeitrackProtocolDecoder extends BaseProtocolDecoder {
         switch (event) {
             case 2:
             case 3:
-            case 4:  position.set(Position.KEY_IGNITION, true);
+            case 4:
+                position.set(Position.KEY_IGNITION, true);
                 break;
 
             case 10:
             case 11:
-            case 12:  position.set(Position.KEY_IGNITION, false);
+            case 12:
+                position.set(Position.KEY_IGNITION, false);
                 break;
 
             default:
                 break;
         }
-        
+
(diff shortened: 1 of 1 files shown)


tananaev:
1. I checked documentation and I don't see where it says that all those events are related to ignition.
2. Before asking for review you have to fix all the issues.

slachtar:
Thank you for your comments,
1. in http://www.meitrack.com/cd-download/Protocols/MEITRACK_GPRS_Protocol.pdf, page 9, IMHO event codes 2,3 and 4 indicate ignition on depending of meitrack tracker model. 10, 11 and 12 events codes indicate ignition off (depending of tracker model).

2. It was a formatting issue.

Thank you

tananaev:
1. Those are generic inputs, as far as I understand it. You can't just assume all of them to be ignition. Sorry, but that's incorrect logic.

matlock08:
Hi

I agree with tananev, in term that those are generic inputs and people use them in different ways, however i have the same need to decode the ignition on meitrack devices.

What i would like to discuss is the option to add and attribute to the device that can be used to determine which input is the ignition on and ignition off.

We might retrieve these fields on the decoder as some other protocol does. 

The idea here is that the attribute ignitionOn fields contains the input used to flag that ignition is on

ignitionOn = device.getString("ignitionOn");
ignitionOff = device.getString("ignitionOff");

attributes


We already use some device field like model in order to adjust some calculation at run time.

I would like to know your opinion on this approach.

tananaev:
You can use computed attributes.

PR #3946, opened 2018-06-25
Added validation of other statuses

kevingoos:
Added validation of other statuses
- R: repeating the message when the package was not received
- P: send when GPS receiver didn't move but there is a valid location avaible)

### DIFF
--- a/src/org/traccar/protocol/LaipacProtocolDecoder.java
+++ b/src/org/traccar/protocol/LaipacProtocolDecoder.java
@@ -115,7 +115,8 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
                 .setTime(parser.nextInt(0), parser.nextInt(0), parser.nextInt(0));
 
         String status = parser.next();
-        position.setValid(status.toUpperCase().equals("A"));
+        String upperCaseStatus = status.toUpperCase();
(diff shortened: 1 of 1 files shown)


kevingoos:
Yes I have documentation about this (I cannot include the full documentation because of confidentiality reasons)
But here is the information about the status:

- ‘R’: The position, speed, course and Current Mileage are not real time data, based on GPS device. As reference, the last time meaningful group of real time data based on GPS device is repeated here. UTC Time and Date are provided by unit own real time clock. It is the time when waypoint or event triggering report is generated.
- ‘r’: Keep its original definition with ‘R’ and ask Server’s acknowledgement
- ‘P’ If device is able to acquire valid real time data from its GPS receiver continuously and without movement for a while, device may turn off its GPS receiver to save power. In this case, when waypoint or event triggering report is generated, it will upload the last set of valid data got from GPS receiver before it turn off the receiver. UTC Time and Date are provided by device own real time clock or get from GSM network. It is the time when waypoint or event triggering report is generated.
- ‘p’: Keep its original definition with ‘P’ and ask Server’s acknowledgement

If you only use A and a, you will not get enough valid location to work with properly so that's why I changed this. And these are actually also valid locations.

tananaev:
Merged, thanks.

PR #3947, opened 2018-06-25
Send queued commands on status online

Abyss777:
You are right, we should not lock to active device adding/removing.
It is fine if we send queued commands when device become online.

There is no good way to determine if udp device on other side, just try to send command to "unknown" device.

I think we can add switch with logic like "queue commands if device status unknown" in future if there will be requests from users.


tananaev:
Merged. Yes, I think we should have some configuration option.

PR #3949, opened 2018-06-25
EasyTrack Status Ignition

michaelloliveira:
Hi! 
I'm back.

Tested in device.

strong hug!


tananaev:
Do you have documentation?

tananaev:
OK. Can you please change the code to use `BitUtil` instead of mask checking. That's what we do it all other decoders.

PR #3950, opened 2018-06-26
Notification refactor to allow custom "notificators"

Abyss777:
Here is continue of #3830 i believe it is worth to be in next release.
Most of the code from @ivanfmartinez 

I fixed conflict, cleaned it up and adjusted to project code style agreements (hope did not forget something). Mostly tested.

Renamed `templates/mail`  to `templates/full` and `template/sms` to `template/short` the idea is that we do not need to lock on transport names. Custom notificator developer can use "full" message formatting if he/she needs a long message with subject, or "short" if he/she needs just a condensed string.

Decided to use term "notificator" everywhere even in Notification model (instead of "transports").

PS: web part is also ready.


setup/default.xml
```diff
@@ -28,6 +28,11 @@
 
     <entry key='media.path'>./media</entry>
 
+    <entry key='notificator.types'>web,mail</entry>
+    <entry key='notificator.web.class'>org.traccar.notification.NotificationWeb</entry>
```
tananaev: I think we should make classes default in the code, but configurable.

src/org/traccar/sms/SMSManager.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.sms;
+
+public interface SMSManager {
```
tananaev: This is wrong naming convention. Should be `SmsManager`.

src/org/traccar/sms/SMSException.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.sms;
+
+
+import org.traccar.notification.NotificationException;
+
+public class SMSException extends NotificationException {
```
tananaev: Naming convention.

src/org/traccar/notification/NotificatorManager.java
```diff
@@ -0,0 +1,77 @@
...
+
+
+    public Set<Typed> getAllNotificatorTypes() {
+        Set<Typed> result = new HashSet<>();
+        for (String notificator : notificators.keySet()) {
+            result.add(new Typed(notificator));
+        }
+        return result;
+    }
+
```
tananaev: Please remove extra spaces here and in other places.

src/org/traccar/notification/NotificatorManager.java
```diff
@@ -0,0 +1,77 @@
...
+        for (String type : types) {
+            final String className = Context.getConfig().getString("notificator." + type + ".class", "");
+            if (className.length() > 0) {
+                try {
+                    final Class<Notificator> clazz = (Class<Notificator>) Class.forName(className);
+                    try {
+                        final Constructor<Notificator> constructor = clazz.getConstructor(new Class[]{String.class});
+                        notificators.put(type, constructor.newInstance(type));
+                    } catch (NoSuchMethodException e) {
+                        // No constructor with String argument
```
tananaev: Unnecessary comment.

src/org/traccar/notification/NotificationSms.java
```diff
@@ -20,33 +20,42 @@
...
+import org.traccar.sms.SMSException;
 
-import com.cloudhopper.smpp.type.RecoverablePduException;
-import com.cloudhopper.smpp.type.SmppChannelException;
-import com.cloudhopper.smpp.type.SmppTimeoutException;
-import com.cloudhopper.smpp.type.UnrecoverablePduException;
+public final class NotificationSms extends Notificator {
 
-public final class NotificationSms {
+    private final SMSManager sms;
```
tananaev: I would recommend to rename variable to `manager` instead. SMS is just a message, so name is confusing.

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -88,16 +88,16 @@ public static Template getTemplate(Event event, String path) {
...
         String formattedMessage = formatMessage(velocityContext, userId, event, position, templatePath);
 
-        return new MailMessage((String) velocityContext.get("subject"), formattedMessage);
+        return new FullMessage((String) velocityContext.get("subject"), formattedMessage);
     }
 
-    public static String formatSmsMessage(long userId, Event event, Position position) {
-        String templatePath = Context.getConfig().getString("sms.templatesPath", "sms");
+    public static String formatShortMessage(long userId, Event event, Position position) {
+        String templatePath = Context.getConfig().getString("message.short.templatesPath", "short");
```
tananaev: Not really related to this pull request, but I think we should just assume "full" and "short" sub-paths here. Base path can still be configurable.

src/org/traccar/sms/SMSException.java
tananaev: Also, do we really need a separate SMS exception class? Can't we just use `NotificationException`?

Abyss777: I'm not sure, we use SMS also for commands. Would it be good to combine it?

tananaev: Then it shouldn't inherit from `NotificationException`. Or maybe we should rename `NotificationException` to something more generic.

Abyss777: I tied myself into knots. Could not make up better than `SendingException`. `CommunicationException` already used somewhere in ldap.
Removing inheritance will produce unneeded try/catch to convert one to another.

tananaev: Is there any specific reason why we need our own exceptions in the first place? Could we just use `RuntimeException` or some other standard `Exception` type?

Abyss777: I think there is no reason... Runtime is to generic I think, what about `IOException` ?

tananaev: Maybe just create a single `MessageException` or `TransmissionException` or something like that?

Abyss777:
Looks like made all changes and others:

Simplified configuration and class instantiation:

`sms.enable` - enables global sms manager used for commands only or both commands and notifications. Used our smpp implementation by default, or can be overridden by `sms.manager.class`

if there is `sms` in `notificator.types` then by default used our `NotificationSms` that can use global sms manager or own overridden `notificator.sms.manager.class`.
Such flexible configuration is reasonable, for example modem can be used to send commands/receive answers (because it has a number) and some web service to send sms notifications because it is cheaper.

src/org/traccar/notification/NotificatorManager.java
Abyss777: I do not remember, have we discussed empty line before last `}` in class? Sometimes it looks better with it.

tananaev: I'm not talking about one line. There were 2 or 3 spaces.

tananaev:
OK, it looks fine now. I'm assuming that we need to wait for web app changes before we can merge this, correct?

PR #3978, opened 2018-07-18
Added integer checking for valid in osmand

kevingoos:
I added a check for 0 or 1, because the `&valid=1`.
Now it will set the valid correct in this case also...


tananaev:
First of all, comparison is incorrect. Second, I would prefer something more elegant and concise.

tananaev:
Merged, thanks.

PR #3996, opened 2018-08-07
Fix for notifications migration on postgresql and mssql

### LINKED ISSUE #3993: PostgreSQL changelog issue
Looks like upgrade doesn't work well on PostgreSQL:

https://www.traccar.org/forums/topic/versiya-40/page/2/#post-33622

@Abyss777, any ideas how to solve it?

Abyss777:
Made separate migrations for h2, mysql, mssql(`common`) and postgresql (`pg`).
Split on `pre`, "main" and `post`.
Added preconditions in case somebody already successfully migrated to 4.0 for future.

It might looks a bit complicated but I hope be universal.

fix #3993

### DIFF
--- a/schema/changelog-4.0.xml
+++ b/schema/changelog-4.0.xml
@@ -53,11 +53,29 @@
 
   </changeSet>
 
-  <changeSet author="author" id="changelog-4.0">
+  <changeSet author="author" id="changelog-4.0-pre">
+    <preConditions onFail="MARK_RAN">
+      <not>
+        <columnExists tableName="tc_notifications" columnName="notificators" />
+      </not>
+    </preConditions>
 
     <addColumn tableName="tc_notifications">
       <column name="notificators" type="VARCHAR(128)" />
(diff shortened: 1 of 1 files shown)


schema/changelog-4.0.xml
```diff
@@ -94,8 +112,69 @@
...
+    </update>
+
+    <update tableName="tc_notifications">
+        <column name="notificators" value="sms" />
+        <where>web = FALSE AND mail = FALSE AND sms = TRUE</where>
+    </update>
+
+  </changeSet>
+
+  <changeSet author="author" id="changelog-4.0-post">
```
tananaev: Can we just name this one `changelog-4.0`? So that we don't have one extra record?

tananaev: It would also mean that we don't need a check.

Abyss777:
You are right, renamed it.

tananaev:
Merged, thanks.

PR #4001, opened 2018-08-09
Resave export templates in MS Office

### LINKED ISSUE #3999: Trips report not working on 4.0
When running a trips report it exports fine but once I open it and says there is a problem with some content in the report and if you try a recovery it doesn't work.

Thanks

Abyss777:
Just resaved in MS Excel 2013

fix #3999


tananaev:
Merged, thanks.

Issue #3707, opened 2018-01-06
Feature Request: implement push notification with FCM


tecufanujacu:
I know that this has been discussed already many times but I'd like to suggest to implement the push notifications as [@pliablepixels](https://github.com/pliablepixels) has implemented the [zmeventserver](https://github.com/pliablepixels/zmeventserver) to add the push notifications in [Zoneminder](https://github.com/ZoneMinder/ZoneMinder) for [zmNinja](https://github.com/pliablepixels/zmNinja).

With the new Google's Firebase Cloud Messaging (FCM) infrastructure should be really simple.

atiqsamtia:
Hey, I have implemented the FCM with single tokens and tested with multiple users check out my pull request here #4002

tecufanujacu:
Hi there, I have just read the documentation to use the the push notifications, and as I have understood I need to compile my own mobile app. No problem for me but I still suggest to implement the push notifications as [@pliablepixels](https://github.com/pliablepixels) has implemented the [zmeventserver](https://github.com/pliablepixels/zmeventserver) to add the push notifications in [Zoneminder](https://github.com/ZoneMinder/ZoneMinder) for [zmNinja](https://github.com/pliablepixels/zmNinja).

I use Zoneminder and for push notifications I sue zmeventserver and zmNinja in which i simply have to add the web address of my own server with Zoneminider, I don't have to compile my own app.

PR #4002, opened 2018-08-09
Integrated FCM Notifications

atiqsamtia:
A much-requested feature to implement **Firebase Cloud Messaging** Notifications

Requested here #3707 , here #3814 

I have implemented the Feature without using FCM library so no extra burden, simple HTTP request just like event forwarding.

**How to use**
o enable, simply put "fcm" in "notificator.types"  and Your Server Key in "notificator.fcm.serverkey" in your configuration file.

debug.xml will look like

```
<entry key='notificator.types'>web,mail,fcm</entry>
<entry key="notificator.fcm.serverkey">YOUR_SERVER_KEY</entry>
```

You can get your Server key from here : [https://stackoverflow.com/a/37427911](https://stackoverflow.com/a/37427911)

A new field is added in user object named  "fcm" This is the Every user token to deliver the notifications to. This can be updated from Rest API just like you update the user.

### DIFF
--- a/schema/changelog-4.0.xml
+++ b/schema/changelog-4.0.xml
@@ -173,4 +173,11 @@
 
   </changeSet>
 
+    <changeSet id="addFcm" author="atiq">
+        <addColumn tableName="tc_users">
+            <column name="fcmtoken" type="VARCHAR(255)">
+            </column>
+        </addColumn>
+    </changeSet>
+
 </databaseChangeLog>
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -28,11 +28,13 @@
 
     <entry key='media.path'>./media</entry>
 
-    <entry key='notificator.types'>web,mail</entry>
+    <entry key='notificator.types'>web,mail,fcm</entry>
 
     <entry key='server.statistics'>https://www.traccar.org/analytics/</entry>
 
     <entry key='commands.queueing'>true</entry>
+
+    <entry key="notificator.fcm.serverkey"></entry>
 
     <!-- DATABASE CONFIG -->
 
--- a/src/org/traccar/model/User.java
+++ b/src/org/traccar/model/User.java
@@ -17,6 +17,7 @@ package org.traccar.model;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
+import org.traccar.Context;
 import org.traccar.database.QueryExtended;
 import org.traccar.database.QueryIgnore;
 import org.traccar.helper.Hashing;
@@ -273,4 +274,29 @@ public class User extends ExtendedModel {
         return Hashing.validatePassword(password, hashedPassword, salt);
     }
 
+    private String fcmToken;
+
+    @JsonIgnore
+    public String getFcmToken() {
+        return fcmToken;
+    }
+
+    public void setFcmToken(String fcmToken) {
+        this.fcmToken = fcmToken;
+    }
+
+    @QueryIgnore
+    public String getFcm() {
+        return fcmToken;
+    }
+
+    public void setFcm(String fcmToken) {
+        if(fcmToken != null && !fcmToken.isEmpty())
+            this.fcmToken = fcmToken;
+        else{
+            User user = Context.getPermissionsManager().getUser(getId());
+            this.fcmToken = user.getFcmToken();
+
+        }
+    }
 }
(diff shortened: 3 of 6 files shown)


tananaev:
Please never ever spam like you did on GitHub. There's no need to comment on every thread that might be related to your pull request.

atiqsamtia:
Sorry, i thought it will help them if they need.

tananaev:
If it's integrated in Traccar, there will be a newsletter with this information for everyone who's interested. Spamming everywhere about a pull request that's not even ready to be merged is clearly not a good idea. Also, some of the threads are not even directly related.

atiqsamtia:
FCM is abstraction over GCM and APN. It delivers the notifications to both Android and IOS only configuration is in FCM console. The implementation is already there in code so no need to update the code. Just configure APN for your App in Firebase console and you are good to go 😊

tananaev:
Point about number of tokens is still valid. Also, unfortunately the code is not ready to be merged.

atiqsamtia:
Yeah, I have pretty much experience in FCM but it is my first pull request on traccar, I'm not sure how do you manage the Mapping multiple items to One. Can you let me know so that I could add the multiple tokens as well?

I have a thought 💭, 

Create a REST API endpoint named FCM with DELETE and PUT operations. 

We could add the token to user mapping from there and also delete the token when they logout.

tananaev:
I think the easiest option is to just list tokens separated by space or comma.

atiqsamtia:
Or if we save the tokens in comma separated values in backend and send, get String Array via Json in User object.?

Abyss777:
Why don't you like simple objects? Like calendars that can be linked to the users.
They can contain registration date, device name, and something other.
User can control their FCM devices, and delete not needed. Like it done in other services.

atiqsamtia:
@Abyss777  Can you give me a little example in the context of traccar? As I have mentioned I'm starting to code for traccar.

Abyss777:
Calendars are good example
You need to create model of token (or how would you call it), then implement Manager like https://github.com/traccar/traccar/blob/master/src/org/traccar/database/CalendarManager.java
and API resource https://github.com/traccar/traccar/blob/master/src/org/traccar/api/resource/CalendarResource.java

I think in code you will need to use `getUserItems(userId)` to retrieve list of user tokens.

Of course you need initialize manager in Context and add code here https://github.com/traccar/traccar/blob/master/src/org/traccar/Context.java#L491
and here https://github.com/traccar/traccar/blob/master/src/org/traccar/database/DataManager.java#L391
Also you need to add schema migration with appropriate tables and columns.

beryindo:
can this still be used? I want to apply it on Android with the Firebase cloud message

beryindo:
ready on v4.0? how use?

tananaev:
Not released yet, but it's already checked in.

beryindo:
must build own? or will release v4.1?

tananaev:
Yes, either.

beryindo:
thank you Anton, I will waiting for update on V4.1

tananaev:
You are welcome to send a pull request, but at the moment we don't have titles for short notifications.

atiqsamtia:
Yeah, I'll do the pull request in a couple of days. 

But I've identified another issue I thought I should confirm from you before opening it as an issue. 

in your latest commit, you have changed the properties of notification and Message to private and you are referencing them directly without setter in sendSync method it is crashing there due to no access if firebase is enabled. 

do check or should I send the pull request directly?

tananaev:
What's the crash stack?

atiqsamtia:
You can reproduce the issue real quick

- enable firebase in the config
- Add Notification for a device online event with firebase channel
- connect the device

the logs show device connected and after a few seconds, it shows disconnected. and the device does not show online in console due to firebase event issue

tananaev:
What is the crash stack? I have just tried and I don't get any issues.

tananaev:
@atiqsamtia, you said there is a crash in the `sendSync` code. Please provide crash stack.

beryindo:
@atiqsamtia can you sent to me traccar with FCM ?

tecufanujacu:
Hello @tananaev , I have just read the documentation to use the the push notifications, and as I have understood I need to compile my own mobile app. No problem for me but I suggest to implement the push notifications via FCM as [@pliablepixels](https://github.com/pliablepixels) has implemented the [zmeventserver](https://github.com/pliablepixels/zmeventserver) to add the push notifications in [Zoneminder](https://github.com/ZoneMinder/ZoneMinder) for [zmNinja](https://github.com/pliablepixels/zmNinja).

I use Zoneminder and for push notifications I sue zmeventserver and zmNinja in which i simply have to add the web address of my own server with Zoneminider, I don't have to compile my own app.

PR #4010, opened 2018-08-15
Added working docker file 👍

danpe:
Will soon add a docker-compose.yml for local fast development.
Just run `./setup/docker/build.sh` to build a docker image and tag it with the latest version.

### DIFF
--- a/setup/docker/Dockerfile
+++ b/setup/docker/Dockerfile
@@ -1,0 +1,43 @@
+# Pull Maven with JDK 7 for build time docker
+FROM maven:3.5.4-jdk-7-slim AS build
+
+MAINTAINER Dan Peleg <dan@globekeeper.com>
+
+COPY . .
+RUN mkdir -p tmp
+RUN mvn package
+RUN cat setup/traccar.xml | awk '/config.default/ && !modif { print;printf("    <entry key=\"web.debug\">true</entry>\n");next; modif=1 } {print}'  > tmp/traccar.xml
+RUN cp -rf /setup/default.xml tmp \
+		&& cp -rf /schema tmp/schema \
+		&& cp -rf /templates tmp/templates \
+		&& cp -rf /target/tracker-server.jar tmp/traccar-server.jar \
+		&& cp -rf /target/lib tmp/lib \
+		&& if [ -d /traccar-web/web ] ; \
+			then cp -rf /traccar-web/web tmp/web ; \
+			else mkdir tmp/web ; fi
+
+# Pull java 8 runtime for running traccar
+FROM java:8-alpine
+RUN apk upgrade --update && \
+    apk add --update curl bash && \
+    rm -rf /var/cache/apk/* && \
+    mkdir -p /opt/traccar/logs && \
+    mkdir -p /opt/traccar/data
+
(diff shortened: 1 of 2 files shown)


tananaev:
Have you seen following project?

https://github.com/traccar/traccar-docker

danpe:
@tananaev I actually missed that, but as I see this only works for an already released version and doesn't let developers build docker images locally for local deployments/testing.

What do you think? is this useful?

tananaev:
It might be useful as an option for developers, but I think using official release should be the primary option for deployment.

Either way, this pull request should be against that repo.

danpe:
@tananaev againsts [traccar-docker](https://github.com/traccar/traccar-docker)?
If so, I should create a development docker overthere that will pull from the [traccar](https://github.com/traccar/traccar/pull/4010) repo and build, right?

Maybe  [traccar-docker](https://github.com/traccar/traccar-docker) should be a sub-repo like [traccar-web](https://github.com/traccar/traccar-web)?

tananaev:
I'm not an expert in Docker. I would recommend to create an issue in that repo and discuss best approach with @magna-z (original author of Traccar docker repo).

setup/docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
+# Pull Maven with JDK 7 for build time docker
+FROM maven:3.5.4-jdk-7-slim AS build
```
magna-z: Maybe build the project on JDK 8 if is using runtime machine?
[maven:3.5.4-jdk-8-alpine](https://hub.docker.com/_/maven/)

setup/docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN apk upgrade --update && \
+    apk add --update curl bash && \
+    rm -rf /var/cache/apk/* && \
+    mkdir -p /opt/traccar/logs && \
+    mkdir -p /opt/traccar/data
+
+ENV JAVA_OPTS -Xms256m -Xmx1024m
+
+# only copy needed binaries 
+COPY --from=build ./tmp/traccar.xml /opt/traccar/traccar.xml
```
magna-z:
```
COPY --from=build ./tmp/traccar.xml /opt/traccar/traccar.xml
COPY --from=build ./tmp/default.xml /opt/traccar/conf/default.xml
COPY --from=build ./tmp/schema /opt/traccar/schema
COPY --from=build ./tmp/templates /opt/traccar/templates
COPY --from=build ./tmp/web /opt/traccar/web
COPY --from=build ./tmp/lib /opt/traccar/lib
COPY --from=build ./tmp/traccar-server.jar /opt/traccar/traccar-server.jar
```
is eq
```
COPY --from=build ./tmp/traccar.xml ./tmp/schema ./tmp/templates ./tmp/web ./tmp/lib ./tmp/traccar-server.jar /opt/traccar/
COPY --from=build ./tmp/default.xml /opt/traccar/conf/default.xml
```

See [Dockerfile reference](https://docs.docker.com/engine/reference/builder/#copy).

tananaev:
We already have a Dockerfile. Closing.

PR #4012, opened 2018-08-16
test: ProgressProtocolDecoderTest should thrown an exception in case …

danglotb:
…of illegal hexa char


tananaev:
You'll have to explain your change with more details.

danglotb:
This test aims at specifying the method `parseHex(string)` of  `org.traccar.helper.DataConverter` in order to ensure that the exception/errors flow is under control.

tananaev:
It still doesn't make sense to me. What is "++"? Where did you get it from?

tananaev:
I think this test is unnecessary.

PR #4019, opened 2018-08-23
Implement Device Engine Hours reset

Abyss777:
Implemented resetting device engine hours with total distance
 
https://www.traccar.org/forums/topic/adjust-hours-and-service/


tananaev:
Should we make values nullable? So that we can reset only hours or only distance.

Abyss777:
Why not. It will be helpful for API users, also we are loading previous values from last position, user can change only one of them, another just will be replaced with same value.

Abyss777:
Made them nullable. If it is omitted in payload it will not go to new position.

tananaev:
Merged, thanks.

PR #4020, opened 2018-08-23
Do not record accuracy, altitude or course if they are not available

oliv3:


tananaev:
1. You need to make sure you thoroughly test all use cases affected by your change. For example, I already see a problem with `WebDataHandler` where `{xxx}` won't be replaced with anything if the value is null.
2. I think we should keep setters as primitive values so that we don't need to change all the protocol decoders.

oliv3:
For 1.: I was not sure what would be the best solution, so I decided to stick with the existing, the {address} case:
```
        if (position.getAddress() != null) {
            request = request.replace(
                    "{address}", URLEncoder.encode(position.getAddress(), StandardCharsets.UTF_8.name()));
        }
```
So {address} is/was already a problem there. What would be a better option ?

tananaev:
With address it's different. Assumption is that use knows if he enabled reverse geocoding or not.

I think we can replace value with empty or zero (for backward compatibility).

oliv3:
For 2.:  https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html considers String (or any object) a primitive type, that defaults to null. If you could elaborate, thanks !
Besides, all protocols have been updated, so adding a new protocol won't add that much work ?

oliv3:
You mean something like "speed=" [nothing] would be ok ? Never tried that. Or I can set to zero, your call.

tananaev:
Have you removed your comments? I don't see them here.

tananaev:
1. Let's go with zeroes.
2. I am talking about keeping setter methods in position class same as they were before the change.

oliv3:
1. Ok
2. Oooh, my bad... The setters. Will fix this and it'll be simpler, right.
Thanks !

oliv3:
@tananaev I'll need to do the same as in https://github.com/traccar/traccar/commit/2c44646f6b574d399dd2d36afa2e88963c7322af, to set altitude, speed and course to null. 

What's the best practice ?
* Add a Position.clear() that sets all three to null
* Add Position.clearAltitude()/Speed()/Course()

Thanks

tananaev:
I think the best option is to add single method to reset position, which resets all nullable values.

tananaev:
And maybe reset coordinates as well, just for consistency.

oliv3:
@tananaev Updated branch with your suggestions. Thanks for reviewing !

tananaev:
It seems to include some of the changes from master branch. Please fix it first. Also, I would recommend to squash your commits into one to avoid future merge conflicts.

oliv3:
Ok, done.

tananaev:
Not sure what you did, but I still see 20 commits.

oliv3:
Should be ok now.

tananaev:
Have you run `mvn site` on it? Please check reports and fix if there are any issues with your changes. Have you tested all functionality? Reports (excel and web), events (web, email) etc?

@Abyss777, what do you think? Do you see any potential issues?

oliv3:
@tananaev Of course I didn't run ```mvn site```. A good reason for this, apart from me not being a Java professional, is that this step is mentioned nowhere, neither in the build instructions, nor in the forum.
So I ran this... "BUILD SUCCESS". May I kindly suggest a documentation update ? Thanks !

oliv3:
Please also add to look for the said report in /path/to/traccar/target/site/index.html also :)

All I can see is 1 bug in org.traccar.protocol.WristbandProtocolDecoder:
```Bug: Format string should use %n rather than \n in org.traccar.protocol.WristbandProtocolDecoder.sendResponse(Channel, String, String, int, String)
Category: BAD_PRACTICE
Details: VA_FORMAT_STRING_USES_NEWLINE
Line: 48
```

I'm pretty sure this is one is not mine :)

tananaev:
If there's nothing else, then it's good. Let's wait and see what Andrey thinks.

oliv3:
Ok, I'm running all the tests I can but please understand I don't have your test fire-power (for example, I own no car with a GPS tracker). Will of course report if I find some. Thanks !

Abyss777:
What is the purpose of the changes?

oliv3:
Well, not to record invalid/misleading 0.0 when data is unavailable.
If nothing is defined storing a NULL value would be more correct.

oliv3:
@Abyss777 For example, in the web interface, if you have no bearing available, then all markers are heading North, which is incorrect and leading the user to confusion.

Abyss777:
I've checked it. I see that in some places we still assuming that speed = 0 when it is null.
I think it is OK for reason to not make it complex, but in general...

Also about web interface, I believe we still will assume null course as North, because implementing circles without arrow will not be simple. I think it is better to postpone it to leaflet migration.

Abyss777:
Just thought, if we assuming speed = null = 0 why we need this changes?

oliv3:
@Abyss777 To make things consistent.

Abyss777:
In this changes we still assuming than null=0, in filters for example or reports.
Why we need this proxy representation 0=null=0 ?!

As for me `null` means there is no value, then why it triggers static filter?
Oh, just noticed
`(position.getAccuracy() == null || position.getAccuracy() > filterAccuracy)`
Is that mean that position without accuracy will be always filtered out?

Also reports... we calculating average speed assuming `null = 0`, may be we should exclude such positions?

oliv3:
Maybe average and max speed should also be nullable, then.

Abyss777:
Yes, I think it is reasonable.

src/org/traccar/WebDataHandler.java
```diff
@@ -95,10 +104,10 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
                 .replace("{valid}", String.valueOf(position.getValid()))
                 .replace("{latitude}", String.valueOf(position.getLatitude()))
                 .replace("{longitude}", String.valueOf(position.getLongitude()))
-                .replace("{altitude}", String.valueOf(position.getAltitude()))
-                .replace("{speed}", String.valueOf(position.getSpeed()))
-                .replace("{course}", String.valueOf(position.getCourse()))
-                .replace("{accuracy}", String.valueOf(position.getAccuracy()))
+                .replace("{altitude}", (position.getAltitude() == null) ? "0" : String.valueOf(position.getAltitude()))
```
Abyss777: Are this substitutions to zero obligatory for protocol? Can it be empty string? Just do not know.

oliv3: @tananaev suggested zeroes, I'm fine with both options. Will let one of you have the final word.

tananaev:
I can see that this becomes more and more complicated for very little end-user value.

oliv3:
Right.. Wrote some unit tests and null speeds will make trips and stops even more complicated.
I'll put null speeds apart, which will make this definitely simpler.

src/org/traccar/reports/model/BaseReport.java
```diff
@@ -52,24 +52,24 @@ public void addDistance(double distance) {
...
+    private Double maxSpeed;
 
-    public double getMaxSpeed() {
+    public Double getMaxSpeed() {
         return maxSpeed;
     }
 
     public void setMaxSpeed(double maxSpeed) {
-        if (maxSpeed > this.maxSpeed) {
+        if (this.maxSpeed == null || maxSpeed > this.maxSpeed) {
```
Abyss777: I believe `&&` should be here also

tananaev:
I think we need to park this pull request for now. There is too much risk and I don't know any users who really need this change.

src/org/traccar/reports/model/BaseReport.java
Abyss777: Or we stopped using this method? Then we should remove it.

oliv3:
@tananaev You're right saying it's too much risk, but I think it's not that risky if those changes are done logically and incrementally. So a good point for you in seeing this as a "massive change", and a bad point for me in being overconfident addressing the four points at the same time. But user experience is also important, that I'll explain in another comment.

tananaev:
I can see that for some reason you are very passionate about this change, but there is no need to exaggerate your case. It doesn't help to convince anyone.

For example take Accuracy. There is no "perfect" accuracy, so zero value can be used as "no value".

tananaev:
We don't have any 3D maps, so we don't really use Altitude anywhere. We only show it as a number. Not many people are interested in it, but those who are (e.g. tracking planes) use devices that always report correct values.

Bearing is a valid case, but I think you hugely overestimate its importance. It was never a critical issue for anyone. Some people complained, but only because they didn't know that network based locations don't have this value.

oliv3:
IHNTA.

oliv3:
Parking for now.

PR #4047, opened 2018-09-06
Add Base Adresse Nationale reverse geocoder

oliv3:


tananaev:
I don't see anything about API key in the documentation. Is it required? How to get one? Any documentation?

oliv3:
It will be possible to get a key in the future, but it will be optional and only for usage tracking purposes.
There is no way to request one at the moment since this feature is not in production yet.
I was confirmed there will be one someday, so this parameter can safely be ignored for now - done by anticipation.
Limitations are 10 requests per second per IP address.

oliv3:
https://adresse.data.gouv.fr/api

tananaev:
OK, if it's not available at the moment, I think we should remove it from the code.

oliv3:
Is commented ok ?

tananaev:
No, it has to be removed.

oliv3:
Hi, code has been removed and branch rebased.

tananaev:
Merged, thanks.

jpmens:
It should be clarified in the documentation that this API returns data for France only.

> ou la France entière

[/via](https://adresse.data.gouv.fr/download)

oliv3:
Indeed. Whole France, including D.O.M.s and T.O.M.s e.g. Kourou, Guyana:
```https://api-adresse.data.gouv.fr/reverse/?lat=5.169189&lon=-52.6813537```
Website is not on Github so I'll leave this to @tananaev .

PR #4079, opened 2018-10-01
Add no_annotations=1 parameter in the call to OpenCage API

freyfogle:
This makes the response slightly faster by not requesting information traccar never uses.

See:
https://opencagedata.com/api#bestpractices
https://opencagedata.com/api#forward-opt


tananaev:
Merged, thanks.

PR #4091, opened 2018-10-05
fix watchframedecoder for TK906 (invalid lenght) device issues

TheBrain8791:
This fixes device issues found with TK906 bike trackers, which tend to
transmit wrong a frame length for some frames very often. 

The fixed Watch protocol frame decoder now not only evaluates the frame
length transmitted from the device itself but also checks the no of
frame bytes till the next "frame end marker" (']') and uses whatever is
correct.

Additionally the frame start detection has been improved to better sync
to valid frames according to the watch protocol spec.

usecase 1) ignore chars which don't belong to a frame (not between '['
and ']'):
```
12345[SG*12345678*0004*LK,,]abcde\n..[SG*12345678*0004*LK,,]222
```
in this example '12345', 'abcde\n..' and '222' stream bytes will be
ignored

usecase 2) partial transmitted frames
```
[SG*12345678[SG*12345678*0002*LK]
```
in this example, the leading bytes '[SG*12345678' will be ignored

The fix does not affact devices which work according to the protocol
spec, but improves handling of devices like the TK906 which fail to
fulfill the protcol spec and otherwise could not be used with traccar.

### DIFF
--- a/src/org/traccar/protocol/WatchFrameDecoder.java
+++ b/src/org/traccar/protocol/WatchFrameDecoder.java
@@ -30,8 +30,34 @@ public class WatchFrameDecoder extends BaseFrameDecoder {
     protected Object decode(
             ChannelHandlerContext ctx, Channel channel, ByteBuf buf) throws Exception {
 
+        int idFollowUpFrameStart;
+        do {
+            int idFrameStart = buf.indexOf(buf.readerIndex(), buf.writerIndex(), (byte) '[');
+            if (idFrameStart < 0) {
+                buf.skipBytes(buf.writerIndex() - buf.readerIndex());
+                return null;
+            } else if (idFrameStart - buf.readerIndex() > 0) {
+                buf.skipBytes(idFrameStart - buf.readerIndex());
+            }
+
+            int idFrameEnd = buf.indexOf(buf.readerIndex(), buf.writerIndex(), (byte) ']');
+            if (idFrameEnd < 0) {
+                return null;
+            }
+
+            idFollowUpFrameStart = buf.indexOf(buf.readerIndex() + 1, buf.writerIndex(), (byte) '[');
+            if (idFollowUpFrameStart > 0) {
+                if (idFollowUpFrameStart < idFrameEnd) {
+                    buf.skipBytes(idFollowUpFrameStart - buf.readerIndex());
+                } else {
+                    break;
+                }
+            }
+        } while (idFollowUpFrameStart >= 0);
+
         int idIndex = buf.indexOf(buf.readerIndex(), buf.writerIndex(), (byte) '*') + 1;
         if (idIndex <= 0) {
+            buf.skipBytes(buf.writerIndex() - buf.readerIndex());
             return null;
         }
 
@@ -56,9 +82,20 @@ public class WatchFrameDecoder extends BaseFrameDecoder {
 
         int length = Integer.parseInt(
                 buf.toString(lengthIndex, payloadIndex - lengthIndex, StandardCharsets.US_ASCII), 16);
-        if (buf.readableBytes() >= payloadIndex + 1 + length + 1) {
+
+        int endmarkerIndex = buf.indexOf(lengthIndex + 5, buf.writerIndex(), (byte) ']');
+        if (endmarkerIndex <= 0) {
+            return null;
+        }
+
+        if (endmarkerIndex < payloadIndex + 1 + length || payloadIndex + 1 + length < endmarkerIndex) {
+            int framelength = length;
+            length = endmarkerIndex - payloadIndex - 1;
+        }
+
+        if (buf.readableBytes() >= payloadIndex + 1 + length + 1 - buf.readerIndex()) {
             ByteBuf frame = Unpooled.buffer();
-            int endIndex = buf.readerIndex() + payloadIndex + 1 + length + 1;
+            int endIndex = payloadIndex + 1 + length + 1;
             while (buf.readerIndex() < endIndex) {
                 byte b = buf.readByte();
                 if (b == 0x7D) {
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.traccar</groupId>
     <artifactId>traccar</artifactId>
-    <version>4.0-SNAPSHOT</version>
+    <version>4.0-SNAPSHOT-WatchFrameDecoderModified_TK906</version>
 
     <name>traccar</name>
     <url>https://www.traccar.org</url>
--- a/src/org/traccar/protocol/WatchProtocolDecoder.java
+++ b/src/org/traccar/protocol/WatchProtocolDecoder.java
@@ -42,23 +42,23 @@ public class WatchProtocolDecoder extends BaseProtocolDecoder {
     }
 
     private static final Pattern PATTERN_POSITION = new PatternBuilder()
-            .number("(dd)(dd)(dd),")             // date (ddmmyy)
-            .number("(dd)(dd)(dd),")             // time (hhmmss)
-            .expression("([AV]),")               // validity
-            .number(" *(-?d+.d+),")              // latitude
+            .number("(dd)(dd)(dd),") // date (ddmmyy)
+            .number("(dd)(dd)(dd),") // time (hhmmss)
+            .expression("([AV]),") // validity
+            .number(" *(-?d+.d+),") // latitude
             .expression("([NS]),")
-            .number(" *(-?d+.d+),")              // longitude
+            .number(" *(-?d+.d+),") // longitude
             .expression("([EW])?,")
-            .number("(d+.?d*),")                 // speed
-            .number("(d+.?d*),")                 // course
-            .number("(d+.?d*),")                 // altitude
-            .number("(d+),")                     // satellites
-            .number("(d+),")                     // rssi
-            .number("(d+),")                     // battery
-            .number("(d+),")                     // steps
-            .number("d+,")                       // tumbles
-            .number("(x+),")                     // status
-            .expression("(.*)")                  // cell and wifi
+            .number("(d+.?d*),") // speed
+            .number("(d+.?d*),") // course
+            .number("(d+.?d*),") // altitude
+            .number("(d+),") // satellites
+            .number("(d+),") // rssi
+            .number("(d+),") // battery
+            .number("(d+),") // steps
+            .number("d+,") // tumbles
+            .number("(x+),") // status
+            .expression("(.*)") // cell and wifi
             .compile();
 
     private void sendResponse(Channel channel, String id, String index, String content) {
--- a/test/org/traccar/ProtocolTest.java
+++ b/test/org/traccar/ProtocolTest.java
@@ -295,4 +295,7 @@ public class ProtocolTest extends BaseTest {
         assertEquals(ByteBufUtil.hexDump(expected), ByteBufUtil.hexDump((ByteBuf) object));
     }
 
+    protected void verifyFrameNull(Object object) {                
+        assertNull(object);
+    }
 }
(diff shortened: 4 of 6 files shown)


tananaev:
OK, so this is the first one?

1. Please provide example from real device with this issue.
2. Partially transmitted frames should not be ignored. You need to wait for the rest.
3. How is your decoder different from delimited-based decoder?

src/org/traccar/protocol/WatchFrameDecoder.java
```diff
@@ -56,9 +82,20 @@ protected Object decode(
 
         int length = Integer.parseInt(
                 buf.toString(lengthIndex, payloadIndex - lengthIndex, StandardCharsets.US_ASCII), 16);
-        if (buf.readableBytes() >= payloadIndex + 1 + length + 1) {
+
```
TheBrain8791: this part fixes invalid encoded frame length by looking up the next frame end maker ']' and comparing the frame length with the no of bytes till this frame end marker.

src/org/traccar/protocol/WatchFrameDecoder.java
```diff
@@ -30,8 +30,34 @@
     protected Object decode(
             ChannelHandlerContext ctx, Channel channel, ByteBuf buf) throws Exception {
 
+        int idFollowUpFrameStart;
```
TheBrain8791: this part removes bytes from the stream which are not placed between a frame start and frame end marker. Also the the usecase of partial frames ( another frame start marker without an frame end marker) is handled here

tananaev:
Unfortunately it doesn't answer my questions.

TheBrain8791:
**1) Please provide example from real device with this issue.**

This is part of a tracker-server.log file showing communication with TK906 with the original (tag v4.0) code.
_Please consider - I don't want to disclose personal gps, lba position data nor device IDs and IP add.  in a public forum,  these have been changed from their original byte value to 0x58 ('X') in the logs below._ 

Keypoint is - **the device often does not send a valid frame length** - so traccar discards most of the position messages. Even worse - sometime it comes to a java.lang.NumberFormatException within the decoder, which causes the connection to the gps tracker to be disconnected !

```
2018-09-20 20:54:09  INFO: [40b89c4e] connected
2018-09-20 20:54:12 DEBUG: [40b89c4e: 5093 < XXX.XXX.XXX.XXX] HEX: 5B53472A393058585858585837382A303041352A55442C3230303931382C3230353431322C562C34582E5858585858582C4E2C31582E5858585858582C452C302E373737382C3030302C39392C30302C58302C58302C302C58302C30303030303030302C352C312C5858582C312C585858322C58585858352C5858392C585858322C58585858352C5858582C585858582C58585858302C3158322C355858322C35585858352C3158582C355832582C33583631582C3158382C2C30305D
2018-09-2020:54:13  INFO: [40b89c4e] id: 90XXXXXX78, time: 2018-09-20 20:54:12, lat: 4X.XXXXX, lon: 1X.XXXXX, speed: 0.4, course: 0.0
2018-09-20 20:54:21 DEBUG: [40b89c4e: 5093 < XXX.XXX.XXX.XXX] HEX: 5B53472A393058585858585837382A303039362A55442C3230303931382C3230353432322C562C34582E5858585858302C4E2C31582E5858585832352C452C302E373737382C3030302C39392C30302C58302C58302C302C58302C30303030303030302C342C312C5833582C312C585858322C35585858302C3158352C355858322C33585858352C3158332C585858322C35585858352C3158582C355858322C33585858352C3158582C2C30305D
2018-09-20 20:54:33 DEBUG: [40b89c4e: 5093 < XXX.XXX.XXX.XXX] HEX: 5B53472A393058585858585837382A303039362A55442C3230303931382C3230353433322C562C34582E5858585858302C4E2C31582E5858585832352C452C302E373737382C3030302C58392C30302C58302C58302C302C58302C30303030303030302C342C312C5858322C312C585858322C58585858352C3158582C355858582C33585858582C3158582C355858582C35585858302C3158312C585832322C58585833352C5858302C2C30305D
2018-09-20 20:54:42 DEBUG: [40b89c4e: 5093 < XXX.XXX.XXX.XXX] HEX: 5B53472A393058585858585837382A303039362A55442C3230303931382C3230353433322C562C34582E5858585832302C4E2C31582E5858585832352C452C302E373737382C3030302C39392C30302C58302C58302C302C58302C30303030303030302C342C312C5858322C312C585858322C58585858352C5858312C355858322C33585858352C3158582C355858582C35585858582C3158582C585858322C58585858352C5858582C2C30305D
2018-09-20 20:54:42  INFO: [40b89c4e] id: 90XXXXXX78, time: 2018-09-20 20:54:22, lat: 4X.XXXXX, lon: 1X.XXXXX, speed: 0.4, course: 0.0
2018-09-20 20:54:51 DEBUG: [40b89c4e: 5093 < XXX.XXX.XXX.XXX] HEX: 5B53472A393058585858585837382A303042342A55442C3230303931382C3230353435322C562C34582E5858585832302C4E2C31582E5858585832352C452C302E373737382C3030302C39392C30302C58302C58302C302C58302C30303030303030302C582C312C5858322C312C585858322C58585858302C3158582C355858582C35585858302C3158582C355858582C33585858582C3158582C585858322C58585858352C5858372C585858312C58585858302C5858362C585858322C58585858352C5858302C2C30305D
2018-09-20 20:54:51  WARN: [40b89c4e] error - java.lang.NumberFormatException: For input string: "UD,200918,205442,V,4X.XXXXXX,N,1X.XXXXXX,E,0.7778,000,99,00,60,90,0,50,00000000,6,1,XXX,1,XXXX,XXXXX,XXX,XXXX,XXXXX,XXX,XXXX,XXXXX,140,XXX2,XXXX5,XX7,XXX1,XXXX0,XX4,XXX2,XXXX5,131,,00][SG" - DecoderException (... < WrapperInboundHandler:53 < ... < BasePipelineFactory:127 < ... < *:104 < ...)
2018-09-20 20:54:51  INFO: [40b89c4e] disconnected
2018-09-20 20:54:51  WARN: [40b89c4e] error - java.lang.NumberFormatException: For input string: "UD,200918,205442,V,4X.XXXXXX,N,1X.XXXXXX,E,0.7778,000,99,00,60,90,0,50,00000000,6,1,XXX,1,XXXX,XXXXX,XXX,XXXX,XXXXX,XXX,XXXX,XXXXX,140,XXX2,XXXX5,XX7,XXX1,XXXX0,XX4,XXX2,XXXX5,131,,00][SG" - DecoderException (... < WrapperInboundHandler:46 < ... < BasePipelineFactory:89 < ...)
2018-09-20 20:54:53  INFO: [06e42a83] connected
2018-09-20 20:55:03 DEBUG: [06e42a83: 5093 < XXX.XXX.XXX.XXX] HEX: 5B53472A393058585858585837382A303042342A55442C3230303931382C3230353530322C562C34582E5858585832302C4E2C31582E5858585832352C452C302E373737382C3030302C39392C30302C58302C58302C302C58302C30303030303030302C582C312C5858322C312C585858322C58585858302C3158582C585858322C58585858302C3158582C585858322C33585858582C3158582C585858582C35585858352C3158582C355858582C35585858302C3158582C585858322C33585858582C5858392C2C30305D
2018-09-20 20:55:03  INFO: [06e42a83] id: 90XXXXXX78, time: 2018-09-20 20:55:02, lat: 4X.XXXXX, lon: 1X.XXXXX, speed: 0.4, course: 0.0
```

This issue is fixed mainly [here](https://github.com/traccar/traccar/pull/4091/files#r223292573).
My patch is based on the redundancy contained in the protocol (frame length is given + frame end marker is used) and uses this to correct the framelength if it is wrong)

Additional sometimes I also saw something like this:

```[SG*90XXXXXX78[SG*90XXXXXX78*0002*LK]```

This happens sometimes after a "forced" reconnect triggered by an decoder exception (as within the example above) - I guess the device firmware does not flush some buffers when the tcp connection is closed by the (portal) server.
Thats why I also added an additional (to the delimiter based decoding) frame sync based on "frame start" and "frame end marker"
see [here](https://github.com/traccar/traccar/pull/4091/files#r223293298)
The problem with the wrong interpreded frame length in combination with the original code also can cause strange behaviour if the delimiter based decoder "gets out of sync" - the new patch ensures that it can resync in any case.

**2) Partially transmitted frames should not be ignored. You need to wait for the rest**

The code waits for partial transmitted frames - it just removes stream bytes which cannot belong to a frame based on the assuption that all framebytes are within frame start and frame end marker.

As already written in the commit msg:

 _usecase 1)_ ignore chars which don't belong to a frame (not between '['
and ']'):

```12345[SG*12345678*0004*LK,,]abcde\n..[SG*12345678*0004*LK,,]222```
in this example '12345', 'abcde\n..' and '222' stream bytes will be
ignored

_usecase 2)_ partial transmitted frames

```[SG*12345678[SG*12345678*0002*LK]```
in this example, the leading bytes '[SG*12345678' will be ignored

In both examples all ignored / discarded bytes cannot belong to a frame - so discarding is save!

This patch will only fail, if a device using the watch protocol would send frames without frame start and frame end markers, but this would happen IMHO with the same probability as the device could miss to send a delimiter (so the methode of the current code would also fail). And non of my devices show such an issue - also google AFAIK does not know such a device issue.

**3) How is your decoder different from delimited-based decoder?**

The difference is as decribed above, that is uses redundancy to correct wrong encoded frame length information and also additionaly to the original only delimited-based decoder used frame start and frame stop markers to discard stream bytes which are not between these markers.

TheBrain8791:
> Unfortunately it doesn't answer my questions.

Hey Anton - slow down *smile* - 1) English is not my first language and 2) it takes time to answer your questions in detail ..

tananaev:
1. In your example I don't see any data outside of the `[` and `]` frame.
3. How do you handle binary data that can have `]` in it?

TheBrain8791:
> 1. In your example I don't see any data outside of the `[` and `]` frame.

As I wrote:

> Additional sometimes I also saw something like this:
> `[SG*90XXXXXX78[SG*90XXXXXX78*0002*LK]`
> 
> This happens sometimes after a "forced" reconnect triggered by an decoder exception ...

I'm not sure if I've preserved the logfile containing that error case - I'll have a look.
Anyway I've discovered this one or two times while testing - that's why I extended the sync part... 

> 2. How do you handle binary data that can have `]` in it?

Frame start / end makers like  ']' = 0x5d, '[' = 0x5b have to be escaped before a binary blob is transmitted - this is handled by the protocol as far as I could see in the existing code (see line 102 to 118) valid for all "special" frame char - e.g. for the delimiter symbol ('*' = 0x2a) otherwise your current code also would not work.
The existing delimiter based decoder also does no special handling for binary data (beside "de - escaping" them)   
_Although this escaping is not described in the protocol spec. (which can be downloaded from your website), this practically works perfect for audio voice msg and images - as my tests with a gps smartwatch showed._

tananaev:
1. That's why I asked for log samples. I have never seen anything like that so far.
2. OK, that makes sense then, but then why do we need length in the first place?

tananaev:
I think we should simplify frame decoder to just use one end-delimiter.

TheBrain8791:
I think this is maybe too simple - in any case that modification would not handler stream bytes out of frame markings. 
Beside that fact - the current code works great (with the exception of the tk906 which has device issues) and is tested  "in the field" for some time now - so I'd suggest better not to remove too much of the frame structure checking within the frame decoder.

In the worst case I'd be afraid if some devices send malformed frames which were simply sorted out till now (unknown - because just in the background), could lead to unknown reactions within the protocol decoder, which would suddenly have to handle them...

Maybe it could be a solution to "move" some of the frame structure checking from the frame decoder to the protocol decoder, but why risk a complete redesign of the frame decoder if new additional use cases  (handle tk906 device issues) can be fixed with smaller adaptions.

Anyway I'll do a test with your suggestion and will have a look if this works and what will be the consequences. I'll report back and I'll have a look for a log as I wrote yesterday and post it...

tananaev:
If there happens to be a bad frame, the worst thing that could happen is one or two messages get corrupted and not decoded. We will still get back to normal after we find end delimiter.

PR #4092, opened 2018-10-05
Add watch framelogging for Debugging

TheBrain8791:
This fixes device issues found with TK906 bike trackers, which tend to
transmit wrong a frame length for some frames very often. 

The fixed Watch protocol frame decoder now not only evaluates the frame
length transmitted from the device itself but also checks the no of
frame bytes till the next "frame end marker" (']') and uses whatever is
correct.

Additionally the frame start detection has been improved to better sync
to valid frames according to the watch protocol spec.

usecase 1) ignore chars which don't belong to a frame (not between '['
and ']'):
```
12345[SG*12345678*0004*LK,,]abcde\n..[SG*12345678*0004*LK,,]222
```
in this example '12345', 'abcde\n..' and '222' stream bytes will be
ignored

usecase 2) partial transmitted frames
```
[SG*12345678[SG*12345678*0002*LK]
```
in this example, the leading bytes '[SG*12345678' will be ignored

The fix does not affact devices which work according to the protocol
spec, but improves handling of devices like the TK906 which fail to
fulfill the protcol spec and otherwise could not be used with traccar.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.traccar</groupId>
     <artifactId>traccar</artifactId>
-    <version>4.0-SNAPSHOT</version>
+    <version>4.0-SNAPSHOT-WatchFrameDecoderModified_TK906</version>
 
     <name>traccar</name>
     <url>https://www.traccar.org</url>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.traccar</groupId>
     <artifactId>traccar</artifactId>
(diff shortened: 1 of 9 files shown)


sonerVrs:
@TheBrain8791 can you please share an e-mail address to communicate with you?

TheBrain8791:
@sonerVrs: I don't want to disclose my mail address public viewable here - you can leave me your mail addr. and I'll reply!

TheBrain8791:
@sonerVrs: I don't want to disclose my mail address public viewable here -
you can leave me your mail addr. and I'll reply!

Am Mo., 22. Okt. 2018 um 12:47 Uhr schrieb sonerVrs <
notifications@github.com>:

tananaev:
Do you have real logs?

PR #4093, opened 2018-10-05
Fix and improve watch voice msg handling

TheBrain8791:
**fix:** 

The cmd was not sent before when triggered from the web frontend if
"real" amr data (taken from a file generated from a smartwatch) was
entered to the "Key_Data" field within the webform, for some reasons:

the call (WatchProtocolEnecoder.java::191): 
cmdtoSend = formatCommand(channel, command, "TK," +
getBinaryData(command));

directly concats the binay data to the cmd content, which causes (with
real data) a handling problem within the class
StringProtocolEncoder::formatCommand
The binary blob also can contain "\" and "$" which cause problems 
when a string replacement methode is called.

According to java doc 'quoteReplacement' should be used in that cases...

Also changed the direct content concating to using the "key -
replacment conecept" all other cmds use.

Also changed the return type of the encode methode to "bytebuf" because
strings cause problems on all non-ASCII data - e.g. as the binary blob
of an amr file.

For the same reason, also the unit test of the watchencoder has been
changed from string to binary assert values and the voice msg test now
covers non-ascii char + char which need to be escaped.

This changes don't affact usual text cmds (which don't contain binary
data) - according to my tests but make the VOICE MSG cmd usable.

**Improvement:** 

Introduced a new cmd type (binary blob) for e.g. voice data.

Also added a way to specify a filename instead of forcing the user to
copy & paste unhandy audio data as hex encoded stream to the web form,
because it should be easier to reference to a file (and read the binary
blob from this file) than to extract this data manually.

To reference an amr - audio file, one needs to enter
"file://[rel.path]/[filename.amr]" within the "KEY_DATA" field in the
webform.
The path of the file is relative to the one which is configured as 
"media.path" 

e.g. if media.path is set to "/mypath" and the audio file is
stored in /mypath/foo/bar.amr then "file://foo/bar.amr" entered
as KEY_DATA will trigger a file read.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.traccar</groupId>
     <artifactId>traccar</artifactId>
-    <version>4.0-SNAPSHOT</version>
+    <version>4.0-SNAPSHOT-WatchFrameDecoderModified_TK906</version>
 
     <name>traccar</name>
     <url>https://www.traccar.org</url>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.traccar</groupId>
     <artifactId>traccar</artifactId>
-    <version>4.0-SNAPSHOT-WatchFrameDecoderModified_TK906</version>
+    <version>4.0-SNAPSHOT-WatchFrameDecoderModified_FrameLogging</version>
 
     <name>traccar</name>
     <url>https://www.traccar.org</url>
--- a/src/org/traccar/StringProtocolEncoder.java
+++ b/src/org/traccar/StringProtocolEncoder.java
@@ -18,6 +18,7 @@ package org.traccar;
 import org.traccar.model.Command;
 
 import java.util.Map;
+import java.util.regex.Matcher;
 
 public abstract class StringProtocolEncoder extends BaseProtocolEncoder {
 
@@ -38,7 +39,7 @@ public abstract class StringProtocolEncoder extends BaseProtocolEncoder {
             if (value == null) {
                 value = entry.getValue().toString();
             }
-            result = result.replaceAll("\\{" + entry.getKey() + "}", value);
+            result = result.replaceAll("\\{" + entry.getKey() + "}", Matcher.quoteReplacement(value));
         }
 
         return result;
--- a/src/org/traccar/database/MediaManager.java
+++ b/src/org/traccar/database/MediaManager.java
@@ -19,6 +19,7 @@ import io.netty.buffer.ByteBuf;
 import org.traccar.helper.Log;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
@@ -46,13 +47,22 @@ public class MediaManager {
         return filePath.toFile();
     }
 
+    private File openFile(String name) throws IOException {
+        Path filePath = Paths.get(path, name);
+        if (!Files.isRegularFile(filePath)) {
+            return null;
+        } else {
+            return filePath.toFile();
+        }
+    }
+
     public String writeFile(String uniqueId, ByteBuf buf, String extension) {
         if (path != null) {
             int size = buf.readableBytes();
             String name = new SimpleDateFormat("yyyyMMddhhmmss").format(new Date()) + "." + extension;
             try (FileOutputStream output = new FileOutputStream(createFile(uniqueId, name));
                     FileChannel fileChannel = output.getChannel()) {
-                    ByteBuffer byteBuffer = buf.nioBuffer();
+                ByteBuffer byteBuffer = buf.nioBuffer();
                 int written = 0;
                 while (written < size) {
                     written += fileChannel.write(byteBuffer);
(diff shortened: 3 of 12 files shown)


tananaev:
Why are there 3 pull requests?

TheBrain8791:
Dear Anton,

Sorry - I planned to leave you this message immediatly after creating the pull requests, but my children had different plans with me yesterday - so I can make that up now :smile: 

First please let me explain a big thank for creating traccar server. It's a great project and enabled me to use cheap Chinese gps tracking devices without being dependent on their tracking servers, because I prefer having control over these very personal data myself instead pushing them to an unknown cloud - so :+1:  for your work.

_Now let me explain the pull requests:_

All my trackers use the watch protocol. Sadly especially the tracker for my bike (Model: tk906) has some device issues - see e.g. this discussion: https://www.traccar.org/forums/topic/tk-906-biketracker-greater-watch-protocoll/ - e.g. the device tends to transmit frames with wrong encoded frame length which currently triggers traccar to discard most of the position reports and makes these trackers unusable for me.

- To solve this - and because it would be much harder to fix bugs in a closed source embedded firmware :wink: , I developed a workaround for this within Traccar and hardened the watch frame decoder to handle some additional "stream conditions" - please find all details for this within pullrequest  https://github.com/traccar/traccar/pull/4091

- To make debugging for watch protocol easier, I also added a possibility to log watch protocol frames as clear text output additional to the existing hex output - please see #4092 for this (to make this change more "compatible" for all users, this option is implemented configurable within Traccar config file.

- Last, I discovered some problems which sending voice messages to another of my devices (also using watch protocol) using Traccar. After some investigation and debugging, I found the reason and developed a patch for this. 
Additional I found it easier to refer to an existing .amr file within the media - folder of the server instead of opening a file in a hex editor, extract the content and paste this within the SendCommand - Key_Data Field. 
But also this change should be "compatible" because the file ref. is an additional possibility for any user - one can still paste the Hex String as before. - all details can be found within #4093 

I created individual pull requests, to make it easier to just merge whatever you find  is useful - but because for me all three changes make sense so they depend on each other. 
If this is a problem, a rebase of #4092 and #4093 to the original traccar v4.0 tag sould be easily possible  - I would do this, on your request.

So finally I just invested some of my rare free time to solve issues, which make Traccar even better to use for me and my devices - and I just would like to make these little improvements available for other users (to give something back to you and the traccar community) 
I tried to use the existing coding style for my changes and did excessive testing with my devices and also using simulated frames through telnet - so I think the changes should work and not brake anything existing. I also read your contribution guide and hope I did everything as expected.

So I hope you are interested in integrating these small "improvements" in your project.
I have some experience in software development but please be patient - java is not my primary language (as Asm, C and C++ is :-)) - so hope my changes comply with your code quality requirements - if not feel free to suggest changes.
Also if you have some more questions or want to discuss the solutions,  feel free to contact me.

Again sorry for just pushing some commits without this explanation before.

best regards from Austria
Gerald

tananaev:
As far as I can tell, your pull requests share same changes. Why? Also, you would need to fix all the conflicts.

TheBrain8791:
Well that's not really correct. As I wrote the changes depend on each other.
#4091 ist based on v4.0 tag and does not cause any conflicts.
#4092 is based on #4091 (so contains these changes too and additional ones) - won't cause any conflict if you accept #4091 before.
The same is true for #4092 ...
That's why I wrote - for me all three changes        make sense - so I used code v4.0 + merged changes from #4091 as base for #4092 a.s.o

sonerVrs:
> **fix:**
> 
> The cmd was not sent before when triggered from the web frontend if
> "real" amr data (taken from a file generated from a smartwatch) was
> entered to the "Key_Data" field within the webform, for some reasons:
> 
> the call (WatchProtocolEnecoder.java::191):
> cmdtoSend = formatCommand(channel, command, "TK," +
> getBinaryData(command));
> 
> directly concats the binay data to the cmd content, which causes (with
> real data) a handling problem within the class
> StringProtocolEncoder::formatCommand
> The binary blob also can contain "" and "$" which cause problems
> when a string replacement methode is called.
> 
> According to java doc 'quoteReplacement' should be used in that cases...
> 
> Also changed the direct content concating to using the "key -
> replacment conecept" all other cmds use.
> 
> Also changed the return type of the encode methode to "bytebuf" because
> strings cause problems on all non-ASCII data - e.g. as the binary blob
> of an amr file.
> 
> For the same reason, also the unit test of the watchencoder has been
> changed from string to binary assert values and the voice msg test now
> covers non-ascii char + char which need to be escaped.
> 
> This changes don't affact usual text cmds (which don't contain binary
> data) - according to my tests but make the VOICE MSG cmd usable.
> 
> **Improvement:**
> 
> Introduced a new cmd type (binary blob) for e.g. voice data.
> 
> Also added a way to specify a filename instead of forcing the user to
> copy & paste unhandy audio data as hex encoded stream to the web form,
> because it should be easier to reference to a file (and read the binary
> blob from this file) than to extract this data manually.
> 
> To reference an amr - audio file, one needs to enter
> "file://[rel.path]/[filename.amr]" within the "KEY_DATA" field in the
> webform.
> The path of the file is relative to the one which is configured as
> "media.path"
> 
> e.g. if media.path is set to "/mypath" and the audio file is
> stored in /mypath/foo/bar.amr then "file://foo/bar.amr" entered
> as KEY_DATA will trigger a file read.

Can you please share an e-mail address to reach you? Thanks

TheBrain8791:
@sonerVrs: I don't want to disclose my mail address public viewable here - you can leave me your mail addr. and I'll reply!

Alternative way - if you like: use this temp. mail address to reach me the first time: praduditaj_at_memeil.top (replace '_at_' with '@') - valid within 48 h

flocsy:
Hi Gerald,
I am also working on the same issues: 1. add textual (non binary) to logs, 2. voice messages
Are you planning to fix these pull requests soon?

PR #4094, opened 2018-10-05
Distanceanddataloss

rahulgadhyan:
changes for the distance algo and data loss algo as discussed... taken care of following:
1) if data missing for over 30 minutes, then comparing the difference of odometer reading and distance between the packets and storing the larger of the 2
2) if distance is larger than max distance that can be covered at speed of 150 km/hr, then treating it as an incorrect packet and storing distance as 0. (Note the lat long of this paccket gets replaced with the lat long of the previous packet as per current code. If we donot want this to happen, then we can make required changes)
3) in line 114 of fuelsensordatahandlerhelp, converting the maximumdistancetravelled to km by dividing by 1000 - as the mileage is in km/hr, we need to convert it to KM
4) In fuelsenssordatahandler, have introduced code in lines 430-474 and 871-877 to check when a packet is received after the lookback period, whether that packet is a normal back dated packet or there has been any fuel fill or drain events - NOTE, Have not implemented the code to send the push notification for corresponding fuel event and code to store the fuel event in corresponding table

you can work on the ccode and put them in a proper manner if you think the changes proposed are correct

I have tested this code using data parts where these issues were observed in first simulation... shall have to re run the simulation to check the correctness of the code...


tananaev:
This is not ready to be merged. If you want any of it to be included, you need to split changes into smaller meaningful pieces and fix any conflict and code issues.

PR #4095, opened 2018-10-07
[Gpsmta] Fix battery level decoding

oliv3:
Protocol analysis reveals Gpsmta client sends battery _level_, not _voltage_.

### DIFF
--- a/src/org/traccar/protocol/GpsmtaProtocolDecoder.java
+++ b/src/org/traccar/protocol/GpsmtaProtocolDecoder.java
@@ -43,7 +43,7 @@ public class GpsmtaProtocolDecoder extends BaseProtocolDecoder {
             .number("(d+) ")                     // accuracy
             .number("(d+) ")                     // altitude
             .number("(d+) ")                     // flags
-            .number("(d+) ")                     // battery
+            .number("(d+) ")                     // battery level
             .number("(d+) ")                     // temperature
             .number("(d)")                       // charging status
             .any()
@@ -77,7 +77,7 @@ public class GpsmtaProtocolDecoder extends BaseProtocolDecoder {
         position.setAltitude(parser.nextInt(0));
 
         position.set(Position.KEY_STATUS, parser.nextInt(0));
-        position.set(Position.KEY_BATTERY, parser.nextInt(0));
+        position.set(Position.KEY_BATTERY_LEVEL, parser.nextInt(0));
         position.set(Position.PREFIX_TEMP + 1, parser.nextInt(0));
         position.set(Position.KEY_CHARGE, parser.nextInt(0) == 1);
 
--- a/test/org/traccar/protocol/GpsmtaProtocolDecoderTest.java
+++ b/test/org/traccar/protocol/GpsmtaProtocolDecoderTest.java
@@ -16,7 +16,7 @@ public class GpsmtaProtocolDecoderTest extends ProtocolTest {
         verifyPosition(decoder, text(
                 "864528021249771 1446116686 49.85073 24.004438 0 217 6 338 00 59 27 0 0"));
 
-        verifyPosition(decoder, text(
+        verifyNotNull(decoder, text(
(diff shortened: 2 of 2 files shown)


tananaev:
Why are you changing test case? It's data from real device.

oliv3:
Are there other devices than the Android client ?

tananaev:
I don't know, but unit tests are always taken from real device or documentation, so that's a valid sample.

oliv3:
Do you have documentation ? I can't find anything on the web.
All I know now is that there is one client, an Android one. And what I can say from network captures is that this client send battery level, how can you get 110% ? 
So this client was broken. 
Other hypothesis would be it would send charging battery voltage (in this case, 110V). But still, the last bit should be 1, shouldn't it ?

tananaev:
Here is the documentation:

[GPSMTA - GPS трекер](http://doro.poltava.ua/gpsm/gpsmta.html#GPSMTA%20%D1%80%D0%B0%D0%B1%D0%BE%D1%82%D0%B0%20%D1%82%D1%80%D0%B5%D0%BA%D0%B5%D1%80%D0%B0%20%D1%81%20%D0%BB%D1%8E%D0%B1%D1%8B%D0%BC,%20%D0%BE%D1%82%D0%BB%D0%B8%D1%87%D0%BD%D1%8B%D0%BC%20%D0%BE%D1%82%20GPSM,%20%D1%81%D0%B5%D1%80%D0%B2%D0%B5%D1%80%D0%BE%D0%BC%20%D0%BC%D0%BE%D0%BD%D0%B8%D1%82%D0%BE%D1%80%D0%B8%D0%BD%D0%B3%D0%B0)

You are right and it should be battery level, according to the documentation, but we should still keep unit test because it represents data from real device.

oliv3:
Then the test suite will fail.

oliv3:
What are the options ?

tananaev:
We can change `verifyPosition` to something else to not validate attributes.

oliv3:
Spliting in two tests and adding a @Ignore looks so bizarre.

tananaev:
That's not what I suggested.

oliv3:
```verifyPosition``` is used pretty much outside of this protocol, and works very well.
I'd just remove the broken case.

tananaev:
No, you shouldn't remove it because it's a real case.

oliv3:
No, you where replying than me. And I can't rebase github comments.

oliv3:
Then what should be done ? Do you really want to change verifyPosition ?

tananaev:
That's not what I said. I said change `verifyPosition` to something else (e.g. `verifyNotNull`). I never said we should change `verifyPosition` method.

oliv3:
If so, I'll let this to you. Don't want to be responsible for this. Tell me when it's done, I'll drop the unit test commit then. Thanks !

oliv3:
Oooh OK. My bad. Please wait.

oliv3:
Couldn't you suggest using verifyNotNull from the very beginning ??

tananaev:
There might be a better option. I don't remember what's available, off the top of my head.

oliv3:
I assume you would give helpful answers.

oliv3:
Or comments.

tananaev:
I think I gave a helpful advice. I even managed to dig up the protocol.

oliv3:
Right on this point.

tananaev:
Fixed the issue in 3226dac7a8831d96a0adbee9740f78693ea601b9. Next time please notify that you have made the change, so we don't duplicate the work. GitHub doesn't send notifications for commits.

oliv3:
https://github.com/traccar/traccar/commit/3226dac7a8831d96a0adbee9740f78693ea601b9 is patchetic.

tananaev:
English please.

tananaev:
I'm not familiar with the term.

oliv3:
Time to learn.

oliv3:
Google for "portmanteau word".

PR #4100, opened 2018-10-11
Fix MSSQL triggers after renaming columns

Abyss777:
Fixed triggers in MSSQL after renaming columns.
Tested migrations 4.0 to current and clean current.
They should be done as separate SQLs because "CREATE TRIGGER needs to be the first in a batch"


tananaev:
Merged, thanks.

PR #4165, opened 2018-12-16
Added group arg

iEvgeny:
This adds a new URL parameter for the forward functionality. In my case, it was necessary to forward information about the route of buses, which is presented as a group. I suppose this is logical and can be included in the upstream.

### DIFF
--- a/src/org/traccar/WebDataHandler.java
+++ b/src/org/traccar/WebDataHandler.java
@@ -24,6 +24,7 @@ import org.traccar.database.IdentityManager;
 import org.traccar.helper.Checksum;
 import org.traccar.model.Device;
 import org.traccar.model.Position;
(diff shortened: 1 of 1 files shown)


src/org/traccar/WebDataHandler.java
```diff
@@ -139,7 +140,16 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
             request = request.replace("{gprmc}", formatSentence(position));
         }
 
-        return request;
+        String deviceGroupName = "";
+
+        if (device.getGroupId() != 0) {
```
tananaev: I think you need to check parameter first. See examples above.

iEvgeny:
Of course. Fixed. Thank!

PR #4181, opened 2019-01-07
small improvements to logging

flocsy:


src/org/traccar/MainEventHandler.java
```diff
@@ -71,6 +71,16 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
             if (position.getAccuracy() > 0) {
                 s.append(", accuracy: ").append(String.format("%.1f", position.getAccuracy()));
             }
+            if (position.getOutdated()) {
```
tananaev: I would prefer to avoid adding more logging.

src/org/traccar/ServerManager.java
```diff
@@ -88,7 +88,7 @@ public void start() throws Exception {
             try {
                 server.start();
             } catch (BindException e) {
-                LOGGER.warn("One of the protocols is disabled due to port conflict");
+                LOGGER.warn("One of the protocols is disabled due to port conflict: " + server.getPort());
```
tananaev: Please change this to "Port XXX is disabled due to conflict".

src/org/traccar/geocoder/JsonGeocoder.java
```diff
@@ -65,7 +65,7 @@ private String handleResponse(
             return formattedAddress;
         } else {
             if (callback != null) {
-                callback.onFailure(new GeocoderException("Empty address"));
+                callback.onFailure(new GeocoderException("Empty address. " + json.getString("error_message")));
```
tananaev: I don't think this is correct. Are you sure every available geocoder use the same key?

src/org/traccar/model/ExtendedModel.java
```diff
@@ -108,14 +108,18 @@ public boolean getBoolean(String key) {
         }
     }
 
-    public int getInteger(String key) {
+    public Integer getInteger(String key, Integer defaultValue) {
```
tananaev: What is this method for?

flocsy: how do you detect that an integer attribute is not present? See in MainEventHandler for example.

src/org/traccar/MainEventHandler.java
flocsy: currently there are updates (many, for example in my case every minute) that don't have position but do have steps and battery. Currently it's displaying outdated position for every update like that. Basically the best would be in my opinion to display the data related to the actual update. If battery was updated, then display it, but not the other things that were not updated. If position was updated then display it. etc...

So what about the following change:

- don't display the location if it's outdated (and no need to display "outdated")

- I would like to display battery, and maybe other things. Actually maybe it would be worth to make it configurable. You will set it to position only, and I will set it to position,battery,steps and anyone else can chose for themselves. What do you think?

src/org/traccar/MainEventHandler.java
```diff
@@ -34,14 +35,17 @@
...
 
     private final Set<String> connectionlessProtocols = new HashSet<>();
+    private final Set<String> logEvents;
 
     public MainEventHandler() {
         String connectionlessProtocolList = Context.getConfig().getString("status.ignoreOffline");
         if (connectionlessProtocolList != null) {
             connectionlessProtocols.addAll(Arrays.asList(connectionlessProtocolList.split(",")));
         }
+        logEvents = Sets.newHashSet(Context.getConfig().getString("logger.events", DEFAULT_LOGGER_EVENTS).split(","));
```
tananaev: I would prefer to use standard Java API instead and minimize dependencies usage when possible.

src/org/traccar/MainEventHandler.java
```diff
@@ -56,24 +60,43 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
             }
 
             String uniqueId = Context.getIdentityManager().getById(position.getDeviceId()).getUniqueId();
-
```
tananaev: Unnecessary change.

src/org/traccar/MainEventHandler.java
```diff
@@ -56,24 +60,43 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
...
-            Object cmdResult = position.getAttributes().get(Position.KEY_RESULT);
-            if (cmdResult != null) {
-                s.append(", result: ").append(cmdResult);
+            if (position.getOutdated() && logEvents.contains("outdated")) {
+                s.append(", outdated");
+            }
+            if (!position.getValid() && logEvents.contains("invalid")) {
+                s.append(", invalid");
+            }
+            if (logEvents.contains(Position.KEY_BATTERY_LEVEL)) {
```
tananaev: Maybe we should make it even more generic and support all attributes? It seems weird to support only some.

src/org/traccar/ServerManager.java
```diff
@@ -88,7 +88,7 @@ public void start() throws Exception {
             try {
                 server.start();
             } catch (BindException e) {
-                LOGGER.warn("One of the protocols is disabled due to port conflict");
+                LOGGER.warn("Port " + server.getPort() + " is disabled due to conflict");
```
tananaev: Please use formatting version of the method instead of concatenating it yourself.

src/org/traccar/geocoder/JsonGeocoder.java
```diff
@@ -113,4 +114,8 @@ public void failed(Throwable throwable) {
 
     public abstract Address parseAddress(JsonObject json);
 
+    protected String getError(JsonObject json) {
+        return "";
```
tananaev: I don't like this default. It should return `null`. Also, I would rename it to `parseError` for consistency.

src/org/traccar/model/ExtendedModel.java
tananaev: Usually we check `containsKey`.

tools/hex.sh
```diff
@@ -27,7 +27,7 @@ send_hex_udp () {
 }
 
 send_hex_tcp () {
-  echo "$hex" | xxd -r -p | nc "$host" "$port"
+  echo "$hex" | xxd -r -p | nc -N "$host" "$port"
```
tananaev: What does `-N` mean? It doesn't seem to be universally supported, so I would recommend removing.

src/org/traccar/model/ExtendedModel.java
flocsy: yeah, but here you have let's say a Device d, then you would need: d.getAttributes().containsKey('foo'), if it was d.containsKey('foo') then I would maybe agree.

src/org/traccar/MainEventHandler.java
flocsy: maybe. I was thinking about doing something even more general, to be able to set a format string similar to log4j or apache, something like:
<entry key='logger.format'>pos: \{{position.lat},{position.lng}\} battery: {batteryLevel}%, {steps} steps</entry>
maybe I'll add it in a later pull request

src/org/traccar/geocoder/JsonGeocoder.java
flocsy: renamed. Not sure about the null, 'cause it'll generate logs like:
"Empty address. null"

src/org/traccar/model/ExtendedModel.java
tananaev: If you want to change API here, it should probably be the other way around. Single argument can return null, but one with default value shouldn't. And we should use it across the app. I don't want to implement a new API just for logging.

flocsy: or another way we can do: rename (refactor) the current function to: int getInt(String key), and add a new function: Integer getInteger(String key) that will return null if it's not contained.

tananaev: I don't see the point of creating a function just to use it in a single place. If you want to change everything to make it consistent, you are welcome, but probably in different PR. For now I would recommend to stick to existing practice.

flocsy: ok

src/org/traccar/MainEventHandler.java
flocsy: made it more general: 1. it's displayed in the same order as it appears in the config, 2. you can choose any attribute

tananaev:
Merged, thanks. Sorry for so many comments. In future I would recommend breaking down PRs into smaller unrelated pieces. Easier to review and easier to get approved.

PR #4183, opened 2019-01-09
added TK2

flocsy:
TK2 behaves similar to TK. The watch tries to send it with TK and if it fails it retries with TK2 until is successfully can send it to the server.


tananaev:
You would need to provide some description before we can merge this.

flocsy:
TK2 behaves similar to TK. The watch tries to send it with TK and if it fails it retries with TK2 until is successfully can send it to the server.

tananaev:
Do you have documentation for it?

PR #4188, opened 2019-01-11
Added logger.rotate configuration parameter for compatibility with logrotate

iEvgeny:
A new option allows to disable native rotation of the log file. This feature allows compatibility with logrotate in unix-based systems.
Logrotate config example /etc/logrotate.d/traccar:
`/var/log/traccar/tracker-server.log {`
`    copytruncate`
`    daily`
`    rotate 7`
`    compress`
`    delaycompress`
`    missingok`
`    notifempty`
`    create 0644 traccar traccar`
`}`


tananaev:
Great, have you tested both cases?

PR #4190, opened 2019-01-13
allow multiple status/alarms

flocsy:
1. sos alarm doesn't override other alarms
2. multiple states/alarms (according to the bits)


src/org/traccar/protocol/WatchProtocolDecoder.java
```diff
@@ -62,6 +64,36 @@ public WatchProtocolDecoder(Protocol protocol) {
             .expression("(.*)")                  // cell and wifi
             .compile();
 
+    private static final String STATE = "State";
+    private static final String ALARM = "Alarm";
+
+    private static final int STATUS_LOW_BATTERY_STATUS_BIT = 0;
```
tananaev: I don't see any good reason for having these constants.

src/org/traccar/protocol/WatchProtocolDecoder.java
```diff
@@ -62,6 +64,36 @@ public WatchProtocolDecoder(Protocol protocol) {
             .expression("(.*)")                  // cell and wifi
             .compile();
 
+    private static final String STATE = "State";
+    private static final String ALARM = "Alarm";
```
tananaev: What is this for?

src/org/traccar/protocol/WatchProtocolDecoder.java
flocsy: the numbers?

src/org/traccar/protocol/WatchProtocolDecoder.java
flocsy: The terminal statement | 00000000 | Represents 			in Hexadecimal string,and the meaning isas following:The high 			16bit is alarming,the low 16 bit is statement. 			Bit(from 			0)        Meaning(1 			effective) 			 			0                  Low-battery 			 			1                  Out of fence  			 			 			2                  In fence 			 			3                  Bracelet statement 			 			16                 SOS alarming 			 			17                 Low-battery alarming 			 			18                 Out of fence alarming 			 			19                 In fence alarming 			20     			            Watch-take-off alarming
-- | -- | --



from here: https://dl.dropboxusercontent.com/s/2hid9vaa2vno4jr/Communication%20Protocol.doc  (linked from https://www.traccar.org/protocols/)

tananaev: Status should be ignored, unless we have special attribute for it. Alarm should only include alarms.

flocsy: So the current code in master is wrong.
Either new constants could be added to Position to accomodate status,
or we can remove bits 0-3. from the above code.
I still think that we should allow multiple alerts at the same time, it's possible to have sos, either entering or exiting geofence (actually it could be both at the same time, depends on how the user draw them:) and watch removed alert at the same time.

tananaev: It's very possible that something is not decoded correctly.

Yes, it would be good to support multiple alerts. There was even some work done for it somewhere.

flocsy: ok, so that part of the patch is ok at least:)
So do you want me to remove STATE, ALARM Strings and bits 0-3 from the Map, leaving only the alerts? Or should we somehow add/keep the information about the status? Either by renaming the method to decodeStatus (I think this is better, and BTW we could add then motion as well) or if you prefer to have an alert only method, then by adding another one for the "rest" of the bits: decodeSatus() ?

tananaev: I would prefer to separate two.

tananaev:
Have you checked that multiple alarms are supported everywhere?

flocsy:
what do you mean everywhere?

tananaev:
I mean all places where we handle events on the server, and in the web app as well.

tananaev:
We already support multiple alarms.

PR #4191, opened 2019-01-13
fixed exception when String.format thought the binary data has formt placeholders

flocsy:


tananaev:
Please include unit test.

flocsy:
the existing unittest covers it, that's why the 1st commit failed

tananaev:
Well, can we have test case for the different behavior?

flocsy:
Added a test. It passes with:
return formatCommand(channel, command, "TK,%s", getBinaryData(command));
but fails with the old:
return formatCommand(channel, command, "TK," + getBinaryData(command));

tananaev:
Merged, thanks.

PR #4192, opened 2019-01-14
Send message (that is displayed on the watch's display)

flocsy:
protocol documentation: https://github.com/tananaev/traccar/files/213814/3g.elec.comm.protocol.docx


tananaev:
Because you are adding a completely new command, you need to implement it in the web app as well.

flocsy:
ok, I'll try to add it there too, but it's in another repo anyway, so any real reason not to include it here already?

tananaev:
Merged, thanks.

PR #4194, opened 2019-01-14
added "debug" to logger.events

flocsy:


tananaev:
Please explain why we need two ways of doing the same thing.

flocsy:
It's not the same. Here you'll see the whole Position dumped, which helps a lot when you don't know it's structure.

tananaev:
I still don't like two way of doing same thing. You can always check database if you want to see full information. Log file is NOT for that.

PR #4206, opened 2019-01-29
fixed message encoding

flocsy:
without this the received message is displayed as gibrish

### DIFF
--- a/src/org/traccar/protocol/WatchProtocolEncoder.java
+++ b/src/org/traccar/protocol/WatchProtocolEncoder.java
@@ -37,7 +37,7 @@ public class WatchProtocolEncoder extends StringProtocolEncoder implements Strin
             DecimalFormat fmt = new DecimalFormat("+#.##;-#.##", DecimalFormatSymbols.getInstance(Locale.US));
             return fmt.format(offset);
         } else if (key.equals(Command.KEY_MESSAGE)) {
-            return DataConverter.printHex(value.toString().getBytes());
+            return DataConverter.printHex(value.toString().getBytes(StandardCharsets.UTF_16));
         }
 
         return null;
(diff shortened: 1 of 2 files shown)


tananaev:
Can I see documentation for the protocol where it says it expects UTF-16?

flocsy:
https://github.com/tananaev/traccar/files/213814/3g.elec.comm.protocol.docx here the example is UTF-16, and for the SMS command it's written that it's "GB232" charset, but the meaning even there is UTF-16.
However the final fix came when today I tested it with a real Q50 watch. Even an ASCII text was displayed as "boxes" until I changed the encoding to UTF-16, and then it appeared as intended.

tananaev:
Can you point me to the specific example with UTF-16. I don't see anything in the document.

flocsy:
33. Phrases Display set instruction
Platform to send:
[CS*YYYYYYYYYY*LEN*MESSAGE, phrase content]
Example:[SG*5678901234*0018*MESSAGE,**597D003100320033**]

flocsy:
6. Send text messages
Platform to send:
[CS*YYYYYYYYYY*LEN*SMS, SMS number, the content of the message]
Example: [SG*8800000015*001C*SMS,00000000000,123ABC everybody good]
Terminal response:
[CS*YYYYYYYYYY*LEN*SMS]
Example:[SG*8800000015*0003*SMS]

Explanation: send a message to the mobile phone number with the command, the content of the message sent by **GB232 code**.

tananaev:
Not sure how you determined that it's UTF-16. But even if it is, it doesn't include BOM, but your code does.

flocsy:
screenshot from 2019-01-30 05-15-31
img_20190130_051244
screenshot from 2019-01-30 05-14-15
img_20190130_051127

flocsy:
You're right, it works without the BOM as well, I changed that in the 2nd commit

flocsy:
and about the other comment: message sending was added by me about a week ago, not sure how many used it since, probably I'm the only one today... So it might or might not work. Even if it will it won't work because of what we "understand" from the google-translated Chinese "documentation" but because we tried it. So choosing not to change the encoding will for sure not work on my watch (I guss that's 100% of the watches who use it currently) and maybe will work on other watches that didn't have this feature 2 weeks ago...

tananaev:
Merged, thanks.

PR #4221, opened 2019-02-12
fixed 8bit voice message

flocsy:
This is the smallest patch I could create that still works. If either one of the 2 places doesn't use ISO_8859_1 then the watch either doesn't even get the voice message or it gets gibrish.

This was tested with Q50 watch.

### DIFF
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -516,6 +516,10 @@ public final class Context {
         identityManager = testIdentityManager;
     }
 
+    public static void initMediaManager(MediaManager mediaManager) {
+        Context.mediaManager = mediaManager;
+    }
+
     public static <T extends BaseModel> BaseObjectManager<T> getManager(Class<T> clazz) {
         if (clazz.equals(Device.class)) {
             return (BaseObjectManager<T>) deviceManager;
--- a/src/org/traccar/Context.java
+++ b/src/org/traccar/Context.java
@@ -508,16 +508,13 @@ public final class Context {
                 Context.getConfig().getBoolean("event.overspeed.preferLowest"));
     }
 
-    public static void init(IdentityManager testIdentityManager) {
+    public static void init(IdentityManager testIdentityManager, MediaManager testMediaManager) {
         config = new Config();
         objectMapper = new ObjectMapper();
         objectMapper.registerModule(new JSR353Module());
         client = ClientBuilder.newClient().register(new ObjectMapperContextResolver());
         identityManager = testIdentityManager;
-    }
-
-    public static void initMediaManager(MediaManager mediaManager) {
-        Context.mediaManager = mediaManager;
+        mediaManager = testMediaManager;
     }
 
     public static <T extends BaseModel> BaseObjectManager<T> getManager(Class<T> clazz) {
--- a/src/org/traccar/protocol/WatchFrameDecoder.java
+++ b/src/org/traccar/protocol/WatchFrameDecoder.java
@@ -33,7 +33,8 @@ public class WatchFrameDecoder extends BaseFrameDecoder {
             while (buf.readerIndex() < endIndex) {
                 byte b = buf.readByte();
                 if (b == '}') {
-                    switch (buf.readByte()) {
+                    byte c = buf.readByte();
+                    switch (c) {
                         case 0x01:
                             frame.writeByte('}');
                             break;
@@ -50,7 +51,8 @@ public class WatchFrameDecoder extends BaseFrameDecoder {
                             frame.writeByte('*');
                             break;
                         default:
-                            throw new IllegalArgumentException();
+                            throw new IllegalArgumentException(String.format(
+                                    "unexpected byte at %d: 0x%02x", buf.readerIndex() - 1, c));
                     }
                 } else {
                     frame.writeByte(b);
--- a/src/org/traccar/protocol/WatchFrameDecoder.java
+++ b/src/org/traccar/protocol/WatchFrameDecoder.java
@@ -31,10 +31,10 @@ public class WatchFrameDecoder extends BaseFrameDecoder {
         if (endIndex > 0) {
             ByteBuf frame = Unpooled.buffer();
             while (buf.readerIndex() < endIndex) {
-                byte b = buf.readByte();
-                if (b == '}') {
-                    byte c = buf.readByte();
-                    switch (c) {
+                byte b1 = buf.readByte();
+                if (b1 == '}') {
+                    byte b2 = buf.readByte();
+                    switch (b2) {
                         case 0x01:
                             frame.writeByte('}');
                             break;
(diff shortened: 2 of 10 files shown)


tananaev:
I don't think it's the right solution. We shouldn't use `StringEncoder` for binary or partially binary protocols.

flocsy:
do you have an example of another protocol that I can look at?

tananaev:
Try Meiligao.

flocsy:
Anton, what do you say now?

src/org/traccar/Context.java
```diff
@@ -516,6 +516,10 @@ public static void init(IdentityManager testIdentityManager) {
         identityManager = testIdentityManager;
     }
 
+    public static void initMediaManager(MediaManager mediaManager) {
```
tananaev: This should be part of `init`.

src/org/traccar/protocol/WatchFrameDecoder.java
```diff
@@ -33,7 +33,8 @@ protected Object decode(
             while (buf.readerIndex() < endIndex) {
                 byte b = buf.readByte();
                 if (b == '}') {
-                    switch (buf.readByte()) {
+                    byte c = buf.readByte();
```
tananaev: We need better names.

src/org/traccar/protocol/WatchProtocolDecoder.java
```diff
@@ -174,41 +182,45 @@ public String getManufacturer() {
         return manufacturer;
     }
 
+    protected int nextIndexOf(ByteBuf buf, char delimiter) {
```
tananaev: Not a big fan of one-line methods. It makes it harder to read the code.

src/org/traccar/protocol/WatchProtocolDecoder.java
```diff
@@ -62,15 +67,18 @@ public WatchProtocolDecoder(Protocol protocol) {
             .expression("(.*)")                  // cell and wifi
             .compile();
 
-    private void sendResponse(Channel channel, String id, String index, String content) {
+    private void sendTextResponse(Channel channel, String id, String index, String content) {
```
tananaev: Do we ever send non-text responses? Do we need to rename it?

src/org/traccar/protocol/WatchProtocolDecoder.java
```diff
@@ -296,6 +308,14 @@ protected Object decode(
 
         } else if (type.equals("TK")) {
 
+            if (buf.readableBytes() == 1) {
+                byte result = buf.readByte();
+                if (result != '1') {
+                    LOGGER.error(type + "," + result);
```
tananaev: This is reserved for some serious errors, not protocols decoding issues. Also, what is the purpose of this check?

src/org/traccar/protocol/WatchProtocolEncoder.java
```diff
@@ -15,6 +15,8 @@
  */
 package org.traccar.protocol;
```
tananaev: This class probably shouldn't extend string encoder. It doesn't really return strings.

test/org/traccar/protocol/WatchProtocolDecoderTest.java
```diff
@@ -101,7 +111,40 @@ public void testDecode() throws Exception {
 
         verifyPosition(decoder, buffer(
                 "[ZJ*014111001350304*0038*008a*UD,070318,021027,V,00.000000,N,000.000000,E,0,0,0,0,100,18,1000,50,00000000,4,255,460,0,9346,5223,42,9346,5214,20,9784,4083,11,9346,5221,5]"));
+    }
+
+    @Test
+    public void testDecodeTK() throws Exception {
```
tananaev: Wrong naming convention. Also probably should have better name.

src/org/traccar/Context.java
flocsy: What do you mean on "part"? It is only used in the unittests. In init we have: mediaManager = new MediaManager(config.getString("media.path")); I could change it to initMediaManager(new MediaManager(config.getString("media.path"))); but I don't think that's what you meant.

src/org/traccar/protocol/WatchFrameDecoder.java
flocsy: like??? d instead of c? :) If b is good enough, then I can change c b2 (it's the byte after b) or ?

src/org/traccar/protocol/WatchProtocolDecoder.java
flocsy: duplicate code makes it harder to maintain :) Besides IMHO nextIndexOf is clearer than to look at the original long lines.

src/org/traccar/protocol/WatchProtocolDecoder.java
flocsy: not really, I was also thinking if I should keep it, we can still overload it later if needed. I'll change it back

src/org/traccar/protocol/WatchProtocolDecoder.java
flocsy: The 1st if is needed because otherwise we are saving 1 byte long amr files when the device responds to a voice message we sent.

src/org/traccar/protocol/WatchProtocolEncoder.java
flocsy: Also thought, but it still uses formatCommand() of StringProtocolEncoder in formatTextCommand. If I don't extend StringProtocolEncoder then I'll have to find another way to do it, which would probably involve code duplication. I could make an inner class in WatchProtocolEncoder that extends StringProtocolEncoder and includes the formatValue. (Thoufh formatCommand is protected...) and we would also have another "copy" of BaseProtocolEncoder inside WatchProtocolEncoder.

test/org/traccar/protocol/WatchProtocolDecoderTest.java
flocsy: Please when you write comments about names give some suggestions. I don't see what's the problem with this name, it explains exactly what it does.

src/org/traccar/protocol/WatchFrameDecoder.java
tananaev: B is short for Byte. C is nothing. Just a random letter. B2 is fine, but then rename first one to B1.

src/org/traccar/protocol/WatchProtocolDecoder.java
tananaev: I don't agree with this. It's one line of code. It's easier to understand code when you know what it's doing and you don't need to refer to a separate method.

test/org/traccar/protocol/WatchProtocolDecoderTest.java
tananaev: Name should be camel-case. Name should be more obvious what it does. I don't know what TK means, so I don't really know what it tests.

src/org/traccar/protocol/WatchProtocolEncoder.java
tananaev: OK, we can probably keep it.

src/org/traccar/Context.java
tananaev: I want to have single `init` for unit tests.

flocsy: why? most tests don't need everything. isn't it a waste of time/resources?

tananaev: It's easier to maintain and creating a single empty object doesn't really affect performance or resource usage.

flocsy:
all resolved

tananaev:
Merged, thanks.

PR #4223, opened 2019-02-14
Fixed geofences processing order

iEvgeny:
If the position simultaneously generates an exit from the previous geofence and an entrance to the next one, then we will get an incorrect sequence of events. I think that in this case the geofenceExit event should be before the geofenceEnter event.
Example: [report.xlsx](https://github.com/traccar/traccar/files/2866664/report.xlsx)


tananaev:
Merged, thanks.

PR #4228, opened 2019-02-22
added event, position data to firebase notifications

flocsy:


tananaev:
Can we use existing event and position models instead?

flocsy:
I thought that most of the things are not relevant for push notifications, because they can be retrieved by the client at the moment it needs it, so I only wanted to send the "minimal" data that is needed to decide how to display the notification on the client.

Also data is Map<String, String> so it wouldn't be easy to fit event and position in there, only by serializing them to json string. But then it's harder to get the data from there, and also it's unnecessarily longer.

tananaev:
I think you arbitrary picked what interests you most, which is obviously not the best approach.

flocsy:
Maybe, but the only difference between this and your picks in MainEventHandler is: who chose them :)

tananaev:
Originally main event handler only logged position fields. No attributes at all.

flocsy:
I tried this:
    public static Map<String, String> buildData(long userId, Event event, Position position) {
        Map<String, String> data = new HashMap<>();
        try {
            if (event != null) {
                String eventJson = Context.getObjectMapper().writeValueAsString(event);
                data.put("eventJson", eventJson);
            }
            if (position != null) {
                String positionJson = Context.getObjectMapper().writeValueAsString(position);
                data.put("positionJson", positionJson);
            }
        } catch (JsonProcessingException e) {
            LOGGER.warn("Notification JSON formatting error", e);
        }
        LOGGER.info("data: " + data);
        return data;
    }

the problem is that for some reason the json doesn't arrive as it is sent. For example the data I send starts like:
data: {eventJson={"id":2136,"attributes":{"audio":"20190222102427.amr"}
as you see eventJson is correct jsonString

but when I receive it it becomes:
data payload: {eventJson={&#34;id&#34;:2136,&#34;attributes&#34;:{&#34;audio&#34;:&#34;20190222102427.amr&#34;}

I'm not sure why. Looks like traccar incorrectly transfers the data before it's sent to firebase.

flocsy:
Actually even body is encoded incorrectly. When I add double quotes to it (Or to the name of the device):

            notification.body = "foo {\"\"} " + NotificationFormatter.formatShortMessage(userId, event, position).trim();

when it arrives it becomes: "foo {&#34;&#34;} Voice message received from"...

flocsy:
maybe we should use: https://firebase.google.com/docs/reference/admin/java/reference/com/google/firebase/messaging/FirebaseMessaging ?

flocsy:
Ok, while it works, it looks that it doesn't work for the use-case I needed it, because of this: https://firebase.google.com/docs/cloud-messaging/concept-options#notification-messages-with-optional-data-payload and what I'd like to do is to customize the notification using the data, but it's not available when the app is in the background. So I'll need to send data-only message, but this is not backwards-compatible, so I'm thinking about some kind of configuration to tell if we need to send notification + data or only data (with the title, body fields part of the data). How would you accept this? Make it a global configuration? Make it configurable per user (who will receive the notification)?

tananaev:
Global configuration should be fine.

flocsy:
is this OK?

tananaev:
Position and event should be proper JSON objects, not strings formatted as JSON.

flocsy:
https://firebase.google.com/docs/cloud-messaging/http-server-ref
Values in string types are recommended. You have to convert values in objects or other non-string data types (e.g., integers or booleans) to string.

tananaev:
I still think we should send nested objects. They don't provide any reasonable justification for why it's not recommended. It works, so I think we should use it. It feels like the right thing to do.

flocsy:
https://firebase.google.com/docs/reference/android/com/google/firebase/messaging/RemoteMessage

Map<String, String> | getData()Gets the message payload data.

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -115,4 +118,20 @@ private static String formatMessage(VelocityContext vc, Long userId, Event event
         return writer.toString();
     }
 
+    public static Map<String, String> buildData(long userId, Event event, Position position) {
+        Map<String, String> data = new HashMap<>();
+        try {
+            if (event != null) {
+                String eventJson = Context.getObjectMapper().writeValueAsString(event);
```
tananaev: Remove unnecessary variable.

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -115,4 +118,20 @@ private static String formatMessage(VelocityContext vc, Long userId, Event event
...
+                data.put("eventJson", eventJson);
+            }
+            if (position != null) {
+                String positionJson = Context.getObjectMapper().writeValueAsString(position);
+                data.put("positionJson", positionJson);
+            }
+        } catch (JsonProcessingException e) {
```
tananaev: I would throw this exception instead.

src/org/traccar/notification/NotificationFormatter.java
```diff
@@ -115,4 +118,20 @@ private static String formatMessage(VelocityContext vc, Long userId, Event event
         return writer.toString();
     }
 
+    public static Map<String, String> buildData(long userId, Event event, Position position) {
+        Map<String, String> data = new HashMap<>();
+        try {
+            if (event != null) {
+                String eventJson = Context.getObjectMapper().writeValueAsString(event);
+                data.put("eventJson", eventJson);
```
tananaev: Rename to "event" and "position". Probably better to use constants for keys.

src/org/traccar/notificators/NotificatorFirebase.java
```diff
@@ -46,29 +49,38 @@
...
+                notification.body = body;
+                message.notification = notification;
+            }
+
+            ClientBuilder.newClient().target(URL).request()
                     .header("Authorization", "key=" + key)
                     .async().post(Entity.json(message), new InvocationCallback<Object>() {
                 @Override
                 public void completed(Object o) {
+                    LOGGER.info("Firebase notification sent to user: {}", userId);
```
tananaev: If you think we really need this, it should not be "info" level. Make it debug or lower.

src/org/traccar/notificators/NotificatorFirebase.java
```diff
@@ -46,29 +49,38 @@
...
+            String body = NotificationFormatter.formatShortMessage(userId, event, position).trim();
+            if (isFormatDataOnly) {
+                message.data.put("body", body);
+            } else {
+                Notification notification = new Notification();
+                notification.body = body;
+                message.notification = notification;
+            }
+
+            ClientBuilder.newClient().target(URL).request()
```
tananaev: Why new client?

tananaev (commented):
OK, added some comments.

src/org/traccar/notificators/NotificatorFirebase.java
flocsy: the problem is that for some reason the json doesn't arrive as it is sent. For example the data I send starts like:
`data: {eventJson={"id":2136,"attributes":{"audio":"20190222102427.amr"}`
as you see eventJson is correct jsonString

but when I receive it it becomes:
`data payload: {eventJson={&#34;id&#34;:2136,&#34;attributes&#34;:{&#34;audio&#34;:&#34;20190222102427.amr&#34;}`

flocsy: this is when it's being sent with the default client

src/org/traccar/notificators/NotificatorFirebase.java
```diff
@@ -46,29 +49,38 @@
...
-
             Message message = new Message();
             message.tokens = user.getString("notificationTokens").split("[, ]");
-            message.notification = notification;
-
-            Context.getClient().target(URL).request()
+            message.data = NotificationFormatter.buildData(userId, event, position);
+            String body = NotificationFormatter.formatShortMessage(userId, event, position).trim();
+            if (isFormatDataOnly) {
+                message.data.put("body", body);
```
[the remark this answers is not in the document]
flocsy: Why not? The client still uses it for the notification title, it just ALSO uses other data from the data field

tananaev:
I'm more and more convinced that we shouldn't include this in the official Traccar at this point. It's not used by Traccar apps, so I think we'll leave it for later.

src/org/traccar/notification/NotificationFormatter.java
flocsy: copied from AsyncSocket::sendData. It might be better to send a dataless notification, than nothing IMHO. Maybe I can refactor NotificatorFirebase so that if it gets an empty data map (or null?) then it falls back to the notification without data?

josuelmm:
I strongly agree that this change be applied to traccar. In order to customize the notifications and not to be established as it comes in the default language. If done in this way allows when creating a mobile application, customize the message to send to the device.

PR #4241, opened 2019-03-07
Add refuel

meteviasa:
15f0744893304a3f2a77d36734cdc2a2706286dd

### DIFF
--- a/src/org/traccar/handler/events/RefuelEventHandler.java
+++ b/src/org/traccar/handler/events/RefuelEventHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 
+ * Copyright 2017 Anton Tananaev (anton@traccar.org)
  *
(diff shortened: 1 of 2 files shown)


gpproton (commented):
Hi @tananaev i feel this refuel event is quite useful, any specific reason you're not considering @meteviasa commits, apart for the minor refactoring required.

tananaev:
The code is not ready and there is no description.

gpproton:
Noted, but do you have any plan on assisting with this. although apart from this event i do feel there's a need to do a check for refuel on (/src/main/java/org/traccar/reports/Summary.java, /src/main/java/org/traccar/reports/Trips.java and /src/main/java/org/traccar/reports/Stops.java)
to give a more accurate result not just a simple subtraction which can easily go wrong in cases of refuel, as i've tried using it with traccar and the results are not impressive.

tananaev:
I believe we already have something similar.

PR #4253, opened 2019-03-19
Modified client proxy servlet to response with access control allow headers

JulioLopezN:


tananaev:
Please provide description on why we would want to include this in the official release.

JulioLopezN:
The problem that I had was that when I was wanting to send a geolocation from a hybrid mobile application I always had the error `No 'Access-Control-Allow-Origin' header is present on the requested resource.` thats why I modified the client proxy servlet to response with those headers. I think I'm not the only one with that problem. Do you have an advice for that problem?

tananaev:
You should should provide much more details. What geolocation? What hybrid app? What does any of that has to do with CORS headers? etc etc etc

JulioLopezN:
I'm developing a mobile application with Ionic 4. The application sends the current position of the device with the plugin `@ionic-native/background-geolocation` via HTTP, as it is a hybrid application everything runs on the mobile browser, that's where the CORS gets angry because the application sends a HTTP request to another domain. The client proxy servlet wasn't prepared to handle CORS requests that's why I modified it.

tananaev:
You shouldn't use web port for sending location data in the first place.

JulioLopezN:
What is the correct way to do that?

tananaev:
You should probably implement another API endpoint if you really want to do it properly.

JulioLopezN:
Ok, I wiil try to do that

PR #4265, opened 2019-04-04
Adding possibility for TOTP using Google Authenticator

mivtdole:
For our own environment I added 2FA login using TOTP and the Google Authenticator.
Hope you like this addition.


jinzo:
As I was looking at various 2FA solutions to implement - this could be a  good starting point. What do you think @tananaev? Are you open to merging features like this?

tananaev:
On a very high level, I think it's a good idea. One thing I would like to change is rename everything to not be "google auth". Instead we should use some other general term. For example, TOTP or something like that.

mivtdole:
@tananaev  I can change the naming from Google to something else. In this case I think 'totp' would be best.

@jinzo Changing it from a global setting to a per user setting is not a problem. In our case it needed to be global. Therefore I also created the config option 'googleauth.notfor' so you can specify one user for which it is not operational to prevent the chicken and egg problem.

Should the user be able to turn it on/off or only the administrator?

I will leave a config option to enable it globally so it is enforced. This will cover our requirement.

Anything else missing?

tananaev:
I thought a bit more about naming and we can probably just use "code" in many places. In other places we can use "totp", unless there is an alternative.

As for configuration, I think following makes sense:
- There should be a flag to disable the feature completely
- If the feature is enabled:
  - It should be possible to enforce it for all users
  - It should be possible to make it optional (each user sets it up if he wants to)

I see that you have a configuration option to ignore it for certain users. I'm not sure that's a good idea.

mivtdole:
I am working on a version which includes above feedback. As I will be busy the next two weeks so it can take a while.

mivtdole:
@tananaev let me know if this version is better?
- I renamed the fields and labels
- Two configuration settings. 1: to enable TOTP and 2: acivate TOTP per user or active for all users

README.md
```diff
@@ -27,6 +27,7 @@ Some of the available features include:
 - Alarms and notifications
 - Account and device management
 - Email and SMS support
+- TOTP multi-factor authentication (Tested with Google Authenticator)
```
tananaev: Please revert this. We can't really list all available features in this list. It's only major ones.

build.gradle
```diff
@@ -60,6 +60,7 @@ dependencies {
     implementation "com.sun.xml.bind:jaxb-core:2.3.0.1"
     implementation "com.sun.xml.bind:jaxb-impl:2.3.2"
     implementation "javax.activation:activation:1.1.1"
+    compile 'com.warrenstrange:googleauth:1.1.2'
```
tananaev: This should be `implementation`.

schema/changelog-4.4b.xml
```diff
@@ -0,0 +1,20 @@
...
+  logicalFilePath="changelog-4.4b">
+
+  <changeSet author="author" id="changelog-4.4b-all">
+    <sql>
+      ALTER TABLE tc_users
+      ADD totpKey VARCHAR(128)
+    </sql>
+    <sql>
+      ALTER TABLE tc_users
+      ADD useTotp BOOLEAN DEFAULT FALSE;
```
tananaev: Already commented on web PR. I think single field is enough.

schema/changelog-master.xml
```diff
@@ -24,4 +24,6 @@
 
   <include file="changelog-4.1.xml" relativeToChangelogFile="true" />
 
+  <include file="changelog-4.4b.xml" relativeToChangelogFile="true" />
```
tananaev: Why is it `4.4b` instead of `4.4`?

src/main/java/org/traccar/model/User.java
```diff
@@ -273,4 +277,56 @@ public boolean isPasswordValid(String password) {
...
+        if (!Context.getConfig().getBoolean("totp.enabled")) {
+            return false;
+        }
+        return this.useTotp;
+    }
+
+    @JsonIgnore
+    @QueryIgnore
+    public boolean isTotpAuthCodeValid(int code)
+    throws GoogleAuthenticatorException {
```
tananaev: Formatting issue.

src/main/java/org/traccar/model/User.java
```diff
@@ -273,4 +277,56 @@ public boolean isPasswordValid(String password) {
...
+        if (!Context.getConfig().getBoolean("totp.enabled")) {
+            return true;
+        }
+
+        GoogleAuthenticator gAuth = new GoogleAuthenticator();
```
tananaev: Please use full words for variable names as much as possible. This can be `authenticator` for example.

schema/changelog-4.4b.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd"
+  logicalFilePath="changelog-4.4b">
+
+  <changeSet author="author" id="changelog-4.4b-all">
+    <sql>
```
tananaev: Why is it SQL instead of proper liquibase definition?

setup/default.xml
```diff
@@ -76,6 +76,10 @@
         SELECT * FROM tc_statistics WHERE captureTime BETWEEN :from AND :to ORDER BY captureTime
     </entry>
 
+    <entry key='totp.enabled'>false</entry>
```
tananaev: If these are defaults, please remove from this file.

src/main/java/org/traccar/api/resource/SessionResource.java
```diff
@@ -99,9 +99,15 @@ public User get(@QueryParam("token") String token) throws SQLException, Unsuppor
     @PermitAll
     @POST
     public User add(
-            @FormParam("email") String email, @FormParam("password") String password) throws SQLException {
+            @FormParam("email") String email, @FormParam("password") String password,
+            @FormParam("code") int code) throws SQLException {
```
tananaev: Why `int` instead of `string`?

src/main/java/org/traccar/api/resource/SessionResource.java
```diff
@@ -99,9 +99,15 @@ public User get(@QueryParam("token") String token) throws SQLException, Unsuppor
...
         User user = Context.getPermissionsManager().login(email, password);
         if (user != null) {
+            if (Context.getConfig().getBoolean("totp.enabled")) {
```
tananaev: Looks like it's checked twice. Here and in the `isTotpAuthCodeValid`.

src/main/java/org/traccar/model/User.java
```diff
@@ -273,4 +277,56 @@ public boolean isPasswordValid(String password) {
...
+        if (Context.getConfig().getString("totp.preference").toLowerCase().equals("user")) {
```
tananaev: What does preference `user` mean? I think we should rename the configuration to be a boolean flag and call it something like `totp.force`.

schema/changelog-4.4b.xml
mivtdole: This field is there when in the server configuration the totp preference is user instead of all. The 'user' settings means we can turn totp on per user while 'all' means it is on for every user it cannot be turned off.
If we only want to have the 'user' option than this checkbox can go.

tananaev: Let's continue discussion in a single place. Please don't duplicate comments.

README.md
mivtdole: I removed the line

build.gradle
mivtdole: done

schema/changelog-master.xml
mivtdole: Modified it to 4.4

src/main/java/org/traccar/model/User.java
mivtdole: This is resolved by catching the exception within the method. This will not throw an exception anymore.

schema/changelog-4.4b.xml
mivtdole: Fixed. Was my first time working with liquibase

setup/default.xml
mivtdole: removed as false is the default. Documentation needs to be updated for these new settings

src/main/java/org/traccar/api/resource/SessionResource.java
mivtdole: Modified to string so when in the future other 2FA solutions are added the code can contain other characters than numbers. Currently I am only aware of numeric 2FA codes.

src/main/java/org/traccar/api/resource/SessionResource.java
mivtdole: This check has been removed is it is indeed also checked within the user class method isTotpAuthCodeValid.

src/main/java/org/traccar/model/User.java
mivtdole: Renamed this setting to 'totp.enforce' By default it is false.

tananaev:
You still haven't fixed the submodule issue.

mivtdole:
I finally got the issue with the traccar-web submodule fixed.

tananaev:
Looks like not all comments are addressed. Also, please squash your commits.

adadnc:
Is this going to be in one of the next releases?

tananaev:
The issue with submodule is still not fixed, so unfortunately we can't merge it as it is.

adadnc:
I'm not sure if the merge requester knows your expectations.

adadnc:
This pull request is a very good idea and also adapted nearly fully. It is too bad to leave it like this. No way on getting the last step with the submodule fixed?

mivtdole:
I have it on my todo list but my work is currently asking to much time.

mivtdole:
Ok fixed the traccar-web submodule commit version. Will no look into squashing the commit history

mivtdole:
I am setting up a new clean fork to see if it solves the problems which still exist.

ryan1969:
What is all this

adadnc:
If someone could explain the problem, maybe we could help?

mivtdole:
I will create a new pull request which is clean and contains everything.

PR #4270, opened 2019-04-06
Updated xexun sos alarm

redowk:
Added "help me" option, for sos alarm alert to the xexun protocol.
Following the discussion and tests described here: https://www.traccar.org/forums/topic/xexun-implement-sos-alarms

### DIFF
--- a/src/main/java/org/traccar/protocol/T800xProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/T800xProtocolDecoder.java
@@ -47,7 +47,7 @@ public class T800xProtocolDecoder extends BaseProtocolDecoder {
 
     private void sendResponse(Channel channel, short header, int type, int index, ByteBuf imei, int alarm) {
         if (channel != null) {
-            ByteBuf response = Unpooled.buffer(15);
+            ByteBuf response = Unpooled.buffer(alarm > 0 ? 16 : 15);
             response.writeShort(header);
             response.writeByte(type);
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/protocol/XexunProtocolDecoder.java
```diff
@@ -79,6 +79,9 @@ private String decodeStatus(Position position, String value) {
                 case "help me!":
                     position.set(Position.KEY_ALARM, Position.ALARM_SOS);
                     break;
+                case "help me":
+                    position.set(Position.KEY_ALARM, Position.ALARM_SOS);
```
tananaev: You should not duplicate code here. Add second switch case for exiting SOS alarm option.

redowk: You mean a new switch case with only one option? Why?
```            switch (value.toLowerCase()) {
                case "help me":
                    position.set(Position.KEY_ALARM, Position.ALARM_SOS);
                    break;
                default:
                    break;
            }```

Comment (tananaev) [reply src/main/java/org/traccar/protocol/XexunProtocolDecoder.java]: This is what I meant:
```java
case "help me!":
case "help me":
    position.set(Position.KEY_ALARM, Position.ALARM_SOS);
    break;
```

redowk: sorry! I understand.
Thanks for your patience!

PR #4294, opened 2019-04-28
support gt06 extend Protocol

juzzs6212:
support gt06 extend Protocol


tananaev:
I was actually just working on the same thing. I think the right way is to map each parameter to a single attribute though.

tananaev:
Merged and refactored. Thanks for pull request.

juzzs6212:
Happy cooperation, thank you for your support.

PR #4309, opened 2019-05-06
Add user,group telegram notification

7hemas7er:
set a variable "notificationTelegramChatId" in group or in user to send a message to a specific chat

Issue #4274, opened 2019-04-10
Added Telegram Notification support


he-leon:
Hi, 

I added Notifications support for Telegram Messenger. See 22436a2434faa8912c8a352e65306d0dc0669b5c

Example config:
```
<entry key='notificator.types'>telegram,web</entry>
<entry key='notificator.Telegram.apikey'>YOUR_BOT_API_KEY</entry>
<entry key='notificator.Telegram.chatid'>CHAT_ID</entry>
```
You can add a bot and get an api key using the BotFather contact in Telegram. 
After doing that you have to get the chatid as described in this link: https://stackoverflow.com/questions/32423837/telegram-bot-how-to-get-a-group-chat-id

Feel free to test it.

Todo: 
- Notification channel shows up as notificatorTelegram in the Web UI. I guess this can be fixed by adding translations to traccar-web. 
- Telegram supports simple HTML formatting in the messages. For now it uses the plain short message format. With a custom formatter (similiar to the Mail formatter) you could make the message look nicer.

Note: I'm not a Java programmer, so the code is probably not perfect. Feel free to correct and improve it. I used the NotificatorFirebase class and modified it to work with Telegram. Hope this is useful for someone.

Best
Leonhard

djbo1982:
Hi...

Excelen solution to get on live notification, can you send us more information about this?

Nitrosito:
+1 ¿How configure it?
I suppose we copy 2 modified files. And in the config file, set 3 entrys (example config you post). 

This must will native feature!! Developers, merge this in the official branch!

he-leon:
@Nitrosito Yep, either copy the modifications to your branch or just clone my fork. Of course you have to compile traccar afterwards to get the new tracker-server.jar file. You can then swap the old tracker-server.jar from your installation with the newly compiled one.

tananaev:
@he-leon can you send a pull request?

Nitrosito:
> 
> 
> @he-leon can you send a pull request?

+1

adadnc:
You may want to consider that the config keys are not
```
notificator.Telegram.apikey
notificator.Telegram.chatid
```

From the current source code it needs to be:
```
notificator.telegram.key
notificator.Telegram.chatid
```

tananaev:
From next release keys will be:

- notificator.telegram.key
- notificator.telegram.chatId

kobisamoray:
Hi,
I believe that sending a private message to a user as other notificators do is more elegant.
[...]

PR #4310, opened 2019-05-06
Send private Telegram messages to users

kobisamoray:
Use the user's phone number attribute to send a private Telegram message
for notification.
To set up, the notificator still requires the chatId attribute - it will
be used to retrieve the user_id by using the user's phone number.
Users will have to activate the bot - this is done in the same manner as
any other Telegram bot.

### DIFF
--- a/src/main/java/org/traccar/notificators/NotificatorTelegram.java
+++ b/src/main/java/org/traccar/notificators/NotificatorTelegram.java
@@ -21,19 +21,29 @@ import org.slf4j.LoggerFactory;
 import org.traccar.Context;
 import org.traccar.model.Event;
 import org.traccar.model.Position;
+import org.traccar.model.User;
 import org.traccar.notification.NotificationFormatter;
 
+import javax.json.JsonObject;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.InvocationCallback;
+import java.util.HashMap;
 
 public class NotificatorTelegram extends Notificator {
-
+    private static final String TELEGRAM_API_PREFIX = "https://api.telegram.org/bot%s/%s";
+    private static final long HOUR_TIME_FACTOR = 1000 * 60 * 60; // mSec to hour factor
     private static final Logger LOGGER = LoggerFactory.getLogger(NotificatorTelegram.class);
 
-    private String url;
+    private String sendMessageUrl;
+    private String sendContactUrl;
     private String chatId;
+    private TelegramCache cache = new TelegramCache();
 
-    public static class Message {
+    public static class TextMessage {
+        public TextMessage(String chatId, String text) {
+            this.chatId = chatId;
+            this.text = text;
+        }
         @JsonProperty("chat_id")
         private String chatId;
         @JsonProperty("text")
@@ -42,21 +52,66 @@ public class NotificatorTelegram extends Notificator {
         private String parseMode = "html";
     }
 
+    public static class ContactMessage {
+        public ContactMessage(String chatId, String phoneNumber) {
+            this.chatId = chatId;
+            this.phoneNumber = phoneNumber;
+        }
+        @JsonProperty("chat_id")
+        private String chatId;
+        @JsonProperty("phone_number")
+        private String phoneNumber;
+        @JsonProperty("first_name")
+        private String firstName = "dummy";
+    }
+
+    public static class TelegramCache {
+        private class CachePayload {
+            CachePayload(String chatId) {
+                this.chatId = chatId;
+                timestamp = System.currentTimeMillis();
+            }
+            private String chatId;
+            private long timestamp;
+        }
+
+        private HashMap<String, CachePayload> cache = new HashMap<>();
+        private long retention;
+
+        public TelegramCache() {
+            retention = Context.getConfig().getLong("notificator.telegram.cacheRetention", 0) * HOUR_TIME_FACTOR;
+        }
+
+        private Boolean isValid(long timestamp) {
+            long currTime = System.currentTimeMillis();
+            return retention < 0 || ((currTime - timestamp) <= retention);
+        }
+
+        public String getChatId(String phoneNumber) {
+            CachePayload payload = cache.get(phoneNumber);
+            if (payload != null && isValid(payload.timestamp)) {
+                return payload.chatId;
+            }
+            return null;
+        }
+
+        public void storeCache(String phoneNumber, String chatId) {
+            cache.put(phoneNumber, new CachePayload(chatId));
+        }
+    }
+
     public NotificatorTelegram() {
-        url = String.format(
-                "https://api.telegram.org/bot%s/sendMessage",
-                Context.getConfig().getString("notificator.telegram.key"));
+        String botKey = Context.getConfig().getString("notificator.telegram.key");
+        sendMessageUrl = String.format(TELEGRAM_API_PREFIX, botKey, "sendMessage");
+        sendContactUrl = String.format(TELEGRAM_API_PREFIX, botKey, "sendContact");
         chatId = Context.getConfig().getString("notificator.telegram.chatId");
     }
 
-    @Override
-    public void sendSync(long userId, Event event, Position position) {
+    private void sendTextMessage(long userId, Event event, Position position, String userChatId) {
+        TextMessage message = new TextMessage(
+                userChatId, NotificationFormatter.formatShortMessage(userId, event, position));
 
-        Message message = new Message();
-        message.chatId = chatId;
-        message.text = NotificationFormatter.formatShortMessage(userId, event, position);
-
-        Context.getClient().target(url).request()
+        Context.getClient().target(sendMessageUrl).request()
                 .async().post(Entity.json(message), new InvocationCallback<Object>() {
             @Override
             public void completed(Object o) {
@@ -64,14 +119,51 @@ public class NotificatorTelegram extends Notificator {
 
             @Override
             public void failed(Throwable throwable) {
-                LOGGER.warn("Telegram API error", throwable);
+                LOGGER.warn("Telegram API error - send message failed", throwable);
             }
         });
+    }
+
+    @Override
+    public void sendSync(long userId, Event event, Position position) {
+
+        final User user = Context.getPermissionsManager().getUser(userId);
+        if (user.getPhone() != null) {
+            String userChatId = cache.getChatId(user.getPhone());
+            if (userChatId != null) {
+                sendTextMessage(userId, event, position, userChatId);
+            } else {
+                ContactMessage message = new ContactMessage(chatId, user.getPhone());
+                Context.getClient().target(sendContactUrl).request()
+                        .async().post(Entity.json(message), new InvocationCallback<JsonObject>() {
+                    @Override
+                    public void completed(JsonObject json) {
+                        JsonObject result = json.getJsonObject("result");
+                        if (json.getBoolean("ok")) {
+                            JsonObject contact = result.getJsonObject("contact");
+                            String userChatId = contact.getJsonNumber("user_id").toString();
+
+                            cache.storeCache(user.getPhone(), userChatId);
+                            sendTextMessage(userId, event, position, userChatId);
+                        } else {
+                            LOGGER.warn("Telegram notificator: contact retrieval failed");
+                        }
+                    }
+
+                    @Override
+                    public void failed(Throwable throwable) {
+                        LOGGER.warn("Telegram API error - user " + user.getName() + "chatId retrieval failed",
+                                throwable);
+                    }
+                });
+            }
+        } else {
+            LOGGER.warn("Telegram notificator: Couldn't find phone number for user " + user.getName());
+        }
     }
 
     @Override
     public void sendAsync(long userId, Event event, Position position) {
         sendSync(userId, event, position);
     }
-
 }


kobisamoray:
Used the following as a method for phone number to chatid resolution:
https://stackoverflow.com/questions/41686463/how-to-get-user-id-of-a-phone-number-in-telegram/41696457

tananaev:
It looks like it's kind of a hack. Basically you send contact first and then send a message. I'm not sure I'm comfortable with such implementation.

kobisamoray:
@tananaev It's hacky, as Telegram doesn't offer any other mean to obtain chat ids.
So both #4310 or #4309 require a hack - one is done within the code, the other requires the admin to hack out the chat id while configuring a user.
In any case, I think that both are more acceptable than the committed change as currently, as the existing notifier doesn't do what other notifiers do - notify the _user_ about an event.

tananaev:
Correct me if I'm wrong, but #4309 just sends to a per-user chat id. Why is it a hack? What hack is required by the admin? User should know chat id, right?

kobisamoray:
Yeah #4309 is good enough IMO. Telegram UI doesn't expose chat id AFAIK.
Obtaining it requires the user to use userinfobot or the admin to retrieve it from JSON with some bot's getUpdates.
#4310 doesn't require additional info from the admin or the user - phone number is enough but yeah it's even more hacky.

tananaev:
Thanks for the information and PR. Let me think about it for a bit.

traccarbr:
kobisamoray Could you give us more details on how to use the attribute, have some print could facilitate

baerni:
#4309 
I think it's even better because we can send notifications to groups as well, not only to private phone numbers.

baerni:
This is a great solution. Will it be implemented in the official application? Do I need to compile the application myself?

kobisamoray:
As both #4309 and #4310 require some setup effort, I believe that it's best to wait until any of these is merged into master branch.

tbernesga:
Didi you test this solution???. I'm doing some tests and for me its impossible to get the userid in the response of the send contact message. Its missing.

kobisamoray:
@tbernesga of course I've tested it - it's working on our test deployment for a few days.
You should start the bot from the user which has its phone number in traccar: send bot url to user, user should start the bot - yep per each user. That's how Telegram bot API works.
Can you try curl as described in the stackoverlow link above?

tbernesga:
@kobisamoray you are rigth. I didn't activate the bot with the /start command, so i never receive the userchatid. 
By the way, i suggest two changes. 
1.- Within the first telegram message you retrieve the userchatid and you save it in the user object, so you don't have to obtain it in every single message you send.
2.- In the database you stored the userchatid and referred to the telephone number. if you change the telephone number you delete the userchatid. 

It's just a suggestion, i'm implementing it. 

Thanks for your job and quick response.

kobisamoray:
@tbernesga yeah I agree, it doesn't make much sense to translate phone number to chatid on every message.
Storing it is a good idea - maybe even a simple solution like a transient cache in a hashmap will suffice.

kobisamoray:
So I've added a simple cache for the chat ids. I believe that this should be good enough but will be happy to get feedback.
Cache retention period is set within the conf with notificator.telegram.cacheRetention (sec).
I guess that it's safe to use a fairly long period here - maybe it's worthwhile to set this in minutes or even larger units? Days?

tbernesga:
@kobisamoray I think the cache retention it's not needed. You only save the Phone number associated with the userchatid. If the phone number changes, the chat id change. If the phone number its the same, the userchatid must be the same. I think its easier and more efficient for lots of notifications (like its my situation). 
What i did: 
1.- Add a userChatId parameter in user object. If userChatId is null i will obtain it and save, if not i just send the message.
2.- In user object, if i set the phone number, i delete the userChatId, so next time i send a message to that user i will obtain the userChatId (because its supoosed to be different). 

In telegrammanager i just separate getUserChatId into a new function and call it only when needed.

kobisamoray:
@tbernesga honestly I used the simplest cache implementation as my knowledge of traccar components is minimal.
Regarding cache retention, I'd rather give an option to expire these entries. I could add a keep forever value (e.g -1).

andsdoug:
does anyone can write a small summary for dummies  showing how we
use/configure telegram in traccar. tks

Em Sex, 10 de mai de 2019 14:58, kobisamoray <notifications@github.com>
escreveu:

kobisamoray:
@andsdoug this PR is not a part of the official traccar repo. 
For the code which is committed into 4.5, follow the instructions here:
https://github.com/traccar/traccar/issues/4274

Nitrosito:
Its possible configure APIKEY and UserChatID per USER in tracar?
It would be interesting if this was configurable by user in the traccar GUI

By the way, have not put the documentation of this function in documents

kobisamoray:
@Nitrosito
There's a PR #4309 which allows the configuration of a per-user chatId.
The APIKEY is the telegram API bot registration - are you really interested in creating one of these per each of your users? What's the use case for this?
This PR (#4310) uses that already-configured phone number to retrieve the user's chatId. That's somewhat easier to maintain than #4309 (you don't need to extract the chatId by yourself) but requires a bit of initial setup and generates some Telegram "noise". #4309 also allows assignment of a group chatId to a tracker.

mrlorenz28:
how to configure it?

kobisamoray:
> how to configure it?

Hi @mrlorenz28,
This change uses 3 parameters:
**notificator.telegram.key**: this is a Telegram bot key, obtained by [using BotFather](https://tutorials.botsfloor.com/creating-a-bot-using-the-telegram-bot-api-5d3caed3266d) contact in Telegram.
**notificator.telegram.chatId**: a Telegram chat-id which includes the bot created as described above. Basically this is used to translate the users' phone numbers into Telegram chat-ids.
You can use the info [here](https://stackoverflow.com/questions/32423837/telegram-bot-how-to-get-a-group-chat-id) to learn how to obtain the chat-id.

Also note that each user who wishes to receive Telegram messages should access the bot and /start it.

mrlorenz28:
how to use the user's phone number for the chat id? 	
i try:
<entry key='notificator.telegram.chatId'>{phone}</entry>
but not working-.

kobisamoray:
Hi @mrlorenz28 
You cannot use the phone number as the chatId - in fact this change is translating the phone number into the chatId (yet it requires a chatId which is accessible to the bot in order to do the translation).
Alternately, use #4309 to use chatIds instead of phone numbers. For that purpose you could use Telegram's [userinfobot](https://telegram.me/userinfobot).

gpproton:
Is there no more consideration for this pull request.

Nitrosito:
It would be interesting if this was an array with multiple ChatID possibility

macano13:
> > how to configure it?
> 
> Hi @mrlorenz28,
> This change uses 3 parameters:
> **notificator.telegram.key**: this is a Telegram bot key, obtained by [using BotFather](https://tutorials.botsfloor.com/creating-a-bot-using-the-telegram-bot-api-5d3caed3266d) contact in Telegram.
> **notificator.telegram.chatId**: a Telegram chat-id which includes the bot created as described above. Basically this is used to translate the users' phone numbers into Telegram chat-ids.
> You can use the info [here](https://stackoverflow.com/questions/32423837/telegram-bot-how-to-get-a-group-chat-id) to learn how to obtain the chat-id.
> 
> Also note that each user who wishes to receive Telegram messages should access the bot and /start it.

Is this official and is it working receiving messages per user not group? Because if we have lets say 10 users and all of them are using same telegram bot all the notifications will be received in the same chat group for all users, please correct me if i'm wrong?

CyberPK:
Sorry, I don't know Telegram very well, but I think this function would be really nice to have.
Because the bot cannot know the user phone number, why don't change something like that:

- generate a key for every user to be shown in the user interface;
- the user uses the bot and identify with a custom command with the provided key;
- bot match the key with the user and saves the chatid.

then should be decided if allow multiple devices chatid per user and manage the revocation from traccar interface, or allow only one device, and store only the latest chatid.
The bot would also implement a command to delete the telegram chatid and stop telegram notifications.

What do you think?
Sorry if there is any error in my idea, as I said I'm not a Telegram expert.

tshelver:
We have implemented a Telegram service bot on data we extract from Telegram and another tracking application, that sends messages directly to users, and also allows users to query vehicle status (status, position, mileage, speed, voltage and so on), as well as run reports (sent as Excel extracts).

First round we implemented this using groups or channels that we created for each set of customer users, that the users could then sign up to.  This was just too much manual work and an ongoing maintenance issue.

Second go-round, the user adds our bot, and sends a registration message to the server with confirming details as to who they are in the system.

We use this message to extract and store their chatid, user name, and other details, and enable their system access automatically.
From there on the user is in a self service state via Telegram and can register / deregister additional groups, set up their required event notifications per vehicle, and pull current status and reports as required.

Several clients use our Telegram app as the primary tool to monitor their fleet via their mobile phones, and only use the web interface if in the office for route tracing and so on.

2 button taps gets them current status or information on all vehicles in the selected group (an additional tap will show the map position per vehicle), and 4 button taps will extract fleet reports (trips,stops, events, travel and time onsite and so on).

status

reports

CyberPK:
@tshelver Really nice. It is open source?

tshelver:
Not at the moment.  It's something I hacked together in Python / pgplswql (Postgresql (and a bit of Redis), based on a reporting database / platform we used to get around issues with a different proprietary tracking system than Traccar. 

It would need some cleaning up and a bit of rework (using conversational mode in the bot) before it's worth releasing.

Also, it's a very different technology to Traccar, in that most of the data processing is done using  Postgresql's pgplsql function language and POSTGis for geo functions, with  Python being used as a front end / middleware. Just one PG function extracts and forms the data for 5 of the above report, for example.  All data mutations are via  PG functions, as are many of the queries.

All of the performance gains in reporting over Traccar (and our other proprietery tracking system) come from using POSTGis functions and datatypes as well as processing complex queries internally in pgplsql before emitting a result set to Python.

Next phases for the system will include moving the existing PG;plsql functions to a different schema to enable a GraphQL API (via Postgraphile) to provide a new smartphone / web interface. 

So to make this open source, I would need to provide the DB schema setup, the parameter setups (currently imports of csv files into postgres), clean up  the Python Traccar extract / load programs, do some minor rework on the reporting Python module, and possibly re-architect the Telegram module for conversational mode.
Probably would also need the admin side of the new web interface working as well, which is a fair bit away given my current workload (and knowledge base... )

kobisamoray:
@tshelver these capabilities are available via Traccar app. What's the advantage of doing these actions via Telegram bot?

tshelver:
@kobisamoray : Traccar is a great product, but as with any application out there, it's built with a number architectural, technical and functionality choices.
We currently use the Traccar Web app unchanged to support newer devices, but we have users for whom it's not as suitable. 

**Issues specific to our company**
A specific advantage in our case is that we have a legacy tracking system in addition to Traccar.
Some companies use both  systems, this gives them 1 unified access  point. 

Another issue specific to our environment is that we don't have Java skills to go ahead and make substantial changes or enhancements to Traccar.  Just about everything in Traccar is highly Java-focused, so that puts us at a major disadvantage.

Another issue is that we operate in multiple geographies (North America, southern Africa) where data costs and reliability / speed are highly variable.

Lastly, we are developing functionality specific to small / medium companies in various verticals.

**App' suitability for purpose'** 
Most of our users are small companies, where the manager in charge of tracking is often out and about using only a phone as communication.  IMHO, for different reasons, neither our legacy system phone app, or the Traccar phone app is very user friendly for people monitoring multiple vehicles.
Also users often don't want to add an additional app to their phones,  Some of our users would need 2 apps on their phones if they are not to be tied down to a laptop when on the road, both of which would be somewhat data intensive.

Versus Traccar, the Telegram app is 'always on' with the menu displayed at the bottom of the input area for the bot.  Network / data overhead is confined to standard Telegram 

Notifications:
The Telegram service sends targeted events to individual users that we would otherwise have to send via email (inour experience, these often get sent to trash / spam folders, thus getting lost) or via SMS, which as a small company operating in a few different geographies (southern Africa, North America) is a substantial expense.

**Ease of user**
No need to log on separately from Telegram, and with the menu system, all functions are never more than 4 button presses away: no data entry required to produce full reports for their whole fleet / by group.
2 button presses will show current status of all vehicles in a group.
Similarly, users can access the seupt tailoring messages / events per vehicle within 4 button presses (from Settings).
Help system: one button press shows a full help facility detailing everything the user needs to know about our Telegram bot.

**Functionality**
Some of the more targeted, useful reports users request are not available in Traccar. The system design has allowed us to spin these additional reports very quickly as most data is first collated and processed by a Postgresql function.

**Performance**
I can run a stop or route /  trip report for around 10 vehicles over a month (some of our reports are substantially more complex) almost an order of magnitude faster than Traccar allows.
For example, reporting on trips / stops in an area uses PostGIS functions, indexes and datatypes to speed access to the relevant data, and only qualifying stops / trips are returned to Python for output formatting and processing.
For address lookups, our cache is currently unlimited, as we first process the geo position against the address area in PostGIS before returning qualified addresses to Python. This has hugely cut down on reverse geocoding processing against external REST APIs, which is one of the slowest aspects of address reporting.
For geo-specific functionality, PostGIS is a LOT faster than other solutions we have tested, including the equivalent functionality in MS SQL Server in a prior iteration of this system, and is a lot more robust and complete than the equivalent facility in MySQL /MariaDB.

**API**
The Traccar API works, but IMHO isn't as fast or as flexible or as targeted as direct access to the database.  I fully understand why Traccar has gone this route, as they have to support multiple databases.
In our case, Postgres has been around a long time, it has all the extensions and performance options that we need, has a strong developer base, and is community driven (unlike about any other DB including MySQL). In particular, it is a standard in the GIS community, being the underlying technology for OpenStreetMap and many others.
So we are very comfortable committing to it.

Going with GraphQL using Postgraphile over Postgres functions allows us more fine-grained permission, plus the ability to make one call to pull back just the data we need for each app function, while not not   We retain most of the performance and flexibility of direct DB calls.

Issue #4277, opened 2019-04-13
[Feature Reqest] Finish Line / Geofence


ac3bf1:
Hello all,

firstly great, fantastic, superb job (everything always seems to work just as it should!)

I'd like to request a "Finish Line" type of feature, where if the path of the position crosses this line at any angle, a notification is triggered.
I have tried making "slim" poligon geofences, but if coordinates do not fall exactly inside poligon, the user is not considered to be "inside" by the notifications.

As an example, imagine the Red Bull Air Race, where airplanes have to travel between pylons and we have a tracker inside the airplane (with 1 second position reporting interval). If airplane crosses a Pylon (Check Point), a notification is triggered.

Thanks in advance, and I hope I made myself clear :-)
Regards,
Jonathan

tananaev:
Why not use bigger geofence?

oliv3:
Side note, for an airplane race, 1s updates will induce a big error :)

ac3bf1:
Larger geofence, would not work, as plane needs to cut a finish line without entering from the "side"
https://www.dropbox.com/s/9u8p9qdescmxbs4/finishLine.png?dl=0

In the image above, the blue arrows, are acceptable crossings, the red arrows are not.
If I made a "box" for the finish line, the red arrows would get points when entering the box, while I would like to have something that is triggered only on "crossing" this line.
Any thoughts perhaps?

ac3bf1:
> Side note, for an airplane race, 1s updates will induce a big error :)

Can I ask why a short update interval would induce a big error?
P.S. let me add that this race is executed at speeds that go from 60 Knots (111 Km/hr) to 110 knots (204 Km/hr) at most

oliv3:
* https://www.calculateme.com/speed/knots/to-meters-per-second/60
* https://www.calculateme.com/speed/knots/to-meters-per-second/110

ac3bf1:
>     * https://www.calculateme.com/speed/knots/to-meters-per-second/60
> 
>     * https://www.calculateme.com/speed/knots/to-meters-per-second/110

Thanks, for the link to the calculation.
Perhaps I will have a delay in the position reported, as 60 m/s is quiete substantial. Agreed.
But when you say "big error" to what type of error do you refer?

Furthermore, any ideas on the finish line?

Thanks and great work! :-)
Jonathan

PR #4316, opened 2019-05-10
Add geofenceCrossed event

jguthrie100:
In repsonse to #4277 , I've added a new Event called _geofenceCrossed_.

It works separately to _geofenceEnter_ and _geofenceExit_ and simply detects if the path from the previous position to the current position crosses over a geofence, without a position being recorded in it.

This could then be used for the finish line situation, or maybe if a road cuts over the corner of a geofence, and the device only records positions either side of it for example.

geofenceCrossed

(You can see it triggers on the circle and the line (twice), but doesn't on the polygon (which registers a position inside)

It also would need a string adding to the traccar-web locale

### DIFF
--- a/gradle/checkstyle.xml
+++ b/gradle/checkstyle.xml
@@ -70,7 +70,10 @@
         <module name="MethodLength">
             <property name="max" value="200"/>
         </module>
-        <module name="ParameterNumber"/>
+        <module name="ParameterNumber">
+            <property name="max" value="8"/>
+            <property name="tokens" value="METHOD_DEF"/>
+        </module>
 
         <!-- Checks for whitespace                               -->
         <!-- See http://checkstyle.sf.net/config_whitespace.html -->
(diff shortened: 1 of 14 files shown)


oliv3:
This is **definitely** a clever solution.

alisaniko:
It would be interesting to come up with an idea, and implement it so that, given the entry into a certain geofence line in the settings and having traveled a certain amount of time or km, the program would identify and assign this tracker to this geofence

jguthrie100:
@tananaev Is there any desire to merge this? I can fix the merge conflicts if so

tananaev:
It doesn't seem like there's enough community interest in this.

bergh-io:
This would be great for me. 
I need to log then cars pass different road tolls. These can sometimes be close to each other witch is a problem it self for the GPS. It really is only the crossing I'm interested in.
image

bergh-io:
Hi again

I have been testing my setup with geofences for some time now and unfortunately it's not accurate enough in some areas. 
The geofence needs to be to big for it to make sense and this event type would pretty much solve my issue 100%.

Please do consider merging this 👍 🥇

jguthrie100:
@bergh-io you'll have to pester tananaev some more heh

bergh-io:
> @bergh-io you'll have to pester tananaev some more heh

I don´t think that I have enough leverage by myself ;)

PR #4327, opened 2019-05-30
Implement New Protocol Inmotion

lcsvcn:
Added support for a new protocol.


tananaev:
What is the difference with GT06 protocol?

lcsvcn:
This protocol is based on GT06, but is not the same. Will cancel this merge and just until I finish editing the changes.

PR #4333, opened 2019-06-14
Fix coordinates.maxError key and coordinates.filter description

Abyss777:
Hello
- Fixed `coordinates.maxError` key string
- Improved `coordinates.filter` description


tananaev:
Merged, thanks. Are you back? :)

Abyss777:
I'm afraid not so much as before. But hope still will be able to help the project and contribute something.

tananaev:
Awesome, thanks.

PR #4342, opened 2019-07-03
Enhance Laipac protocol decoder and encoder

edvalley:
Enhance Laipac protocol decoder and encoder

### DIFF
--- a/src/main/java/org/traccar/database/DeviceManager.java
+++ b/src/main/java/org/traccar/database/DeviceManager.java
@@ -115,14 +115,9 @@ public class DeviceManager extends BaseObjectManager<Device> implements Identity
         return devicesByUniqueId.get(uniqueId);
     }
 
-    public String getDevicePassword(long id) {
-
-        return getById(id).getString(Command.KEY_DEVICE_PASSWORD);
-    }
-
     public String getDevicePassword(long id, String defaultPassword) {
 
-        String password = getDevicePassword(id);
+        String password = getById(id).getString(Command.KEY_DEVICE_PASSWORD);
         if (password != null) {
             return password;
         }
--- a/src/main/java/org/traccar/database/DeviceManager.java
+++ b/src/main/java/org/traccar/database/DeviceManager.java
@@ -115,6 +115,7 @@ public class DeviceManager extends BaseObjectManager<Device> implements Identity
         return devicesByUniqueId.get(uniqueId);
     }
 
+    @Override
     public String getDevicePassword(long id, String defaultPassword) {
 
         String password = getById(id).getString(Command.KEY_DEVICE_PASSWORD);
--- a/src/main/java/org/traccar/database/DeviceManager.java
+++ b/src/main/java/org/traccar/database/DeviceManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 - 2018 Anton Tananaev (anton@traccar.org)
+ * Copyright 2016 - 2019 Anton Tananaev (anton@traccar.org)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,6 +29,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.traccar.config.Config;
 import org.traccar.Context;
+import org.traccar.model.Command;
 import org.traccar.model.Device;
 import org.traccar.model.DeviceState;
 import org.traccar.model.DeviceAccumulators;
@@ -112,6 +113,21 @@ public class DeviceManager extends BaseObjectManager<Device> implements Identity
         updateDeviceCache(forceUpdate);
 
         return devicesByUniqueId.get(uniqueId);
+    }
+
+    public String getDevicePassword(long id) {
+
+        return getById(id).getString(Command.KEY_DEVICE_PASSWORD);
+    }
+
+    public String getDevicePassword(long id, String defaultPassword) {
+
+        String password = getDevicePassword(id);
+        if (password != null) {
+            return password;
+        }
+
+        return defaultPassword;
     }
 
     public Device getDeviceByPhone(String phone) {
--- a/src/main/java/org/traccar/database/IdentityManager.java
+++ b/src/main/java/org/traccar/database/IdentityManager.java
@@ -26,8 +26,6 @@ public interface IdentityManager {
 
     Device getByUniqueId(String uniqueId) throws Exception;
 
-    String getDevicePassword(long id);
-
     String getDevicePassword(long id, String defaultPassword);
 
     Position getLastPosition(long deviceId);
--- a/src/main/java/org/traccar/database/IdentityManager.java
+++ b/src/main/java/org/traccar/database/IdentityManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Anton Tananaev (anton@traccar.org)
+ * Copyright 2015 - 2019 Anton Tananaev (anton@traccar.org)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,10 @@ public interface IdentityManager {
     Device getById(long id);
 
     Device getByUniqueId(String uniqueId) throws Exception;
+
+    String getDevicePassword(long id);
+
+    String getDevicePassword(long id, String defaultPassword);
 
     Position getLastPosition(long deviceId);
 
--- a/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
@@ -223,7 +223,7 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
         position.set(Position.KEY_ALARM, decodeAlarm(event));
         position.set(Position.KEY_EVENT, decodeEvent(event, position));
         position.set(Position.KEY_BATTERY, Double.parseDouble(parser.next().replaceAll("\\.", "")) * 0.001);
-        position.set(Position.KEY_ODOMETER, parser.nextDouble());
+        position.set(Position.KEY_ODOMETER, parser.nextDouble() * 1000);
         position.set(Position.KEY_GPS, parser.nextInt());
         position.set(Position.PREFIX_ADC + 1, parser.nextDouble() * 0.001);
         position.set(Position.PREFIX_ADC + 2, parser.nextDouble() * 0.001);
--- a/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
@@ -167,8 +167,13 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
 
     }
 
-    private Object handleEchk(
-            String sentence, Parser parser, Channel channel, SocketAddress remoteAddress) {
+    private Object decodeEchk(
+            String sentence, Channel channel, SocketAddress remoteAddress) {
+
+        Parser parser = new Parser(PATTERN_ECHK, sentence);
+        if (!parser.matches()) {
+            return null;
+        }
 
         if (channel != null) {
             channel.writeAndFlush(new NetworkMessage(sentence + "\r\n", remoteAddress));
@@ -177,8 +182,13 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
         return null;
     }
 
-    protected Object handleAvrmc(
-            String sentence, Parser parser, Channel channel, SocketAddress remoteAddress) {
+    protected Object decodeAvrmc(
+            String sentence, Channel channel, SocketAddress remoteAddress) {
+
+        Parser parser = new Parser(PATTERN_AVRMC, sentence);
+        if (!parser.matches()) {
+            return null;
+        }
 
         DeviceSession deviceSession =
             getDeviceSession(channel, remoteAddress, parser.next());
@@ -245,15 +255,10 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
         String sentence = (String) msg;
 
         if (sentence.startsWith("$ECHK")) {
-            Parser parser = new Parser(PATTERN_ECHK, sentence);
-            if (parser.matches()) {
-                return handleEchk(sentence, parser, channel, remoteAddress);
-            }
-        } else if (sentence.startsWith("$AVRMC")) {
-            Parser parser = new Parser(PATTERN_AVRMC, sentence);
-            if (parser.matches()) {
-                return handleAvrmc(sentence, parser, channel, remoteAddress);
-            }
+            return decodeEchk(sentence, channel, remoteAddress);
+        }
+        if (sentence.startsWith("$AVRMC")) {
+            return decodeAvrmc(sentence, channel, remoteAddress);
         }
 
         return null;
--- a/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
@@ -40,8 +40,7 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
 
     public LaipacProtocolDecoder(Protocol protocol) {
         super(protocol);
-        defaultDevicePassword = Context.getConfig().getString(
-            getProtocolName() + ".defaultPassword", "00000000");
+        defaultDevicePassword = "00000000";
     }
 
     private static final Pattern PATTERN_ECHK = new PatternBuilder()
--- a/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
@@ -163,16 +163,6 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
 
     }
 
-    private Object decodeEchk(
-            String sentence, Channel channel, SocketAddress remoteAddress) {
-
-        if (channel != null) {
-            channel.writeAndFlush(new NetworkMessage(sentence + "\r\n", remoteAddress));
-        }
-
-        return null;
-    }
-
     protected Object decodeAvrmc(
             String sentence, Channel channel, SocketAddress remoteAddress) {
 
@@ -246,9 +236,10 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
         String sentence = (String) msg;
 
         if (sentence.startsWith("$ECHK")) {
-            return decodeEchk(sentence, channel, remoteAddress);
-        }
-        if (sentence.startsWith("$AVRMC")) {
+            if (channel != null) {
+                channel.writeAndFlush(new NetworkMessage(sentence + "\r\n", remoteAddress));
+            }
+        } else if (sentence.startsWith("$AVRMC")) {
             return decodeAvrmc(sentence, channel, remoteAddress);
         }
 
--- a/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
@@ -99,17 +99,12 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
 
     private String decodeEvent(String event, Position position) {
 
-        position.set(Position.KEY_IGNITION, false);
-
         if (event.length() == 1) {
             char inputStatus = event.charAt(0);
             if (inputStatus >= 'A' && inputStatus <= 'D') {
                 int inputStatusInt = inputStatus - 'A';
                 position.set(Position.PREFIX_IN + 1, inputStatusInt & 1);
                 position.set(Position.PREFIX_IN + 2, inputStatusInt & 2);
-                if ((inputStatusInt & 1) != 0) {
-                    position.set(Position.KEY_IGNITION, true);
-                }
                 return null;
             }
         }
--- a/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
@@ -26,8 +26,6 @@ import org.traccar.helper.DateBuilder;
 import org.traccar.helper.Parser;
 import org.traccar.helper.PatternBuilder;
 import org.traccar.model.CellTower;
-import org.traccar.model.Device;
-import org.traccar.model.Command;
 import org.traccar.model.Network;
 import org.traccar.model.Position;
 
@@ -121,21 +119,6 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
         }
 
         return event;
-    }
-
-    private String getDevicePassword(DeviceSession deviceSession) {
-
-        String devicePassword = DEFAULT_DEVICE_PASSWORD;
-
-        Device device = Context.getIdentityManager().getById(deviceSession.getDeviceId());
-        if (device != null) {
-            String password = device.getString(Command.KEY_DEVICE_PASSWORD);
-            if (password != null) {
-                devicePassword = password;
-            }
-        }
-
-        return devicePassword;
     }
 
     private void sendEventResponse(
@@ -247,7 +230,8 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
 
             sendAcknowledge(status, event, checksum, channel, remoteAddress);
 
-            String devicePassword = getDevicePassword(deviceSession);
+            String devicePassword = Context.getIdentityManager()
+                .getDevicePassword(deviceSession.getDeviceId(), DEFAULT_DEVICE_PASSWORD);
             sendEventResponse(event, devicePassword, channel, remoteAddress);
         }
 
@@ -260,18 +244,19 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
 
         String sentence = (String) msg;
 
-        Parser parser = new Parser(PATTERN_ECHK, sentence);
-        if (parser.matches()) {
-            return handleEchk(sentence, parser, channel, remoteAddress);
-        }
-
-        parser = new Parser(PATTERN_AVRMC, sentence);
-        if (parser.matches()) {
-            return handleAvrmc(sentence, parser, channel, remoteAddress);
+        if (sentence.startsWith("$ECHK")) {
+            Parser parser = new Parser(PATTERN_ECHK, sentence);
+            if (parser.matches()) {
+                return handleEchk(sentence, parser, channel, remoteAddress);
+            }
+        } else if (sentence.startsWith("$AVRMC")) {
+            Parser parser = new Parser(PATTERN_AVRMC, sentence);
+            if (parser.matches()) {
+                return handleAvrmc(sentence, parser, channel, remoteAddress);
+            }
         }
 
         return null;
     }
-
 
 }
--- a/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
@@ -30,11 +30,9 @@ import org.traccar.model.Device;
 import org.traccar.model.Command;
 import org.traccar.model.Network;
 import org.traccar.model.Position;
-import org.traccar.database.ConnectionManager;
 
 import java.net.SocketAddress;
 import java.util.regex.Pattern;
-import java.util.Date;
 
 public class LaipacProtocolDecoder extends BaseProtocolDecoder {
 
@@ -72,9 +70,11 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
             .number(",(d)")                      // gps status
             .number(",(d+)")                     // adc1
             .number(",(d+)")                     // adc2
-            .number(",(x{1}|x{8})")              // lac+cid
-            .number(",(d{1}|d{6})")              // mcc+mnc
-            .optional(2)
+            .number(",(x{4}|x{1})")              // lac | lac+cid = 0
+            .number("(x{4})")                    // cid | nothing
+            .number(",(d{3}|d{1})")              // mcc | mcc+mnc = 0
+            .number("(d{3})")                    // mnc | nothing
+            .optional(4)
             .expression(",([^*]*)")              // anything remaining (be forward compatible)
             .optional(1)
             .text("*")
@@ -140,28 +140,64 @@ public class LaipacProtocolDecoder extends BaseProtocolDecoder {
         return devicePassword;
     }
 
+    private String getEventResponse(String event, String devicePassword) {
+
+        String responseCode = null;
+
+        switch (event) {
+            case "3":
+                responseCode = "d";
+                break;
+            case "S":
+            case "T":
+                responseCode = "t";
+                break;
+            case "X":
+            case "4":
+                responseCode = "x";
+                break;
+            case "Y":
+                responseCode = "y";
+                break;
+            case "Z":
+                responseCode = "z";
+                break;
+            default:
+                break;
+        }
+
+        if (responseCode != null) {
+            String response = "$AVCFG," + devicePassword + "," + responseCode;
+            response += Checksum.nmea(response) + "\r\n";
+            return response;
+        }
+
+        return null;
+    }
+
+    private String getStatusResponse(String status, String event, String checksum) {
+
+        if (Character.isLowerCase(status.charAt(0))) {
+            String response = "$EAVACK," + event + "," + checksum;
+            response += Checksum.nmea(response) + "\r\n";
+            return response;
+        }
+
+        return null;
+    }
+
     private Object handleEchk(
-            String sentence, int checksum, Parser parser, Channel channel, SocketAddress remoteAddress) {
+            String sentence, Parser parser, Channel channel, SocketAddress remoteAddress) {
 
         if (channel != null) {
             channel.writeAndFlush(new NetworkMessage(sentence + "\r\n", remoteAddress));
-        }
-
-        DeviceSession deviceSession =
-            getDeviceSession(channel, remoteAddress, parser.next());
-        if (deviceSession != null) {
-            ConnectionManager cm = Context.getConnectionManager();
-            if (cm != null) {
-                cm.updateDevice(deviceSession.getDeviceId(),
-                    Device.STATUS_ONLINE, new Date());
-            }
         }
 
         return null;
     }
 
     protected Object handleAvrmc(
-            String sentence, int checksum, Parser parser, Channel channel, SocketAddress remoteAddress) {
+            String sentence, Parser parser, Channel channel, SocketAddress remoteAddress) {
 
         DeviceSession deviceSession =
             getDeviceSession(channel, remoteAddress, parser.next());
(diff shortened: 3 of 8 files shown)


src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
```diff
@@ -0,0 +1,71 @@
...
+    protected String getProtocolName() {
+        return protocol.getName();
+    }
+
+    @Override
+    protected Object encodeCommand(Command command) {
+
+        initDevicePassword(command, defaultDevicePassword);
+
+        String cmd = "";
```
tananaev: Use full words.

src/test/java/org/traccar/protocol/LaipacProtocolDecoderTest.java
```diff
@@ -120,6 +120,13 @@ public void testDecode() throws Exception {
         verifyPosition(decoder, text(
                 "$AVRMC,358174067149865,143747,R,5050.1124,N,00420.0542,E,1.34,161.96,190318,a,3416,119,1,0,0*7D"));
 
+        //Zero LAC, CID, MCC, MNC
```
tananaev: Space is missing after `//`.

src/test/java/org/traccar/protocol/LaipacProtocolDecoderTest.java
```diff
@@ -120,6 +120,13 @@ public void testDecode() throws Exception {
...
+        verifyPosition(decoder, text(
+                "$AVRMC,358174067149865,143747,P,5050.1124,N,00420.0542,E,1.34,161.96,190318,A,3416,119,1,0,0,0,0*5F"));
+
+        //New unknown parameters
```
tananaev: Same.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
```diff
@@ -0,0 +1,71 @@
...
+import org.traccar.helper.Checksum;
+
+public class LaipacProtocolEncoder extends StringProtocolEncoder {
+
+    private final Protocol protocol;
+    private final String defaultDevicePassword;
+
+    public LaipacProtocolEncoder(Protocol protocol) {
+        this.protocol = protocol;
+        defaultDevicePassword = Context.getConfig().getString(
```
tananaev: What about per-device password?

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright 2015 - 2019 Anton Tananaev (anton@traccar.org)
```
tananaev: This should be updated.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -164,4 +268,46 @@ protected Object decode(
...
+            return null;
+        }
+
+        // Handle ECHK sentences
+        Parser parser = new Parser(PATTERN_ECHK, sentence);
+        if (parser.matches()) {
+            return handleEchk(sentence, checksum, parser, channel, remoteAddress);
+        }
+
+        // Handle AVRMC sentences
```
tananaev: Please remove all unnecessary comments, which is almost all comments in this class.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -164,4 +268,46 @@ protected Object decode(
...
+        if (sentence.charAt(0) != '$') {
+            return null;
+        }
+        if (sentence.charAt(slen - 3) != '*') {
+            return null;
+        }
+
+        // Verify sentence checksum
+        int checksum = Integer.parseInt(sentence.substring(slen - 2), 16);
+        if (checksum != Checksum.xor(sentence.substring(1, slen - 3))) {
```
tananaev: We don't usually verify checksums. I would recommend removing to make it consistent with other decoders.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -164,4 +268,46 @@ protected Object decode(
...
+        int slen = sentence.length();
+        if (slen <= 5) {
+            return null;
+        }
+
+        // Validate sentence format
+        if (sentence.charAt(0) != '$') {
+            return null;
+        }
+        if (sentence.charAt(slen - 3) != '*') {
```
tananaev: Why do we need all this validation? That's the whole point of regex.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -81,81 +100,166 @@ private String decodeAlarm(String event) {
...
+            } else if (event.equals("Y")) {
+                response = "$AVCFG," + devicePassword + ",y";
             } else if (event.equals("Z")) {
-                channel.writeAndFlush(new NetworkMessage("$AVCFG,00000000,z*2F\r\n", remoteAddress));
-            } else if (Character.isLowerCase(status.charAt(0))) {
-                String response = "$EAVACK," + event + "," + checksum;
+                response = "$AVCFG," + devicePassword + ",z";
+            }
+
+            if (response.length() > 0) {
```
tananaev: `isEmpty`?

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -81,81 +100,166 @@ private String decodeAlarm(String event) {
...
+                ack += Checksum.nmea(ack) + "\r\n";
+                channel.writeAndFlush(new NetworkMessage(ack, remoteAddress));
+            }
+
+            String response = "";
+            String devicePassword = getDevicePassword(deviceSession);
+
             if (event.equals("3")) {
-                channel.writeAndFlush(new NetworkMessage("$AVCFG,00000000,d*31\r\n", remoteAddress));
+                response = "$AVCFG," + devicePassword + ",d";
```
tananaev: I would probably make a separate method instead.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -81,81 +100,166 @@ private String decodeAlarm(String event) {
...
             position.setNetwork(new Network(CellTower.from(mcc, mnc, lac, cid)));
         }
 
-        String checksum = parser.next();
+        // Skip remaining parameters
+        String unused = parser.next();
+
+        // Checksum
+        String checksumStr = parser.next();
+        if (checksum != Integer.parseInt(checksumStr, 16)) {
```
tananaev: Another checksum validation?

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -81,81 +100,166 @@ private String decodeAlarm(String event) {
...
-        Integer lac = parser.nextHexInt();
-        Integer cid = parser.nextHexInt();
-        Integer mcc = parser.nextInt();
-        Integer mnc = parser.nextInt();
-        if (lac != null && cid != null && mcc != null && mnc != null) {
+        // LAC, CID, MCC, MNC
+        Integer laccid = parser.nextHexInt();
+        Integer mccmnc = parser.nextInt();
+        if (laccid != null && laccid != 0 && mccmnc != null && mccmnc != 0) {
+            Integer lac = (laccid >> 16) & 0xFFFF;
```
tananaev: I believe all this should be handled by pattern.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -81,81 +100,166 @@ private String decodeAlarm(String event) {
...
 
+        // Time
         DateBuilder dateBuilder = new DateBuilder()
                 .setTime(parser.nextInt(0), parser.nextInt(0), parser.nextInt(0));
 
+        // Status [ V: Invalid | A: Valid | R: Not realtime | P: Parked ]
         String status = parser.next();
         String upperCaseStatus = status.toUpperCase();
-        position.setValid(upperCaseStatus.equals("A") || upperCaseStatus.equals("R") || upperCaseStatus.equals("P"));
+        position.setValid(upperCaseStatus.equals("A"));
```
tananaev: Why?

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -81,81 +100,166 @@ private String decodeAlarm(String event) {
...
+
+        if (channel != null) {
+            channel.writeAndFlush(new NetworkMessage(sentence + "\r\n", remoteAddress));
         }
 
-        DeviceSession deviceSession = getDeviceSession(channel, remoteAddress, parser.next());
+        DeviceSession deviceSession =
+            getDeviceSession(channel, remoteAddress, parser.next());
+        if (deviceSession != null) {
+            ConnectionManager cm = Context.getConnectionManager();
```
tananaev: Don't do this.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
edvalley: Ok

src/test/java/org/traccar/protocol/LaipacProtocolDecoderTest.java
edvalley: I used the same format used all around the file. No problem, I'll fix it all.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
edvalley: That's the job of initDevicePassword call on line 42

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
edvalley: Are you suggesting that I use my name there?

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Good. I'll fix it.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: The thing is that you can configure these devices to send just a heartbeat between reports. For example, you can set the device to send heartbeats every 5 minutes, and reports every 15 minutes. I thought this should be enough to state the device as on-line on every heartbeat, or on anything it sends to the server. What do you think about this? If you don't agree with me, no problem, I'll remove that code. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: V: Invalid | A: Valid | R: Not realtime | P: Parked
In tests I've done, I confirmed that the only reports that always are real (the device is in the position the reports says), are those marked with a 'a' or 'A'. You can take the device away to a real far place, removing the GPS antenna, and it will still send reports marked with an 'r' or 'R', using the same original position and creating false positives. What do you think about this? Thanks.

src/test/java/org/traccar/protocol/LaipacProtocolDecoderTest.java
edvalley: Ok. I'll fix it.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Ok. I'll fix it. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: This verification is valid for any kind of frames or sentences received, and I think it is sane to have it there. It will avoid any fake position or info in a rare case of a transmission error or manipulation of the data. Anyway, that's why checksum is there no? If you don't agree, I'll remove it, just tell me. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: This validation ensures that substring indexes in the code bellow it, does not go out of bounds. Regular expressions are being used to parse specific kind of sentences or frames, without taking cheksums into account. If you don't agree, just tell me, I'll remove the code. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Can be empty if no code above modifies it, because no response needs to be sent. I'll be move this code to a method, as you suggested bellow. Thank you.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Just double checking that the parser is in the right position, not verifying checksums. You tell me. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Ok. I'll work on it. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
tananaev: So, what is the point of this then?

src/test/java/org/traccar/protocol/LaipacProtocolDecoderTest.java
tananaev: I see. That's probably needs to be fixed as well. Someone else must have added it.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
tananaev: You can use your name if you want, but I'm referring to the years.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
tananaev: We don't usually do any validation other than regex. Please remove all other validation.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
tananaev: Same, please remove validation.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
tananaev: My point here was to use `!response.isEmpty()` instead of `response.length() > 0`, but moving everything to a separate method is the preferred option.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
tananaev: Sorry, but this explanation doesn't really make sense. Your expression is static. You don't need to verify position.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
tananaev: Even if it's an old but valid GPS location, it should be marked as valid, assuming that timestamp is correct for the fix.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
tananaev: This is already handled for you. You don't need to do anything like this in the decoder.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
edvalley: Just to set a default password (in the configuration file, can be another place) in case no per-device password is configured using web interface

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
edvalley: Oh, I see, this is a new file. Sorry, I'll fix it. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Ok. I'll do it that way. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Ok. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Sorry, I see your point now. Moving code to a method...

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Ok. Removing this code. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: I understand. Undoing this change. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Ok. Removing code. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
tananaev: I don't think it's the right way to do with. If you want password to be configurable in the config file, we should probably use lookup attribute in the `initDevicePassword` method.

edvalley: Can it be a "Server" wide attribute?

tananaev: It can be, but if you want to change it, probably create a separate PR. In this one I would keep default password static.

edvalley: Ok. Doing it the way you suggest. Thanks.

edvalley:
I've made suggested changes. Tell me what's next...

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
```diff
@@ -0,0 +1,69 @@
...
+
+import org.traccar.Protocol;
+import org.traccar.StringProtocolEncoder;
+import org.traccar.model.Command;
+import org.traccar.helper.Checksum;
+
+public class LaipacProtocolEncoder extends StringProtocolEncoder {
+
+    private final Protocol protocol;
+    private final String defaultDevicePassword;
```
tananaev: Same question here. Also, it should probably be defined as a constant in a single place.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
```diff
@@ -0,0 +1,69 @@
...
+package org.traccar.protocol;
+
+import org.traccar.Protocol;
+import org.traccar.StringProtocolEncoder;
+import org.traccar.model.Command;
+import org.traccar.helper.Checksum;
+
+public class LaipacProtocolEncoder extends StringProtocolEncoder {
+
+    private final Protocol protocol;
```
tananaev: We don't usually pass protocol into the encoders. Maybe we should, but then it should probably be in the base class.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -81,37 +99,114 @@ private String decodeAlarm(String event) {
...
+        if (responseCode != null) {
+            String response = "$AVCFG," + devicePassword + "," + responseCode;
+            response += Checksum.nmea(response) + "\r\n";
+            return response;
+        }
+
+        return null;
+    }
+
+    private String getStatusResponse(String status, String event, String checksum) {
```
tananaev: The common practice we have is to have a method that sends response instead of returning something. Please check other protocols for examples.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -81,37 +99,114 @@ private String decodeAlarm(String event) {
...
+            }
+        }
 
-        if (sentence.startsWith("$ECHK") && channel != null) {
-            channel.writeAndFlush(new NetworkMessage(sentence + "\r\n", remoteAddress)); // heartbeat
-            return null;
+        return event;
+    }
+
+    private String getDevicePassword(DeviceSession deviceSession) {
```
tananaev: This like code duplication with what we have in the encoder.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -33,33 +36,46 @@
 
 public class LaipacProtocolDecoder extends BaseProtocolDecoder {
 
+    private final String defaultDevicePassword;
```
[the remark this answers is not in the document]
edvalley: Just a residue of my idea of a server wide configurable default device/protocol password. I'll remove it. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -33,33 +36,46 @@
...
+            .text("$ECHK")
+            .expression(",([^,]+)")             // identifier
+            .number(",(d+)")                    // sequence number
+            .text("*")
+            .number("(xx)")                     // checksum
+            .compile();
+
+    private static final Pattern PATTERN_AVRMC = new PatternBuilder()
+            .text("$AVRMC")
+            .expression(",([^,]+)")              // identifier
```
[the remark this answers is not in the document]
edvalley: My motivation for this change is that this way, if we add a new parameter to the regex, the coma will be part of the new parameter (one line addition), we won't need to modify the previous parameter part of the expression. If you don't like the idea, no problem, I'll roll it back. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
edvalley: Do you think it can be a public member of LaipacProtocol class? It needs to be usable from both the encoder and the decoder. I don't want it to be a constant because I'm planning to make it configurable (if you like the idea, using a server wide attribute)

edvalley: For now, would you like it to be a constant in LaipacProtocol accesible by a method?

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
edvalley: Not needed anymore, for now. Removing that code...

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
tananaev: No, usually constants go to decoder.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Going that way... Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
edvalley: So... What about the first idea?

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Good. I'll fix it. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
tananaev: Default should not be configurable. We discussed it.

edvalley: So, just to know if I understand. You are saying that we should not have a configurable default password for every device of a type (using the same protocol), and we should configure the password for each device, even if it is the same non hard-coded default?

tananaev: Default is default. It should be a constant. Password should be configurable.

edvalley: Sorry, now I remember you said to keep it static for this PR. But, what do you suggest not being a private static constant in LaipacProtocol, accessible through a method from both the encoder and the decoder?

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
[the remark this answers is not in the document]
edvalley: Do you like the idea or should I roll it back? Thanks.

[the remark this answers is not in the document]
tananaev: Please roll it back.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
tananaev: Make it a static constant in decoder, like all other constants there.

edvalley: And what about the code in the encoder? It needs the password too to send commands to the device.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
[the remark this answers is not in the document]
edvalley: Ok, no problem. Thanks...

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
tananaev: What's the problem with accessing constant from encoder.

edvalley: No problem at all, but we need to declare the same constant in both places. Is that ok for you?

tananaev: Please define it in a single place, as I mentioned previously.

edvalley: Sorry bothering you. That's my point... What single place (common to both encoder and decoder) do you suggest not being LaipacProtocol class? Thanks...

tananaev: I said it twice - decoder!

edvalley: Sorry again. Do you mean to declare the constant as a public static constant in the decoder class?

tananaev: Yes, please check how it's done in other decoders.

tananaev: In general follow existing practice. Don't try to invent anything.

edvalley: Oh, I see your point now. I'll follow your guideline. I'll be doing some tests and come with more commits. Thank you for your patience.

edvalley:
Do I need to comment after I make a commit so it can be seen? Sorry, I'm new to GitHub. Thanks.

tananaev:
I usually get commit notifications, but I wouldn't know if you fixed everything unless you comment.

edvalley:
Good. So, comment my changes whenever you want. Thanks again.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+package org.traccar.protocol;
+
+import org.traccar.StringProtocolEncoder;
+import org.traccar.model.Command;
+import org.traccar.helper.Checksum;
+
+public class LaipacProtocolEncoder extends StringProtocolEncoder {
+
+    public LaipacProtocolEncoder() {
```
tananaev: What's the point of empty constructor?

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
```diff
@@ -0,0 +1,59 @@
...
+            case Command.TYPE_REBOOT_DEVICE:
+                commandSentence = formatCommand(command, "$AVRESET,{%s},{%s}",
+                    Command.KEY_UNIQUE_ID, Command.KEY_DEVICE_PASSWORD);
+                break;
+            default:
+                break;
+        }
+
+        if (commandSentence != null) {
+            commandSentence += Checksum.nmea(commandSentence) + "\r\n";
```
tananaev: You can probably override `formatCommand` and add checksum there. That way you won't need null check and this whole method will be more concise because you can just return directly in switch.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -145,23 +239,39 @@ protected Object decode(
...
         return position;
     }
 
+    @Override
+    protected Object decode(
+        Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
+
+        String sentence = (String) msg;
+
+        Parser parser = new Parser(PATTERN_ECHK, sentence);
```
tananaev: Looks like you can just check header using `startsWith` instead of trying to match all patterns.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
tananaev: Sorry, what is the fix here?

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
edvalley: Just another residue. Sorry, I missed it too.

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
edvalley: Good. I'll do it that way.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Each regex processing will stop when the beginning of the pattern doesn't match (almost the same, I think, that if we use `startsWith`). The only regex that will be fully parsed, and will be the last one, is the one that fully matches. In this case both code will be executed, `startsWith` code and regex matching code. I can be wrong, of course, I'm not an experienced Java programmer. If you don't agree, no problem, I'll do it your way. Thanks.

tananaev: Regex matching is still much slower than `startsWith` even if it fails early. Also, we do the same logic in other places, so it would be nice to have consistent implementations.

edvalley: I see your point. I'll do it that way then. Thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Sorry, there is something I'm missing. We don't have code like this anymore in the encoder. Thank you.

tananaev: Isn't in the base encoder?

edvalley: Are you talking about the `initDevicePassword` method on `BaseProtocolEncoder` class? The one that is protected and receives a `Command` object that I don't need here?. It can't be accessed directly from the decoder. Maybe we can put both methods on the base encoder, unify shared code and provide a public method to get the password directly from a `Device` object or a device ID string. What do you propose? Thanks.

tananaev: Yes, we need to unify the code. One option is to define it in the `IdentityManager`.

edvalley: Great idea. I'll work on it. Thanks...

edvalley:
I've made new changes. Comments are welcomed. Thanks for your time...

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
```diff
@@ -0,0 +1,53 @@
...
+                return formatCommand(command, "AVREQ,{%s},1",
+                    Command.KEY_DEVICE_PASSWORD);
+            case Command.TYPE_REBOOT_DEVICE:
+                return formatCommand(command, "AVRESET,{%s},{%s}",
+                    Command.KEY_UNIQUE_ID, Command.KEY_DEVICE_PASSWORD);
+            default:
+                break;
+        }
+
+        return null;
```
tananaev: Move this to default.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -145,23 +222,41 @@ protected Object decode(
...
     }
 
+    @Override
+    protected Object decode(
+        Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
+
+        String sentence = (String) msg;
+
+        if (sentence.startsWith("$ECHK")) {
+            Parser parser = new Parser(PATTERN_ECHK, sentence);
```
tananaev: This can probably be moved to corresponding methods. Also instead of `handleEchk` we usually names like `decodeEchk`.

src/main/java/org/traccar/database/IdentityManager.java
```diff
@@ -26,6 +26,10 @@
 
     Device getByUniqueId(String uniqueId) throws Exception;
 
+    String getDevicePassword(long id);
+
+    String getDevicePassword(long id, String defaultPassword);
```
tananaev: What's the reason for having two versions of `getDevicePassword`?

src/main/java/org/traccar/protocol/LaipacProtocolEncoder.java
edvalley: Done.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Done.

src/main/java/org/traccar/database/IdentityManager.java
edvalley: Nothing special. Keeping just one. Thanks.

edvalley:
Made changes again. Thanks a lot for your feedback.

edvalley:
I've confirmed that my Laipac device odometer parameter unit is kilometers (checked documentation too). The related code is taking it as meters, displaying it incorrectly at web interface. Should I correct it (maybe in this PR) or do I need to do something else to confirm other users doesn't have any problem with this change. Thanks.

src/test/java/org/traccar/TestIdentityManager.java
```diff
@@ -29,6 +29,14 @@ public Device getByUniqueId(String uniqueId) {
         return createDevice();
     }
 
+    public String getDevicePassword(long id) {
```
tananaev: You still have this method here. That's why you should annotate methods as `@Override`.

src/main/java/org/traccar/database/DeviceManager.java
```diff
@@ -114,6 +115,16 @@ public Device getByUniqueId(String uniqueId) throws SQLException {
         return devicesByUniqueId.get(uniqueId);
     }
 
+    public String getDevicePassword(long id, String defaultPassword) {
```
tananaev: Add `@Override` as well.

src/test/java/org/traccar/TestIdentityManager.java
edvalley: Oh, you are right. Sorry I missed that. Thanks.

src/main/java/org/traccar/database/DeviceManager.java
edvalley: Same here, sorry...

edvalley:
Fixed things up. Awaiting for review. Thanks again for your time.

edvalley:
I would like to know which is the right place or way to talk about feature implementations before making a pull request? Thank you very much.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -37,29 +38,41 @@ public LaipacProtocolDecoder(Protocol protocol) {
         super(protocol);
     }
 
-    private static final Pattern PATTERN = new PatternBuilder()
+    public static final String DEFAULT_DEVICE_PASSWORD = "00000000";
+
+    private static final Pattern PATTERN_ECHK = new PatternBuilder()
```
tananaev: I just realized that you don't seem to use data from this pattern anywhere. What's the point of this regex then?

tananaev:
> I would like to know which is the right place or way to talk about feature implementations before making a pull request? Thank you very much.

The best option is to create an issue to discuss it.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: You're right, I'm not using it right now. I think I was just trying to keep the code uniform. No problem, removing it...

edvalley:
Here again, sorry about all this back and forth. Ready for review. Thank you very much...

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -81,37 +88,108 @@ private String decodeAlarm(String event) {
...
+
+        if (Character.isLowerCase(status.charAt(0))) {
+            String response = "$EAVACK," + event + "," + checksum;
+            response += Checksum.nmea(response) + "\r\n";
+            channel.writeAndFlush(new NetworkMessage(response, remoteAddress));
+        }
+
+    }
+
+    private Object decodeEchk(
```
tananaev: OK, I think now it doesn't make sense to have this method. All it does is this:
```java
if (channel != null) {
    channel.writeAndFlush(new NetworkMessage(sentence + "\r\n", remoteAddress));
}
```

edvalley: Again, just trying to keep the code uniform. No problem, removing it. Thanks...

edvalley:
Fixes done. I avoided directly handling ignition attribute. This devices doesn't have a dedicated input for this purpose. Also, anyway it was giving false positives because input status cannot be determined from every position report. I think using the copy attributes feature together with the computed attributes feature, will give the bests results. Thanks...

tananaev:
Merged, thanks.

edvalley:
Than you very much. I'll be doing some more pull request. You are doing a great job.

PR #4349, opened 2019-07-07
Fix, rssi info is deleting when locations get from geolocation services

truebest:
We losing data from device when location captured via geoservices.


tananaev:

PR #4355, opened 2019-07-16
Make device password a group configurable attribute

edvalley:
As the title says... Thanks.

### DIFF
--- a/src/main/java/org/traccar/database/DeviceManager.java
+++ b/src/main/java/org/traccar/database/DeviceManager.java
@@ -116,11 +116,18 @@ public class DeviceManager extends BaseObjectManager<Device> implements Identity
     }
 
     @Override
-    public String getDevicePassword(long id, String defaultPassword) {
+    public String getDevicePassword(long id, String protocol, String defaultPassword) {
 
-        String password = getById(id).getString(Command.KEY_DEVICE_PASSWORD);
+        String password = lookupAttributeString(id, Command.KEY_DEVICE_PASSWORD, null, false);
         if (password != null) {
             return password;
+        }
+
+        if (protocol != null) {
+            password = Context.getConfig().getString(protocol + "." + Command.KEY_DEVICE_PASSWORD);
+            if (password != null) {
+                return password;
+            }
         }
 
         return defaultPassword;
--- a/src/main/java/org/traccar/database/DeviceManager.java
+++ b/src/main/java/org/traccar/database/DeviceManager.java
@@ -118,7 +118,7 @@ public class DeviceManager extends BaseObjectManager<Device> implements Identity
     @Override
     public String getDevicePassword(long id, String protocol, String defaultPassword) {
 
-        String password = lookupAttributeString(id, Command.KEY_DEVICE_PASSWORD, null, false);
+        String password = lookupAttributeString(id, Command.KEY_DEVICE_PASSWORD, null, false, false);
         if (password != null) {
             return password;
         }
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolEncoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolEncoder.java
@@ -1,4 +1,3 @@
-
 /*
  * Copyright 2015 - 2019 Anton Tananaev (anton@traccar.org)
  *
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolEncoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolEncoder.java
@@ -34,7 +34,7 @@ public class Gt06ProtocolEncoder extends BaseProtocolEncoder {
     private ByteBuf encodeContent(long deviceId, String content) {
 
         boolean language = Context.getIdentityManager()
-            .lookupAttributeBoolean(deviceId, "gt06.language", false, false, true);
+            .lookupAttributeBoolean(deviceId, getProtocolName() + ".language", false, false, true);
 
         ByteBuf buf = Unpooled.buffer();
 
@@ -67,13 +67,17 @@ public class Gt06ProtocolEncoder extends BaseProtocolEncoder {
     protected Object encodeCommand(Command command) {
 
         boolean alternative = Context.getIdentityManager().lookupAttributeBoolean(
-                command.getDeviceId(), "gt06.alternative", false, false, true);
+                command.getDeviceId(), getProtocolName() + ".alternative", false, false, true);
+
+        initDevicePassword(command, "123456");
 
         switch (command.getType()) {
             case Command.TYPE_ENGINE_STOP:
-                return encodeContent(command.getDeviceId(), alternative ? "DYD,123456#" : "Relay,1#");
+                return encodeContent(command.getDeviceId(),
+                    alternative ? "DYD," + Command.KEY_DEVICE_PASSWORD + "#" : "Relay,1#");
             case Command.TYPE_ENGINE_RESUME:
-                return encodeContent(command.getDeviceId(), alternative ? "HFYD,123456#" : "Relay,0#");
+                return encodeContent(command.getDeviceId(),
+                    alternative ? "HFYD," + Command.KEY_DEVICE_PASSWORD + "#" : "Relay,0#");
             case Command.TYPE_CUSTOM:
                 return encodeContent(command.getDeviceId(), command.getString(Command.KEY_DATA));
             default:
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolEncoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolEncoder.java
@@ -1,3 +1,4 @@
+
 /*
  * Copyright 2015 - 2019 Anton Tananaev (anton@traccar.org)
  *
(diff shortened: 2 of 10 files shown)


src/main/java/org/traccar/database/DeviceManager.java
```diff
@@ -117,13 +117,7 @@ public Device getByUniqueId(String uniqueId) throws SQLException {
...
     @Override
     public String getDevicePassword(long id, String defaultPassword) {
-
-        String password = getById(id).getString(Command.KEY_DEVICE_PASSWORD);
-        if (password != null) {
-            return password;
-        }
-
-        return defaultPassword;
+        return lookupAttributeString(id, Command.KEY_DEVICE_PASSWORD, defaultPassword, false);
```
tananaev: I think it should be configurable in the configuration, not server. That's for security reasons.

Also, I think it should be configurable per protocol.

edvalley:
I understand, but, what if I have same protocol / different model devices and I have a different password per model. Or, same protocol / different users / different password, thinking on device management delegation. I'm not saying my approach is the right one. Thanks.

src/main/java/org/traccar/database/DeviceManager.java
edvalley: Sorry, commented bellow...

tananaev: How does your version solve any of the mentioned problems?

edvalley: You can create a group of devices, lets say all same model devices, or all devices of a specific user, even if not the same protocol, then add the `devicePassword` attribute for that group, and then delegate management of that group to an specific administrator.

edvalley: Maybe the attribute can be named `protocol.devicePassword` to make it a protocol specific, server wide attribute, but I think the proposed approach is more generic.

edvalley: As you said, it can be a file configuration, in the same form `protocol.devicePassword`, but again, the proposed approach I think is more generic. Thanks.

tananaev: I think there is some misunderstanding. Your approach is already server-wide.

edvalley: > Maybe the attribute can be named `protocol.devicePassword` to make it a protocol specific, server wide attribute, but I think the proposed approach is more generic.

Yes, I meant we can implement it only as a per protocol, server wide attribute, instead of a server wide attribute for all protocols. But, sorry for repetition, the proposed approach is not restricted to a protocol, you can make it as wide as you like using device groups, and solves the problem that for now, we need to define a per device password if we don't want to use the hard-coded protocol default. Am I wrong about the last part? Is that what you are trying to say?

edvalley: Of course, for this to work we need to enable `deviceManager.lookupGroupsAttribute` in config

tananaev: I'm not saying that it shouldn't be configurable per group. I'm saying that it also should be configurable per protocol and in the config file. It's more flexible than your approach.

edvalley: Oh, I see. So, are you saying that we must keep this approach I'm proposing, but also we need to make it configurable via `traccar.xml` using somethink like `[protocol].devicePassword`?

tananaev: Yes.

edvalley: Great! We are now on the same page. I'll come back. Thank you very much.

edvalley:
Just before I make a mess from this. We need a protocol name string in `IdentityManager.getDevicePassword` to build the configuration key `[protocol].devicePassword`. We use that method from both, `BaseProtocolEncoder` and `BaseProtocolDecoder`. The decoder already receives a `Protocol` object in the constructor, but the encoder doesn't. Now, I remember you said maybe we should pass it in the base class (I think that time has come) which will require to change all protocol encoder's constructors. Do you want me to do that right now or you can think of another approach? Thank you.

tananaev:
Ideally we add it to the base encoder, same way we do for decoder. If you do it, please make a separate pull request for it.

edvalley:
Done. Let me split things up to make another pull request. Then, we can finish this one. Thanks a lot.

edvalley:
Well, waiting for review on PR [https://github.com/traccar/traccar/pull/4356](https://github.com/traccar/traccar/pull/4356). Thanks...

edvalley:
Changes done. Thanks...

tananaev:
There is something wrong with the PR. It shows all the encoder changes.

edvalley:
Oh no... I see I made a mistake merging changes. I don't know if I can fix it, I'll try. Can I make another PR if I can't? Thank you very much.

tananaev:
Yeah, you can make another PR or just replace changes here.

edvalley:
Great! Thanks again.

edvalley:
It seems something went wrong. Travis exited with code 0 but here says still pending.

src/main/java/org/traccar/database/DeviceManager.java
```diff
@@ -116,13 +116,20 @@ public Device getByUniqueId(String uniqueId) throws SQLException {
     }
 
     @Override
-    public String getDevicePassword(long id, String defaultPassword) {
+    public String getDevicePassword(long id, String protocol, String defaultPassword) {
 
-        String password = getById(id).getString(Command.KEY_DEVICE_PASSWORD);
+        String password = lookupAttributeString(id, Command.KEY_DEVICE_PASSWORD, null, false);
```
tananaev: This still checks server-wide settings for the password.

edvalley: Yes, this way we have in descending order of preference:
- a device-specific web configurable value
- a group-wide web configurable value
- a server-wide web configurable value
- a protocol-wide file `(traccar.xml)` configurable  value
- a protocol-wide hard-coded default

I think it is very flexible, and I thought we agreed on that from here:
https://github.com/traccar/traccar/pull/4355#discussion_r304123522

Maybe there's something I misunderstood, please, correct me if that's the case.
Thank you.

tananaev: I don't want it to be server-wide settings configurable because server confirm is public. Anyone can see it.

edvalley: Anyway, if there's no server-wide password configured, it won't be used. Am I right? Thanks.

edvalley: I wrote before seeing your comment, but, isn't that the solution?

tananaev: I don't think we should allow it because not everyone realizes that server attributes are public.

edvalley: I see your point, and totally agree. So, are you saying that we can keep every step in this lookup for the attribute, except the server-wide attribute? Thanks.

tananaev: Yes. Or just use option with config with protocol prefix.

edvalley: Would you like if we make optional the lookup for server attributes with an additional argument to the private method `lookupAttribute` of `DeviceManager` class?

tananaev: I guess it makes sense.

edvalley: Let's do it. I'll make another PR. Thanks.

edvalley:
Waiting for comments on PR [https://github.com/traccar/traccar/pull/4357](https://github.com/traccar/traccar/pull/4357). Thanks a lot.

edvalley:
Well, here we are again. Waiting for comments. Thanks a lot.

edvalley:
I don't know what am I doing wrong about merge that all changes from the other PR are appearing here...

edvalley:
If I use `git diff master` the only changes that I see are those intended

tananaev:
Next time you should probably rebase instead of merging.

edvalley:
That's what I'm doing, rebasing my branch on upstream/master. I think I made some mistake in the process.

tananaev:
It doesn't seem like you are rebasing correctly. When you rebase you shouldn't have my merges on your branch.

edvalley:
You're right, I'm doing something wrong, let me try to fix it. If I can't, will reopen. Sorry wasting your time. Thanks.

edvalley:
Can you please check if everything is ok now? Thanks.

tananaev:
Looks good. Do we need to update any other places where we have hardcoded protocol name in encoders?

edvalley:
Let me check...

edvalley:
You were right, some encoders, were using hard-coded protocol names as well as non-configurable default passwords. I hope no one escaped. Check whenever you want. Thanks.

edvalley:
Sorry, wait, I made a mistake replacing passwords...

edvalley:
I realized that `formatCommand` is not available for binary protocols. Tell me if you want to do something different about my change. Thank you.

src/main/java/org/traccar/protocol/Gt06ProtocolEncoder.java
```diff
@@ -1,3 +1,4 @@
+
```
tananaev: Please remove this new line.

edvalley: Oh, what a typo! Done.

edvalley:
Fixed.

edvalley:
Did you review this? Thanks.

tananaev:
Merged, thanks.

edvalley:
Thank you very much.

PR #4357, opened 2019-07-17
Make optional server attributes lookup

edvalley:
Make optional the server-wide attributes lookup. It's already avoided when a lookup up is performed on the configuration file, I'm keeping it that way...

### DIFF
--- a/src/main/java/org/traccar/database/DeviceManager.java
+++ b/src/main/java/org/traccar/database/DeviceManager.java
@@ -387,13 +387,12 @@ public class DeviceManager extends BaseObjectManager<Device> implements Identity
                     }
                 }
             }
-            if (result == null) {
-                if (lookupConfig) {
-                    result = Context.getConfig().getString(attributeName);
-                } else if (lookupServer) {
-                    Server server = Context.getPermissionsManager().getServer();
-                    result = server.getAttributes().get(attributeName);
-                }
+            if (result == null && lookupConfig) {
+                result = Context.getConfig().getString(attributeName);
+            }
+            if (result == null && lookupServer) {
+                Server server = Context.getPermissionsManager().getServer();
+                result = server.getAttributes().get(attributeName);
             }
         }
         return result;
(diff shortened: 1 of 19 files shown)


src/main/java/org/traccar/database/DeviceManager.java
```diff
@@ -390,7 +390,7 @@ private Object lookupAttribute(long deviceId, String attributeName, boolean look
             if (result == null) {
                 if (lookupConfig) {
                     result = Context.getConfig().getString(attributeName);
-                } else {
+                } else if (lookupServer) {
```
tananaev: I think it would make sense to remove `else`. In future we might potentially want to use both.

edvalley: I thought the same, but as I said in my first comment, just wanted to keep same behavior. No problem, removing it. Thanks.

tananaev: It's not a public method, so we don't have to keep existing behavior for this one.

edvalley: You're right. Working on that...

edvalley:
Fixed. Waiting for comments. Thanks again.

edvalley:
One thing I think about now, will configuration file settings precede server attributes, in case we lookup both? Or we need to reverse that behavior?

src/main/java/org/traccar/database/DeviceManager.java
```diff
@@ -325,7 +325,7 @@ public Position getLastPosition(long deviceId) {
     @Override
     public boolean lookupAttributeBoolean(
             long deviceId, String attributeName, boolean defaultValue, boolean lookupConfig) {
-        Object result = lookupAttribute(deviceId, attributeName, lookupConfig);
+        Object result = lookupAttribute(deviceId, attributeName, true, lookupConfig);
```
tananaev: Passing `true` every time changes behavior of these functions.

edvalley: You're right, I coded it that way for the first commit, and never looked back for the second. I think `!lookupConfig` will make the trick.

edvalley: Now, do you think we should override the methods that calls `lookupAttribute`, or modify every call to it?

tananaev: I think modify every call.

edvalley: I agree. Ok, let's do it. Thanks.

edvalley:
Changes done. Waiting for review. Thanks for your time.

edvalley:
Thanks.

PR #4371, opened 2019-07-30
Add leafspy app protocol and decoder

jesserockz:
This PR adds a protocol and decoder to decode the live data that can be sent from the [LeafSpy Pro](http://www.electricvehiclewiki.com/wiki/leaf-spy-pro/) application which pulls information from the OBD port of a Nissan Leaf.

### DIFF
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -269,5 +269,6 @@
     <entry key='starcom.port'>5190</entry>
     <entry key='mictrack.port'>5191</entry>
     <entry key='plugin.port'>5192</entry>
+    <entry key='leafspy.port'>5193</entry>
 
 </properties>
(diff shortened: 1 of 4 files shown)


tananaev:
Please make sure you include at least one unit test. Also, do you have a link to protocol/format documentation?

jesserockz:
Hi Anton, Thanks for taking a look. 
Documentation for the app is available [here](http://www.leafspypro.com/wp-content/uploads/2017/09/LeafSpy-Help-1_2_1-iOS.pdf) with server data details on pages 39-40. 
Will get a/some tests in when I can next.

tananaev:
Documentation says that lat/lon should be in the DD MM.MMM format.

jesserockz:
I believe that is the CSV file format, but the http client always sends decimal format

```
Lat=xxx.xxxxx
Long=xxx.xxxx
```

tananaev:
But doc says it should be the same format. Is the documentation incorrect?

jesserockz:
Im not sure. Im just a user of the app and my local server always receives decimal format.

tananaev:
Merged, thanks.

PR #4374, opened 2019-08-01
EGTS protocol support for oid authentication (auto-detect)

7erg:
EGTS in some retranslator systems uses OID containing last part of imei, without authentication package. In this mode OID is used as deviceId. DeviceId cache in this mode is disabled, because of many devices per channel.


tananaev:
It looks like you have many more changes that you mentioned. Can we please separate PR into smaller parts. If you want to refactor decoder, it has to be separate from any new functionality. Also, you have some commented out code that would have to be fixed.

7erg:
Just fixed commented code. 
There is no refactoring. It may look like many changes because of whitespaces (large part of code moved into `while` over package length).

tananaev:
Why do we need this extra while loop?

7erg:
Sorry, it is my mistake - I faced to data package with multiple frames is not read completely when running tests, but forgot that EgtsFrameDecoder is not applied there.

src/main/java/org/traccar/BaseProtocolDecoder.java
```diff
@@ -143,7 +143,13 @@ private long findDeviceId(SocketAddress remoteAddress, String... uniqueIds) {
     }
 
     public DeviceSession getDeviceSession(Channel channel, SocketAddress remoteAddress, String... uniqueIds) {
+        return getDeviceSession(channel, remoteAddress, false, uniqueIds);
+    }
+
+    public DeviceSession getDeviceSession(Channel channel, SocketAddress remoteAddress, boolean ignoreCache,
+                                          String... uniqueIds) {
```
tananaev: We don't use formatting like this. Please move all argument to a new line.

src/main/java/org/traccar/protocol/EgtsProtocolDecoder.java
```diff
@@ -121,19 +127,29 @@ protected Object decode(
...
+        short headerLength = buf.getUnsignedByte(buf.readerIndex() + 3);
+        int frameDataLength = buf.getUnsignedShortLE(buf.readerIndex() + 5);
         int index = buf.getUnsignedShort(buf.readerIndex() + 5 + 2);
-        buf.skipBytes(buf.getUnsignedByte(buf.readerIndex() + 3));
+        short packetType = buf.getUnsignedByte(buf.readerIndex() + 5 + 2 + 2);
+        buf.skipBytes(headerLength);
 
-        List<Position> positions = new LinkedList<>();
+        if (packetType == PT_RESPONSE) {
+            // some retranslator systems do send response on response packets - just skip them
```
tananaev: I think comment is unnecessary.

src/main/java/org/traccar/protocol/EgtsProtocolDecoder.java
```diff
@@ -121,19 +127,29 @@ protected Object decode(
...
+            return null;
+        }
 
-        while (buf.readableBytes() > 2) {
+        int frameDataEnd = buf.readerIndex() + frameDataLength;
```
tananaev: Why do we need this change?

src/main/java/org/traccar/protocol/EgtsProtocolDecoder.java
```diff
@@ -121,19 +127,29 @@ protected Object decode(
...
+        long oid = 0L;
```
tananaev: Please use full words for variables. This should be something like `objectId`.

src/main/java/org/traccar/protocol/EgtsProtocolDecoder.java
```diff
@@ -121,19 +127,29 @@ protected Object decode(
...
+        while (buf.readerIndex() < frameDataEnd) {
 
             int length = buf.readUnsignedShortLE();
             int recordIndex = buf.readUnsignedShortLE();
             int recordFlags = buf.readUnsignedByte();
 
             if (BitUtil.check(recordFlags, 0)) {
-                buf.readUnsignedIntLE(); // object id
+                oid = buf.readUnsignedIntLE(); // object id
```
tananaev: Comment is unnecessary now.

src/main/java/org/traccar/protocol/EgtsProtocolDecoder.java
```diff
@@ -172,12 +189,14 @@ protected Object decode(
                         buf.readUnsignedShortLE(); // home dispatcher identifier
                     }
                     if (BitUtil.check(flags, 1)) {
-                        getDeviceSession(
-                                channel, remoteAddress, buf.readSlice(15).toString(StandardCharsets.US_ASCII).trim());
+                        String imei = buf.readSlice(15).toString(StandardCharsets.US_ASCII).trim();
+                        LOGGER.trace("[{}] Using IMEI1 as deviceId: {}", channel == null ? "" : channel.id(), imei);
```
tananaev: I don't really understand the need for logging here. I would recommend reverting all 3 places.

src/main/java/org/traccar/protocol/EgtsProtocolDecoder.java
```diff
@@ -254,8 +274,17 @@ protected Object decode(
                 buf.readerIndex(end);
             }
 
-            if (serviceType == SERVICE_TELEDATA && deviceSession != null) {
-                positions.add(position);
+            if (serviceType == SERVICE_TELEDATA && position.getFixTime() != null) {
```
tananaev: Why are we checking `position.getFixTime() != null` here?

src/test/java/org/traccar/protocol/EgtsProtocolDecoderTest.java
```diff
@@ -6,7 +6,52 @@
...
+        verifyNull(decoder, binary(
+                "0100000b0003006c430004010000acfb"));
+        verifyNull(decoder, binary(
+                "0100000b0003006e430088030000cc95"));
+        verifyNull(decoder, binary(
```
tananaev: Seems like a lot of duplication in unit tests. Do we need it?

Also, please fix spacing.

src/test/java/org/traccar/protocol/EgtsProtocolDecoderTest.java
```diff
@@ -6,7 +6,52 @@
 public class EgtsProtocolDecoderTest extends ProtocolTest {
 
     @Test
-    public void testDecode() throws Exception {
+    public void testDecodeWithOid() throws Exception {
```
tananaev: Use `objectId` instead of `oid` everywhere please..

src/main/java/org/traccar/protocol/EgtsProtocolDecoder.java
```diff
@@ -254,8 +274,17 @@ protected Object decode(
...
+                if (useOidAsDeviceId && oid != 0L) {
+                    LOGGER.debug("[{}] Using OID as deviceId: {}", channel == null ? "" : channel.id(), oid);
+                    deviceSession = getDeviceSession(channel, remoteAddress, true, String.valueOf(oid));
+                    if (deviceSession != null) {
+                        position.setDeviceId(deviceSession.getDeviceId());
+                    }
+                }
+                if (position.getDeviceId() != 0L) {
```
tananaev: I think it's better to check `deviceSession`.

src/main/java/org/traccar/protocol/EgtsProtocolDecoder.java
7erg: position is created before loop, and when loop ends position is empty some times. Should we change this to position.isValid()?

src/main/java/org/traccar/protocol/EgtsProtocolDecoder.java
7erg: I see it also was usable with unsplitted frames, reverted

tananaev:
Merged, thanks.

PR #4388, opened 2019-08-24
Enhance Wialon protocol decoder

edvalley:
Enhancing Wialon protocol decoder, mainly extending it according to specifications to decode messages carrying the device id, allowing to:
- Use UDP protocol as transport
- Receive messages for multiple devices using the same TCP connection (of course, disabling session cache for the protocol)

Waiting for comments...
Thanks a lot


tananaev:
Merged, thanks.

edvalley:
Great!
Thanks.

PR #4389, opened 2019-08-26
Add an optional queue to web data handler

edvalley:
Added an optional (configurable) queue, actually a ring buffer, to forwarded positions in web data handler. The size of the queue can be specified in the configuration file. If the queue is enabled, HTTP responses needs to be 200 OK to get positions removed from the queue. I'm planning to do the same with event forwarding.
Waiting for comments...
Thanks.

### DIFF
--- a/src/main/java/org/traccar/MainModule.java
+++ b/src/main/java/org/traccar/MainModule.java
@@ -378,7 +378,7 @@ public class MainModule extends AbstractModule {
 
     @Singleton
     @Provides
-    public static Timer provideGlobalTimer() {
+    public static Timer provideTimer() {
         return GlobalTimer.getTimer();
     }
 
--- a/src/main/java/org/traccar/MainModule.java
+++ b/src/main/java/org/traccar/MainModule.java
@@ -73,6 +73,7 @@ import org.traccar.reports.model.TripsConfig;
 
 import javax.annotation.Nullable;
 import javax.ws.rs.client.Client;
+import io.netty.util.Timer;
 
 public class MainModule extends AbstractModule {
 
@@ -373,6 +374,12 @@ public class MainModule extends AbstractModule {
     @Provides
     public static DriverEventHandler provideDriverEventHandler(IdentityManager identityManager) {
         return new DriverEventHandler(identityManager);
+    }
+
+    @Singleton
+    @Provides
+    public static Timer provideGlobalTimer() {
+        return GlobalTimer.getTimer();
     }
 
     @Override
--- a/src/main/java/org/traccar/WebDataHandler.java
+++ b/src/main/java/org/traccar/WebDataHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 - 2019 Anton Tananaev (anton@traccar.org)
+ * Copyright 2015 - 2020 Anton Tananaev (anton@traccar.org)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,9 @@ package org.traccar;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.netty.channel.ChannelHandler;
+import io.netty.util.Timeout;
+import io.netty.util.TimerTask;
+
 import org.traccar.config.Config;
 import org.traccar.config.Keys;
 import org.traccar.database.IdentityManager;
@@ -41,6 +44,7 @@ import java.util.Calendar;
 import java.util.Formatter;
 import java.util.Locale;
 import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 @ChannelHandler.Sharable
@@ -57,12 +61,11 @@ public class WebDataHandler extends BaseDataHandler {
     private final String header;
     private final boolean json;
 
-    private final Boolean retryEnabled;
-    private final Integer retryDelayMin;
-    private final Integer retryDelayMax;
-    private final Integer deliveryPendingLimit;
+    private final boolean retryEnabled;
+    private final int retryDelayMin;
+    private final int retryDelayMax;
+    private final int deliveryPendingLimit;
 
-    private AtomicInteger deliveryFaliuresInRow;
     private AtomicInteger deliveryPendingCurrent;
 
     @Inject
@@ -77,17 +80,10 @@ public class WebDataHandler extends BaseDataHandler {
         this.json = config.getBoolean(Keys.FORWARD_JSON);
 
         this.retryEnabled = config.getBoolean(Keys.FORWARD_RETRY_ENABLE);
+        this.retryDelayMin = config.getInteger(Keys.FORWARD_RETRY_DELAY_MIN, 1);
+        this.retryDelayMax = config.getInteger(Keys.FORWARD_RETRY_DELAY_MAX, 10);
+        this.deliveryPendingLimit = config.getInteger(Keys.FORWARD_RETRY_PENDING_LIMIT, 100);
 
-        Integer retryDelayMin = config.getInteger(Keys.FORWARD_RETRY_DELAY_MIN, 1);
-        this.retryDelayMin = (retryDelayMin > 0 && retryDelayMin < 3600) ? retryDelayMin : 1;
-
-        Integer retryDelayMax = config.getInteger(Keys.FORWARD_RETRY_DELAY_MAX, 10);
-        this.retryDelayMax = (retryDelayMax > retryDelayMin && retryDelayMax < 3600) ? retryDelayMax : retryDelayMin;
-
-        Integer deliveryPendingLimit = config.getInteger(Keys.FORWARD_RETRY_PENDING_LIMIT, 100);
-        this.deliveryPendingLimit = (retryDelayMax > 0) ? deliveryPendingLimit : 100;
-
-        this.deliveryFaliuresInRow = new AtomicInteger(0);
         this.deliveryPendingCurrent = new AtomicInteger(0);
     }
 
@@ -183,15 +179,15 @@ public class WebDataHandler extends BaseDataHandler {
 
         class AsyncRequestAndCallback implements InvocationCallback<Response> {
 
-            private Integer delay = retryDelayMin;
-            private Map<String, Object> jsonPayload;
+            private int delay = retryDelayMin;
+            private Map<String, Object> payload;
             private Invocation.Builder requestBuilder;
 
             AsyncRequestAndCallback(Position position) {
 
                 String formattedUrl;
                 try {
-                    formattedUrl = (json) ? url : formatRequest(position);
+                    formattedUrl = json ? url : formatRequest(position);
                 } catch (UnsupportedEncodingException | JsonProcessingException e) {
                     throw new RuntimeException("Forwarding formatting error", e);
                 }
@@ -205,7 +201,7 @@ public class WebDataHandler extends BaseDataHandler {
                 }
 
                 if (json) {
-                    jsonPayload = prepareJsonPayload(position);
+                    payload = prepareJsonPayload(position);
                 }
 
                 deliveryPendingCurrent.incrementAndGet();
@@ -215,7 +211,7 @@ public class WebDataHandler extends BaseDataHandler {
 
             private void send() {
                 if (json) {
-                    requestBuilder.async().post(Entity.json(jsonPayload), this);
+                    requestBuilder.async().post(Entity.json(payload), this);
                 } else {
                     requestBuilder.async().get(this);
                 }
@@ -223,36 +219,35 @@ public class WebDataHandler extends BaseDataHandler {
 
             private void retry() {
                 try {
-                    deliveryFaliuresInRow.incrementAndGet();
-                    if (!retryEnabled || deliveryPendingCurrent.get() > deliveryPendingLimit) {
-                        deliveryPendingCurrent.decrementAndGet();
-                    } else {
-                        Integer i = 0;
-                        for ( ; i < delay; i++) {
-                            Thread.sleep(1000);
-                            if (deliveryFaliuresInRow.get() == 0) {
-                                delay = retryDelayMin;
-                                break;
+                    if (retryEnabled && deliveryPendingCurrent.get() <= deliveryPendingLimit) {
+                        GlobalTimer.getTimer().newTimeout(new TimerTask() {
+                            @Override
+                            public void run(Timeout timeout) {
+                                if (!timeout.isCancelled()) {
+                                    if (delay < retryDelayMax) {
+                                        delay++;
+                                    }
+                                    send();
+                                }
                             }
-                        }
-                        if (i >= delay && delay < retryDelayMax) {
-                            delay++;
-                        }
-                        send();
+                        }, delay, TimeUnit.SECONDS);
+                        return;
                     }
                 } catch (Exception e) {
                 }
+                deliveryPendingCurrent.decrementAndGet();
             }
 
+            @Override
             public void completed(Response response) {
                 if (response.getStatus() == 200) {
-                    deliveryFaliuresInRow.set(0);
                     deliveryPendingCurrent.decrementAndGet();
                 } else {
                     retry();
                 }
             }
 
+            @Override
             public void failed(Throwable throwable) {
                 retry();
             }
--- a/src/main/java/org/traccar/WebDataHandler.java
+++ b/src/main/java/org/traccar/WebDataHandler.java
@@ -27,9 +27,11 @@ import org.traccar.model.Position;
 import org.traccar.model.Group;
 
 import javax.inject.Inject;
+import javax.ws.rs.core.Response;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.Invocation;
+import javax.ws.rs.client.InvocationCallback;
 import java.util.HashMap;
 import java.util.Map;
 import java.io.UnsupportedEncodingException;
@@ -39,12 +41,22 @@ import java.util.Calendar;
 import java.util.Formatter;
 import java.util.Locale;
 import java.util.TimeZone;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.apache.commons.collections.Buffer;
+import org.apache.commons.collections.BufferUtils;
+import org.apache.commons.collections.buffer.CircularFifoBuffer;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @ChannelHandler.Sharable
 public class WebDataHandler extends BaseDataHandler {
 
     private static final String KEY_POSITION = "position";
     private static final String KEY_DEVICE = "device";
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(WebDataHandler.class);
 
     private final IdentityManager identityManager;
     private final ObjectMapper objectMapper;
@@ -53,6 +65,13 @@ public class WebDataHandler extends BaseDataHandler {
     private final String url;
     private final String header;
     private final boolean json;
+
+    private final Integer queueSize;
+
+    private Buffer positionsQueue = null;
+    private AtomicBoolean sendInProgress;
+
+    private Invocation.Builder requestBuilder;
 
     @Inject
     public WebDataHandler(
@@ -63,6 +82,12 @@ public class WebDataHandler extends BaseDataHandler {
         this.url = config.getString(Keys.FORWARD_URL);
         this.header = config.getString(Keys.FORWARD_HEADER);
         this.json = config.getBoolean(Keys.FORWARD_JSON);
+        Integer queueSize = (config.getInteger(Keys.FORWARD_QUEUE_SIZE, 0));
+        this.queueSize = (queueSize > 0) ? queueSize : 0;
+        if (this.queueSize > 0) {
+            this.positionsQueue = BufferUtils.synchronizedBuffer(new CircularFifoBuffer(this.queueSize));
+        }
+        this.sendInProgress = new AtomicBoolean(false);
     }
 
     private static String formatSentence(Position position) {
@@ -152,9 +177,7 @@ public class WebDataHandler extends BaseDataHandler {
         return request;
     }
 
-    @Override
-    protected Position handlePosition(Position position) {
-
+    protected void sendPosition(Position position) {
         String url;
         if (json) {
             url = this.url;
@@ -166,19 +189,63 @@ public class WebDataHandler extends BaseDataHandler {
             }
         }
 
-        Invocation.Builder requestBuilder = client.target(url).request();
-
-        if (header != null && !header.isEmpty()) {
-            for (String line: header.split("\\r?\\n")) {
-                String[] values = line.split(":", 2);
-                requestBuilder.header(values[0].trim(), values[1].trim());
+        if (!json || requestBuilder == null) {
+            requestBuilder = client.target(url).request();
+            if (header != null && !header.isEmpty()) {
+                for (String line: header.split("\\r?\\n")) {
+                     String[] values = line.split(":", 2);
+                     requestBuilder.header(values[0].trim(), values[1].trim());
+                }
             }
         }
 
+        InvocationCallback<Response> callback = new InvocationCallback<Response>() {
+            public void completed(Response response) {
+                if (positionsQueue != null) {
+                    if (response.getStatus() == 200) {
+                        positionsQueue.remove();
+                    }
+                    logQueueStatus();
+                    sendInProgress.set(false);
+                    sendQueuedPositions();
+                }
+            }
+            public void failed(Throwable throwable) {
+                if (positionsQueue != null) {
+                    logQueueStatus();
+                    sendInProgress.set(false);
+                    sendQueuedPositions();
+                }
+            }
+        };
+
         if (json) {
-            requestBuilder.async().post(Entity.json(prepareJsonPayload(position)));
+            requestBuilder.async().post(Entity.json(prepareJsonPayload(position)), callback);
         } else {
-            requestBuilder.async().get();
+            requestBuilder.async().get(callback);
+        }
+    }
+
+    protected void logQueueStatus() {
+        LOGGER.info(String.format("Position forwarding queue: %d/%d",
+            positionsQueue.size(), queueSize));
+    }
+
+    protected void sendQueuedPositions() {
+        if (!positionsQueue.isEmpty() && !sendInProgress.get()) {
+            sendInProgress.set(true);
+            sendPosition((Position) positionsQueue.get());
+        }
+    }
+
+    @Override
+    protected Position handlePosition(Position position) {
+
+        if (positionsQueue == null) {
+            sendPosition(position);
+        } else {
+            positionsQueue.add(position);
+            sendQueuedPositions();
         }
 
         return position;
--- a/src/main/java/org/traccar/WebDataHandler.java
+++ b/src/main/java/org/traccar/WebDataHandler.java
@@ -41,7 +41,7 @@ import java.util.Calendar;
 import java.util.Formatter;
 import java.util.Locale;
 import java.util.TimeZone;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.collections.Buffer;
 import org.apache.commons.collections.BufferUtils;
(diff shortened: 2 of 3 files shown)


edvalley:
Back again. I've modified the code to have configurable concurrency and auto-throttling.
Waiting for comments...
Thanks a lot.

edvalley:
Please, suggest what to do with queue status logging (on every forwarded position). Should I change the log level? To what level? Completely remove it?. Thanks.

edvalley:
Sorry, I know you should be very busy, but can you please review my code? Thank you very much in advance.

tananaev:
It's still a queue. I don't really understand reasoning for that.

edvalley:
Thanks. You're right, it's still a queue:

- Implemented using a ring buffer
- We can configure if we use it or not (we can keep legacy behavior if we want)
- We can make it as short as a 1 position buffer (better if we make it at least a 10 one)
- We can configure how many positions at most, at any time, can be pending, undelivered, retrying, this time in parallel (throttling)
- Pending deliveries are reduced to one if the queue becomes full (throttling)
- Positions are only queued if there are pending requests
- Allow us to implement (as you said) a retry logic for positions not correctly delivered
- Correct delivery is assured, by checking the HTTP result code

From my point of view, there is no reason to keep the legacy behavior, but as I said before, that's up to you. This implementation (simple in my opinion) can be tuned to **ensure no single forwarded position get lost**, at a really very low cost, and that's the only reasoning behind this idea.

If you like the idea, but don't like the implementation, please just tell me, suggest another approach, I will do my best to follow you.

Thanks again for your time. Sorry bothering you.

tananaev:
Can we please discuss the reasoning for queueing the data. You can have much simpler retry without queuing, which has pretty much the same advantages as you listed. I want to discuss the differences.

By the way, you are using atomic counter incorrectly, but let's settle on the overall solution first.

edvalley:
For the second point, I've no doubt you're right, as I said once, I'm not an experienced Java programmer, maybe I just need to read/study a little more. Sorry about that. Corrections are happily welcomed.

For the first point: I can think of another way to implement retrying, we can just use the callbacks to wait and retry. This way:

- We can configure how many times we want to retry
- We can configure how long we want to retry
- We can configure how long to wait before retry
- We can also limit how many positions can be pending at most (discarding older ones)
- I guess the underlying HTTP client connection pooling limits how many TCP connections can be established at any time
- No doubt, it will be simpler to code
- We don't need a queue

But:

- I roughly guess we use more resources (memory and CPU, by keeping callback objects instead of positions in a queue)
- We absolutely loss control about delivery order (with a queue, we reduce possible disorder to the configured concurrency)

The last one, is the main reason why I used a queue, just to do the best effort to keep the order. My first implementation kept it perfectly, but also perfectly I understood why you rejected it. This can be helpful or not, depending on receiver implementations, just thinking about systems generating events based on speed value (not moving before, moving now, can be the opposite). But, thinking about it now, with the current implementation, nobody could implement something like this.

What do you think about resource usage and delivery disorder? Do you think this approach will be better? Or do you have something else in mind?

Thanks a lot again.

tananaev:
1. Resources usage shouldn't be a problem. Callbacks should not have any data other than position reference.
2. The order is not guaranteed in Traccar itself. If you want to keep order, we need to fix problems in other places first (e.g. geocoding).

edvalley:
Good point, so we forget about disorder for now. So, can you think of something else? Or should we stick to keep callbacks waiting and retrying for a while? Have a configurable maximum pending requests and retry interval?

tananaev:
Callbacks seems like a reasonable solution. Probably exponential retry time makes sense with some configuration on max retries.

edvalley:
I agree about exponential retries (I'm thinking of a global value for all callbacks, incremented on failures and reset to zero after a successful delivery). But, we must also put a configurable limit on how many callbacks can be pending right? I'm just thinking in callbacks being accumulated because of high frequency input, zero output, and the retry time and max retries not causing faster discard. In this case, memory usage can be a problem.

tananaev:
Well, memory usage is the same whether it's a queue or not.

edvalley:
Yes, of course memory usage is the same, I'm just saying that we must limit it in this implementation, the same way it's limited by the queue size.

I already agreed in not using a queue, I understand it's simpler, and without taking order into account, there's really no reason to use it.

So, anything else suggest before getting to work?

tananaev:
Not really.

edvalley:
I'm back again.  Sorry for the almost two months delay, I've been very busy.
Here's what I got now:

When position forwarding retrying is enabled, additional attempts are made to deliver positions. If initial delivery fails, because of an unreachable server or an HTTP response different from '200 OK', the software waits for 'forward.retry.delay.min' seconds to retry delivery. On subsequent failures, this  delay is incremented by 1 second up to 'forward.retry.delay.max'. On successful delivery, the delay is reset to 'forward.retry.delay.min'. Pending positions to be delivered are limited to 'forward.retry.pending.limit'. If this limit is reached, positions are discarded before next retry.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -54,15 +57,38 @@
...
         this.objectMapper = objectMapper;
         this.client = client;
         this.url = config.getString(Keys.FORWARD_URL);
         this.header = config.getString(Keys.FORWARD_HEADER);
         this.json = config.getBoolean(Keys.FORWARD_JSON);
+
+        this.retryEnabled = config.getBoolean(Keys.FORWARD_RETRY_ENABLE);
+
+        Integer retryDelayMin = config.getInteger(Keys.FORWARD_RETRY_DELAY_MIN, 1);
+        this.retryDelayMin = (retryDelayMin > 0 && retryDelayMin < 3600) ? retryDelayMin : 1;
```
tananaev: I would recommend to remove the validations for value ranges. Don't see any good reason to have additional code for it. Configuration is intended for admin users, which should know what they are doing. For other configuration we never do validation, so also for consistency we should remove.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -54,15 +57,38 @@
...
         this.identityManager = identityManager;
         this.objectMapper = objectMapper;
         this.client = client;
         this.url = config.getString(Keys.FORWARD_URL);
         this.header = config.getString(Keys.FORWARD_HEADER);
         this.json = config.getBoolean(Keys.FORWARD_JSON);
+
+        this.retryEnabled = config.getBoolean(Keys.FORWARD_RETRY_ENABLE);
+
+        Integer retryDelayMin = config.getInteger(Keys.FORWARD_RETRY_DELAY_MIN, 1);
```
tananaev: After removing validations this will simplify.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -54,15 +57,38 @@
     private final String header;
     private final boolean json;
 
+    private final Boolean retryEnabled;
```
tananaev: Use primitives for all these configuration parameters.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +181,86 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+
+            private Integer delay = retryDelayMin;
+            private Map<String, Object> jsonPayload;
+            private Invocation.Builder requestBuilder;
+
+            AsyncRequestAndCallback(Position position) {
+
+                String formattedUrl;
+                try {
+                    formattedUrl = (json) ? url : formatRequest(position);
```
tananaev: Please remove unnecessary parentheses everywhere.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +181,86 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
-            url = this.url;
-        } else {
-            try {
-                url = formatRequest(position);
-            } catch (UnsupportedEncodingException | JsonProcessingException e) {
-                throw new RuntimeException("Forwarding formatting error", e);
+        class AsyncRequestAndCallback implements InvocationCallback<Response> {
+
+            private Integer delay = retryDelayMin;
+            private Map<String, Object> jsonPayload;
```
tananaev: I would just rename this to `payload`.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +181,86 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                        if (i >= delay && delay < retryDelayMax) {
+                            delay++;
+                        }
+                        send();
+                    }
+                } catch (Exception e) {
+                }
+            }
+
+            public void completed(Response response) {
```
tananaev: Any overrides should be properly annotated.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +181,86 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
-                requestBuilder.header(values[0].trim(), values[1].trim());
+            private void retry() {
+                try {
+                    deliveryFaliuresInRow.incrementAndGet();
+                    if (!retryEnabled || deliveryPendingCurrent.get() > deliveryPendingLimit) {
+                        deliveryPendingCurrent.decrementAndGet();
+                    } else {
+                        Integer i = 0;
+                        for ( ; i < delay; i++) {
+                            Thread.sleep(1000);
```
tananaev: I don't like what's going on here. Please use proper async scheduling, ideally with existing timer that we have.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Good.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Ok.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Ok.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Ok.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Ok.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Sorry, I missed that.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Sorry, should we use primitives also on `src/main/java/org/traccar/config/Keys.java`?
I just remembered that not using primitives there, was the reason why I didn't use primitives here.

tananaev: No, there we use class types.

edvalley: Hm... Ok then. Thanks.

src/main/java/org/traccar/WebDataHandler.java
edvalley: To be honest, now that I see it again, I don't like it either. Are you talking about using `GlobalTimer`? I hadn't seen it before. Thank you.

tananaev: Yes.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Do you think we should not check even for negative values?

tananaev: I don't think we need to. Configuration is for admin users and it's usually done once. I don't see why someone would enter negative value, unless they want to try and break their own server.

edvalley: Jajaja... That's a good one. Ok.

edvalley:
Just one more thing...
Do you think we should log pending positions status?
Maybe only on failures?
What log level?

tananaev:
I would log only errors with warning level.

edvalley:
Ok. I will push what I got now. In the mean time, I'll add the code to log errors. Thanks.

edvalley:
Well, tell me what do you think. Thank you very much.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +181,95 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                            + ". Delivery will not be retried.");
+                    }
+                } catch (Exception e) {
+                }
+                deliveryPendingCurrent.decrementAndGet();
+            }
+
+            @Override
+            public void completed(Response response) {
+                if (response.getStatus() == 200) {
```
tananaev: All 2xx codes should be a success. There might be a helper method somewhere to check it.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +181,95 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                                        send();
+                                    }
+                                }
+                            }, delay, TimeUnit.SECONDS);
+                            return;
+                        }
+                        LOGGER.warn(message + " Pending: " + pending
+                            + ". Delivery will not be retried.");
+                    }
+                } catch (Exception e) {
```
tananaev: Why do we need to catch generic exception.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +181,95 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                }
+            }
 
-        if (header != null && !header.isEmpty()) {
-            for (String line: header.split("\\r?\\n")) {
-                String[] values = line.split(":", 2);
-                requestBuilder.header(values[0].trim(), values[1].trim());
+            private void retry() {
+                try {
+                    String message = "Position forwarding failed.";
```
tananaev: Doesn't make sense to have a variable for logging message. Duplicate it if needed, but I would just make all logging less verbose. Make text shorter.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +181,95 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                    if (!retryEnabled) {
+                        LOGGER.warn(message);
+                    } else {
+                        int pending = deliveryPendingCurrent.get();
+                        if (pending <= deliveryPendingLimit) {
+                            LOGGER.warn(message + " Pending: " + pending
+                                + ". Retrying in " + delay + " seconds.");
```
tananaev: I think this fits in one line.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +181,95 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                                            delay++;
+                                        }
+                                        send();
+                                    }
+                                }
+                            }, delay, TimeUnit.SECONDS);
+                            return;
+                        }
+                        LOGGER.warn(message + " Pending: " + pending
+                            + ". Delivery will not be retried.");
```
tananaev: Same here.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +181,95 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                try {
+                    String message = "Position forwarding failed.";
+                    if (!retryEnabled) {
+                        LOGGER.warn(message);
+                    } else {
+                        int pending = deliveryPendingCurrent.get();
+                        if (pending <= deliveryPendingLimit) {
+                            LOGGER.warn(message + " Pending: " + pending
+                                + ". Retrying in " + delay + " seconds.");
+                            GlobalTimer.getTimer().newTimeout(new TimerTask() {
```
tananaev: Please use injection.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +181,95 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                                @Override
+                                public void run(Timeout timeout) {
+                                    if (!timeout.isCancelled()) {
+                                        if (delay < retryDelayMax) {
```
tananaev: I thought the idea was that it will stop retying after it reaches max. Otherwise I don't see any reason to have a max. Also, usually interval is doubled, not incremented.

src/main/java/org/traccar/WebDataHandler.java
edvalley: You were absolutely right. Done!

src/main/java/org/traccar/WebDataHandler.java
edvalley: The only purpose is to ensure that if forwarding will not be retried (also the case if any exception is thrown), `deliveryPendingCurrent` gets decremented. If you can see a different or better approach, it will be very welcomed.

src/main/java/org/traccar/WebDataHandler.java
edvalley: The point is that I don't want positions to be delivered, lets say half an hour, after the problem that caused the failures, has been resolved (at minute 16), and even worst, newer positions to be dropped after the first try, just because the pending limit has been reached by positions that again, are waiting for half an hour to be delivered.
What if instead of a min and a max, we use a configurable retry interval, and only retry if a successful delivery happened while waiting? In this case the only limit will be the configurable value of how much pending deliveries we can have. If this limit is reached, the first delivery to be retried will be dropped (It's maybe not the oldest, but will never be the newest)
Maybe I'm seeing this wrong, but please, tell me what do you think.
Thanks.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Maybe we just need to notify about the error and print how many positions are pending.
Something like:
Position forwarding failed: 10 pending.
Would this be ok?

src/main/java/org/traccar/WebDataHandler.java
edvalley: Oh, my old habit of not reaching column 80.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Ok

src/main/java/org/traccar/WebDataHandler.java
tananaev: Maybe you can use finally block for anything that needs to be handled.

src/main/java/org/traccar/WebDataHandler.java
tananaev: Sounds good to me. We don't need a lot of logging. Also don't add "." at the end.

src/main/java/org/traccar/WebDataHandler.java
tananaev: Standard way to implement something like this is retry X number of times with doubling interval. For example:

5s -> 10s -> 20s -> 40s -> 80s -> ....

src/main/java/org/traccar/WebDataHandler.java
edvalley: Good. Done.

src/main/java/org/traccar/WebDataHandler.java
edvalley: That will be ok, but actually, there's no specific exception I need to handle there. In that case I cannot use a finally block. Do you think I should go that way? Or maybe catch anything, do the job, then re-throw? All of this just to ensure the counter gets decremented when forwarding wont be retried.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Sorry, can you please be more specific?

src/main/java/org/traccar/WebDataHandler.java
edvalley: Ok. Let's do it that way. Do you think we should configure "max retries" or stick to "max delay"?

src/main/java/org/traccar/WebDataHandler.java
tananaev: Sorry, why are you not able to use finally?

src/main/java/org/traccar/WebDataHandler.java
tananaev: Don't use `GlobalTimer.getTimer()`. Use injection for any dependencies.

src/main/java/org/traccar/WebDataHandler.java
tananaev: Max retries make more sense to me.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Oh, I see..., I think I got it. Sorry, coming from C++

src/main/java/org/traccar/WebDataHandler.java
edvalley: Ok, working on that. Thanks.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Sorry again, coming from C++ where we have RAII, I have never used before a try-finally clause (without a catch). Great, now I know it can be done. Thanks.

src/main/java/org/traccar/MainModule.java
```diff
@@ -375,6 +376,12 @@ public static DriverEventHandler provideDriverEventHandler(IdentityManager ident
         return new DriverEventHandler(identityManager);
     }
 
+    @Singleton
+    @Provides
+    public static Timer provideGlobalTimer() {
```
tananaev: Please rename to `provideTimer`.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -39,13 +47,17 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 @ChannelHandler.Sharable
 public class WebDataHandler extends BaseDataHandler {
 
     private static final String KEY_POSITION = "position";
     private static final String KEY_DEVICE = "device";
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(WebDataHandler.class);
```
tananaev: This should be the first line in the class.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +182,92 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
             }
-        }
 
-        if (json) {
-            requestBuilder.async().post(Entity.json(prepareJsonPayload(position)));
-        } else {
-            requestBuilder.async().get();
         }
 
+        new AsyncRequestAndCallback(position);
```
tananaev: It seems weird that constructor calls `send` on its own. I think you should probably call send here instead.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +182,92 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
 
-        String url;
-        if (json) {
-            url = this.url;
-        } else {
-            try {
-                url = formatRequest(position);
-            } catch (UnsupportedEncodingException | JsonProcessingException e) {
-                throw new RuntimeException("Forwarding formatting error", e);
+        class AsyncRequestAndCallback implements InvocationCallback<Response> {
```
tananaev: Why is the class defined inline in a method? I would recommend to move it out to the class scope.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +182,92 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                    requestBuilder.async().get(this);
+                }
+            }
 
-        if (header != null && !header.isEmpty()) {
-            for (String line: header.split("\\r?\\n")) {
-                String[] values = line.split(":", 2);
-                requestBuilder.header(values[0].trim(), values[1].trim());
+            private void retry() {
+                boolean ok = false;
```
tananaev: Please rename to something better that indicates the meaning of this variable. What does `ok` mean? Is it sent? Does it mean no errors? Does it mean accepted? It's not clear.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +182,92 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                try {
+                    if (retryEnabled && deliveryPending.get() <= retryLimit && retries < retryCount) {
+                        Main.getInjector().getInstance(Timer.class).newTimeout(new TimerTask() {
+                            @Override
+                            public void run(Timeout timeout) {
+                                boolean ok = false;
```
tananaev: Same here.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +182,92 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                }
+            }
 
-        if (header != null && !header.isEmpty()) {
-            for (String line: header.split("\\r?\\n")) {
-                String[] values = line.split(":", 2);
-                requestBuilder.header(values[0].trim(), values[1].trim());
+            private void retry() {
+                boolean ok = false;
+                try {
```
tananaev: Why do we need tries everywhere? Where do you expect exceptions?

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -155,32 +182,92 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
+                    if (retryEnabled && deliveryPending.get() <= retryLimit && retries < retryCount) {
+                        Main.getInjector().getInstance(Timer.class).newTimeout(new TimerTask() {
+                            @Override
+                            public void run(Timeout timeout) {
+                                boolean ok = false;
+                                try {
```
tananaev: Same question here?

src/main/java/org/traccar/MainModule.java
edvalley: Ok. I see your point. Now, do you think it's ok to use the class `Timer` for this purpose?

src/main/java/org/traccar/WebDataHandler.java
edvalley: Ok, I agree.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Ok. Let's do it that way.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Would you like to move it before `WebDataHandler`'s constructor or just before `handlePosition` method?

src/main/java/org/traccar/WebDataHandler.java
edvalley: You're right. Fixed.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Fixed.

src/main/java/org/traccar/WebDataHandler.java
edvalley: I'm just ensuring `deliveryPending` gets decremented if anything "unexpected" happens. Do you think we don't really need it on the `retry` or `run` methods?

src/main/java/org/traccar/WebDataHandler.java
edvalley: Please, read above.

src/main/java/org/traccar/MainModule.java
tananaev: I think it makes sense.

src/main/java/org/traccar/WebDataHandler.java
tananaev: I guess it makes sense. Is there a way to have a single block though instead of nesting?

src/main/java/org/traccar/MainModule.java
edvalley: Ok then.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Maybe I'm wrong, but I think the overridden `run` method for the new `TimerTask` object, the one that will run asynchronously, is at a different context, so we are not actually nesting try blocks. Please correct me.
Other way could be to create a new class, override the run method there, then use a new object of that class here. Do you really think we need that kind of stuff here? If you prefer that approach, no problem, I'll code it. Thank you.

tananaev: Correct, but it's just hard to read when code is nested so much. I think the easiest option to clean this up is make your class extend `TimerTask`.

src/main/java/org/traccar/WebDataHandler.java
edvalley: What about this?

tananaev: What about this?

edvalley: Would you like to move `AsyncRequestAndCallback` class implementation before `WebDataHandler`'s constructor, just bellow member variables declaration, or just before `handlePosition` method? Or anywhere else? Thanks.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Interesting. Do you mean to make `AsyncRequestAndCallback` to implement both, `InvocationCallback` and `TimerTask`, and then use the same object (`this`) as the first parameter to `newTimeout`? That should do the job. Thanks.

src/main/java/org/traccar/WebDataHandler.java
tananaev: I would move it above `handlePosition` or below everything.

src/main/java/org/traccar/WebDataHandler.java
tananaev: Yes, `AsyncRequestAndCallback` to implement both.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Ok. Thanks.

src/main/java/org/traccar/WebDataHandler.java
edvalley: Great. Let's do it. Thanks.

tananaev:
Merged, thanks.

edvalley:
Great. Thank you very much.

PR #4412, opened 2019-10-08
Adds JC100 camera alarms

rsvanda:
Adds JC100 camera alarms as described in [document](https://www.dropbox.com/s/gysbq0mgugri01u/20170830%20Communication%20Protocol%20of%20JC100.pdf?dl=0), linked from a [traccar support forum topic](https://www.traccar.org/forums/topic/jc100-image-and-video-protocol/).

The camera communicates using *gt06* protocol but the alerts differ from what has been already implemented in `Gt06ProtocolDecoder` in traccar server.

### DIFF
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
@@ -103,6 +103,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
     public static final int MSG_MULTIMEDIA = 0x21;
     public static final int MSG_BMS_2 = 0x40;
     public static final int MSG_MULTIMEDIA_2 = 0x41;
+    public static final int MSG_ALARM = 0x95;
 
     private static boolean isSupported(int type) {
         return hasGps(type) || hasLbs(type) || hasStatus(type);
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -750,6 +751,39 @@ private Object decodeBasicOther(
...
+                    break;
+                case 0x92:
+                    position.set(Position.KEY_ALARM, Position.ALARM_CORNERING);
+                    break;
+                default:
+                    position.set(Position.KEY_ALARM, Position.ALARM_GENERAL);
+                    break;
+            }
+
+            position.set("alarmValue", buf.readShort());
```
tananaev: What's alarm value?

src/test/java/org/traccar/protocol/Gt06ProtocolDecoderTest.java
```diff
@@ -300,6 +300,26 @@ public void testDecode() throws Exception {
         verifyNull(decoder, binary(
                 "7878058A000688290D0A"));
 
+        verifyAttribute(decoder, binary(
+                "78780c95130a0209321c90000112800d0a"),
+                "alarm", "hardAcceleration");
```
tananaev: Please use constants from `Position`. Also, we probably don't need a test for each single alarm. 1-2 samples would be more than enough.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
rsvanda: If you look in the specification linked above, on pages 23-24, section 5.4.1. they say that the message contains

* Alarm Type (1 byte)
* Alarm Value (N bytes) 

In sections 5.4.1.6 and 5.4.1.7 they say it can contain speed value for over-speed alert. 

I tested all the alerts while driving the car and the "Alarm Value" is always 2 byte number - the mentioned speed for overspeed alert or a number for the rest. I assume those values correspond with alert sensitivity, because the appliance can be set to 3 different levels of sensitivity for acceleration/deceleration/turn (see commands settings on page 43+).

The specification does not say much more about it and is a bit chaotic, probably machine translated from Chinese.

src/test/java/org/traccar/protocol/Gt06ProtocolDecoderTest.java
rsvanda: OK, no prob.

PR #4413, opened 2019-10-11
swagger: Fix type's to avoid deserialize issues.

mujeerhashmi:
Refer to this [post](https://www.traccar.org/forums/topic/does-position-model-needs-correction/) for discussion.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1503,7 +1503,8 @@
                     "type": "number"
                 },
                 "network": {
-                    "type": "string"
+                    "type": "object",
+                    "additionalProperties": true
                 },
                 "attributes": {
                     "type": "object",


swagger.json
```diff
@@ -1503,11 +1503,16 @@
                     "type": "number"
                 },
                 "network": {
-                    "type": "string"
+                    "type": "object",
+                    "additionalProperties": {
+                       "type": "object"
+					}
```
tananaev: Please make sure you use consistent formatting.

swagger.json
```diff
@@ -1503,11 +1503,16 @@
...
                 },
                 "attributes": {
                     "type": "object",
-                    "additionalProperties": true
```
tananaev: This seems to be correct way.

swagger.json
```diff
@@ -1503,11 +1503,16 @@
                     "type": "number"
                 },
                 "network": {
-                    "type": "string"
+                    "type": "object",
+                    "additionalProperties": {
+                       "type": "object"
```
tananaev: I don't this is correct.

swagger.json
mujeerhashmi: With this change, The type for attributes generated is 

```
@SerializedName("attributes")
  private Object attributes = null;
```

swagger.json
mujeerhashmi: With this change, The type for attributes generated is 
```
@SerializedName("attributes")
  private Map<String, Object> attributes = null;
```
I feel this is a better way to read and manage stuff at the client side.

swagger.json
tananaev: Are you sure it works correctly for other languages? Because properties are usually not objects, but primitive values.

mujeerhashmi: I have not checked for other languages. But then I see attributes to be map of string & primitive types.
Ex:
1. GT06N Position attributes
`attributes: {iccid=8991000901255216, distance=0.0, totalDistance=9968.19, motion=false}`
2. Teltonika Position attributes
`attributes: {priority=0.0, sat=4.0, event=0.0, ignition=false, motion=true, rssi=3.0, io200=0.0, gpsStatus=1.0, pdop=1.4000000000000001, hdop=1.4000000000000001, power=0.0, io24=7.0, battery=3.6910000000000003, io68=0.0, operator=40445.0, io199=0.0, io16=1492.0, distance=35.03, totalDistance=1705.4}`

tananaev: That's only valid in Java. It might not be valid for some other language.

mujeerhashmi: I could check for javascript and python, this is how the client code is generated

**Javascript:**

```javascript
if (data.hasOwnProperty('attributes')) {
        obj['attributes'] = ApiClient.convertToType(data['attributes'], {'String': Object});
}

 /**
   * @member {Object.<String, Object>} attributes
   */
  exports.prototype['attributes'] = undefined;
```

**Python:**

```python
swagger_types = {
        'id': 'int',
        'device_id': 'int',
        'protocol': 'str',
        'device_time': 'datetime',
        'fix_time': 'datetime',
        'server_time': 'datetime',
        'outdated': 'bool',
        'valid': 'bool',
        'latitude': 'float',
        'longitude': 'float',
        'altitude': 'float',
        'speed': 'float',
        'course': 'float',
        'address': 'str',
        'accuracy': 'float',
        'network': 'dict(str, object)',
        'attributes': 'dict(str, object)'
    }
```

tananaev: That doesn't seem right. Primitive types are not objects.

mujeerhashmi: Agreed & Updated the commit. But the primitive type is different for each attribute so wanted to make it an object so it can be cast to desired primitive type after fetching at the client side.

tananaev:

Issue #4408, opened 2019-09-30
Feature request: Time-distance Matrix


RafaelMiquelino:
Hello, is there any plan to implement this feature on Traccar? That would be really valuable for fleet managers who needs to define which is the most suitable vehicle to send to a specific location. It could be implemented as a report, taking a list of vehicles and the destination coordinates as input. There are some open API providers based on OSM, I know at least the following that provide an API for this:

- https://openrouteservice.org/dev/#/api-docs/matrix/
- https://valhalla.readthedocs.io/en/latest/api/matrix/api-reference/

RafaelMiquelino:
https://github.com/traccar/traccar/issues/2629

RafaelMiquelino:
LocationIQ also provide this service: https://locationiq.com/docs/#matrix

kirantpatil:
> @kirantpatil, did you have any progress on the implementation of this feature?

@RafaelMiquelino, we haven't done any implementation.

RafaelMiquelino:
> @RafaelMiquelino, we haven’t done any implementation

Ok, @kirantpatil. Thanks!

@tananaev, would you accept a PR for this feature? I’m thinking on adding this as a new report. The output would be something like that:

Device name | Distance to the destination | Time to the destination
--- | ---| ---
Device 1 | 20 km | 0 h 15 m
Device 2 | 22 km | 0 h 35 m
Device 3 | 15 km | 0 h 20 m

The inputs to be set by the user would be Device/Group and Destination coordinates/Address.

If so, let me know if you have any specific requirement.

tananaev:
How would you calculate distance and time?

RafaelMiquelino:
> How would you calculate distance and time?

Hi, you can get it using the API's I described in the beginning of this thread. The destination coordinates/address would be set by the user and I would take the current (or last known) coordinates of the selected vehicles as input parameters. E.g.: follow a screenshot of the output of the Graphhopper API with some dummy data (https://graphhopper.com/api/1/examples/#matrix):

<img width="834" alt="STAI-Matrix" src="https://user-images.githubusercontent.com/31367475/66142330-97ddba00-e5db-11e9-862b-eb6abb226db8.png">

tananaev:
OK, sounds good.

RafaelMiquelino:
Check bellow a sketch of what I intend to implement. Let me know what you think.

[...]

PR #4414, opened 2019-10-12
Report matrix

RafaelMiquelino:
Implemented time-distance matrix report at backend level for [LocationIQ](https://locationiq.com/docs#matrix) and [OpenRouteService](https://openrouteservice.org/dev/#/api-docs/v2/matrix/{profile}/post).

Traccar configuration file `traccar.xml` must be set at least with the following parameters:

```xml
<entry key='matrix.enable'>true</entry>
<entry key='matrix.type'>provider</entry>
<entry key='matrix.key'>api_key</entry> 
```

Implemented provider options are the following:
**- `locationiq`**
**- `openrouteservice`**

API request example: http://localhost:8082/api/reports/matrix?deviceId=16&deviceId=17&deviceId=18&destLat=48.8630487&destLon=2.2871436

API response example: 
```json
[{
    "deviceId": 16,
    "deviceName": "VeÃ­culo-2",
    "devLat": 48.85478,
    "devLon": 2.34223,
    "destLat": 48.8630487,
    "destLon": 2.2871436,
    "time": 336000,
    "distance": 4601.2
}, {
    "deviceId": 17,
    "deviceName": "VeÃ­culo-3",
    "devLat": 48.854265,
    "devLon": 2.3448486666666666,
    "destLat": 48.8630487,
    "destLon": 2.2871436,
    "time": 403000,
    "distance": 5243.7
}, {
    "deviceId": 18,
    "deviceName": "VeÃ­culo-4",
    "devLat": 48.85657033333334,
    "devLon": 2.346719333333333,
    "destLat": 48.8630487,
    "destLon": 2.2871436,
    "time": 443000,
    "distance": 5651.0
}]
```

**Time in ms and distance in m.**


RafaelMiquelino:
@tananaev, this is related to the discussion at https://github.com/traccar/traccar/issues/4408.

tananaev:
A few high level comments:

- Please don't use shortened names like "dev", "dist" etc
- Please use better naming than payload1, 2 etc
- Matrix can mean many things, so we probably need more specific name
- There is some disaster with request formatting
- Result should not have json classes
- If destination is shared, why we need to include in every result?

RafaelMiquelino:
Thanks for the comments. They do make sense. I’m not sure if I got the following two, could you please give more details to what exactly you refer to?
> * There is some disaster with request formatting
> * Result should not have json classes

tananaev:
> There is some disaster with request formatting

For example, why are you formatting using `String.format`? You should use proper JSON conversion.

> Result should not have json classes

`MatrixResponse` should have proper Java data structures (e.g. List), not JSON array objects.

RafaelMiquelino:
> > There is some disaster with request formatting
> 
> For example, why are you formatting using `String.format`? You should use proper JSON conversion.

Yes, sure, this is the best way to do that. However it makes sense only for OpenRouteService API since LocationIQ's does not use JSON as input, it gets a string on a specific formatting style implemented on the URL, so keeping String.format is still the best strategy for this last case in my opinion.

>> Result should not have json classes
>
>MatrixResponse should have proper Java data structures (e.g. List), not JSON array objects.
Ok.

I made available a new version with some improvements since last time. It is still missing the comments bellow but I'll go trough it as soon as I have some time. Let me kwon what you think for now.
>* Please don't use shortened names like "dev", "dist" etc
>* Matrix can mean many things, so we probably need more specific name

tananaev:
It looks like you are manually building JSON instead of using object mapping. I think it needs to be changed. As for formatting query parameters, it's ok to use String.format, but the way you format other stuff seems unnecessary complicated. How about using string join?

RafaelMiquelino:
I was avoiding adding new dependencies to the project but I just realised you already have Jackson Databind installed. I'll change that.

>but the way you format other stuff seems unnecessary complicated.

Could you give an example?

RafaelMiquelino:
> > but the way you format other stuff seems unnecessary complicated.
> 
> Could you give an example?

Let me know if you are talking about the for loop that builds the location coordinates for LocationIQ. If so, I think string join can be a good idea.

tananaev:
Yes, looping through coordinates is what I'm talking about. There has to be a way to make it cleaner and easier with join.

RafaelMiquelino:
Yes, there are some other options. We could use join if we had a list of strings on the coordinates but as we have a 2D array of Doubles I think the best way would be using StringJoiner. We would still keep the loop but in a cleaner and more efficient way then building manually the string with += as before. Something like this, where locationsArray is the 2D array of coordinates:

```java
StringJoiner locationsStringJoiner = new StringJoiner(";");
        for (List<Double> coordinates : locationsArray) {
            StringJoiner locationStringJoiner = new StringJoiner(",");
            for (Double point : coordinates) {
                locationStringJoiner.add(Objects.toString(point));
            }
            locationsStringJoiner.add(locationStringJoiner.toString());
        }
        String locations = locationsStringJoiner.toString();
```

Let me know what you think.

tananaev:
Doesn't really make sense to me. Why have loops when we can do everything in a single line?

RafaelMiquelino:
And how would that be?

RafaelMiquelino:
The only other way that I see would be with Stream but then I would prefer the loop. Something like this:
```java
        String locations = locationsArray
                .stream()
                .map(l -> l.stream()
                        .map(Objects::toString)
                        .collect(Collectors.joining(",")))
                .collect(Collectors.joining(";"));
```
Pure String.join would not be suitable since our array is made of an array of Doubles and not Strings.

tananaev:
This cannot be replaced with one line?
```java
            StringJoiner locationStringJoiner = new StringJoiner(",");
            for (Double point : coordinates) {
                locationStringJoiner.add(Objects.toString(point));
            }
```

RafaelMiquelino:
I don't know if that's a rhetorical question but I don't see a way that would work for the string formatting required by the API.

RafaelMiquelino:
This new push has the following updates:

* Removal of shortened names like "dev", "dist", "devCoord", etc.;
* Class and methods naming improving from "Matrix" to "TimeDistance";
* Cleanup in the result.

Follow bellow the new result pattern:
```json
[
  {
    "deviceId": 16,
    "deviceName": "VeÃ­culo-2",
    "time": 336000,
    "distance": 4601.2
  },
  {
    "deviceId": 17,
    "deviceName": "VeÃ­culo-3",
    "time": 403000,
    "distance": 5243.7
  },
  {
    "deviceId": 18,
    "deviceName": "VeÃ­culo-4",
    "time": 443000,
    "distance": 5651
  }
]
```

The last thing I'm considering doing is the exception handling during the whole process.

src/main/java/org/traccar/Context.java
```diff
@@ -41,6 +41,7 @@
 import org.traccar.database.NotificationManager;
 import org.traccar.database.PermissionsManager;
 import org.traccar.database.UsersManager;
+import org.traccar.directions.timeDistance.TimeDistanceMatrix;
```
tananaev: Incorrect package naming convention. It should be all lowercase. Also, I'm not sure why we need sub-package.

src/main/java/org/traccar/Context.java
```diff
@@ -153,6 +154,10 @@ public static Geocoder getGeocoder() {
         return Main.getInjector() != null ? Main.getInjector().getInstance(Geocoder.class) : null;
     }
 
+    public static TimeDistanceMatrix getMatrix() {
```
tananaev: Method name still says just Matrix. Please fix all places.

src/main/java/org/traccar/MainModule.java
```diff
@@ -201,6 +204,25 @@ public static GeolocationProvider provideGeolocationProvider(Config config) {
...
+            String type = config.getString(Keys.MATRIX_TYPE, "locationiq");
+            String url = config.getString(Keys.MATRIX_URL);
+            String key = config.getString(Keys.MATRIX_KEY);
+
+            switch (type) {
+                case "openrouteservice":
+                    return new OpenRouteServiceTimeDistance(url, key);
+                case "locationiq":
+                default:
+                    return new LocationIQTimeDistance(url, key);
```
tananaev: Incorrect naming convention again. We use camel-case.

src/main/java/org/traccar/api/resource/ReportResource.java
```diff
@@ -43,13 +43,15 @@
 import org.traccar.model.Event;
 import org.traccar.model.Position;
 import org.traccar.reports.Events;
+import org.traccar.reports.Route;
 import org.traccar.reports.Summary;
 import org.traccar.reports.Trips;
+import org.traccar.reports.Stops;
```
tananaev: Please don't re-order existing imports. It causes unnecessary merge conflicts.

src/main/java/org/traccar/directions/timeDistance/JsonTimeDistance.java
```diff
@@ -0,0 +1,61 @@
...
+        JsonObject resultJson = getTimeDistanceResponse(this.url, this.key, timeDistanceRequest);
+
+        TimeDistanceResponse result = new TimeDistanceResponse();
+
+        try {
+            result = JsonTimeDistanceObjectMapper
+                        .getObjectMapper()
+                        .readValue(resultJson.toString(), TimeDistanceResponse.class);
+        } catch (IOException e) {
+            e.printStackTrace();
```
tananaev: Don't do this. Use proper logging.

src/main/java/org/traccar/directions/timeDistance/OpenRouteServiceTimeDistance.java
```diff
@@ -0,0 +1,40 @@
...
+            e.printStackTrace();
+        }
+
+        Entity<String> requestBodyEntity = Entity.json(requestBodyString);
+
+        Response request = Context.getClient().target(url)
+                .request()
+                .header("Authorization", key)
+                .header("Accept",
+                        "application/json, application/geo+json, application/gpx+xml, img/png; charset=utf-8")
```
tananaev: Why do we need all these accept options here?

src/main/java/org/traccar/directions/timeDistance/OpenRouteServiceTimeDistance.java
```diff
@@ -0,0 +1,40 @@
...
+    public JsonObject getTimeDistanceResponse(String url, String key, TimeDistanceRequest timeDistanceRequest) {
+        if (url == null) {
+            url = "https://api.openrouteservice.org/v2/matrix/driving-car";
+        }
+
+        String requestBodyString = null;
+        try {
+            requestBodyString = JsonTimeDistanceObjectMapper.getObjectMapper().writeValueAsString(timeDistanceRequest);
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
```
tananaev: Logging.

src/main/java/org/traccar/directions/timeDistance/TimeDistanceRequest.java
```diff
@@ -0,0 +1,46 @@
+package org.traccar.directions.timeDistance;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class TimeDistanceRequest {
+    private List<List<Double>> locations = new ArrayList<>();
```
tananaev: Why do we need default empty lists?

src/main/java/org/traccar/directions/timeDistance/TimeDistanceRequest.java
```diff
@@ -0,0 +1,46 @@
+package org.traccar.directions.timeDistance;
```
tananaev: Don't forget to add copyright header to each file (except unit tests). You can put your name if you want.

src/main/java/org/traccar/reports/model/TimeDistanceReport.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public double getDistance() {
+        return distance;
+    }
+
+    public void setDistance(double distance) {
+        this.distance = distance;
+    }
+}
+
```
tananaev: Remove extra space.

src/main/java/org/traccar/reports/TimeDistance.java
```diff
@@ -0,0 +1,75 @@
...
+            timeDistanceReport.setTime(timeDistanceResponse.getDuration(deviceIndex) * 1000);
+            result.add(timeDistanceReport);
+            deviceIndex++;
+        }
+
+        return result;
+    }
+
+    public static Collection<TimeDistanceReport> getObjects(long userId, Collection<Long> deviceIds,
+                                                            Collection<Long> groupIds,
```
tananaev: This is not the way we format arguments. Please use consistent formatting with existing code.

src/main/java/org/traccar/reports/TimeDistance.java
```diff
@@ -0,0 +1,75 @@
...
+            deviceLocation.add(Context.getIdentityManager().getLastPosition(deviceId).getLatitude());
+            sourceLocations.add(deviceLocation);
+        }
+
+        TimeDistanceResponse timeDistanceResponse = Context.getMatrix().getTimeDistanceMatrix(sourceLocations,
+                destinationLocation);
+
+        int deviceIndex = 0;
+
+        for (long deviceId: ReportUtils.getDeviceList(deviceIds, groupIds)) {
```
tananaev: I think standard indexed for loop makes more sense here.

src/main/java/org/traccar/directions/timeDistance/TimeDistanceRequest.java
tananaev: Same for all other places.

src/main/java/org/traccar/directions/timeDistance/TimeDistanceMatrix.java
```diff
@@ -0,0 +1,11 @@
+package org.traccar.directions.timeDistance;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public interface TimeDistanceMatrix {
+
+    TimeDistanceResponse getTimeDistanceMatrix(List<List<Double>> sourceLocations,
```
tananaev: Argument formatting.

src/main/java/org/traccar/directions/timeDistance/OpenRouteServiceTimeDistance.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.ws.rs.core.Response;
+
+public class OpenRouteServiceTimeDistance extends JsonTimeDistance {
+    public OpenRouteServiceTimeDistance(String url, String key) {
+        super(url, key);
+    }
+
+    @Override
+    public JsonObject getTimeDistanceResponse(String url, String key, TimeDistanceRequest timeDistanceRequest) {
+        if (url == null) {
```
tananaev: Why not once in constructor?

src/main/java/org/traccar/directions/timeDistance/LocationIQTimeDistance.java
```diff
@@ -0,0 +1,54 @@
...
+
+        String finalUrl = String.format(
+                "%s%s?destinations=%s&sources=%s&annotations=%s&key=%s",
+                url,
+                locations,
+                destinations, sources,
+                metrics,
+                key);
+
+        Invocation.Builder request = Context.getClient().target(finalUrl)
```
tananaev: Why do you need this variable? If you use builder, might as well chain everything directly in return.

src/main/java/org/traccar/directions/timeDistance/LocationIQTimeDistance.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public JsonObject getTimeDistanceResponse(String url, String key, TimeDistanceRequest timeDistanceRequest) {
+        if (url == null) {
+            url = "https://us1.locationiq.com/v1/matrix/driving/";
+        }
+        String metrics = String.join(",", timeDistanceRequest.getMetrics());
+
+        StringJoiner locationsStringJoiner = new StringJoiner(";");
+        for (List<Double> coordinates : timeDistanceRequest.getLocations()) {
+            StringJoiner locationStringJoiner = new StringJoiner(",");
```
tananaev: This still doesn't make sense to me. Why simple `coordinates.join` doesn't work here?

src/main/java/org/traccar/directions/timeDistance/JsonTimeDistanceObjectMapper.java
```diff
@@ -0,0 +1,16 @@
...
+
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public final class JsonTimeDistanceObjectMapper {
+
+    private JsonTimeDistanceObjectMapper() { };
+
+    private static ObjectMapper objectMapper = new ObjectMapper()
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
tananaev: I think it's better to use same mapper for everything. Also, isn't there a way to specify it for the class instead of globally?

tananaev (commented):
Added comments. Quite a bit of small issues and inconsistencies.

src/main/java/org/traccar/directions/timeDistance/LocationIQTimeDistance.java
RafaelMiquelino: String.join only accepts List<String>.

RafaelMiquelino:
Still missing to deal with some comments. Will do that on the next days.

src/main/java/org/traccar/reports/TimeDistance.java
RafaelMiquelino: I need the deviceId, so I'd rather use this enhanced for loop.

src/main/java/org/traccar/directions/timeDistance/OpenRouteServiceTimeDistance.java
RafaelMiquelino: Didn't get this one.

RafaelMiquelino:
Ok. All clear. Just some comments missing better understanding.

src/main/java/org/traccar/reports/TimeDistance.java
tananaev: I understand, but in this case the common practice would be to use index.

src/main/java/org/traccar/directions/timeDistance/OpenRouteServiceTimeDistance.java
tananaev: Why is this code not in constructor?

src/main/java/org/traccar/reports/TimeDistance.java
RafaelMiquelino: Done.

src/main/java/org/traccar/directions/timeDistance/OpenRouteServiceTimeDistance.java
RafaelMiquelino: Done.

RafaelMiquelino:
Ok. Everything reviewed according to the comments.

src/main/java/org/traccar/timedistancematrix/JsonTimeDistance.java
```diff
@@ -0,0 +1,83 @@
...
+        } catch (IOException e) {
+            LOGGER.warn("Time distance response error", e);
+        }
+
+        return result;
+    }
+
+    JsonObject getTimeDistanceResponse(TimeDistanceRequest timeDistanceRequest)
+            throws ClientErrorException {
+        return null;
```
tananaev: What's the point of this implementation?

src/main/java/org/traccar/timedistancematrix/JsonTimeDistance.java
```diff
@@ -0,0 +1,83 @@
...
+        TimeDistanceRequest timeDistanceRequest = new TimeDistanceRequest();
+        timeDistanceRequest.setLocations(locations);
+        timeDistanceRequest.setSources(sourceIndexes);
+        timeDistanceRequest.setDestinations(destinationIndexes);
+        timeDistanceRequest.setMetrics(metrics);
+
+        JsonObject resultJson = null;
+        try {
+            resultJson = getTimeDistanceResponse(timeDistanceRequest);
+        } catch (ClientErrorException e) {
```
tananaev: Why are we catching this error?

src/main/java/org/traccar/timedistancematrix/JsonTimeDistance.java
```diff
@@ -0,0 +1,83 @@
...
+            LOGGER.warn("Time distance network error", e);
+        }
+
+        TimeDistanceResponse result = null;
+        try {
+            result = Context.getObjectMapper().readerFor(TimeDistanceResponse.class)
+                    .without(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
```
tananaev: Why is this not a global configuration?

src/main/java/org/traccar/timedistancematrix/JsonTimeDistance.java
```diff
@@ -0,0 +1,83 @@
...
+        JsonObject resultJson = null;
+        try {
+            resultJson = getTimeDistanceResponse(timeDistanceRequest);
+        } catch (ClientErrorException e) {
+            LOGGER.warn("Time distance network error", e);
+        }
+
+        TimeDistanceResponse result = null;
+        try {
+            result = Context.getObjectMapper().readerFor(TimeDistanceResponse.class)
```
tananaev: Are you expecting all providers to return same format? I feel like that's too optimistic.

tananaev (commented):
Added more comments. Also, please fix copyright notice. New files should not have past dates. Also, If there is only one year, there should be no dash.

gpproton:
It would be nice if this is clicked and route is shown on map, overall this is quite awesome.

Issue #4387, opened 2019-08-22
[Feature request] delete position API endpoint


maximilianloy:
I would like to request a DELETE operation for the positions endpoint with parameters to delete a positions based on id or timespan.

deXetrous:
Hi @maximilianloy I want to work on this if that's fine. Can you please help me to get started and understand the code base?

maximilianloy:
Sorry, I don't know the code base. You need to ask Anton.

Harshit Patel <notifications@github.com> schrieb am Sa., 24. Aug. 2019,
10:45:

deXetrous:
Hi @tananaev  I want to work on this if that's fine. I need your help to understand the code base?

tananaev:
What kind of help do you need?

deXetrous:
I want to work on this feature request, but I am not sure how to proceed?

tananaev:
Study the code, look at other APIs that already implemented.

rovertzxd:
Hello @tananaev I have continued the work that Alexander started for adding the DELETE positions endpoint. my implementation requires all arguments and can only delete positions based on timespan. Please let me know if this is in good shape for merge, or if you have any additional requirements for this feature.

Thank you for your time!

PR #4421, opened 2019-10-20
Implement DELETE api endpoint for Positions

alexjmeech:
Implements a DELETE endpoint on the PositionResource to handle position deletion by ID or timestamp. Addresses feature request #4387

### DIFF
--- a/src/main/java/org/traccar/api/resource/PositionResource.java
+++ b/src/main/java/org/traccar/api/resource/PositionResource.java
@@ -23,6 +23,7 @@ import org.traccar.web.CsvBuilder;
 import org.traccar.web.GpxBuilder;
 
 import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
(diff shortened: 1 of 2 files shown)


tananaev:
If we want to delete a time range, I don't think it's very efficient to delete positions one by one.

alexjmeech:
> If we want to delete a time range, I don't think it's very efficient to delete positions one by one.
Good point. I'll implement a more performant solution and add to this PR.

tananaev:
There's also another problem. What if you delete position that is references from devices?

alexjmeech:
> There's also another problem. What if you delete position that is references from devices?

What do you mean by that?

tananaev:
Positions are references from devices table. What would happen to the reference?

alexjmeech:
> Positions are references from devices table. What would happen to the reference?

I'll work on a fix for that problem as well this weekend and add to this PR

PR #4429, opened 2019-10-29
Allow negative number in cell data (signal strength?)

IanCassTwo:


src/main/java/org/traccar/protocol/H02ProtocolDecoder.java
```diff
@@ -223,7 +223,7 @@ private Position decodeBinary(ByteBuf buf, Channel channel, SocketAddress remote
             .number("(d+),")                     // mnc
             .number("d+,")                       // gsm delay time
             .number("d+,")                       // count
-            .number("((?:d+,d+,d+,)+)")          // cells
+            .number("((?:d+,d+,-?d+,)+)")          // cells
```
tananaev: Please fix the comment alignment.

tananaev:
Merged, thanks.

PR #4434, opened 2019-11-08
Allow the URL of an OpenCellId like service to be set via configuration.

neuharthr:
Added a new type of geolocation named "opencellid-standalone" which allows a user to pass in an OpenCellId like URL via the configuration in order to dictate the end point to use. This flexibility enables users to have an alternative host for this service.

### DIFF
--- a/src/main/java/org/traccar/MainModule.java
+++ b/src/main/java/org/traccar/MainModule.java
@@ -192,6 +192,8 @@ public class MainModule extends AbstractModule {
                     return new GoogleGeolocationProvider(key);
                 case "opencellid":
                     return new OpenCellIdGeolocationProvider(key);
+                case "opencellid-standalone":
+                    return new OpenCellIdGeolocationProvider(url, key);
                 case "unwired":
                     return new UnwiredGeolocationProvider(url, key);
                 default:


tananaev:
Why do we need a new key for it?

neuharthr:
I did not see a way to pass in a different URL via configuration for it. Thus, this seemed to be the best way to allow for the URL to be changed via configuration versus using the hardcoded on inside the constructor of OpenCellIdGeolocationProvider.java.

tananaev:
Why not just use same `opencellid` key?

neuharthr:
It is not the key I need to change, but rather the URL. I have setup my own server with coordinates and I need Traccar to hit that server host. The return results are the same (lat, lon, range) as OpenCellID so that code works as is. The only thing needed is to allow the URL to be configurable via the geolocation.url property.

tananaev:
If I'm not clear, I want type name to the same for default URL and for custom URL.

neuharthr:
I am not sure that I completely understand what you are asking. But, I think you would like to keep the same type (opencellid) and change the parameter so that url is always passed into OpenCellIdGeolocationProvider constructor. Then change the OpenCellIdGeolocationProvider to use the URL if set, else default to what is already in the code (http://opencellid.org/cell/get). Is that correct?

tananaev:
Yes.

PR #4435, opened 2019-11-09
Allow users to set the OpenCellID URL via the config file

neuharthr:
Updated OpenCellId geolocation to utilize the URL parameter from the
configuration file in order to allow for changes to the end point. If
not set, default URL is used for backwards compatibility.


tananaev:
Merged, thanks.

tananaev:
Did a little bit of refactoring to make constructor more readable.

PR #4436, opened 2019-11-09
added language to watch protocol's set_timezone command

flocsy:


src/main/java/org/traccar/StringProtocolEncoder.java
```diff
@@ -33,6 +36,7 @@ protected String formatCommand(Command command, String format, ValueFormatter va
 
         String result = String.format(format, (Object[]) keys);
 
+        Set<String> missingKeys = new HashSet<>(Arrays.asList(keys));
```
tananaev: This feel kind of hacky. How about we refactor everything to just do one `String.format` with values.

For the input would be something like `"LZ,,%s", Command.KEY_TIMEZONE` and then instead of formatting with keys first, we convert keys to value and do a single `String.format` with values.

This would require removing `{}` from all encoders, but feel like better solution.

flocsy: What about making the current one less hacky by saying that there are required parameters: "{%s}" and optional parameters: "{%s?}" This is more flexible, and if we're missing a required key, then at least we'll still see (as befor emy change) the missing key, because we'll have: "foo {missing_key} bar" in the logs.

    protected String formatCommand(Command command, String format, ValueFormatter valueFormatter, String... keys) {

        String result = String.format(format, (Object[]) keys);

        Set<String> missingKeys = new HashSet<>(Arrays.asList(keys));
        result = result.replaceAll("\\{" + Command.KEY_UNIQUE_ID + "}", getUniqueId(command.getDeviceId()));
        for (Map.Entry<String, Object> entry : command.getAttributes().entrySet()) {
            String value = null;
            if (valueFormatter != null) {
                value = valueFormatter.formatValue(entry.getKey(), entry.getValue());
            }
            if (value == null) {
                value = entry.getValue().toString();
            }
            result = result.replaceAll("\\{" + entry.getKey() + "\\??}", value);
            missingKeys.remove(entry.getKey());
        }
        for (String key : missingKeys) {
            result = result.replaceAll("\\{" + key + "\\?}", "");
        }

        return result;
    }

tananaev: This just complicates everything instead of making it simpler.

flocsy: Why complicates? Every encoder stays as it is, and in WatchEncoder we change 1 line:

return formatTextCommand(channel, command, "LZ,{%s?},{%s}", Command.KEY_LANGUAGE, Command.KEY_TIMEZONE);

Clean and easy and more explicit.

flocsy: and one more thing: doing "only" the keys we passed won't work, it breaks other encoders, because in some places you add password or other stuff to the command, so we need to keep the 1st loop as it is.

tananaev: What's an example?

flocsy: MiniFinderProtocolEncoderTest. testEncode
Pt502ProtocolEncoderTest. testEncodeAlarmSpeed
Pt502ProtocolEncoderTest. testEncodeCustom
Pt502ProtocolEncoderTest. testEncodeOutputControl
WatchProtocolEncoderTest. testEncode
WatchProtocolEncoderTest. testEncodeTimezone

tananaev: You should probably elaborate on how password changes anything. I don't see any problem with it.

flocsy: org.junit.ComparisonFailure: expected:<[123456]L+01> but was:<[null]L+01>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.traccar.protocol.MiniFinderProtocolEncoderTest.testEncode(MiniFinderProtocolEncoderTest.java:21)

org.junit.ComparisonFailure: expected:<#SPD[120]
> but was:<#SPD[null]
>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.traccar.protocol.Pt502ProtocolEncoderTest.testEncodeAlarmSpeed(Pt502ProtocolEncoderTest.java:65)

org.junit.ComparisonFailure: expected:<[#PTI300]
> but was:<[null]
>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.traccar.protocol.Pt502ProtocolEncoderTest.testEncodeCustom(Pt502ProtocolEncoderTest.java:21)

org.junit.ComparisonFailure: expected:<#OPC[2,1]
> but was:<#OPC[null,null]
>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.traccar.protocol.Pt502ProtocolEncoderTest.testEncodeOutputControl(Pt502ProtocolEncoderTest.java:36)

tananaev: Removed the commend because it's completely without any context. Also, there is no need for duplication.

flocsy: I replaced the loop to this:

        for (String key : keys) {
            Object value = command.getAttributes().get(key);
            if (valueFormatter != null && value != null) {
                value = valueFormatter.formatValue(key, value);
            }
            result = result.replaceAll("\\{" + key + "}", String.valueOf(value));
        }

what this shows is that the above tests fail if we only search for the keys we passed. That's why I left the loop over the command's arguments.

tananaev: I think we are talking about completely different things. I'm suggesting complete refactoring of the formatting, including ALL encoders.

flocsy: Good idea. Let me know when you finished with the refactoring and I'll send you a new pull request for the WatchProtocol.

tananaev: Done.

flocsy:
here it is

tananaev:
Merged, thanks.

PR #4445, opened 2019-11-19
Update Pattern

lcsvcn:
Add the missing option to update geolocation with D1. Must accept 0 and 1, not just 0!


tananaev:
You should probably test your changes at least once before sending a pull request.

PR #4449, opened 2019-11-25
Xrb28 protocol 

lcsvcn:
From Documentation: https://drive.google.com/file/d/1o7cquL131Jpjs7s8P8YzhL6sBOB4uKoL/view?usp=sharing

Add regex verification to get 0 and 1 from the device response.

It can receive 0 or 1, depending of which position command we send. If it is D1, Time it receives D0,1 with position info. If is D0 it receives D0,0 with position info


lcsvcn:
Follow your advice I have uploaded the Test and the changes. With the documentation which we used to identify the problem and make the change.

src/test/java/org/traccar/protocol/Xrb28ProtocolDecoderTest.java
```diff
@@ -27,6 +27,12 @@ public void testDecode() throws Exception {
 
         verifyPosition(decoder, text(
                 "*SCOR,OM,863158022988725,D0,0,124458.00,A,2237.7514,N,11408.6214,E,6,0.21,151216,10,M,A#"));
+           
+         verifyPosition(decoder, text(
```
tananaev: Wrong indentation. Also, I think two tests are completely unnecessary here. Once should be enough. Also, use a sample from real device, not just manually modified sample from other test.

src/main/java/org/traccar/protocol/Xrb28ProtocolDecoder.java
```diff
@@ -47,7 +47,7 @@ public Xrb28ProtocolDecoder(Protocol protocol) {
             .expression("..,")                   // vendor
             .number("d{15},")                    // imei
             .expression("..,")                   // type
-            .number("0,")                        // reserved
+            .number("[01],")                        // reserved
```
tananaev: Comment indentation is wrong.

lcsvcn:
Hello Anton.

I have review everything that you point.

See if there is anything left.

tananaev:
Can you please test locally what you are pushing? Your code is not even compiling.

lcsvcn:
It was missing the ; at the end.

tananaev:
Merged, thanks.

PR #4454, opened 2019-12-12
Alternative method to Create schema in Oracle DB

anjumsohail:
Schema creation in Oracle with Traccar has always issues.
if Problems persist while creating schema using changelog-xxx.XML with liquibase, this the the alternative method.
1. Modify the detault.xml in at Path traccar/setup/default.xml
Remove the given below entry.
<entry key='database.changelog'>./schema/changelog-master.xml</entry>
2. Create Schema Traccar in Oracle Database (If you change the Name change the above script also)
3. Grant CONNECT,RESOURCE to Traccar schema.
4. Connect with Traccar User.
5. Execute the above script.
6. Run the Traccar.


anjumsohail:
Corrections with Trigger creation scripts

tananaev:
Why are there issues with Liquibase?

anjumsohail:
1. Oracle Do not support Auto Generated Primary Keys, 
2. Oracle do not support Varchar type more 4000 Length.
I experimented a lot but Liquibase do not accommodate these differences.
Now i have Modified the Version of Your Traccar, such that it can Run with Oracle using both thin or thick (OCI) clients.
but problems is that 
How design Traccar solutions Generalized as it is rather then Specialized version for Oracle only.

tananaev:
1. We can disable it for Oracle and use triggers, like you do. It can still be part of Liquibase config.
2. We can limit it to 4000. I thought we already have limited it.

anjumsohail:
May be my expertise with Liquibase config is limited, if same could be done, Plz mentioned but even then same may also need to be accommodated in you Code as well in QueryBuilder.java as given below

Line of code in above class need to check the auto generated keys support first before executing otherwise this will cause exception for after ward.

import java.sql.DatabaseMetaData
...
if (supportsGetGeneratedKeys) {
statement = connection.prepareStatement(parsedQuery, Statement.RETURN_GENERATED_KEYS);
}
else
{
int[] ColIndex={1};
statement = connection.prepareStatement(parsedQuery, ColIndex);
}
see issue link.
https://github.com/traccar/traccar/issues/4456

tananaev:
Well, code contributions are welcome. I'm not sure how your SQL script solves that problem though.

anjumsohail:
Problems using Oracle are 
1. Connection Issue : JDBC unable to Locate appropriate Drivers / Class file / DB installation at expected Path, You can resolve such issues if you have expertise with Oracle.
2. Schema Generation : using Liquibase
3. DML operations in the code that expect primary keys are autogenerated.

The SQL script i have contributed address the 2nd issue, anyone that is an Oracle Expert do not need Liquibase but the Script and disable Auto schema generated using ChangeLog.xml

The Code Contribution is Logic is simple.
First Check AutoKey Generation is supported ? if No then Use Column index for primary key, and in Traccar case Luckly Primary key is always at Index =1

PR #4459, opened 2019-12-29
Provision of Duplicate Database

anjumsohail:
After Configuration changes in debug.xml, now Traccar can simultaneously update two different database (i.e H2 , Sql Server , Oracle) one will be the Master while other will be the slave.

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -18,7 +18,7 @@
     <entry key='database.url'>jdbc:h2:./target/database</entry>
     <entry key='database.user'>sa</entry>
     <entry key='database.password'></entry>
-
+    <entry key='database.Duplicate'>true</entry>
     <entry key='atrack.custom'>true</entry>
 
 </properties>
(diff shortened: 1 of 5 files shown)


anjumsohail:
Any suggestions or improvements, actually this modification is running successfully in my setup without issue.

tananaev:
Formatting is off everywhere. Styling and naming is off. And in general I don't really understand the use case.

anjumsohail:
Well actually in my there was a requirement to update two different databases systems, Replication is one option but restricted to related databases.
Above modification have provision to update to different databases like Oracle, My Sql etc

tananaev:
I understand that it's your requirement, but you need to explain why it would be useful for everyone. I think just using a database cluster or replication makes much more sense, but maybe you know something that I don't.

gpproton:
I personally see a use case for this, first a long time migration plan and then for users without much knowledge of setting up cluster but want to keep almost identical copies of database.

tananaev:
What's the reason for keeping identical copies of the database?

Issue #4393, opened 2019-09-03
Login attempts


gabronick:
failed login attempts are not registering in the log. It would be important to register them with the IP so that through fail2ban they can be blocked.

PR #4460, opened 2020-01-02
Failed Login Attempts logs IP Addresses

DeveloperKurt:
As requested in issue #4393, added logging IP address functionality when login attempts fail.


build.gradle
```diff
@@ -75,6 +75,7 @@ dependencies {
     implementation "com.sun.xml.bind:jaxb-impl:2.3.2"
     implementation "javax.activation:activation:1.1.1"
     testImplementation "junit:junit:4.12"
+    testImplementation group: 'org.springframework', name: 'spring-test', version: '5.2.2.RELEASE'
```
tananaev: I don't think it makes sense to include spring just for a simple unit test.

src/main/java/org/traccar/helper/IpRetriever.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * If the server is behind a reverse proxy, the header value will also
+     * contain the IP's from load balancer and reverse proxy.
+     * This method gets rid of them.
+     *
+     * @param ipAddress IP address value from the header
+     * @return IP address of the client
+     */
+    private static String removeUnwantedData(String ipAddress) {
+        return ipAddress.contains(",") ? ipAddress.split(",")[0] : ipAddress;
```
tananaev: If you just need first value before comma, this can be much more efficient without using split.

src/main/java/org/traccar/helper/IpRetriever.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * If the server is behind a reverse proxy, the header value will also
+     * contain the IP's from load balancer and reverse proxy.
+     * This method gets rid of them.
+     *
+     * @param ipAddress IP address value from the header
+     * @return IP address of the client
+     */
+    private static String removeUnwantedData(String ipAddress) {
```
tananaev: I don't think it make sense to have this one liner as a separate method, especially because it's used in a single place.

src/main/java/org/traccar/helper/IpRetriever.java
```diff
@@ -0,0 +1,50 @@
...
+public final class IpRetriever {
+
+    /**
+     * Retrieves the client's IP address.
+     * Handles the cases like whether the server is behind a proxy/reverse proxy server or a load balancer
+     *
+     * @param request {@link HttpServletRequest} instance
+     * @return client's IP address
+     */
+    public static String retrieveIP(HttpServletRequest request) {
```
tananaev: I would can this `retrieveRemoteAddress` or something like that.

src/main/java/org/traccar/helper/IpRetriever.java
```diff
@@ -0,0 +1,50 @@
+package org.traccar.helper;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * Gets the client's IP address regardless of whether
+ * the server is behind a proxy/reverse proxy server or a load balancer.
+ */
+public final class IpRetriever {
```
tananaev: Maybe `ServletHelper` instead or something more generic?

src/main/java/org/traccar/helper/IpRetriever.java
```diff
@@ -0,0 +1,50 @@
+package org.traccar.helper;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * Gets the client's IP address regardless of whether
```
tananaev: Please remove javadoc comments from everywhere. We only comment code where it's absolutely necessary.

src/test/java/org/traccar/helper/IpRetrieverTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.junit.Test;
+import org.springframework.mock.web.MockHttpServletRequest;
+
+import static org.junit.Assert.assertEquals;
+
+public class IpRetrieverTest {
+
+    private static final String NORMAL_IP_ADDRESS = "231.23.45.65";
+    private static final String GATEWAY_IP_ADDRESS = "147.120.1.5";
+    private static final String IP_ADDRESS_BEHIND_REVERSE_PROXY = "231.23.45.65, 10.20.10.33, 10.20.20.34";
```
tananaev: Doesn't make sense to have constants defined for unit tested values.

build.gradle
DeveloperKurt: If we want to test the edge cases, we need a MockHttpServletRequest that allows us to set custom headers and remote address. 
If you want I can remove the tests overall but I don't see any significant downside of adding the dependency.

src/main/java/org/traccar/helper/IpRetriever.java
DeveloperKurt: Alright, but in my opinion having such code without any comments as well would lead to confusion since it is not obvious why we would even need such an operation in the first place.

DeveloperKurt: What exactly can be more efficient?

src/main/java/org/traccar/helper/IpRetriever.java
DeveloperKurt: Alright, but in my opinion, it would be confusing to have such code without any comments as well since it is not obvious why we need such an operation in the place.

src/main/java/org/traccar/helper/IpRetriever.java
DeveloperKurt: Hmm, yeah, sounds better.

src/main/java/org/traccar/helper/IpRetriever.java
DeveloperKurt: Noted.

src/main/java/org/traccar/helper/IpRetriever.java
DeveloperKurt: Okay but just out of curiosity, why do you apply such policy?

src/test/java/org/traccar/helper/IpRetrieverTest.java
DeveloperKurt: Perhaps I can define them in the tests where they are being used at but using the values without variables would lead to confusion.

build.gradle
tananaev: You don't need a framework to have a mock implementation of a class.

src/main/java/org/traccar/helper/IpRetriever.java
tananaev: `indexOf(",")`

src/main/java/org/traccar/helper/IpRetriever.java
tananaev: You can put a short comment where you use the code if you really want to explain it.

src/main/java/org/traccar/helper/IpRetriever.java
tananaev: My opinion is that code should be self-documenting. Adding comments adds extra maintenance. I think it's pretty general consensus nowadays that if your code requires a lot of comments, you are doing something wrong. One possible exception is frameworks/libraries javadoc where you need to generate API documentations. And even in that case, usually only public methods have comments.

src/test/java/org/traccar/helper/IpRetrieverTest.java
tananaev: I don't really see why it would lead to a confusion. It's pretty obvious what the values are. You can name test cases appropriately to explain what you are testing.

build.gradle
DeveloperKurt: How about creating my own mock implementation as an inner class of the test class?

tananaev: Sounds good to me.

Voppis:
Is there a solution for linux?

tananaev:
Merged, thanks.

PR #4464, opened 2020-01-06
[fix] temperature not parsing on upro protocol

mvalee:
`data.readableBytes() == 6` only works if it is exactly equal to 6.
Modified to check for `data.readableBytes() >= 6`


tananaev:
Please add a unit test for the case.

src/test/java/org/traccar/protocol/UproProtocolDecoderTest.java
```diff
@@ -12,7 +12,12 @@ public void testDecode() throws Exception {
         UproProtocolDecoder decoder = new UproProtocolDecoder(null);
 
         verifyPosition(decoder, buffer(
-                "*HQ201861909268000132,BA&A1820223307024309650492530000311019&B0100000000&F0000&V0036&R0500&J000182&M0052&W00000091&I231026027BD39090827BD5ACA04&X(501E0)(B0000)(E0136)(J01E0)(L3)(k8937204016201240376F)&K00200&T85&N01#"));
+                "*HQ201861909268000132,BA&A1820223307024309650492530000311019&B0100000000&F0000&V0036&R0500&J000182&M0052&W00000091&I231026027BD39090827BD5ACA04&X(501E0)(B0000)(E0136)(J01E0)(L3)(k8937204016201240376F)&K00200&T85&N01#"),
+                );
```
tananaev: I think this is going to fail. Have you tried compiling?

mvalee: Whoops, my bad!

mvalee:
Hmm, seems like this does not solve my issue after all..

PR #4472, opened 2020-01-20
here geocoder url config parameter

jcardus:
Looks like it was hardcoded.

The here url has changed.


jcardus:
checkStyle failing...

tananaev:
What's the new URL? I think we should set at least to be a default.

jcardus:
https://reverse.geocoder.api.here.com/6.2/reversegeocode.json

I put it on the unit test.

My api_key doesn't work with the old url.

By the way I'm having problems trying to fix the unit tests, It has to be run with the Conext initialized because of the "concrete map type error". But we need a settings file for that and I think it's too much for such a small test. Any idea on this?

jcardus:
sorry,
this is the new url:

https://reverse.geocoder.ls.hereapi.com/6.2/reversegeocode.json

src/main/java/org/traccar/geocoder/HereGeocoder.java
```diff
@@ -19,20 +19,24 @@
...
-        url += "&prox=%f,%f,0";
-        url += "&app_id=" + id;
-        url += "&app_code=" + key;
-        if (language != null) {
-            url += "&language=" + language;
-        }
-        return url;
+    private static String formatUrl(String url, String id, String key, String language) {
+      if (url == null) {
+        url = "https://reverse.geocoder.ls.hereapi.com/6.2/reversegeocode.json";
```
tananaev: Please fix indentation here.

src/main/java/org/traccar/geocoder/HereGeocoder.java
```diff
@@ -19,20 +19,24 @@
...
+      if (language != null) {
+          url += "&language=" + language;
+      }
+      return url;
     }
 
-    public HereGeocoder(String id, String key, String language, int cacheSize, AddressFormat addressFormat) {
-        super(formatUrl(id, key, language), cacheSize, addressFormat);
+    public HereGeocoder(String url, String id, String key, String language,
+                        int cacheSize, AddressFormat addressFormat) {
```
tananaev: And here. We don't align parameters with open bracket. See any other classes for formatting examples.

src/test/java/org/traccar/geocoder/GeocoderTest.java
```diff
@@ -73,7 +73,7 @@ public void testBan() {
     @Ignore
     @Test
     public void testHere() {
-        Geocoder geocoder = new HereGeocoder("", "", null, 0, new AddressFormat());
+        Geocoder geocoder = new HereGeocoder("https://reverse.geocoder.ls.hereapi.com/6.2/reversegeocode.json", "", "", null, 0, new AddressFormat());
```
tananaev: URL here can be removed.

src/test/java/org/traccar/geocoder/GeocoderTest.java
```diff
@@ -73,7 +73,7 @@ public void testBan() {
     @Ignore
     @Test
     public void testHere() {
-        Geocoder geocoder = new HereGeocoder("", "", null, 0, new AddressFormat());
+        Geocoder geocoder = new HereGeocoder("", "", "", null, 0, new AddressFormat());
```
tananaev: This doesn't look right. Shouldn't the URL be `null`?

tananaev:
Merged, thanks.

PR #4494, opened 2020-03-01
add support for pushover.net

floetenbaer:
add four new traccar.xml entries:

notificator.pushover.token  // (required) token from pushover.net
notificator.pushover.user // (required) user from pushover.net
notificator.pushover.device // optional: your user's device name to send the message directly to that device, rather than all of the user's devices (multiple devices may be separated by a comma)
notificator.pushover.title // optional: your message's title, otherwise your app's name is used

### DIFF
--- a/src/main/java/org/traccar/notificators/NotificatorPushover.java
+++ b/src/main/java/org/traccar/notificators/NotificatorPushover.java
@@ -54,10 +54,10 @@ public class NotificatorPushover extends Notificator {
         // see https://pushover.net/api
         token = Context.getConfig().getString("notificator.pushover.token"); // (required) token from pushover.net
         user = Context.getConfig().getString("notificator.pushover.user"); // (required) user from pushover.net
-        device = Context.getConfig().getString("notificator.pushover.device",""); // optional:
+        device = Context.getConfig().getString("notificator.pushover.device", ""); // optional:
         // your user's device name to send the message directly
         // to that device, rather than all of the user's devices (multiple devices may be separated by a comma)
-        title = Context.getConfig().getString("notificator.pushover.title",""); // optional: your message's title,
+        title = Context.getConfig().getString("notificator.pushover.title", ""); // optional: your message's title,
         // otherwise your app's name is used
     }
 
--- a/src/main/java/org/traccar/notificators/NotificatorPushover.java
+++ b/src/main/java/org/traccar/notificators/NotificatorPushover.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2019 Anton Tananaev (anton@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


floetenbaer:
comment was too long -> fixed

tananaev:
Can each user configure their own devices?

floetenbaer:
Yes. the ‚User‘ and the ‚key‘ are personalized values. So, the messages the massages are private.

tananaev:
So, how each user can set it up?

floetenbaer:
add the following in traccar.xml: (see https://pushover.net/api)

    <entry key='notificator.types'>web,mail,pushover</entry>
    (required) <entry key='notificator.pushover.token'>xxxx</entry>
    (required) <entry key='notificator.pushover.user'>xxxx</entry>
    (optional) <entry key='notificator.pushover.title'>demo.traccar.org</entry>
    (optional) <entry key='notificator.pushover.device'>myiPhone10, myiMac</entry>

tananaev:
How would regular user access the config file? I think you misunderstood my question.

floetenbaer:
No, the regular user can not access the properties. I did not figure out, how to change the properties in der web ui, like attributeMailSmtpHost. Do you have a documentation for that or a hint for me?

tananaev:
You can look at some existing attributes for example.

floetenbaer:
with pull request in traccar-web user can set the four conf. values, with overrides the traccar.xml settings

tananaev:
I think now you went a little bit overboard with it. My idea was to replicate what we have for firebase. Basically only device token is configured per user. Sorry.

floetenbaer:
No problem, tell me just, what you would accept. We have the pair user, key which identifies a account, normally a family or a small business. A account has one or more devices (normale a mobile phone or a computer), which are owned by a user. Fours we have a title of a message, which can be configured in the account by the account's admin. So I think: user and key goes in the traccar.xml, the device in the users properties and we drop the title. Whats your opinion?

tananaev:
Yes, I think user and key goes into the config. Device(s) is configured per user. As for title, ideally it comes from the template.

floetenbaer:
ok, finished (without title).

tananaev:
It looks very similar to `NotificatorFirebase`. Can we unify the code? Also, it doesn't make sense to use `PropertiesProvider` for single attribute.

floetenbaer:
Ok, done

src/main/java/org/traccar/notificators/NotificatorPushover.java
```diff
@@ -0,0 +1,104 @@
+/*
+ * Copyright 2019 Anton Tananaev (anton@traccar.org)
```
tananaev: Update the date here.

floetenbaer: done

floetenbaer:
but how to revert the submodule?

tananaev:
Just revert it to the original commit.

floetenbaer:
Im back on master branch in traccar-web but there is no change in the main repro to commit. So I don't know what to do :-(

tananaev:
I don't even know how you managed to commit it because it's excluded. You likely have to temporary modify `.gitmodules` file.

floetenbaer:
seems to be ok now :-)

tananaev:
Merged, thanks.

thenib:
Is it possible to add a parameter for pushovers critical alerts?

PR #4495, opened 2020-03-04
Allowing custom content type header in location forward

presidentio:
At the moment it's impossible to forward location as a json to some url with custom `Content Type` header. This is a required to forward location to Openhab using `rest/items/location_raw_mazda/state` endpoint. This changes take content type from the headers defined through `forward.header` parameter,`application/json` used as a default

### DIFF
--- a/src/main/java/org/traccar/WebDataHandler.java
+++ b/src/main/java/org/traccar/WebDataHandler.java
@@ -33,6 +33,8 @@ import org.traccar.model.Position;
 import org.traccar.model.Group;
 
 import javax.inject.Inject;
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.Entity;
@@ -184,6 +186,7 @@ public class WebDataHandler extends BaseDataHandler {
         private int retries = 0;
         private Map<String, Object> payload;
         private Invocation.Builder requestBuilder;
+        private MediaType mediaType = MediaType.APPLICATION_JSON_TYPE;
 
         AsyncRequestAndCallback(Position position) {
 
@@ -198,7 +201,12 @@ public class WebDataHandler extends BaseDataHandler {
             if (header != null && !header.isEmpty()) {
                 for (String line: header.split("\\r?\\n")) {
                     String[] values = line.split(":", 2);
-                    requestBuilder.header(values[0].trim(), values[1].trim());
+                    String headerName = values[0].trim();
+                    String headerValue = values[1].trim();
+                    requestBuilder.header(headerName, headerValue);
+                    if (headerName.equals(HttpHeaders.CONTENT_TYPE)) {
+                        mediaType = MediaType.valueOf(headerValue);
+                    }
                 }
             }
 
@@ -211,7 +219,11 @@ public class WebDataHandler extends BaseDataHandler {
 
         private void send() {
             if (json) {
-                requestBuilder.async().post(Entity.json(payload), this);
+                try {
+                    requestBuilder.async().post(Entity.entity(objectMapper.writeValueAsString(payload), mediaType), this);
+                } catch (JsonProcessingException e) {
+                    throw new RuntimeException("Failed to serialize payload to json: " + payload);
+                }
             } else {
                 requestBuilder.async().get(this);
             }


src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -198,7 +201,12 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
             if (header != null && !header.isEmpty()) {
                 for (String line: header.split("\\r?\\n")) {
                     String[] values = line.split(":", 2);
-                    requestBuilder.header(values[0].trim(), values[1].trim());
+                    String headerName = values[0].trim();
+                    String headerValue = values[1].trim();
+                    requestBuilder.header(headerName, headerValue);
```
tananaev: This should probably be in else statement.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -211,7 +219,11 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
 
         private void send() {
             if (json) {
-                requestBuilder.async().post(Entity.json(payload), this);
+                try {
+                    requestBuilder.async().post(Entity.entity(objectMapper.writeValueAsString(payload), mediaType), this);
+                } catch (JsonProcessingException e) {
+                    throw new RuntimeException("Failed to serialize payload to json: " + payload);
```
tananaev: I don't think we need to append payload here. It's too long if it's the content or doesn't make sense if it's just a reference.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -184,12 +186,13 @@ public String formatRequest(Position position) throws UnsupportedEncodingExcepti
...
         private Map<String, Object> payload;
         private Invocation.Builder requestBuilder;
+        private MediaType mediaType = MediaType.APPLICATION_JSON_TYPE;
 
         AsyncRequestAndCallback(Position position) {
 
             String formattedUrl;
             try {
-                formattedUrl = json ? url : formatRequest(position);
+                formattedUrl = formatRequest(position);
```
tananaev: There is a performance implication of doing `formatRequest`. I don't think it's a good idea to do it for JSON. It should at least be configurable option.

src/main/java/org/traccar/config/Keys.java
```diff
@@ -100,6 +100,13 @@
     public static final ConfigKey FORWARD_JSON = new ConfigKey(
             "forward.json", Boolean.class);
 
+    /**
+     * Boolean value to enable URL parameters in json mode. For example, {uniqueId} for device identifier,
+     * {latitude} and {longitude} for coordinates.
+     */
+    public static final ConfigKey FORWARD_JSON_URL_VARIABLES = new ConfigKey(
+        "forward.json.url.variables", Boolean.class);
```
tananaev: I think we should make this shorter. Something like `forward.urlVariables`.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -67,6 +67,7 @@
     private final String url;
     private final String header;
     private final boolean json;
+    private final boolean jsonUrlVariables;
```
tananaev: Just `urlVariables`.

src/main/java/org/traccar/config/Keys.java
presidentio: I thought about it, but it will mean that this property enables URL variables for both modes (post, get). Is it ok to rename property to `forward.urlVariables` and change logic with the following
`formattedUrl = urlVariables ? url : formatRequest(position);` ?
Notice: it will change logic for existing users

tananaev: No, it shouldn't change non JSON. Static URL with no payload makes no sense and it will break existing users.

presidentio: Done

tananaev:
Merged, thanks.

presidentio:
@tananaev Can I ask you to rebuild docker image and push it to docker registry, as I use it for my deployment?

P.S. thank you for the great product

tananaev:
You would have to ask @magna-z about the docker image. He maintains it.

presidentio:
@magna-z Could you rebuild docker image for traccar and push it to docker registry? I use it for my deployment and I need this fixe there

presidentio:
@tananaev When are you going to release a new version?

tananaev:
Not sure yet.

PR #4504, opened 2020-03-25
Compression on jetty

jcardus:


jcardus:
2 extra files wasn't supposed to be there...

PR #4505, opened 2020-03-25
Jcardus

jcardus:
Compression on jetty


src/main/java/org/traccar/web/WebServer.java
```diff
@@ -97,7 +98,9 @@ protected void handleErrorPage(
         HandlerList handlers = new HandlerList();
         initClientProxy(config, handlers);
         handlers.addHandler(servletHandler);
-        server.setHandler(handlers);
+        GzipHandler gzipHandler = new GzipHandler();
+        gzipHandler.setHandler(handlers);
+        server.setHandler(gzipHandler);
```
tananaev: Can we just use `addHandler` for `GzipHandler`?

src/main/java/org/traccar/web/WebServer.java
```diff
@@ -95,12 +95,13 @@ protected void handleErrorPage(
             }
         });
 
+        GzipHandler gzipHandler = new GzipHandler();
```
tananaev: Any reason you need this variable instead of just inlining it?

tananaev:
Merged, thanks.

PR #4516, opened 2020-04-19
adding jetty log

jcardus:
This needs configuration. Just making the PR to get some guidance.

### DIFF
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -403,13 +403,15 @@ public final class Keys {
             "location.longitudeHemisphere", Boolean.class);
 
     /**
-    * Enable Jetty Request Log.
-    */
+     * Enable Jetty Request Log.
+     */
     public static final ConfigKey WEB_REQUEST_LOG_ENABLE = new ConfigKey(
             "web.requestLog.enable", Boolean.class);
 
     /**
      * Jetty Request Log Path.
+     * The path must include the string "yyyy_mm_dd", which is replaced with the actual date when creating and rolling over the file.
+     * Example: ./logs/jetty-yyyy_mm_dd.request.log
      */
     public static final ConfigKey WEB_REQUEST_LOG_PATH = new ConfigKey(
             "web.requestLog.path", String.class);
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -412,13 +412,13 @@ public final class Keys {
      * Jetty Request Log Path.
      */
     public static final ConfigKey WEB_REQUEST_LOG_PATH = new ConfigKey(
-              "web.requestLog.path", String.class);
+            "web.requestLog.path", String.class);
 
     /**
      * Set the number of days before rotated request log files are deleted.
      */
     public static final ConfigKey WEB_REQUEST_LOG_RETAIN_DAYS = new ConfigKey(
-              "web.requestLog.retainDays", Integer.class);
+            "web.requestLog.retainDays", Integer.class);
 
     private Keys() {
     }
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -402,6 +402,24 @@ public final class Keys {
     public static final ConfigKey LOCATION_LONGITUDE_HEMISPHERE = new ConfigKey(
             "location.longitudeHemisphere", Boolean.class);
 
+    /**
+    * Enable Jetty Request Log.
+    */
+    public static final ConfigKey WEB_REQUEST_LOG_ENABLE = new ConfigKey(
+            "web.requestLog.enable", Boolean.class);
+
+    /**
+     * Jetty Request Log Path.
+     */
+    public static final ConfigKey WEB_REQUEST_LOG_PATH = new ConfigKey(
+              "web.requestLog.path", String.class);
+
+    /**
+     * Set the number of days before rotated request log files are deleted.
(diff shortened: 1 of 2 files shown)


iceberg1369:
I still see no key in config file

src/main/java/org/traccar/web/WebServer.java
```diff
@@ -100,6 +101,16 @@ protected void handleErrorPage(
         handlers.addHandler(servletHandler);
         handlers.addHandler(new GzipHandler());
         server.setHandler(handlers);
+
+        if (config.getBoolean("web.logRequests")) {
```
tananaev, line 105: You have to use configuration keys.

src/main/java/org/traccar/web/WebServer.java
```diff
@@ -100,6 +101,16 @@ protected void handleErrorPage(
...
+          NCSARequestLog requestLog = new NCSARequestLog("./logs/jetty-yyyy_mm_dd.request.log");
```
tananaev, line 106: Log file location has to be configurable.

src/main/java/org/traccar/web/WebServer.java
```diff
@@ -100,6 +101,16 @@ protected void handleErrorPage(
...
+          requestLog.setAppend(true);
+          requestLog.setExtended(true);
+          requestLog.setLogTimeZone("GMT");
+          requestLog.setLogLatency(true);
```
tananaev, line 110: What does this one do?

src/main/java/org/traccar/web/WebServer.java
```diff
@@ -100,6 +101,16 @@ protected void handleErrorPage(
         handlers.addHandler(servletHandler);
         handlers.addHandler(new GzipHandler());
         server.setHandler(handlers);
+
+        if (config.getBoolean("web.logRequests")) {
+          NCSARequestLog requestLog = new NCSARequestLog("./logs/jetty-yyyy_mm_dd.request.log");
+          requestLog.setAppend(true);
+          requestLog.setExtended(true);
+          requestLog.setLogTimeZone("GMT");
```
tananaev, line 109: Why GMT?

src/main/java/org/traccar/web/WebServer.java
```diff
@@ -100,6 +101,16 @@ protected void handleErrorPage(
...
+          requestLog.setLogLatency(true);
+          requestLog.setRetainDays(7);
```
tananaev, line 111: This should also be configurable.

src/main/java/org/traccar/web/WebServer.java
jcardus: ok

src/main/java/org/traccar/web/WebServer.java
jcardus: ok

src/main/java/org/traccar/web/WebServer.java
jcardus: duration or request

src/main/java/org/traccar/web/WebServer.java
jcardus: standard

src/main/java/org/traccar/web/WebServer.java
jcardus: ok

src/main/java/org/traccar/web/WebServer.java
tananaev: Can you please provide a reference.

jcardus: I say this because I never saw a request log that it's not on utc. But we can't take it off and I suppose it will use the servers timezone.

tananaev: Let's remove it then, unless it's actually some kind of standard. I think a lot of servers are configured to be in UTC, so that might explain why you only see UTC.

jcardus: I removed it. No, that's not why, IIS for instance it's utc by default independently of the server timezone.

src/main/java/org/traccar/config/Keys.java
```diff
@@ -402,6 +402,24 @@
...
             "location.longitudeHemisphere", Boolean.class);
 
+    /**
+    * Enable Jetty Request Log.
+    */
+    public static final ConfigKey WEB_REQUEST_LOG_ENABLE = new ConfigKey(
+            "web.requestLog.enable", Boolean.class);
+
+    /**
+     * Jetty Request Log Path.
```
tananaev, line 412: Please provide more documentation here, ideally with example.

tananaev:
Merged, thanks.

tananaev:
@jcardus looks like `NCSARequestLog` is deprecated. Would you be interested in upgrading the code?

jcardus:
Yes, I think it was already deprecated when I implemented it. I didn't find
any good documentation for an alternative...

tananaev:
There is a comment in the Javadoc:
```
@deprecated use {@link CustomRequestLog} given format string {@link CustomRequestLog#EXTENDED_NCSA_FORMAT} with a {@link RequestLogWriter}
```

jcardus:
on it

jcardus:
I upgraded the code,

The pull request goes to an old one I didn't have any response:

https://github.com/traccar/traccar/pull/4576

tananaev:
Please create a separate pull request.

jcardus:
https://github.com/traccar/traccar/pull/4618

PR #4517, opened 2020-04-20
Feature/filter users by id

kgrosvenor:
Creates a way for REST API's using admin tokens on /user endpoint to retrieve one or many users by id and also hooks into existing caching system.

This PR adds a new method to UserManager for validating a set of longs and then some extra logic within UserResource to return the expected result, i am making use of some available methods here too.

Usage for example `http://localhost:8082/api/users?id=1&id=33`

### DIFF
--- a/src/main/java/org/traccar/api/resource/UserResource.java
+++ b/src/main/java/org/traccar/api/resource/UserResource.java
@@ -46,13 +46,20 @@ public class UserResource extends BaseObjectResource<User> {
     }
 
     @GET
-    public Collection<User> get(@QueryParam("userId") long userId) throws SQLException {
+    public Collection<User> get(
+            @QueryParam("userId") long userId,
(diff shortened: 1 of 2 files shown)


tananaev:
I think it makes sense to implement something more generic.

kgrosvenor:
What kind of scope do you think that could have so i can understand, something generic to enrich id's similar to this on all object types, how should it work?

tananaev:
Well, if we implement this functionality, we should implement it for all models. Maybe add something to base classes.

kgrosvenor:
Ok i'll close this PR and open another one at a later date when i've spent some more time on it, i'm getting to grips with the code base at least. 

Does it makes sense to add a method on ManageableObjects interface, then its up to us if we want to override and implement it on each manager?

kgrosvenor:
Ah yep, i see - could have just used getItems straight from the BaseObjectManager

PR #4518, opened 2020-04-20
Simplistic get user by id method with permission checks and strip null devices from response

kgrosvenor:
This should solve what i mean about user filtering and being able to load data on a one to one basis from other back ends with the admin tokens

I also noticed when passing &id=1&id=2 to /api/devices it would return 2 as null as it does not exist, this may as well be removed from the response


tananaev:
First of all, can we split two things. We can merge the bug fix, but the issue with users API is still the same.

kgrosvenor:
I'm basically just after minimal way to load a user by id with admin permissions, I've noticed other mentioning in the forum. Would this not be a good addition? Please explain what is wrong with it so i can improve.

tananaev:
As I mentioned on the forum, it has to be a more generic solution that works not only for users, but for other endpoints as well.

PR #4519, opened 2020-04-20
Ensure item is not null before adding to the collection  in getItems

kgrosvenor:
Noticed that if you use the id parameter and pass an item  that does not exist null items are also returned in the response array when enriching devices. This PR ensures an item is not null before adding into the collection


tananaev:
Merged, thanks for fixing it.

Issue #4524, opened 2020-05-03
Monitoring traccar and process manager


britisharmy:
I am running traccar on a vps and i am concerned about memory usage and what happens when there is no more memory.

I would like to monitor memory usage and also be able to automatically restart if need be. What is the recommended process manager like PM2 but for java environment should one use?.

Also, what monitoring tool would you recommend?.

jcardus:
You should use the forum.

In my case, I start the process with 3G heap size.
 
I'm starting to implement https://github.com/jmxtrans to send monitor data to cloudwatch (aws). 
I can share the configuration scripts when it's done.

Here's my procfile:
web: java -Xms3g -Xmx3g -Dcom.amazonaws.sdk.enableDefaultMetrics -javaagent:lib/jmxtrans-agent-1.2.6.jar=jmxtrans-agent.xml -jar tracker-server.jar conf/traccar.xml

jcardus:
It's implemented. Works like a charm :-)

I made a pull request for it. You will need the jmxtrans agent.

Check the image.

Here's my configuration:

`<jmxtrans-agent>
  <queries>
    
    <query objectName="java.lang:type=OperatingSystem" attribute="SystemLoadAverage" resultAlias="os.systemLoadAverage"/>

    
    <query objectName="java.lang:type=Memory" attribute="HeapMemoryUsage" key="used"
           resultAlias="jvm.heapMemoryUsage.used"/>
    <query objectName="java.lang:type=Memory" attribute="HeapMemoryUsage" key="committed"
           resultAlias="jvm.heapMemoryUsage.committed"/>
    <query objectName="java.lang:type=Memory" attribute="NonHeapMemoryUsage" key="used"
           resultAlias="jvm.nonHeapMemoryUsage.used"/>
    <query objectName="java.lang:type=Memory" attribute="NonHeapMemoryUsage" key="committed"
           resultAlias="jvm.nonHeapMemoryUsage.committed"/>
    <query objectName="java.lang:type=Threading" attribute="ThreadCount" resultAlias="jvm.thread"/>

    
    <query objectName="org.eclipse.jetty.server.session:context=ROOT,type=defaultsessioncache,id=0"
           attribute="sessionsCurrent"
           resultAlias="traccar.sessionsCurrent"/>
    <query objectName="org.eclipse.jetty.server.session:context=ROOT,type=defaultsessioncache,id=0"
           attribute="sessionsMax"
           resultAlias="traccar.sessionsMax"/>
[...]

PR #4525, opened 2020-05-03
adding Jetty JMX and connection statistics

jcardus:


jcardus:
Is this going to be approved? The only reason I need this is to get the number of currently connected users

tananaev:
There is no description why we need it. Obviously it won't be approved like this.

jcardus:
This started in the forum.

jcardus:
#4524

tananaev:
I still want to see proper description. What code does, what information is provided by JMX, what are the security implication... etc.

jcardus:
I'm just enabling a configuration which is already available on jetty configuration files. We have to do it this way because we use the embedded version.

All that it does and all the information jetty puts on JMX is available in the [link](https://www.eclipse.org/jetty/documentation/current/jmx-chapter.html) I used for the documentation.

This allows you to monitor your traccar instance a lot easier. In my case I use a java agent which puts a lot of metrics in amazon cloudwatch for me.

This is not opening any port or any connection, it just makes jetty put jmx data on the local java virtual machine its being ran. How you expose your java virtual machine to allow jmx connection is another story and should have security concerns yes. But that's not what we're doing here.

PR #4532, opened 2020-05-12
Update Freematics API

KalenXI:
Added support for standard PIDs returned by Freematics firmware. Prefixed all other PID attributes to maintain compatibility with JEXL.

### DIFF
--- a/src/main/java/org/traccar/protocol/FreematicsProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/FreematicsProtocolDecoder.java
@@ -134,6 +134,9 @@ public class FreematicsProtocolDecoder extends BaseProtocolDecoder {
                 case 0xF:
                     position.set(Position.KEY_SATELLITES, Integer.parseInt(value));
                     break;
+                case 0x12:
+                    position.set(Position.KEY_HDOP, Integer.parseInt(value));
+                    break;
                 case 0x20:
                     position.set(Position.KEY_ACCELERATION, value);
                     break;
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/protocol/FreematicsProtocolDecoder.java
```diff
@@ -146,8 +146,23 @@ private Object decodePosition(
...
+                    break;
+                case 0x10d:
+                    position.set(Position.KEY_OBD_SPEED, UnitsConverter.knotsFromKph(Integer.parseInt(value)));
+                    break;
+                case 0x111:
+                    position.set(Position.KEY_THROTTLE, Integer.parseInt(value));
+                    break;
                 default:
-                    position.set(data[0], value);
+                    position.set(Position.PREFIX_IN + data[0], value);
```
tananaev, line 165: I think this should be `Position.PREFIX_IO + key` to be consistent with other decoders. Everything else looks good.

KalenXI: Works for me, I've updated the pull request and also added the PID for HDOP.

src/main/java/org/traccar/protocol/FreematicsProtocolDecoder.java
```diff
@@ -146,8 +149,23 @@ private Object decodePosition(
...
+                    break;
+                case 0x10d:
+                    position.set(Position.KEY_OBD_SPEED, UnitsConverter.knotsFromKph(Integer.parseInt(value)));
+                    break;
+                case 0x111:
+                    position.set(Position.KEY_THROTTLE, Integer.parseInt(value));
+                    break;
                 default:
-                    position.set(data[0], value);
+                    position.set(Position.PREFIX_IO + data[0], value);
```
tananaev, line 168: Looks like you forgot to replace `data[0]` with `key`.

KalenXI: I left it as the hex representation because I felt like OBD PIDs are more often referred to by their hex ID rather than the integer equivalent, but if having it as the integer is more consistent with the other decoders I'll change that as well.

tananaev: Yes, it's for consistency.

tananaev:
Merged, thanks.

PR #4534, opened 2020-05-15
Add AutoNavi  Geocoder

0xbw:
AutoNavi is popular in China.
It reverse geocoding is very helpful for the chinese user.
The "formatted_address" contains "state", "city", "street", "building" and more.
So, the Confirgurations required to be added in the config file
"traccar.xml" are
```
    <entry key='geocoder.type'>autonavi</entry>
    <entry key='geocoder.requestLocationReverse'>true</entry>
    <entry key="geocoder.format">%f, %c</entry>
    <entry key='geocoder.url'>http://restapi.amap.com/v3/geocode/regeo</entry>
    <entry key='geocoder.key'>YOUR API KEY</entry>
```

### DIFF
--- a/src/main/java/org/traccar/geocoder/JsonGeocoder.java
+++ b/src/main/java/org/traccar/geocoder/JsonGeocoder.java
@@ -27,6 +27,7 @@ import java.util.AbstractMap;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import org.traccar.config.Keys;
 
 public abstract class JsonGeocoder implements Geocoder {
 
@@ -34,12 +35,14 @@ public abstract class JsonGeocoder implements Geocoder {
 
     private final String url;
     private final AddressFormat addressFormat;
+    private final boolean requestLocationReverse;
 
     private Map<Map.Entry<Double, Double>, String> cache;
 
     public JsonGeocoder(String url, final int cacheSize, AddressFormat addressFormat) {
         this.url = url;
         this.addressFormat = addressFormat;
+        requestLocationReverse = Context.getConfig().getBoolean(Keys.GEOCODER_REQUEST_LOCATION_REVERSE);
         if (cacheSize > 0) {
             this.cache = Collections.synchronizedMap(new LinkedHashMap<Map.Entry<Double, Double>, String>() {
                 @Override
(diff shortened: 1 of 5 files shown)


mac-zhou:
AutoNavi webstie: https://www.autonavi.com
API Documentation: https://lbs.amap.com/api/webservice/guide/api/georegeo

src/main/java/org/traccar/geocoder/JsonGeocoder.java
```diff
@@ -88,7 +88,14 @@ public String getAddress(
             }
         }
 
-        Invocation.Builder request = Context.getClient().target(String.format(url, latitude, longitude)).request();
+        String urlFormat = String.format(url, latitude, longitude);
+        // The AutoNavi web API(V4.2.0)'s location Must Format "longitude,latitude".
+        boolean isAutoNavi = url.contains("restapi.amap.com/v3/geocode/regeo");
+        if (isAutoNavi){
+            urlFormat = String.format(url, longitude, latitude);
```
tananaev, line 95: What is this for?

mac-zhou: traccar's default “location” format is "latitude,longitude", but AutoNavi must be "longitude, latitude".
AutoNavi API Documentation like this:
```
传入内容规则：经度在前，纬度在后，经纬度间以“,”分割，经纬度小数点后不要超过 6 位。
```
translate:
```
Incoming content rules: longitude first, latitude last, and longitude and latitude are separated by ",", no more than 6 digits after the decimal point
```

tananaev: I see, but the solution is not generic. We would need something better.

mac-zhou: Add a  "UrlFormat" class like "AddressFormat"?

tananaev: I think we can either have placeholders like `{latitude}` or have a boolean flag. Formatter class is probably overkill here.

mac-zhou: ok. i will add a boolean flag.  do you hava a better key name?
```
    /**
     * Boolean flag to apply reverse location. 
     * Traccar's default locatin is "latitude,longitude", but some Geolocation Provider is "longitude, latitude".
     * Such as AutoNavi. Default value is false.
     */
    public static final ConfigKey GEOCODER_REQUEST_LOCATION_REVERSE= new ConfigKey(
            "geocoder.requestLocationReverse", Boolean.class);
```

tananaev: I think you misunderstood. It doesn't have to be configurable in the config file. Subclass should just provide this flag because it doesn't change.

tananaev: As for the name, I would prefer something more explicit like `latitudeFirst`.

mac-zhou: > I think you misunderstood. It doesn't have to be configurable in the config file. Subclass should just provide this flag because it doesn't change.

well. i think that would be a big change, involving too much.

tananaev: Why would it be a big change? Seems like a pretty trivial change to me.

PR #4536, opened 2020-06-01
Teltonika protocol - change fuel level from io89 to io84

lukasha12:
Teltonika protocol - change fuel level from io89 (in percents) to io84 (in litres). 
Io84 this is a fuel in litres * 0.1


tananaev:
Merged, thanks.

PR #4539, opened 2020-06-04
Teltonika protocol - change Ecodriving io and movement

lukasha12:
Teltonika protocol:
changed  Ecodriving from io238 to io253.
removed io237.


tananaev:
Merged.

PR #4540, opened 2020-06-05
Kafka notificator

mvniekerk:
Send events to a Kafka stream.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,6 +74,7 @@ dependencies {
     implementation "com.sun.xml.bind:jaxb-core:2.3.0.1"
     implementation "com.sun.xml.bind:jaxb-impl:2.3.2"
     implementation "javax.activation:activation:1.1.1"
+    implementation 'org.apache.kafka:kafka-clients:2.5.0'
     testImplementation "junit:junit:4.13"
 }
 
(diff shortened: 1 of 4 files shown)


tananaev:
Please provide some details on why we need this. Also, why you decided to use notificator option.

mvniekerk:
When viewed as a stream of events to be processed, Kafka is suitable. There are other Apache projects that have Kafka adapters like Flink and Pulsar, which opens more stream processing options post fact.

As for the choice on using notificator - are there other options?

tananaev:
1. Please elaborate on "which opens more stream processing options post fact". A few concrete examples would be helpful.

2. As for notificator, it is designed to send notifications per user. It doesn't really make sense in this case. I think you want something more global. Check for example event forwarding.

3. Why not use an HTTP bridge for Kafka to forward events without integrating Kafka library directly into Traccar?

mvniekerk:
I'll look at the event forwarding.

As for the Kafka lib -
What is more favourable - setting up another server (http bridge) or having the client talk Kafka over the wire?

tananaev:
I think if you set up an HTTP bridge, you won't need to make any changes to Traccar code.

PR #4542, opened 2020-06-09
[teltonika] protocol update

geduxas:
Add CNG readouts.
Add door status readouts.
Fix incorrect obd speed value KPH->KNOTS.


src/main/java/org/traccar/model/Position.java
```diff
@@ -94,6 +95,10 @@
 
     public static final String KEY_DRIVER_UNIQUE_ID = "driverUniqueId";
 
+    public static final String KEY_CNG_STATUS = "cngStatus"; // in use
```
tananaev, line 98: New parameters are added to `Position` class only after very careful consideration and only if at very minimum 2 protocols use it.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -235,6 +235,10 @@ private void decodeOtherParameter(Position position, int id, ByteBuf buf, int le
             case 67:
                 position.set(Position.KEY_BATTERY, readValue(buf, length, false) * 0.001);
                 break;
+            case 68:
```
tananaev, line 238: Are you absolutely sure all the parameters are used by all Teltonika devices? Please provide documentation as a proof.

geduxas: Hello @tananaev, as i know teltonika opened they're information to public sources.. yes there may be compatibility problems with old and very old devices. Here is AVL procols:

https://wiki.teltonika-gps.com/view/FMB640_AVL_ID

For CNG values, i just thought that it could be usable in feature for other devices.. don't know for now about that.

tananaev: The policy is to add only if multiple protocols already need it and it's a generic enough attribute. Only then we add it to a list of known attributes. Also, when you add it there, you should update the web app.

Otherwise I'm ok with adding custom attributes. I only have a problem with door info. That's too many attributes. I would rather keep it as a mask for now. You can always use computed attributes to decode the mask.

geduxas: ok, so do i need to cancel pull request and update ? :)

tananaev: You can update it.

geduxas:
Looks better ?

src/main/java/org/traccar/model/Position.java
```diff
@@ -50,6 +50,7 @@
     // The preferred units of measure are specified in the comment for each.
     public static final String KEY_POWER = "power"; // volts
     public static final String KEY_BATTERY = "battery"; // volts
+    public static final String KEY_BATTERY_CURRENT = "batteryCurrent"; // amps
```
tananaev, line 53: Why are there still changes in the position file?

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -269,7 +273,6 @@ private void decodeOtherParameter(Position position, int id, ByteBuf buf, int le
                 break;
             case 90:
                 position.set(Position.KEY_DOOR, readValue(buf, length, false));
-                break;
```
tananaev, line 272: This would simply break the decoder. Please make sure you test everything thoroughly before asking for final review.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -308,6 +311,18 @@ private void decodeOtherParameter(Position position, int id, ByteBuf buf, int le
             case 199:
                 position.set(Position.KEY_ODOMETER_TRIP, readValue(buf, length, false));
                 break;
+            case 200:
+                position.set("sleepMode", readValue(buf, length, false));
+                break;
+            case 232:
+                position.set("cngStatus", readValue(buf,length,false) == 1); // CNG Status 1=driving on CNG
```
tananaev, line 318: Remove comments for these parameters and add missing spacing in `readValue`. Same for other.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -235,6 +235,10 @@ private void decodeOtherParameter(Position position, int id, ByteBuf buf, int le
             case 67:
                 position.set(Position.KEY_BATTERY, readValue(buf, length, false) * 0.001);
                 break;
+            case 68:
+                 //Battery current in A  * 0.001
```
tananaev, line 239: Remove comment.

geduxas:
Sorry for problems, i have fixed my mistakes, also have tested for couple of days. Seams working ok for now.

src/main/java/org/traccar/model/Position.java
```diff
@@ -142,6 +142,7 @@
     public static final String ALARM_TAMPERING = "tampering";
     public static final String ALARM_REMOVING = "removing";
 
+
```
tananaev, line 145: Looks like there are no changes in this file. Can you please revert this.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -253,7 +256,7 @@ private void decodeOtherParameter(Position position, int id, ByteBuf buf, int le
                 position.set("workMode", readValue(buf, length, false));
                 break;
             case 81:
-                position.set(Position.KEY_OBD_SPEED, readValue(buf, length, false));
+                position.set(Position.KEY_OBD_SPEED, UnitsConverter.knotsFromKph(readValue(buf, length, false)));
```
tananaev, line 259: This is not correct. For OBD speed we actually use kph, as far as I remember.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -325,6 +340,26 @@ private void decodeOtherParameter(Position position, int id, ByteBuf buf, int le
             case 241:
                 position.set(Position.KEY_OPERATOR, readValue(buf, length, false));
                 break;
+            case 246:
+                if (readValue(buf, length, false) == 1) {
+                    position.set(Position.KEY_ALARM, Position.ALARM_TOW);
+                }
+                break;
+            case 247:
+                if (readValue(buf, length, false) >= 1) {
```
tananaev, line 349: Why is this one different from other (`>=` instead of `==`).

iceberg1369: according to docs

    1 – Crash Detected
    2 – Crash Trace Record, (begins 5 sec before crash, and ends 5
      sec after crash

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
geduxas: Yes it is in knots! It was strange to me, after testing for OBD speed i have noticed that it was incorect in web, so diging arround i have found comment to KEY_OBD_SPEED that it is in KNOTS don't know why.. also diging around in other protocols there is some which converts kph to knots to store in DB.. if you could replace whole logic to use kph instead of knots it would be great :)

src/main/java/org/traccar/model/Position.java
geduxas: Removed that.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
tananaev: Can you please provide specific examples.

geduxas: At first I have noticed that in web page obd speed values shows strange numbers, they are not related to any speed sources available in protocol.. just didn't know why it was wrong.. thought that maybe ID is wrong, or readings from CAN incorrect... then I tried to dig where could be problem. So first i have noticed comment:
https://github.com/traccar/traccar/blob/d834c23f8602fd8f3321d8cb8ff9648b163cdafd/src/main/java/org/traccar/model/Position.java#L90

It shows that value is in knots.. ok thought maybe it's only leftovers from something old.. and why it should be knots ?!? So tried to see which devices uses same value, found couple of them, and only some have conversation in knots from kph 
`FreeMatics, GI2000, XT2400` 

So I Just tried to implement the same thing on my testings.. and OBD speed value start to show correct speed. 

Diggig deeper:

https://github.com/traccar/traccar/blob/d834c23f8602fd8f3321d8cb8ff9648b163cdafd/src/main/java/org/traccar/model/Position.java#L249

again -  knots.
 
and from there i went back to teltonika protocol 
https://github.com/traccar/traccar/blob/d834c23f8602fd8f3321d8cb8ff9648b163cdafd/src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java#L509

so it looks like speed value need to be saved in knots.. sorry didn't look deeper..

tananaev: OK, so the "speed" value should be in knots, but "obd speed" should not.

geduxas: It's just works after using that conversion.. 

Look:
https://github.com/traccar/traccar-web/blob/d67213556dcc87453c0ad5cd6f9e8851c477a2f3/web/app/store/PositionAttributes.js#L227-L232

This is from web frontend, obdspeed, is using data type of speed, also same is for speed value so i am 100% Shure that front end uses knots for speed value.

Also here:
https://github.com/traccar/traccar-web/blob/d67213556dcc87453c0ad5cd6f9e8851c477a2f3/web/app/store/SpeedUnits.js#L18-L35

Knots have factor of 1.. so agen all speed values in database is expected to be in knots!

tananaev: That's an issue in the web app.

geduxas: Omg so what should we do? One value is saved in form where web understands.. another value is not like that.. but both treaded as knots.. some protocol decoder's converts kphToKnots to work with frontend.. anothers leaves as is.. it's strange to me that nobody noticed that?

tananaev: I see two options:
- Change everything to use knots.
- Change everything to use km.

For this specific PR, most places in the code use km/h as the unit for OBD speed, correct? If that's true, then you should also use km/h here.

geduxas: For my opinion it should be saved in km/h, knots used only in boats or planes.  It is possible to change everything to km/h but it will be pain to change saved historical values. It's looks that might bee 50/50 using in knots or kph..

tananaev: In this case it was already km/h.

geduxas: So to work in current situation it need's to be in knots. If it's ok then there will be issue in whole project.

tananaev: Once again, "speed" and "OBD speed" are two completely different concepts in Traccar.

geduxas: Yes speed is stored in column, obdspeed stored as attribute.. so what? web front-end still treats it as speed value, and will use converter to show speed. By default it thinks that it stored in knots.

tananaev: I'm not sure what we are arguing about. I already explained options.

geduxas: Ok, web frontend shows speed values, GPS (main speed value shows correct) value taken from OBD show's incorrect. This is an issue which I would like to fix with along others.. so situation is that web interface treat's any speed to be stored in database as knots. In that sittuation only GPS speed is shown in web site correct becouse it is converted to knots, but obd speed is stored as kmph, so in web frontend it will be converted becouse front-end thinks that obd speed is stored in knots.. i know that it is web problem but to solve that issue we need to store obd value in knots!

iceberg1369:
@geduxas is it possible to reopen this PR without obd speed. There are lots of good changes that alreardy are missing

geduxas:
@tananaev will you accept thees small changes? I would like to get readings of that device in your releases.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -227,14 +227,17 @@ private void decodeOtherParameter(Position position, int id, ByteBuf buf, int le
             case 26:
             case 27:
             case 28:
-                position.set(Position.PREFIX_TEMP + (id - 24), readValue(buf, length, true) * 0.1);
+                position.set("bleTemp" + (id - 24), readValue(buf, length, true) * 0.1);
```
tananaev, line 230: I don't agree with this part. Just use standard prefix and different indexes.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -227,14 +227,17 @@ private void decodeOtherParameter(Position position, int id, ByteBuf buf, int le
...
                 break;
             case 66:
                 position.set(Position.KEY_POWER, readValue(buf, length, false) * 0.001);
                 break;
             case 67:
                 position.set(Position.KEY_BATTERY, readValue(buf, length, false) * 0.001);
                 break;
+            case 68:
+                position.set("batteryCurrent", readValue(buf, length,false) * 0.001);
```
tananaev, line 239: Space missing.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
geduxas: Just find out that ID's 25-28 is different in FM*640 devices.
https://wiki.teltonika-gps.com/view/FMM640_AVL_ID

What do you think will be best option for that? Remove BLE temperature implementation? 
It would be great option for teltonika device to have selection of device types, and implement different protocol for each.

tananaev: If it's not consistent between different models, we should remove it.

geduxas: also my changes to CNG, conflict's with FMx640.. so i think i will cancel my PR, and will use my fork till someone figures out how to implement different protocols.

geduxas:
changes conflict's with FM*640 device protocol.

iceberg1369:
@geduxas  you can omit conflicting parts and just push missing parts ( eco driving towing power cut)

geduxas:
PowerCut ID252 will conflict with FM*640 Authorized Driving. 

Sorry for that but it seams not worth to update main code with couple changes. I just want to have features which i am using to get auto-updated from repository.. but i see were will be much more problems, also i still need to make changes to be happy :) So i think i will stick to my forked code.

geduxas:
Also towing, power cut and etc you can implement in "Computed attributes" :)

PR #4544, opened 2020-06-15
Update WatchProtocolDecoder.java

mattti-post:
Implementing "TKQ2" message(analogical to already implemented "TKQ" message) based on documentation for watch protocol:

比电协议V1.8（20160926）.doc

### DIFF
--- a/src/main/java/org/traccar/protocol/WatchProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/WatchProtocolDecoder.java
@@ -262,13 +262,9 @@ public class WatchProtocolDecoder extends BaseProtocolDecoder {
 
             return position;
 
-        } else if (type.equals("TKQ")) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/protocol/WatchProtocolDecoder.java
```diff
@@ -266,6 +266,10 @@ protected Object decode(
 
             sendResponse(channel, id, index, "TKQ");
 
+        } else if (type.equals("TKQ2")) {
```
tananaev, line 269: Can we please combine TKQ and TKQ2 cases.

tananaev:
Merged, thanks.

src/main/java/org/traccar/protocol/WatchProtocolDecoder.java
mattti-post: I commited the combined condition for these cases.

PR #4545, opened 2020-06-15
Update WatchProtocolEncoder.java

mattti-post:
Based on protocol documentation, the implementation of Command.TYPE_POSITION_SINGLE is wrong.

"RG" message is supposed to be issued when requesting location data from BTS (no gps signal)

"CR" message instructs the terminal to wake up gps module immediately and send gps positions in 10 sec intervals during the timespan of 3 minutes.

Communication Protocol.doc
比电协议V1.8（20160926）.doc

https://github.com/tananaev/traccar/files/213814/3g.elec.comm.protocol.docx


tananaev:
Merged, thanks. I hope it doesn't break some other devices.

PR #4548, opened 2020-06-23
Added TOPIN Update Time + Status Response

Fluxpy:
I have added the update time call for the TOPIN protocol. This allows the traccar server to set the correct UTC time of the tracker. The time update package is send right after login and when requested by the tracker. 
The time update response after the login prevents receiving data entries with wrong, non-utc timestamps.
Further I fixed the status response according to the documentation. Only the interval is needed as part of the response.


tananaev:
Please provide documentation.

Fluxpy:
The protocol documentation was taken from the Traccar Protocol Archives ([ZhongXun Topin Locator Communication Protocol.pdf](https://github.com/traccar/traccar/files/4820786/ZhongXun.Topin.Locator.Communication.Protocol.pdf)). The implemented behaviour of the protocol was verified with a Topin ZX303 and compared to the Topin default Server communication.

The "0x30 update time" instruction (e.g. 7878 01 30 0D0A) asks the traccar server for a time update. The server should respond with the current time (e.g. 7878 07 30 07E00705053718 0D0A --> July 5, 2016 5:55:24).

For the status (0x13 status package) response of the server, the Topin protocol specifies only the interval (e.g 7878 02 13 03 0D0A) as contents for the server answer.

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
```diff
@@ -60,6 +63,21 @@ private void sendResponse(Channel channel, int length, int type, ByteBuf content
         }
     }
 
+    private void updateTime(Channel channel, int msgType){
+        ByteBuf dateBuffer = Unpooled.buffer();
+
+        Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+
+        dateBuffer.writeBytes(BigInteger.valueOf(calendar.get(Calendar.YEAR)).toByteArray());
```
tananaev, line 71: Can we use `writeShort` instead of this?

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
```diff
@@ -60,6 +63,21 @@ private void sendResponse(Channel channel, int length, int type, ByteBuf content
         }
     }
 
+    private void updateTime(Channel channel, int msgType){
```
tananaev, line 66: Rename `msgType` to `type`. Please use consistent naming.

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
```diff
@@ -122,13 +150,7 @@ protected Object decode(
             position.set(Position.KEY_VERSION_FW, firmware);
 
             ByteBuf content = Unpooled.buffer();
-            content.writeByte(battery);
```
tananaev, line 125: Why are we removing all this?

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
```diff
@@ -78,6 +96,10 @@ protected Object decode(
             ByteBuf content = Unpooled.buffer();
             content.writeByte(deviceSession != null ? 0x01 : 0x44);
             sendResponse(channel, length, type, content);
+
+            //update time directly after login
+            updateTime(channel, MSG_TIME_UPDATE);
```
tananaev, line 101: This doesn't seem like a requirement from the protocol, so this shouldn't be here.

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
Fluxpy: As stated in the documentation, the TOPIN protocol expects only the heartbeat interval as an answer to the 0x13 status message and not all other parameters.

**0x13 status package**
The device is sent to the server
Start bit 2byte packet length 1byte protocol number 1byte battery power 1byte software version
1byte time zone 1byte status packet upload interval 1byte end bit 2byte
Eg.7878 06 13 55 23 08 03 0D0A
55 means that the battery power is 85%, 23 said the software version number 35 08 that is the
East 8 area (only integer), 03 for the state packet upload interval of 3 minutes
Start bit 2byte packet length 1byte protocol number 1byte status packet upload interval 1byte
end bit 2byte
This data is sent to the server for the server, used to set the status packet upload interval
**Eg.7878 02 13 03 0D0A**
03 to set the heartbeat packet upload interval of 3 minutes to upload once, in minutes

tananaev: Have you checked all other documents?

Fluxpy: As there is only one document for the specific TOPIN protocol and the decoder in traccar is specific for this given protocol and it matches with the TOPIN default server behaviour, there should not be an issue!

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
Fluxpy: It is not in the documentation, but the TOPIN default server does this to ensure that no data with incorrect timestamp is send to the server. This is the reason why this change was added! If the power to the tracker was lost the first few packages will have an incorrect timestamp (non-utc). This leads to incorrect data in the traccar system.

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
Fluxpy: Fixed in commit 6fc7210.

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
Fluxpy: Fixed in commit 6fc7210.

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
```diff
@@ -60,6 +63,21 @@ private void sendResponse(Channel channel, int length, int type, ByteBuf content
         }
     }
 
+    private void updateTime(Channel channel, int type){
```
tananaev, line 66: Just noticed another space missing here.

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
tananaev: OK, let's keep it, but please remove comment.

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
```diff
@@ -122,13 +150,7 @@ protected Object decode(
             position.set(Position.KEY_VERSION_FW, firmware);
 
             ByteBuf content = Unpooled.buffer();
```
tananaev, line 152: If content is not required here, please remove it and allow null value in the `sendResponse` method.

Fluxpy: The content is needed in this place as the interval is written to it and send back to the tracker. It is just not listed in this extracted view. The content is listed in line 153!

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
Fluxpy: Fixed in 1d7d66a.

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
Fluxpy: Fixed in 1d7d66a.

src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
```diff
@@ -60,13 +63,28 @@ private void sendResponse(Channel channel, int length, int type, ByteBuf content
...
+        sendResponse(channel, dateBuffer.readableBytes(), type, dateBuffer);
+    }
+
     @Override
     protected Object decode(
             Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
 
         ByteBuf buf = (ByteBuf) msg;
 
-        buf.skipBytes(2); // header
```
tananaev, line 69: Please leave the comments that were already in the code.

Fluxpy: Done! c990bf7

tananaev:
Merged, thanks.

Fluxpy:
Always a pleasure!

PR #4549, opened 2020-06-25
Add support to custom event fowarding

RodolfoSilva:
Just like #1534 added support for position fowarding, this PR add support for event fowarding withou extra changes to traccar code, so people can plug their own handlers for events.

### DIFF
--- a/src/main/java/org/traccar/Context.java
+++ b/src/main/java/org/traccar/Context.java
@@ -55,9 +55,7 @@ import org.traccar.model.Group;
 import org.traccar.model.Maintenance;
 import org.traccar.model.Notification;
 import org.traccar.model.User;
-import org.traccar.notification.EventForwarder;
 import org.traccar.notification.EventForwarderManager;
-import org.traccar.notification.JsonTypeEventForwarder;
 import org.traccar.notification.NotificatorManager;
 import org.traccar.reports.model.TripsConfig;
 import org.traccar.sms.SmsManager;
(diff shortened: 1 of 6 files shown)


tananaev:
The custom handler, that you are referring to, has been removed, as far as I know. Can you please explain your use case. I feel like if you want something custom, you can have an external script that would take JSON from standard forwarder and convert it to whatever you need.

RodolfoSilva:
The custom handler is still in the code (https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/BasePipelineFactory.java#L160-L171). 

I implemented this handler so I can use protocols other than HTTP to forward events. I'm actually using MQTT, but this handler is generic and could be used with any protocol.

tananaev:
But someone would still have to implement that other handler, right?

RodolfoSilva:
I can also add the protocol specific code for MQTT.

tananaev:
That would make more sense, but I'm also not convinced we need the whole `EventForwarderManager`. Maybe an interface with different implementations is a better option.

RodolfoSilva:
Hey @tananaev, thanks for feedback.

How about I create a new `NotificatorMQTT` to handle events and a `MQTTDataHandler` to handle positions, so we can take advantage of the structure already in place?

And I would also, add the MQTT connection params to the `traccar.xml`.

tananaev:
It's already in place for both positions and events. There's forwarding for both.

J-Paul0815:
@tananaev 
Forwarding the position data via MQTT is exactly what I am looking for. Is there more information on how and what needs to be entered in the configuration? Thanks in advance.

tananaev:
@J-Paul0815, I don't really understand your question. There is no MQTT support for it yet, so what configuration are you talking about?

J-Paul0815:
@tananaev @
Sorry Anton, I must have misunderstood the previous posts. I am looking for a way to get current position data in Iobroker.

PR #4552, opened 2020-06-29
Fix multiple alarms types comparison

andreabreu-me:


tananaev:
What are the cases where we get multiple alarms?

andreabreu-me:
We have custom alarms in production, and some of them have some suffix like "vibrationWithPowerOff". So, when it compares the event "vibration" it will fail because, `"vibrationWithPowerOff".contains("vibration")` will be true, and must be false. In that way we are getting false entries on notificators array.

tananaev:
Are you trying to merge in a change that is only useful for your custom implementation? I don't think we can accept that, sorry.

andreabreu-me:
Not really. If you tomorrow implement some alarm type that suffix any of what you have now, you will get caught too. You are just waiting for a bug if you not fix.

tananaev:
It's not a bug. It's a known limitation. And the fix involves more than this single change.

andreabreu-me:
You are not understanding the issue, maybe I am not explaining it right.

andreabreu-me:
I realised now that you already can have been caught for this bug, because you have `ALARM_LOCK` and `ALARM_UNLOCK`, and the word `LOCK` is a subset of `UNLOCK`

tananaev:
What is the bug?

andreabreu-me:
Create two notifications for:
ALARM_UNLOCK -> SMS,WEB
ALARM_LOCK -> WEB

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -92,9 +92,10 @@ public void updateEvent(Event event, Position position) {
                     if (getById(notificationId).getType().equals(event.getType())) {
                         boolean filter = false;
                         if (event.getType().equals(Event.TYPE_ALARM)) {
-                            String alarms = notification.getString("alarms");
-                            if (alarms == null || !alarms.contains(event.getString(Position.KEY_ALARM))) {
```
tananaev, line 96: What happened to the null check?

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -92,9 +92,10 @@ public void updateEvent(Event event, Position position) {
...
-                                filter = true;
+                            String alarmsStr = notification.getString("alarms");
```
tananaev, line 95: Please rename. We don't use shortened words with some rare historical exceptions.

src/main/java/org/traccar/database/NotificationManager.java
andreabreu-me: The null check is still there `if (alarmsAttribute != null)` if non null get the array of alarm types. Since split will never return null, we are safe

src/main/java/org/traccar/database/NotificationManager.java
andreabreu-me: done

src/main/java/org/traccar/database/NotificationManager.java
tananaev: Yeah, but before the logic was to set `filter = true` when the string is null. You removed that logic and now if it's null, it will always be false. Am I missing something?

andreabreu-me: You are right, we need to check when it is `null` and then set `filter=true`

andreabreu-me: fixed

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -92,9 +92,12 @@ public void updateEvent(Event event, Position position) {
                     if (getById(notificationId).getType().equals(event.getType())) {
                         boolean filter = false;
                         if (event.getType().equals(Event.TYPE_ALARM)) {
-                            String alarms = notification.getString("alarms");
-                            if (alarms == null || !alarms.contains(event.getString(Position.KEY_ALARM))) {
+                            String alarmsAttribute = notification.getString("alarms");
+                            if (alarmsAttribute == null) {
                                 filter = true;
+                            } else {
+                                List<String> alarms = Arrays.asList(alarmsAttribute.split(","));
```
tananaev, line 99: One last thought. Should we use `Set` instead of `Array`?

andreabreu-me: Good question. Since it is redundant in this case, I would prefer the less verbose option here for simplicity. For java 9+ we can use something like `Set<String> alarms = Set.of(alarmsAttribute.split(","));`. What do you think?

tananaev: I think we can keep as is for now.

tananaev:
Merged, thanks.

andreabreu-me:
Thank you too

tananaev:
Have you tried compiling your code?

andreabreu-me:
Fixed, really sorry. I thought that we had some sort of ci/di.

andreabreu-me:
Should I create another pull request?

tananaev:
I already fixed it myself, but I'm just shocked that someone would submit a PR without even trying to compile it.

andreabreu-me:
You should be thankful and not shocked by someone that is trying to help. Unfortunately I created the pull request from Github and did not compiled the code. Mistakes happen everywhere even for experienced developers.

PR #4562, opened 2020-07-19
[TELTONIKA] Fix OBD_SPEED - change kmh to knots

lukasha12:
[TELTONIKA] Fix OBD_SPEED - change kmh to knots.
Teltonika gives OBD_SPEED in km/h, Traccar wants in knots.

### DIFF
--- a/src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
@@ -256,7 +256,7 @@ public class TeltonikaProtocolDecoder extends BaseProtocolDecoder {
                 position.set("workMode", readValue(buf, length, false));
                 break;
             case 81:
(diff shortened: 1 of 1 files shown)


tananaev:
Please see conversations here:

https://github.com/traccar/traccar/pull/4542

lukasha12:
Thanks Anton.  I have read. I think geduksy is right about the obd speed - the frontend expects values ​​in knots.  You can either change the frontend so that it doesn't convert from knots to km / h.  Or you can save data to the database in knots, then converting to km / h will give a good value.  I suggested /1,852 but the solution suggested by  geduksy - UnitsConverter.knotsFromKph (if this function exists in traccar) looks better.
Now it looks that:

tananaev:
As I mentioned there, the correct fix I expect is to use km/h for OBD speed.

lukasha12:
You're right, it will result in less computation.  Unfortunately, I have too little knowledge to make this fix.

PR #4571, opened 2020-09-03
Expanded teltonika device decoder

ldm72:
Based on data from official website: https://wiki.teltonika-gps.com/view/FMB_AVL_ID


tananaev:
This has been discussed multiple times before. Unless your change is valid for ALL Teltonika devices, including legacy ones, we won't be able to accept it.

PR #4572, opened 2020-09-05
Chrome Cross-Site Cookie

jcardus:
Fixing [chrome samesite cookie problem](https://www.traccar.org/forums/topic/chrome-cross-site-cookie/)

This only works on https


tananaev:
Why are there so many commits? Does it fix the issue?

jcardus:
I still had the changes from my last PR about JMX that wasn't approved. Also had sensitive information committed by mistake. So I had to "git reset" to an old commit and then there were some merging issues.

Yes, it fixes the issue.

tananaev:
I've updated configuration a little bit. Can you please check if it still works: https://github.com/traccar/traccar/commit/03bd0f0d0945a80f5af19d06d37ff31a52d294ed.

jcardus:
Yeap, still works.

PR #4573, opened 2020-09-10
Handle Freematics client sending double

callumj:
Found some cases using latest v5 firmware telelogger from Freematics where it was sending `55.00000` as a device temp causing `parseInt` to fail. Since it appears that Traccar supports doubles, switching to parsing in these cases.


tananaev:
Which of the values can be double? For example, it doesn't make sense for satellites to be double.

callumj:
@tananaev I looked at the client more closely and it appears to be a bug on the author's end. It was initialized as a `float` but reads only returned a `int` - I'll patch the client and close this until a fix is needed.

PR #4576, opened 2020-09-28
Log device id on SqlException and log invalid websocket sessions

jcardus:
This helped me debug these two issues:

I'm getting sql deadlock exceptions with some units when the device status is updated at the same time as the last position.

Because of a bug on my front end I was having clients trying to make a websocket connection with an invalid session.

log


jcardus:
I updated the code for the

PR #4582, opened 2020-10-14
Fix gl200 deviceTime parsing

### LINKED ISSUE #4577: PatternBuilder.any() consumes too much input if next token is optional
**Describe the bug**

Whenever `PatternBuilder.any()` (or the equivalent variations such as `.expression("(.*)")`) is followed by optional content, the greedy nature of `.*` consumes too much input and causes the next field to be parsed as absent.

For example, messages parsed with `Gl200TextProtocolDecoder.PATTERN_BASIC` wrongly skip `hdop` and `deviceTime`.  Since `hdop` influences position validity, that is affected also.

A possible solution is to replace `.*` and variants with a reluctant quantifier `.*?`, which will consume as little input as possible.

**To Reproduce**

Feed traccar with a GL200 GTSTT text message.  `hdop` will be absent, and `deviceTime` will not reflect device time in message text.

**Expected behavior**

Traccar should correctly parse `hdop` and `deviceTime`.

**Additional context**

Device: Queclink GV55
Protocol: GL200

rogeriomgatto:
Fixes #4577 for GL200 protocol.  Other uses of `any` or similar constructs would have to be reviewed and changed separately.

* added `readDelimited` and `skipDelimited` methods to `PatternBuilder` as safer alternatives do `any` (capturing and non-capturing version)
* updated patterns in `Gl200TextProtocolDecoder` to use `readDelimited` and `skipDelimited` where applicable
* updated `Gl200TextProtocolDecoder` to decode deviceTime for all parsed positions
* added `verify*` overloads do `ProtocolTest` to allow checking multiple parameter values at once
* updated `Gl200TextProtocolDecoderTest` to check `deviceTime` of all messages


tananaev:
OK, I think this needs to be broken down into separate PRs. Also, I'm not sure it makes sense to add so many new APIs. Happy to discuss options before you make any changes.

rogeriomgatto:
Ok, let's go through the changes and I'll adjust/split the PR accordingly.

Modifying `PatternBuilder.any()` would impact all code that uses it, and I considered it too risky.  The new methods are just a way to abstract constructs like `.expression("[^,]*,")` or `.expression("([^,]*),")` and to make intention clearer (read or skip).  I can live without them and revert to `.expression`, replacing instances of `.any` with an appropriate expression.

In `Gl200TextProtocolDecoder`, there were three types of changes:
* Changes in patterns, to fix the original issue and use the new `readDelimited`/`skipDelimited` methods.  Some of those changes could be reverted, and some could be replaced with modified expressions.
* Because of the changes in the patterns, some usages of `.split()` started to fail.  The new pattern might return `null` in some cases, so I used `Optional` to deal with that.  Also, there were cases that only worked because the pattern captured extra content. Content like `".....xxx,,,20201015151732"` became `".....xxx,,"`, and `split` by default omits empty trailing segments.  Calling `.split(-1)` fixed that case.
* Some decoding paths didn't try to read deviceTime at all.  I added calls to `parser.nextDateTime()` or `decodeDeviceTime` as required.  I also added a variation of `decodeDeviceTime` that can handle multiple positions (for messages that parse into multiple positions).

Finally, the tests. I wanted to check all those messages, so I added variations of `verify*` that could check for deviceTime, and modified each one of them.  That helped me find all cases where deviceTime wasn't being parsed or set.  Now that I know, I could replace all those changes with a minimal set of `verify` calls using existing methods.

tananaev:
Let's start with the first part. Have you tried changing `any`? Do unit tests pass? If not, maybe we can have a boolean parameters.

rogeriomgatto:
> Let's start with the first part. Have you tried changing `any`? Do unit tests pass? If not, maybe we can have a boolean parameters.

Changing `any` to use a reluctant quantifier caused a single failure in `FifotrackProtocolDecoderTest`:

```
java.lang.NumberFormatException: For input string: "190220085833"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:652)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at org.traccar.helper.Parser.nextInt(Parser.java:66)
	at org.traccar.protocol.FifotrackProtocolDecoder.decode(FifotrackProtocolDecoder.java:221)
	at org.traccar.ProtocolTest.verifyNull(ProtocolTest.java:101)
	at org.traccar.protocol.FifotrackProtocolDecoderTest.testDecode(FifotrackProtocolDecoderTest.java:28)
```

tananaev:
Just a single failed tests? I think we can probably fix that protocol instead of adding new API then.

rogeriomgatto:
I instrumented Parser to save its results, so I could compare match groups before and after the change in `any` (see files in [parser_dumps.zip](https://github.com/traccar/traccar/files/5402391/parser_dumps.zip) with a visual diff tool like meld).  The following protocols were affected:

`Tt8850ProtocolDecoderTest`: changes seem OK, a field that was being skipped is now correctly parsed.
`Gl200TextProtocolDecoderTest`: almost all changes were OK, but the message in line 49 now skips network fields.
`FifotrackProtocolDecoderTest`: message in line 28 fails, and the next test is skipped because the protocol class is stateful (does that work?)
`VisiontekProtocolDecoderTest`: messages in lines 13 and 22 wrongly match some "V" in the middle instead of the final ",A".
`XexunProtocolDecoderTest`: changes seem OK, only the final digit of the first field was captured before.

With the exception of that test in Fifotrack that throws an exception, none of the changes were detected by the test cases because they don't check every parsed attribute.

How do you want to proceed? Should I start a new PR from scratch, and we work through each of those issues?

tananaev:
Thanks for doing all that work. I think it would be nice to add test cases for all those scenarios. And then fix each protocol and migrate to updated `any` version.

I expect that the won't be too many changes, so we can split it into 2 PR:

1. Add test cases without changing any decoders
2. Fix decoders and update parser

What do you think?

rogeriomgatto:
Ok, I'll add tests that cover the failing cases to `Gl200TextProtocolDecoderTest`, `FifotrackProtocolDecoderTest`, and `VisiontekProtocolDecoderTest`.  Those tests will pass, because `any` will be modified in a different PR.

I'll not add tests to `Tt8850ProtocolDecoderTest` or `XexunProtocolDecoderTest` in the first PR, because those would fail with current implementation of `any`.  See you on next PRs, feel free to reject this one.

tananaev:
Thanks. Closing this one.

PR #4584, opened 2020-10-18
Anonymous types are replaced with lambda.

kamaci:


src/main/java/org/traccar/sms/smpp/SmppClient.java
```diff
@@ -44,6 +31,17 @@
 import com.cloudhopper.smpp.type.SmppChannelException;
 import com.cloudhopper.smpp.type.SmppTimeoutException;
 import com.cloudhopper.smpp.type.UnrecoverablePduException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.traccar.Context;
+import org.traccar.notification.MessageException;
+import org.traccar.sms.SmsManager;
+
+import java.util.concurrent.ExecutorService;
```
tananaev, line 40: Please don't rearrange imports. It might cause merge conflicts later on.

src/main/java/org/traccar/web/CsvBuilder.java
```diff
@@ -16,23 +16,22 @@
  */
 package org.traccar.web;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
```
tananaev, line 19: Same here.

src/main/java/org/traccar/sms/smpp/SmppClient.java
kamaci: Fixed it!

src/main/java/org/traccar/web/CsvBuilder.java
kamaci: Fixed it!

tananaev:
Merged, thanks.

PR #4586, opened 2020-10-18
BigDecimal.valueOf(double) is used because of floating point imprecision.

kamaci:


tananaev:
What is the difference. Can you please provide details.

kamaci:
> What is the difference. Can you please provide details.

Sure! It's been mentioned at [Javadocs](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#BigDecimal-double-):

> The results of this constructor can be somewhat unpredictable. One might assume that writing new BigDecimal(0.1) in Java creates a BigDecimal which is exactly equal to 0.1 (an unscaled value of 1, with a scale of 1), but it is actually equal to 0.1000000000000000055511151231257827021181583404541015625. This is because 0.1 cannot be represented exactly as a double (or, for that matter, as a binary fraction of any finite length). Thus, the value that is being passed in to the constructor is not exactly equal to 0.1, appearances notwithstanding.

tananaev:
How is `valueOf` different from that?

kamaci:
`valueOf ` translates a double into a BigDecimal, using the double's canonical string representation provided by the Double.toString(double) method. 

It is due to the fact that the `String` constructor is perfectly predictable: writing `new BigDecimal("0.1")` creates a `BigDecimal` which is `exactly equal to 0.1`, as one would expect. Therefore, it is generally recommended that the `String` constructor be used in preference to this one.

tananaev:
Merged, thanks for explanation.

PR #4587, opened 2020-10-20
test case improvements

rogeriomgatto:
Introduces test assertions in preparation for the refactoring of `PatternBuilder.any()`


src/test/java/org/traccar/protocol/FifotrackProtocolDecoderTest.java
```diff
@@ -25,9 +34,35 @@ public void testDecode() throws Exception {
...
 
+        byte[] expectedImage = new byte[13152];
+        new Random().nextBytes(expectedImage);
+        int imageSize = expectedImage.length;
+        int chunkSize = 1024;
+
         verifyNull(decoder, buffer(
                 "$$79,868345037864709,382,D05,190220085833,22.643210,114.018176,1,1,1,13152,23FFD339*25"));
 
+        for (int i = 0; i < imageSize; i += chunkSize) {
```
tananaev, line 45: Can we just test that we get the image attribute? We also probably don't need such a big buffer. We can maybe have 2-3 and then hardcode everything.

tananaev: And also we should probably separate it into its own test case.

src/test/java/org/traccar/protocol/Gl200TextProtocolDecoderTest.java
```diff
@@ -48,6 +52,18 @@ public void testDecode() throws Exception {
 
         verifyAttributes(decoder, buffer(
                 "+BUFF:GTSTC,410301,864802030022424,,,0,,,,,,,0228,0002,4EE8,1BFF489,00,20181207134332,EC90$"));
+        
+        Position position = (Position) decoder.decode(null, null, buffer(
```
tananaev, line 56: For this test case I think it makes sense to add a new testing method. Something like `verifyNetwork`. I think we can start with an option that takes a single `CellTower`. That's the most common case.

src/test/java/org/traccar/protocol/FifotrackProtocolDecoderTest.java
rogeriomgatto: Something like this?

```
verifyNull(... "photo start")
verifyNull(... "large hardcoded string")
Position position = decoder.decode(... "final string")
assertNotNull(position)
assertNotNull(position.getAttributes().get(Position.KEY_IMAGE)
```

tananaev: Something like that. It would be ideal if we can do
```
verifyNull(... "photo start")
verifyNull(... "large hardcoded string")
verifyAttribute(... "final string", ...)
```

PR #4589, opened 2020-10-21
Add 2FA using TOTP

mivtdole:
This update will add 2FA using TOTP to traccar

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,6 +76,7 @@ dependencies {
     implementation "com.sun.xml.bind:jaxb-core:2.3.0.1"
     implementation "com.sun.xml.bind:jaxb-impl:2.3.3"
     implementation "javax.activation:activation:1.1.1"
+    implementation 'com.warrenstrange:googleauth:1.1.2'
     testImplementation "junit:junit:4.13.1"
 }
 
--- a/schema/changelog-4.4.xml
+++ b/schema/changelog-4.4.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd"
+  logicalFilePath="changelog-4.4">
+
+  <changeSet author="author" id="changelog-4.4-all">
+    <addColumn tableName="tc_users">
+        <column name="totpKey" type="varchar(128)"/>
+    </addColumn>
+    <addColumn tableName="tc_users">
+        <column name="useTotp" type="BOOLEAN" defaultValueBoolean="false"/>
+    </addColumn>
+  </changeSet>
+
+</databaseChangeLog>
--- a/src/main/java/org/traccar/model/User.java
+++ b/src/main/java/org/traccar/model/User.java
@@ -20,8 +20,12 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import org.traccar.database.QueryExtended;
 import org.traccar.database.QueryIgnore;
 import org.traccar.helper.Hashing;
+import org.traccar.Context;
 
 import java.util.Date;
+
+import com.warrenstrange.googleauth.GoogleAuthenticator;
+import com.warrenstrange.googleauth.GoogleAuthenticatorException;
 
 public class User extends ExtendedModel {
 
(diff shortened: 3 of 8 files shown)


build.gradle
```diff
@@ -76,6 +76,7 @@ dependencies {
     implementation "com.sun.xml.bind:jaxb-core:2.3.0.1"
     implementation "com.sun.xml.bind:jaxb-impl:2.3.3"
     implementation "javax.activation:activation:1.1.1"
+    implementation 'com.warrenstrange:googleauth:1.1.2'
```
tananaev, line 79: Have you done any analysis of available libraries? It looks like this one hasn't been updated for a while. Also, have you considered embedding implementation? It looks like it's a really simple algorithm that can be implemented easily on our side.

schema/changelog-4.4.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd"
+  logicalFilePath="changelog-4.4">
+
+  <changeSet author="author" id="changelog-4.4-all">
+    <addColumn tableName="tc_users">
```
tananaev, line 10: I don't remember if we discussed this, but can we use attributes for this instead?

src/main/java/org/traccar/model/User.java
```diff
@@ -273,4 +277,64 @@ public boolean isPasswordValid(String password) {
...
+    public boolean getUseTotp() {
+        if (!Context.getConfig().getBoolean("totp.enabled")) {
+            return false;
+        }
+        return this.useTotp;
+    }
+
+    @JsonIgnore
+    @QueryIgnore
+    public boolean isTotpAuthCodeValid(String code) {
```
tananaev, line 313: It would be nice to move most of the validation code our of this class.

schema/changelog-4.4.xml
mivtdole: This was not discussed before yet.  
The enable totp could be an attribute. 
The totp key field is server side generated. And there needs to be an option to clear/regenerate it. I do not know if this is possible with the attributes. If so do you have an attribute which I can use as an example?

src/main/java/org/traccar/model/User.java
mivtdole: I can look if this can be moved to a new class.

build.gradle
mivtdole: I did not consider it as back when I needed it I needed it fast. And using existing code was faster then writing it myself.

mivtdole:
Some extra info on how to use:

In your traccar.xml file you need to add two extra entries:
    <entry key="totp.enabled">true</entry>
    <entry key="totp.enforce">false</entry>

With the "totp.enabled" all the totp fields in the UI become visible. Also you can enable the use of totp for each user/account. A totp key will be generated once the new or updated account is saved. When you reopen you can see the totp key. This is because it is server side generated.

When "totp.enforce" is "false" you can use a mix in accounts which have totp enabled and which not. When you set this to "true" all accounts require to use a totp code for login. So if you did not enable totp on an account you can no longer use to login with this account until totp is enabled on it. 

Procedure to enable on an new system is to first set "totp.enabled" to true and "totp.enforce" to false. Once everything is setup and tested you can set "totp.enforce" to true.

Following are the updates to the UI:
login_and_totp

user_and_totp

schema/changelog-4.4.xml
Abyss777: I believe using attribute in this case is similar to NotificatorFirebase.

mivtdole: I will have a look at this class and report back

build.gradle
tananaev: OK, can you please investigate this.

schema/changelog-4.4.xml
tananaev: I think it's nice to use attributes here because it's optional functionality.

mivtdole: What to do with the situation where the admin needs to enforce the use of it so it is not optional for a user? In our case the user should not be able to change it only the admin.

tananaev: I don't see any case where admin would need to enforce it on user by user basis.

iceberg1369: @tananaev no update on this?

tananaev: Waiting for response, as you can see.

adadnc:
This is a very useful PR. Any news on this, maybe it can make it to the next version?

adadnc:
Anything missing on the PR, can it make into next release?

tananaev:
Comments are not addressed.

adadnc:
> Comments are not addressed.
Maybe with a direct mention? @mivtdole

mivtdole:
As I have little time to work on this there was no update. Second @tananaev requested two changes:

1. Use attributes to store the settings. Which should not be to complicated but there is one open issue as mentioned above where, in our case, the admin enforces to use 2fa for every user.
2. Currently I am using a module/library from another party for the TOTP implementation but @tananaev requested to write the TOTP functionality from scratch. Which is odd as there are multiple libraries from third parties used. But if this is a requirement by @tananaev then this will take more time to develop. But if someone else would like to create this TOTP functionality from scratch you are welcome.

mivtdole:
@SargonofAssyria yes see top of this merge request or https://github.com/mivtdole/traccar/tree/totp

tananaev:
Closing for now as it seems like the code is outdated.

zeeshanhameed-projectmarker:
@tananaev Is there any plan to update the code and merge this PR?

tananaev:
No near term plans. Don't really have time to work on this.

vanderlpp:
UP

any intention of rechecking this code for inclusion?

tananaev:
I believe we already have TOTP support.

PR #4592, opened 2020-10-22
Preserve existing traccar.xml on update

mivtdole:
When you update an existing traccar installation with the standard installer it will overwrite an existing "traccar.xml" config file. 
As this file is used for local changes to your installation you currently have to manually copy it aside before running the update. And restore it after the update.

This change will preserve the existing "traccar.xml" when it finds one and stores the new "traccar.xml" template from the installer as "traccar.xml.new".

### DIFF
--- a/setup/setup.sh
+++ b/setup/setup.sh
@@ -1,8 +1,20 @@
 #!/bin/sh
 
+PRESERVECONFIG=0
+if [ -f /opt/traccar/conf/traccar.xml ]
+then
+    cp /opt/traccar/conf/traccar.xml /opt/traccar/conf/traccar.xml.saved
+    PRESERVECONFIG=1
+fi
+
 mkdir -p /opt/traccar
 cp -r * /opt/traccar
 chmod -R go+rX /opt/traccar
+
+if [ ${PRESERVECONFIG} -eq 1 ] && [ -f /opt/traccar/conf/traccar.xml.saved ]
+then
+    mv -f /opt/traccar/conf/traccar.xml.saved /opt/traccar/conf/traccar.xml
+fi
 
 mv /opt/traccar/traccar.service /etc/systemd/system
 chmod 664 /etc/systemd/system/traccar.service


setup/setup.sh
```diff
@@ -1,9 +1,22 @@
 #!/bin/sh
 
+PRESERVECONFIG=0
+if [ -f /opt/traccar/conf/traccar.xml ]
```
tananaev, line 4: Are you sure generic shell supports all the expressions that you are using?

mivtdole: I am pretty sure any Linux flavour or shell supports the same ["Evaluate conditional expression" test](http://linuxcommand.org/lc3_man_pages/testh.html).

As test platforms I have Ubuntu which uses [dash](https://linux.die.net/man/1/dash) and Fedora which uses [bash](https://linux.die.net/man/1/bash) as the default shell. Both support this. 

I also tested with [Korn Shell (ksh)](https://linux.die.net/man/1/ksh) which is default on OpenBSD and also supports it.

tananaev: Sounds good.

setup/setup.sh
```diff
@@ -1,9 +1,22 @@
...
+    PRESERVECONFIG=1
+fi
+
 mkdir -p /opt/traccar
 cp -r * /opt/traccar
 chmod -R go+rX /opt/traccar
 
+if [ ${PRESERVECONFIG} -eq 1 ] && [ -f /opt/traccar/conf/traccar.xml.saved ]
+then
+    mv /opt/traccar/conf/traccar.xml /opt/traccar/conf/traccar.xml.new
```
tananaev, line 16: I think we don't need to save the new file. It's unlikely to change, but can cause confusion for some users.

tananaev (commented):
Added one more comment. Otherwise looks good.

setup/setup.sh
adadnc: Same for [zsh](http://zsh.sourceforge.net/Doc/Release/Conditional-Expressions.html)

setup/setup.sh
mivtdole: Removed this move to save new template

setup/setup.sh
```diff
@@ -1,9 +1,21 @@
...
+    PRESERVECONFIG=1
+fi
+
 mkdir -p /opt/traccar
 cp -r * /opt/traccar
 chmod -R go+rX /opt/traccar
 
+if [ ${PRESERVECONFIG} -eq 1 ] && [ -f /opt/traccar/conf/traccar.xml.saved ]
+then
+    mv /opt/traccar/conf/traccar.xml.saved /opt/traccar/conf/traccar.xml
```
tananaev, line 16: One other thing I just thought about. What if `mv` asks for confirmation before replacing the file? Is it possible?

adadnc: Many distributions have aliases, but the correct switch would be "-f"

mivtdole: According to man page of mv:

`-f, --force
              do not prompt before overwriting`

`-i, --interactive
              prompt before overwrite`

In my experience just plain 'mv' never asks for permission. If we want to circumvent this then we should use full path to mv to exclude alias using. e.g:  /bin/mv.  Or add a 'rm' to remove any existing file first when it exists. This will keep alias usage in place.

So there are different solutions but as said in my experience the current has never asked questions and a default Ubuntu or Fedora installation.

tananaev: Is there any reason we can't use `mv -f`?

mivtdole: No. I will add it an push the change

tananaev:
Merged, thanks.

PR #4595, opened 2020-10-26
Added support for JC400 alarm messages

radek-svanda-elder:
This is a follow up for #4412, this time supports JC400 (and possibly JC500) devices.
Product homepage: https://www.jimilab.com/products/jc400-aivision-cam.html

The device uses `gt06` in a similar way as JC100, carries a bit more information in alarm packets.

Other changes made:

* refactored `decodeGps` method so the code does not have to be copy pasted
* removed `alarmValue` attribute. That field is not present in JC100 message, the JC400 puts zeros there.


src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -222,9 +222,19 @@ public static boolean decodeGps(Position position, ByteBuf buf, boolean hasLengt
 
         position.set(Position.KEY_SATELLITES, BitUtil.to(buf.readUnsignedByte(), 4));
 
+        decodeLocation(position, buf, true);
+
+        return true;
+    }
+
+    private static void decodeLocation(Position position, ByteBuf buf, boolean withSpeed) {
```
tananaev, line 230: Instead of extracting a new method, can we just pass some boolean flags?

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -779,16 +788,20 @@ private Object decodeBasicOther(
...
 
         } else if (type == MSG_ALARM) {
-
             DateBuilder dateBuilder = new DateBuilder(deviceSession.getTimeZone())
                     .setDate(buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte())
                     .setTime(buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte());
-
-            getLastLocation(position, dateBuilder.getDate());
-
+            if (dataLength == 7) {
```
tananaev, line 794: Is this a way to identify device model? Maybe we can have a boolean variable for it? So it's more readable.

tananaev (commented):
Added some comments. Can you also please attach relevant protocol documentation.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
radek-svanda-elder: Thanks for the review.

The `decodeGps` method would have to have one flag for each:

* date
* length
* altitude (the part in `decodeExtendedModular` uses that)
* satelites
* speed

Looks like too many switches to be honest. Another solution could be an adapter on top of the buffer but that is a bit more effort.

Anyway, your project, your rules, your call - if you would like to have those switches, let me know and I will refactor this pull request.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
radek-svanda-elder: Not a nice solution, is it?

JC100 sends this packet for harsh braking: `78780c95130a0209321c91000112800d0a`

Same packet for JC400: `78788b95140a060e2208055D4A800209D9C014c59100004556454e545f3335333337363131303032333139365f30303030303030305f323032305f31305f30365f31365f33345f30385f33322e6d70342c4556454e545f3335333337363131303032333139365f30303030303030305f323032305f31305f30365f31365f33345f30385f33312e6d70340000da360d0a`

The long part of JC400 message starting with `455645...` are names of uploaded video files - the device does that the moment an event happens.

So the important parts of both packets, stripped of header/length/type:

```
130a0209321c 91
140a060e2208 055D4A80 0209D9C0 14c5 91
```

It is date/time the the beginning for both, JC400 continues with GPS location and event type while the JC100 has only the event type there.

The packet length was the obvious solution for me here. Any idea of a better solution?

radek-svanda-elder:
Regarding the documentation - this work is partially based on an internal document I cannot publish and partially on reverse engineering packets.

I can share packet break ups and event types though, if that is enough?

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
tananaev: Yep, let's do the switches. I think we just need 2 new arguments, not 3. But let's see.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
tananaev: I'm ok with the solution. I think we just need add a variable to make it more explicit.

tananaev:
Please share as much as you can. It will help with any troubleshooting later.

radek-svanda-elder:
### Alert packet breakdown

Packet example:

```
78785195140a020c2914055D4A800209D9C014009300004556454e545f3335333337363131303032333139365f30303030303030305f323032305f31305f30325f31345f34315f32305f30352e6d70340004e3a60d0a
```

Breakdown:

```
7878          <- header
51            <- length
95            <- message type
140a02 0c2914 <- date/time
05461cb7      <- lat
01b6bb2c      <- long
1400          <- Course, Status
93            <- alarm type
0000          <- alarm value
4556454E545F33353333373631313030303030305F30303030303030305F323032305F31305F30325F31345F34315F32305F30352E6D7034
              <- file names (EVENT_35337611000000_00000000_2020_10_02_14_41_20_05.mp4)
0004          <- message serial number
e3a6          <- crc
0d0a          <- end
```

#### Packet fields

Field values are in same format as in JC100. See the [document](https://www.dropbox.com/s/gysbq0mgugri01u/20170830%20Communication%20Protocol%20of%20JC100.pdf?dl=0), linked from a [traccar support forum topic](https://www.traccar.org/forums/topic/jc100-image-and-video-protocol/).

##### Message type

`0x95` identifies an alarm packet

##### GPS Info

* *Date/Time* - 6B
* *Latitude* - 4B
* *Longitude* - 4B
* *Course, Status* - 2B

##### Alarm Info

* *Alarm Type* - 1B

| Alarm ID | Description |
| --- | --- |
| 0x80 | Vibrating alert |
| 0x81 | SIM exceeds traffic limit |
| 0x82 | Device restarted |
| 0x84 | Camera 1 exception |
| 0x85 | Camera 2 exception |
| 0x86 | TF card unidentifiable |
| 0x87 | Speeding alert |
| 0x88 | Power cut alert |
| 0x89 | No USB camera |
| 0x90 | Harsh acceleration |
| 0x91 | Harsh braking |
| 0x92 | Sharp cornering |
| 0x93 | Collision alert |
| 0x8A | Fuel cut alert enabled |
| 0x8B | Fuel cut alert disabled |
| 0x8D | Switched to land transport |
| 0x8E | Environment exception |
| 0x95 | Switched to waterborne transport mode |
| 0x96 | Switched to parking mode |
| 0x8C | Fatigue driving |
| 0x97 | Driver in a call |
| 0x9A | Driver smoking |
| 0x8F | Driver distracted |
| 0x94 | Driver strange |
| 0x98 | Active capture |
| 0x99 | Driver changed |

* *Alarm Value* - 2B - contains only speed for overspeed alarm, otherwise zeros
* *File names* - NB - comma separated names of video files the device will upload to configured _upload endpoint_

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -210,6 +210,11 @@ private void sendPhotoRequest(Channel channel, int pictureId) {
     }
 
     public static boolean decodeGps(Position position, ByteBuf buf, boolean hasLength, TimeZone timezone) {
+        return decodeGps(position, buf, hasLength, true, true, timezone);
+    }
+
+    public static boolean decodeGps(Position position, ByteBuf buf, boolean hasLength, boolean hasSatellites,
+                                    boolean hasSpeed, TimeZone timezone) {
```
tananaev, line 217: We don't do this formatting. Please move everything to the next line with 2x indentation. Check other code to see formatting style.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -804,13 +818,13 @@ private Object decodeBasicOther(
...
                     break;
+                case 0x93:
+                    position.set(Position.KEY_ALARM, Position.ALARM_ACCIDENT);
+                    break;
                 default:
                     position.set(Position.KEY_ALARM, Position.ALARM_GENERAL);
                     break;
             }
-
-            position.set("alarmValue", buf.readShort());
```
tananaev, line 812: Wha happened to this value?

radek-svanda-elder: As I wrote above - it is not there for JC100 at all and is zero for JC400. Only over speed alarm from JC400 has the actual speed value there but you get that from the location update already.
I was wrong in #4412 and read the packet serial number by mistake.

src/test/java/org/traccar/protocol/Gt06ProtocolDecoderTest.java
```diff
@@ -334,12 +334,21 @@ public void testDecode() throws Exception {
                 Position.KEY_ALARM, Position.ALARM_ACCELERATION);
 
         verifyAttribute(decoder, binary(
-                "78780c95130a0209321c90000112800d0a"),
-                "alarmValue", 1);
+                "78780c95130a0209321c91000112800d0a"),
```
tananaev, line 337: What is the reason for changing existing test cases?

radek-svanda-elder: It is a git merge thing - I removed the `alarmValue` above, also removed the test case. Then added new test fro JC400 braking. It looks like I have changed the test case in diff but it is not.

tananaev:
Merged, thanks.

PR #4598, opened 2020-10-29
H02FrameDecoder  for tracker protocol

ilockit-CFR:
Adjust frame length according to new protocol version. Distinguish between LBS, Wifi and GPS data in binary messages.
[GPRS PROTOCOL-20201021 PET02.pdf](https://github.com/traccar/traccar/files/5458160/GPRS.PROTOCOL-20201021.PET02.pdf)

Not sure if generally applicable for other H02 devices. I set messagelength to 37 in config file, because this is the shortest message I received. It only contains GPS data. Other message lengths are calculated according to the content.


tananaev:
Doesn't look correct to me. Also PR checks failed.

PR #4599, opened 2020-10-30
Stop running traccar service when performing update

mivtdole:
When you perform an install (update) op traccar on a linux system which already has traccar installed and the traccar service is running the installer produces the following error:

```
# ./traccar.run 
Creating directory out
Verifying archive integrity...  100%   All good.
Uncompressing traccar  100%  
cp: cannot create regular file '/opt/traccar/jre/bin/java': Text file busy
Created symlink /etc/systemd/system/multi-user.target.wants/traccar.service → /etc/systemd/system/traccar.service.
```

This update/patch fixes this issue by stopping the service when it is active before copying the new files into place.

### DIFF
--- a/setup/setup.sh
+++ b/setup/setup.sh
@@ -7,8 +7,14 @@ then
     PRESERVECONFIG=1
 fi
 
+ISACTIVE=`systemctl is-active traccar.service`
+if [ "${ISACTIVE}" == "active" ]
+then
+    systemctl stop traccar
+fi
+
 mkdir -p /opt/traccar
-cp -r * /opt/traccar
+cp -r -f * /opt/traccar
 chmod -R go+rX /opt/traccar
 
 if [ ${PRESERVECONFIG} -eq 1 ] && [ -f /opt/traccar/conf/traccar.xml.saved ]
@@ -20,7 +26,7 @@ mv /opt/traccar/traccar.service /etc/systemd/system
 chmod 664 /etc/systemd/system/traccar.service
 
 systemctl daemon-reload
-systemctl enable traccar.service
+systemctl enable traccar
 
 rm /opt/traccar/setup.sh
 rm -r ../out


setup/setup.sh
```diff
@@ -7,6 +7,12 @@ then
...
 fi
 
+ISACTIVE=`systemctl is-active traccar.service`
+if [ "${ISACTIVE}" == "active" ]
+then
+    systemctl stop traccar.service
+fi
+
 mkdir -p /opt/traccar
 cp -r * /opt/traccar
```
tananaev, line 17: Maybe we should add `-f` here as well?

setup/setup.sh
```diff
@@ -7,6 +7,12 @@ then
     PRESERVECONFIG=1
 fi
 
+ISACTIVE=`systemctl is-active traccar.service`
```
tananaev, line 10: Can we remove `.service` suffix?

tananaev (commented):
Are there any more changes to this script coming?

setup/setup.sh
mivtdole: According to the man page '-f' will not work for the stop command:

```
       -f, --force
           When used with enable, overwrite any existing conflicting symlinks.

           When used with edit, create all of the specified units which do not already exist.

           When used with halt, poweroff, reboot or kexec, execute the selected operation without shutting down all units. However, all processes will be killed forcibly and all file systems are unmounted or
           remounted read-only. This is hence a drastic but relatively safe option to request an immediate reboot. If --force is specified twice for these operations (with the exception of kexec), they will be
           executed immediately, without terminating any processes or unmounting any file systems. Warning: specifying --force twice with any of these operations might result in data loss. Note that when
           --force is specified twice the selected operation is executed by systemctl itself, and the system manager is not contacted. This means the command should succeed even when the system manager has
           crashed.
```
There are two solutions:

1. Keep this as is. I never had problems that the traccar service never stopped.
2. After the stop check if the process are still running and then bail and throw an error so the user can solve it.

My preference would be option 1.

setup/setup.sh
mivtdole: This can be removed. Should I remove it also for the 'enable' line?

mivtdole:
This will be last change as far as I can see for now.

setup/setup.sh
tananaev: Yep.

setup/setup.sh
tananaev: I meant for the `cp` command.

PR #4601, opened 2020-11-11
Jt701 implementation

laudarch:
I have done a quick implementation for Jt701, this has status and alarms that conflict with Jt600 when you try to combine them


tananaev:
If those are the only conflicts, I don't think we need to duplicate everything else.

laudarch:
> If those are the only conflicts, I don't think we need to duplicate everything else.

Am looking at adding the other missing features for JT701, documentation is here https://we.tl/t-uTZa9e74XX

tananaev:
Can you please elaborate on the conflicts.

laudarch:
> Can you please elaborate on the conflicts.

I just updated the comment

tananaev:
Unfortunately it doesn't answer my question - why do you want to duplicate the rest of the code?

laudarch:
> Unfortunately it doesn't answer my question - why do you want to duplicate the rest of the code?

What do you suggest?

tananaev:
I suggest reusing the code.

laudarch:
> I suggest reusing the code.

let me attempt and test

laudarch:
Reuse is not working bro, tried extending even, am going to implement JT707A because I have a need for it, might have to visit this later unless you can fix it.

tananaev:
OK, closing this for now. Feel free to send a new PR later if you find a way.

Issue #4102, opened 2018-10-14
[feature]what 3 words address


icf20:
i let this videos here so you get the idea

https://www.youtube.com/watch?v=a4ZBzM3L6ws

https://www.youtube.com/watch?v=Fus75eaLK5A

is there any interest in having an `what 3 words ` next to show address ?

api: https://docs.what3words.com/api/v2/

tananaev:
I've heard about it and I don't really like that it's a proprietary solution. Also, I don't think it's popular enough to be considered at this point.

whatnick:
This is an open/free equivalent technology from Google - https://plus.codes/ . Less memorable than words, but similar advantages apply.

icf20:
i dont see exactly how is exactly proprietary since are just words, one could get the list and rehost it and would totally be OK

tananaev:
How do you get the list/database?

icf20:
that is quite a task since you need to use a api https://what3words.com/developers/ to get access to their database and there is no way to get the whole thing, as far i understand you can save and copy everything you get from the api

tananaev:
If you crunch the numbers, it's an impossible task. That's why it's proprietary solution. There is no database. There is a formula that they don't disclose.

PR #4619, opened 2020-12-29
Add What3Words Geocoder

TwoWheelDev:
I saw that there had been some interest in having a Geocoder for What3Words previously. The service is now a lot more popular and in use by a number of emergency services and couriers.

I have therefore created a Geocoder for What3Words and tested it against the current Github version. The address will display a What3Words address with a link; that links to their map tile.

w3w_address

I have checked the API guidelines and for converting coordinates to a w3w address this is only subject to fair use policy. In keeping with the other geocoders, the API key is configurable from the configuration file, however.

The required configuration is as follows: 
```
<entry key='geocoder.type'>what3words</entry>
<entry key='geocoder.key'>**APIKEY**</entry>
```


tananaev:
Can you please provide some references for this:
> I saw that there had been some interest in having a Geocoder for What3Words previously

dlintott:
The only one I can now find is a feature request from 2018, #4102. I seemed to recall seeing it elsewhere as well but am unable to find it at this current time!

tananaev:
I feel like it has very limited value. Let's see if anyone else is interested in it. There was a hype around it a couple years back, but it's still not widely used.

covertbert:
I think whether it's widely used may depend on where you live. For example it's used by all mountain rescue organisations in the UK (I can't speak for other countries of course)

tsmgeek:
> I feel like it has very limited value. Let's see if anyone else is interested in it. There was a hype around it a couple years back, but it's still not widely used.

In the UK we have adverts about it on main TV, also many emergency services are using it.

andywarren1969:
Thanks for developing this it's really useful, however there is a small bug as sometimes in place of the address link I get the following displayed as html

<span style='color: #E11F26'>///</span><a href='https://w3w.co/earplugs.saved.reserving' target='_blank'>earplugs.saved.testing</a>

marcgran:
I'm interested in try this geocoder with my project but I don't know how to do.
I put the config in the traccar.xml :
<entry key='geocoder.type'>what3words</entry>
<entry key='geocoder.key'>**APIKEY**</entry>
But only show the normal direction, how I can install this geocoder?

tananaev:
You have to checkout the code and compile it.

netvader:
@marcgran would also have liked to recompile it, unfortunately the pull request no longer works with the new traccar versions

PR #4625, opened 2021-01-23
2169 fixes

starblazr:
Fixes:
Date decoding
GPS location decoding

### DIFF
--- a/src/main/java/org/traccar/handler/events/BehaviorEventHandler.java
+++ b/src/main/java/org/traccar/handler/events/BehaviorEventHandler.java
@@ -41,7 +41,7 @@ public class BehaviorEventHandler extends BaseEventHandler {
     public void onPosition(Position position, Callback callback) {
 
         Position lastPosition = cacheManager.getPosition(position.getDeviceId());
-        if (lastPosition != null && position.getFixTime().equals(lastPosition.getFixTime())) {
+        if (lastPosition != null && !position.getFixTime().equals(lastPosition.getFixTime())) {
             double acceleration = UnitsConverter.mpsFromKnots(position.getSpeed() - lastPosition.getSpeed()) * 1000
                     / (position.getFixTime().getTime() - lastPosition.getFixTime().getTime());
             if (accelerationThreshold != 0 && acceleration >= accelerationThreshold) {


tananaev:
Please fix formatting issues.

starblazr:
fixed, and also adds the battery multiplication

src/main/java/org/traccar/protocol/XirgoProtocolDecoder.java
```diff
@@ -370,3 +382,5 @@ private Object decodeFixed(
     }
 
 }
+
```
tananaev, line 385: Undo this please.

src/main/java/org/traccar/protocol/XirgoProtocolDecoder.java
```diff
@@ -233,10 +233,18 @@ private Object decodeCustom(
                     dateBuilder.setSecond(Integer.parseInt(time[2]));
                     break;
                 case "LT":
-                    position.setLatitude(Double.parseDouble(values[i]));
+                    if (values[i].contains(".")) {
+                        position.setLatitude(Double.parseDouble(values[i]));
+                    } else {
+                        position.setLatitude(Double.parseDouble(values[i]) / 1000000);
```
tananaev, line 239: This can probably be optimized by using `parseInt` instead. Same for other places.

.gitignore
```diff
@@ -11,3 +11,4 @@ nbactions.xml
 out
 build
 src/main/java/org/traccar/protobuf
+debug.xml
```
tananaev, line 14: Why is this added to ignore file?

tananaev (commented):
Added some comments. Also, please add a unit test.

.gitignore
starblazr: Shouldn't have pushed, that's just for my local repo.

src/main/java/org/traccar/protocol/Xt2400ProtocolDecoder.java
```diff
@@ -132,66 +154,84 @@ protected Object decode(
                     }
                     position.setDeviceId(deviceSession.getDeviceId());
                     break;
-                case 0x04:
+                case 0x04: // Reason Code (Old-school stuff)
```
aimiebell, line 157: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

PR #4630, opened 2021-02-04
Allow per-device forward url override.

osery:
This is to allow overriding target urls for forwarding position information and events per device or per device group.

Our use case is 1 Traccar server that needs to forward events to one of N target servers depending on a device group.

@tananaev, please let me know if this is something useful for the core Traccar and any changes needed to get it in :v:.

### DIFF
--- a/src/main/java/org/traccar/Context.java
+++ b/src/main/java/org/traccar/Context.java
@@ -324,7 +324,7 @@ public final class Context {
         serverManager = new ServerManager();
         scheduleManager = new ScheduleManager();
 
-        if (config.hasKey(Keys.EVENT_FORWARD_URL)) {
+        if (config.getBoolean(Keys.EVENT_FORWARD_ENABLE)) {
             eventForwarder = new EventForwarder();
         }
 
--- a/src/main/java/org/traccar/MainModule.java
+++ b/src/main/java/org/traccar/MainModule.java
@@ -264,7 +264,7 @@ public class MainModule extends AbstractModule {
     @Provides
     public static WebDataHandler provideWebDataHandler(
             Config config, IdentityManager identityManager, ObjectMapper objectMapper, Client client) {
-        if (config.hasKey(Keys.FORWARD_URL)) {
+        if (config.getBoolean(Keys.FORWARD_ENABLE)) {
             return new WebDataHandler(config, identityManager, objectMapper, client);
         }
         return null;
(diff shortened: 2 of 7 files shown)


tananaev:
I'm not sure if it's going to be useful for a wider audience. There is also a bit of security concern. This means that people can "disable" global forwarding by using incorrect URL or their own URL. But maybe let's keep the PR open and see if anyone else is interested in similar functionality.

jcardus:
I have the same need and implemented also with device attributes being checked on the code running behind the url being invoked by traccar.

lukazi:
I have the same need.  People can overide global forwarding for owned devices only. Global forwarding could run in parallel if You wory about that.

tananaev:
Maybe if the global URL is not set, we can use per device. If it is set, we use that..

jcardus:
Thats even better for me, because right now I have to forward all the
devices and only a few are processed

Anton Tananaev <notifications@github.com> escreveu em sex., 5/02/2021 às
16:29 :

osery:
Good point with the possible security issue :thinking:.
Let me change the PR as suggested, probably tomorrow :+1:.

osery:
@tananaev so I have changed the behavior as suggested above, i.e., giving precedence to the global `forward.url` config option when set.

To make that happen, I have added `forward.enable` and `event.forward.enable` config options defaulting to `true`. The motivation is to allow disabling the forwarding overhead when not needed (which now cannot be done solely by the emptiness of the global urls), but to keep this from breaking older config files (missing value would not disable(=break) the forwarding on upgrade).

PTAL and let me know if this is something worth merging and what changes would have to be done.

An alternative is to keep the original behavior and add a config option to allow per-device overrides, but that does not easily cover the "forward only for a subset of devices" scenario described by @jcardus above :shrug:.

tananaev:
We just removed those keys a couple of weeks ago. Please don't add them back. Just use URL.

osery:
Not sure how tbh. If it is filled in, it will have precedence over the per-device settings. If it is not, the components `WebDataHandler` and `EventForwarder` don't get even initialized. Do you mean to always have these initialized then? I can surely do that, though it will then mean unnecessary overhead for each position processing. Is that ok? Happy to change if so. Let me know :+1:.

tananaev:
I see what you mean. Let's leave it for now. I need to think about it. We are actually in the process of changing device/user configuration.

osery:
Hi @tananaev, any thoughts on this one?

tananaev:
Haven't had time to work on configuration yet. Swarmed with other work.

jpmens:
I would like to see this implemented, and I like the idea of the global `forward.url` config taking precedence.

PR #4635, opened 2021-02-11
Add UDP support for GPSLogger and OpenGTS

AuroraRAS:
You can select the UDP mode in GPSLogger now.

Signed-off-by: AuroraRAS <chplee@gmail.com>


tananaev:
Please include unit tests. Is the format completely different for UDP?

AuroraRAS:
> Please include unit tests. Is the format completely different for UDP?

mostly different. They are shared $GPRMC field, but not other.

tananaev:
If it's NMEA, can't we just use T55 protocol for it?

AuroraRAS:
it's complex message, include account and device id. port 5005 can't handle it.

AuroraRAS:
You are right, port 5005 is working fine.

PR #4636, opened 2021-02-13
CI: Validate Gradle Wrapper

EwoutH:
Validates the gradle-wrapper.jar file on each builds, which is a binary blob of executable code. This action ensures it's legit and doesn't execute malicious code.

See https://github.com/gradle/wrapper-validation-action

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -8,14 +8,14 @@ on:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
-
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


tananaev:
Not sure I understand the purpose. What is the threat it protects from?

EwoutH:
If an outside developer or bot wants to update the Gradle Wrapper, it validates that the gradle-wrapper.jar is legit.

See [Verifying Gradle Wrappers with GitHub Actions](https://blog.gradle.org/gradle-wrapper-checksum-verification-github-action)

tananaev:
We don't let outside developers to update gradle.

PR #4637, opened 2021-02-13
Add workflow which checks weekly for Gradle updates

EwoutH:
This workflow checks every Monday morning if a new version of Gradle is available. If so, it opens a PR to update the Gradle wrapper.

Uses https://github.com/gradle-update/update-gradle-wrapper-action


tananaev:
Again, I'm not sure I understand the value.

EwoutH:
It helps keep the Gradle wrapper up to date and reduces the risk that it isn't updated for a long time. Just helps a bit with maintenance.

tananaev:
Honestly I don't see much value. I prefer to not update it often to avoid a lot of binary blobs in the repo increasing the size.

PR #4638, opened 2021-02-13
CI: Only pass if there are no Gradle warnings

EwoutH:
Adds the CLI option `--warning-mode=fail` to the Gradle build job. This ensures the build only passes if no warnings are cast.

See: https://docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_warnings


tananaev:
This seems reasonable. Merged, thanks.

PR #4650, opened 2021-03-04
Feature sns integration

subodhranadive:
**added**
- Amazon SNS SMS support

**Documentation**

Configuration added

- `aws.sns.enabled` : Use this key to send message using SNS. Accepted value - enable/disable
- `aws.access` : AWS Access Key with SNS permission
- `aws.secret` : AWS Secret Access Key with SNS permission
- `aws.region` : AWS Region for SNS service. **NOTE** : Make sure to use regions that are supported for SNS messaging.

To enable sns messaging, we need to make sure that the value `sms.enable` is `true` and `notificator.types` contains `sms`.


subodhranadive:
@tananaev Please can you take a look at this PR.

Let me know if there are any changes required from my end.

Thanks,
Subodh

build.gradle
```diff
@@ -77,6 +77,7 @@ dependencies {
     implementation "com.sun.xml.bind:jaxb-core:2.3.0.1"
     implementation "com.sun.xml.bind:jaxb-impl:2.3.3"
     implementation "javax.activation:activation:1.1.1"
+    implementation 'com.amazonaws:aws-java-sdk-sns:1.11.692'
```
tananaev, line 80: Is there an API we can use instead of integrating a library?

subodhranadive: Thanks for responding @tananaev.
It would be better if we use SDK instead of using bare API itself. As we underlying authentication/authorisation and other complexities are taken care by the SDK itself. Also, we wont have to deal with the security aspect as well in case of SDK.

tananaev: Is the SDK open source?

subodhranadive: Yes the SDK is open source.

**Ref**: [AWS Java SDK](https://github.com/aws/aws-sdk-java)

src/main/java/org/traccar/config/Keys.java
```diff
@@ -687,6 +687,36 @@ private Keys() {
             "sms.http.template",
             Collections.singletonList(KeyType.GLOBAL));
 
+    /**
+     * AWS Access Key with SNS permission.
+     */
+    public static final ConfigKey<String> AWS_ACCESS_KEY = new ConfigKey<>(
+            "aws.access",
```
tananaev, line 694: Let's name all parameters with a prefix `sms.aws.`.

src/main/java/org/traccar/config/Keys.java
```diff
@@ -687,6 +687,36 @@ private Keys() {
...
+    public static final ConfigKey<String> AWS_SECRET_KEY = new ConfigKey<>(
+            "aws.secret",
+            Collections.singletonList(KeyType.GLOBAL));
+
+    /**
+     * SNS Enable switch.
+     * Use 'enable' to send message with SNS or 'disable' to stop using SNS
+     */
+    public static final ConfigKey<String> AWS_SNS_ENABLED = new ConfigKey<>(
+            "aws.sns.enabled",
```
tananaev, line 709: We are trying to reduce number of config parameters. Let's use presence of one of the other keys as "enabled" flag.

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,69 @@
+package org.traccar.sms;
```
tananaev, line 1: Please add license notice here.

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,69 @@
...
+        snsStatus = Context.getConfig().getString(Keys.AWS_SNS_ENABLED);
+        region = Context.getConfig().getString(Keys.AWS_REGION);
+        snsClient = awsSNSClient(accessKey, secretKey, region);
+
+        if (!snsStatus.equals("true") || accessKey == null || secretKey == null || region == null) {
+            LOGGER.error("SNS Not Configured Properly. Please provide valid config.");
+        }
+    }
+
+    public AmazonSNS awsSNSClient(String accessKey, String secretKey, String region) {
```
tananaev, line 42: This can probably be inlined and we won't need to keep config parameters as class variables.

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,69 @@
...
+        BasicAWSCredentials awsCreds = new BasicAWSCredentials(accessKey, secretKey);
+        return AmazonSNSClientBuilder.standard().withRegion(region)
+                .withCredentials(new AWSStaticCredentialsProvider(awsCreds)).build();
+    }
+
+    public void sendSNSMessage(String message, String destAddress) {
```
tananaev, line 48: Why do we need this method? Can't we just put this code in `sendMessageSync`?

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,69 @@
...
+        Map<String, MessageAttributeValue> smsAttributes = new HashMap<>();
+        smsAttributes.put("AWS.SNS.SMS.SenderID",
+                new MessageAttributeValue().withStringValue("SNS").withDataType("String"));
+        smsAttributes.put("AWS.SNS.SMS.SMSType",
+                new MessageAttributeValue().withStringValue("Transactional").withDataType("String"));
+
+        PublishResult result = this.snsClient.publish(new PublishRequest().withMessage(message)
```
tananaev, line 55: Remove `this`.

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,69 @@
...
+                .withPhoneNumber(destAddress).withMessageAttributes(smsAttributes));
+    }
+
+    @java.lang.Override
```
tananaev, line 59: Why do you need `java.lang.`?

src/main/java/org/traccar/config/Keys.java
subodhranadive: I have gone through the config variables but I couldn't find one which represents the specific/meaningful description to enable/disable this feature.

src/main/java/org/traccar/sms/SnsSmsClient.java
subodhranadive: To minimize the amount of code duplication I have implemented this method, as we are using the same code in `sendMessageSync` and `sendMessageAsync`.

tananaev: You can just put it directly in the `sendMessageSync`. And call that method from `sendMessageAsync`. I think we do it in some other places.

src/main/java/org/traccar/config/Keys.java
tananaev: I meant to use other keys, like `region` or something.

subodhranadive:
@tananaev I have made changes as requested. Can you please review it and let me know if there are any changes required.

Thanks,
Subodh

src/main/java/org/traccar/config/Keys.java
```diff
@@ -687,6 +687,28 @@ private Keys() {
             "sms.http.template",
             Collections.singletonList(KeyType.GLOBAL));
 
+    /**
+     * AWS Access Key with SNS permission.
+     */
+    public static final ConfigKey<String> AWS_ACCESS_KEY = new ConfigKey<>(
```
tananaev, line 693: Please make sure variable names match keys. For example this should be `SMS_AWS_ACCESS`.

src/main/java/org/traccar/protocol/HuabaoProtocolDecoder.java
```diff
@@ -171,7 +172,7 @@ protected Object decode(
                         formatMessage(MSG_TERMINAL_REGISTER_RESPONSE, id, false, response), remoteAddress));
             }
 
-        } else if (type == MSG_TERMINAL_AUTH) {
+        } else if (type == MSG_TERMINAL_AUTH || type == MSG_HEARTBEAT) {
```
tananaev, line 175: What is this change?

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -838,7 +838,9 @@ private void decodeHeartbeat(ByteBuf buf, Position position) {
         buf.readUnsignedShort(); // satellites
         buf.readUnsignedByte(); // alarm
         buf.readUnsignedByte(); // language
-        buf.readUnsignedByte(); // battery
+
+        position.set(Position.KEY_BATTERY_LEVEL, buf.readUnsignedByte());
```
tananaev, line 842: And this?

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Map;
+
+
+public class SnsSmsClient implements SmsManager {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SnsSmsClient.class);
+
+    private final AmazonSNS snsClient;
+
+    public SnsSmsClient() {
+        if (Context.getConfig().getString(Keys.AWS_REGION) == null
```
tananaev, line 40: There's `hasKey` that you should be using instead.

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,77 @@
...
+                || Context.getConfig().getString(Keys.AWS_ACCESS_KEY) == null
+                || Context.getConfig().getString(Keys.AWS_SECRET_KEY) == null) {
+            throw new RuntimeException("SNS Not Configured Properly. Please provide valid config.");
+        }
+        snsClient = awsSNSClient();
+    }
+
+    public AmazonSNS awsSNSClient() {
```
tananaev, line 48: What's the reason for this method?

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,77 @@
...
+                .withPhoneNumber(destAddress).withMessageAttributes(smsAttributes));
+    }
+
+    @Override
+    public void sendMessageAsync(String destAddress, String message, boolean command) {
+        try {
+            sendMessageSync(destAddress, message, command);
+        } catch (InterruptedException interruptedException) {
+            LOGGER.warn("SMS send failed", interruptedException.getMessage());
+        } catch (MessageException messageException) {
```
tananaev, line 73: Why do you need two catch blocks that do the same thing?

src/test/java/org/traccar/protocol/Gt06ProtocolDecoderTest.java
```diff
@@ -17,6 +17,10 @@ public void testDecode() throws Exception {
         verifyNull(decoder, binary(
                 "78780D01086471700328358100093F040D0A"));
 
+        verifyAttribute(decoder, binary(
```
tananaev, line 20: This also doesn't seem to be related.

tananaev (commented):
Added some comments. Looks like you have some unrelated changes included.

src/main/java/org/traccar/sms/SnsSmsClient.java
subodhranadive: To abstract the AWS SNS client so that we can use just the same client in future if we need to add new feature and we can use the same method.

subodhranadive:
@tananaev I have updated the PR. Please can you review it and let me know if there are any changes required.

Thanks,
Subodh

src/main/java/org/traccar/sms/SnsSmsClient.java
tananaev: This part doesn't really make sense to me. Please inline for now.

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public void sendMessageSync(String destAddress, String message, boolean command)
+            throws InterruptedException, MessageException {
+        Map<String, MessageAttributeValue> smsAttributes = new HashMap<>();
+        smsAttributes.put("AWS.SNS.SMS.SenderID",
+                new MessageAttributeValue().withStringValue("SNS").withDataType("String"));
+        smsAttributes.put("AWS.SNS.SMS.SMSType",
+                new MessageAttributeValue().withStringValue("Transactional").withDataType("String"));
+        snsClient.publish(new PublishRequest().withMessage(message)
```
tananaev, line 64: Wait. Is this method asynchronous or not?

subodhranadive: Yes, this method is `asynchronous`. While publishing the message AWS returns request id and behind the scenes use `asynchronous` processing for sending the actual message. (as far as i know, AWS is not providing any `synchronous` method for sending sns messages)

tananaev: OK, can you please provide some reference to confirm that it's `asynchronous`. Also, I'm not sure how you asynchronous method can return those exceptions.

subodhranadive: **Reference** : [AWS SNS Publish Message](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sns/AmazonSNSClient.html#publish-com.amazonaws.services.sns.model.PublishRequest-)

As described in the documentation, it just returns messageId and processes the request.

`When a messageId is returned, the message has been saved and Amazon SNS will attempt to deliver it shortly.`

tananaev: Just want to clarify. I'm talking about network request to Amazon SNS. I'm not talking about flow on the Amazon server. It sounds like the documentation you are referencing is referring to the latter.

subodhranadive: When it comes to network call, it is synchronous.

What we can do is, we can create another method and make it as Async and call `sendMessageSync(...)` method from the newly created method. - For this we will have to use `@Async` annotation which requires `spring-context` dependency.

Please share your thoughts on the same.

tananaev: I don't think we need to include Sprint to just make a single asynchronous method. I think just starting a new thread would be enough.

subodhranadive: I have updated PR with required changes.

Thanks,
Subodh

tananaev:
Looks good. We'll merge it before the next release.

subodhranadive:
> Looks good. We'll merge it before the next release.

Thanks @tananaev

subodhranadive:
**Documentation Update**

Configuration added

- `aws.region` : AWS Region for SNS service.
- `aws.access` : AWS Access Key with SNS permission
- `aws.secret` : AWS Secret Access Key with SNS permission

To enable sns messaging, we need to make sure that the value `sms.enable` is `true` and `notificator.types` contains `sms`.

**NOTE** : 
- Make sure to use regions that are supported for SNS messaging in `aws.region`.
- `aws.region` : This key will enable/disable SNS messaging. (If this key is present, then only SNS messaging will work).

jcardus:
Just for the record, there's an async class
<https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sns/AmazonSNSAsyncClient.html>
for that.

tananaev:
Great, then we have to use that instead. @subodhranadive can you please update the PR.

subodhranadive:
> Great, then we have to use that instead. @subodhranadive can you please update the PR.

@tananaev I have updated the PR. Please can you review it.

jcardus:
hi @subodhranadive, looks like you're ignoring the result and possibly exceptions

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,70 @@
...
+        if (Context.getConfig().hasKey(Keys.SMS_AWS_REGION)
+                && Context.getConfig().hasKey(Keys.SMS_AWS_ACCESS)
+                && Context.getConfig().hasKey(Keys.SMS_AWS_SECRET)) {
+            snsClient = awsSNSClient();
+        } else {
+            throw new RuntimeException("SNS Not Configured Properly. Please provide valid config.");
+        }
+    }
+
+    public AmazonSNSAsync awsSNSClient() {
```
tananaev, line 48: Please inline this.

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public void sendMessageSync(String destAddress, String message, boolean command) {
+        Map<String, MessageAttributeValue> smsAttributes = new HashMap<>();
+        smsAttributes.put("AWS.SNS.SMS.SenderID",
+                new MessageAttributeValue().withStringValue("SNS").withDataType("String"));
+        smsAttributes.put("AWS.SNS.SMS.SMSType",
+                new MessageAttributeValue().withStringValue("Transactional").withDataType("String"));
+        snsClient.publishAsync(new PublishRequest().withMessage(message)
```
tananaev, line 62: If there's an error, we should log something.

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright 2021 Anton Tananaev (anton@traccar.org)
```
tananaev, line 2: Feel free to put your name/email here if you want.

src/main/java/org/traccar/sms/SnsSmsClient.java
subodhranadive: Can you please elaborate a little more. Perhaps an example.

tananaev: Remove this method. I already had a comment about it, but it got outdated, but not addressed.

subodhranadive:
@tananaev I have updated the PR with suggested changes.

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,73 @@
...
+    public void sendMessageSync(String destAddress, String message, boolean command) {
+        Map<String, MessageAttributeValue> smsAttributes = new HashMap<>();
+        smsAttributes.put("AWS.SNS.SMS.SenderID",
+                new MessageAttributeValue().withStringValue("SNS").withDataType("String"));
+        smsAttributes.put("AWS.SNS.SMS.SMSType",
+                new MessageAttributeValue().withStringValue("Transactional").withDataType("String"));
+        try {
+            snsClient.publishAsync(new PublishRequest().withMessage(message)
+                    .withPhoneNumber(destAddress).withMessageAttributes(smsAttributes));
+        } catch (Exception exception) {
```
tananaev, line 64: I don't think that's how you handle asynchronous errors. This seems incorrect. Also, why is it a generic `Exception`?

tananaev (commented):
Found an issues with error handling.

src/main/java/org/traccar/sms/SnsSmsClient.java
subodhranadive: Hi @tananaev 
I have changed the generic `Exception` to `RuntimeException`.

Regarding handling of asynchronous errors, I took the reference of `publishAsync()` method of `AmazonSNSAsyncClient.java` as below 

[https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-sns/src/main/java/com/amazonaws/services/sns/AmazonSNSAsyncClient.java#L1337-L1362](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-sns/src/main/java/com/amazonaws/services/sns/AmazonSNSAsyncClient.java#L1337-L1362)

Here they are throwing the exception and as we have added catch block in our code and as a result that will be handled.

tananaev: But it is thrown in a function that is presumably executed asynchronously. You cannot catch that here.

subodhranadive: Hi @tananaev 

Can we use something like below:

```
Thread.setDefaultUncaughtExceptionHandler((thread, exception) ->
                LOGGER.error("SMS send failed", exception));

        new Thread(() -> {
            Map<String, MessageAttributeValue> smsAttributes = new HashMap<>();
            smsAttributes.put("AWS.SNS.SMS.SenderID",
                    new MessageAttributeValue().withStringValue("SNS").withDataType("String"));
            smsAttributes.put("AWS.SNS.SMS.SMSType",
                    new MessageAttributeValue().withStringValue("Transactional").withDataType("String"));
            snsClient.publishAsync(new PublishRequest().withMessage(message)
                    .withPhoneNumber(destAddress).withMessageAttributes(smsAttributes));
        });
```
If not, Please can you provide a small example/snippet.

tananaev: I would think that because it's already `publishAsync`, you don't need a new thread. I don't have a snippet obviously.

subodhranadive: Hi @tananaev 

I have added AWS `AsyncHandler` for exception handling of `publishAsync()` method.

src/main/java/org/traccar/sms/SnsSmsClient.java
```diff
@@ -0,0 +1,83 @@
...
+                .withPhoneNumber(destAddress).withMessageAttributes(smsAttributes);
+
+        snsClient.publishAsync(publishRequest, new AsyncHandler<PublishRequest, PublishResult>() {
+            @Override
+            public void onError(Exception exception) {
+                LOGGER.error("SMS send failed", exception);
+            }
+            @Override
+            public void onSuccess(PublishRequest request, PublishResult result) {
+                LOGGER.info("SMS sent successfully");
```
tananaev, line 74: Please don't log anything on success.

tananaev (commented):
Looks much better. Just one comment.

src/main/java/org/traccar/sms/SnsSmsClient.java
subodhranadive: Hi @tananaev 

I have updated the PR.

tananaev:
Merged, thank you.

subodhranadive:
> Merged, thank you.

Thanks @tananaev

PR #4653, opened 2021-03-13
starting user email validation

jcardus:
Checking if starting in the right direction. 

Maybe makes more sense to serialise the whole user model as json and then hash it. Not sure if the velocity template will handle the url encoding if I put a big string on the query string.

I see Response.serverError is not used anywhere else but I didn't want to add a new exception to the add method signature on BaseObjectResource.

Where does it make sense to put the new api method?

### DIFF
--- a/src/main/java/org/traccar/api/resource/UserResource.java
+++ b/src/main/java/org/traccar/api/resource/UserResource.java
@@ -15,15 +15,20 @@
  */
 package org.traccar.api.resource;
 
+import org.apache.velocity.VelocityContext;
 import org.traccar.Context;
 import org.traccar.api.BaseObjectResource;
 import org.traccar.config.Keys;
 import org.traccar.database.UsersManager;
+import org.traccar.helper.Hashing;
 import org.traccar.helper.LogAction;
 import org.traccar.model.ManagedUser;
 import org.traccar.model.User;
+import org.traccar.notification.FullMessage;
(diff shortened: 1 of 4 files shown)


tananaev:
I would put the new API on user as well. Can be sub-path.

Some issues that I see so far in what you have:
- Hashing can't be used; you need a proper signature
- Verification should only be needed for self-registering users, but the ones that admin/manager registers

jcardus:
I gave that example on the issue, why shouldn't we need verification for users registered by the admin / register?

tananaev:
Because they should be able to register users directly without verification.

jcardus:
that's my question, why should they? You're not solving the problem if you're not fixing it to the whole user base.

tananaev:
It would be silly if administrator can't register a user. That's not how most systems work. Email verification is intended for external regular users to make sure they don't abuse the system, can restore their password etc. It is not intended to restrict administrative privileges.

jcardus:
I stated the motivation for this feature on the issue and has nothing to do with that. 
Most systems that allow you to create users, also require email validation for it.
That's why my first approach wouldn't deny user login before email validation.
I'm not saying an administrator can't register or update a user. I'm saying the user he creates should have his email validated. I'm not restricting any privilege.

tananaev:
In case we don't agree, we should probably hold on.

jcardus:
ok

PR #4654, opened 2021-03-15
Added missing protocol decodings for both teltonika and pretrace protocol

geocept:
New Teltonika devives like FMM640 are sending beacon values as ID=10500 and no longer as ID=385 - see https://wiki.teltonika-gps.com/view/FMM640_Teltonika_Data_Sending_Parameters_ID . For pretrace protocol the parsing of position type (LBS or GPS) had been missing and the parsing of the state bits. The pretrace state bits have different meaning depending on the firmware and device, but anyway parsing helps to enable the displaying of states in the front end.

If you have question about pretrace changes you can discuss it with ben@pretrace.com - I did the changes together with Ben.

I added logger to help me dealing with the data - the logger import can of course be left out - I forgot this.

Would be nice if you could use these small changes or even make a better one.

### DIFF
--- a/src/main/java/org/traccar/protocol/PretraceProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/PretraceProtocolDecoder.java
@@ -16,6 +16,8 @@
 package org.traccar.protocol;
 
 import io.netty.channel.Channel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.traccar.BaseProtocolDecoder;
 import org.traccar.DeviceSession;
 import org.traccar.Protocol;
@@ -37,7 +39,7 @@ public class PretraceProtocolDecoder extends BaseProtocolDecoder {
             .text("(")
             .number("(d{15})")                   // imei
             .number("Uddd")                      // type
-            .number("d")                         // gps type
+            .number("(d)")                         // gps type
             .expression("([AV])")                // validity
             .number("(dd)(dd)(dd)")              // date (yymmdd)
             .number("(dd)(dd)(dd)")              // time (hhmmss)
@@ -75,8 +77,8 @@ public class PretraceProtocolDecoder extends BaseProtocolDecoder {
         Position position = new Position(getProtocolName());
         position.setDeviceId(deviceSession.getDeviceId());
 
+        position.set(Position.KEY_GPS, parser.nextInt(0));
         position.setValid(parser.next().equals("A"));
-
         position.setTime(parser.nextDateTime());
 
         position.setLatitude(parser.nextCoordinate());
@@ -90,7 +92,16 @@ public class PretraceProtocolDecoder extends BaseProtocolDecoder {
         position.set(Position.KEY_HDOP, parser.nextInt(0));
         position.set(Position.KEY_RSSI, parser.nextInt(0));
 
-        parser.next(); // state
+        String stateValue = parser.next(); // state
+        int bitName = 0;
+        for (int bytePos=7; bytePos>=0; bytePos--)
+        {
+            for (int bitPos=0; bitPos<4; bitPos++)
+            {
+                if (((stateValue.charAt(bytePos) - '0') & (1 << bitPos)) == (1 << bitPos)) position.set("State_Bit"+bitName, true);
+                bitName++;
+            }
+        }
 
         if (parser.hasNext()) {
             for (String value : parser.next().split(",")) {
--- a/src/main/java/org/traccar/protocol/PretraceProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/PretraceProtocolDecoder.java
@@ -16,8 +16,6 @@
 package org.traccar.protocol;
 
 import io.netty.channel.Channel;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/PretraceProtocolDecoder.java
```diff
@@ -90,7 +92,16 @@ protected Object decode(
         position.set(Position.KEY_HDOP, parser.nextInt(0));
         position.set(Position.KEY_RSSI, parser.nextInt(0));
 
-        parser.next(); // state
+        String stateValue = parser.next(); // state
```
tananaev, line 95: We usually save state as a single value. Also, if you want to break it down, you should probably use standard prefixes. There are also a bunch of formatting issues here.

src/main/java/org/traccar/protocol/PretraceProtocolDecoder.java
```diff
@@ -37,7 +39,7 @@ public PretraceProtocolDecoder(Protocol protocol) {
             .text("(")
             .number("(d{15})")                   // imei
             .number("Uddd")                      // type
-            .number("d")                         // gps type
+            .number("(d)")                         // gps type
```
tananaev, line 42: Comment indentation is off

src/main/java/org/traccar/protocol/PretraceProtocolDecoder.java
```diff
@@ -75,8 +77,8 @@ protected Object decode(
         Position position = new Position(getProtocolName());
         position.setDeviceId(deviceSession.getDeviceId());
 
+        position.set(Position.KEY_GPS, parser.nextInt(0));
```
tananaev, line 80: What is this?

src/main/java/org/traccar/protocol/PretraceProtocolDecoder.java
```diff
@@ -75,8 +77,8 @@ protected Object decode(
...
         position.setValid(parser.next().equals("A"));
-
```
tananaev, line 79: undo

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -559,6 +561,7 @@ private void decodeLocation(Position position, ByteBuf buf, int codec) {
                         if (BitUtil.check(flags, 2)) {
                             position.set("beacon" + index + "Temp", data.readUnsignedShort());
                         }
+
```
tananaev, line 564: undo

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -538,7 +540,7 @@ private void decodeLocation(Position position, ByteBuf buf, int codec) {
                 int length = buf.readUnsignedShort();
                 if (id == 256) {
                     position.set(Position.KEY_VIN, buf.readSlice(length).toString(StandardCharsets.US_ASCII));
-                } else if (id == 385) {
+                } else if ((id == 385) || (id == 10500)) {
```
tananaev, line 543: You don't need nested brackets.

tananaev (commented):
Added a bunch of comments. You should also provide unit tests. And also please fix build issues.

src/main/java/org/traccar/protocol/PretraceProtocolDecoder.java
geocept: This is a bitvalue indicating the GPS type:
image

src/main/java/org/traccar/protocol/PretraceProtocolDecoder.java
geocept: Store it in whatever value you want. Not storing it is not a good option I think. This is a bitfield with different states - I did not fiend a good existing standard prefix. If you know a good standard prefix please use it.

src/main/java/org/traccar/protocol/PretraceProtocolDecoder.java
geocept: This is needed to use the gps type - see below.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
geocept: Feel free to remove the brackets - it will not change much I think. Some like brackets some don't.

tananaev: Just to clarify, are you interested in fixing this PR or do you want us to take over?

geocept:
> Added a bunch of comments. You should also provide unit tests. And also please fix build issues.

You would not like it if I add unit tests. I am not a java developer - sorry I am the wrong person to do this.

tananaev:
Unfortunately we won't be able to merge the PR until it's fixed.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
geocept: Take over please. It works for me like this - perhaps I have different java version. You are the one that could make it work for everybody.

PR #4656, opened 2021-03-26
Implements MapboxGeocoder

RafaelMiquelino:
Add Mapbox reverser geocoder support


src/main/java/org/traccar/geocoder/MapboxGeocoder.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright 2020 Anton Tananaev (anton@traccar.org)
```
tananaev, line 2: You should probably remove this line.

RafaelMiquelino: Done

tananaev:
Merged, thanks.

PR #4664, opened 2021-04-20
Support native location messages for Telegram

soshial:
1. I decided it would be very convenient, if Telegram bots would send location native to Telegram chats.
2. Also, I think users would benefit if bot sent full message instead on short one. It shouldn't be that short as Firebase notification. (There I forgot to add `getBody()` to the code to convert to string.)

What do you think, @tananaev ?

Screenshot_20210420-184756

### DIFF
--- a/src/main/java/org/traccar/notificators/NotificatorTelegram.java
+++ b/src/main/java/org/traccar/notificators/NotificatorTelegram.java
@@ -25,6 +25,7 @@ import org.traccar.model.Event;
 import org.traccar.model.Position;
 import org.traccar.notification.NotificationFormatter;
 
+import javax.annotation.Nullable;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.InvocationCallback;
 
@@ -36,11 +37,14 @@ public class NotificatorTelegram extends Notificator {
     private final String urlSendLocation;
     private final String chatId;
 
-    public static class Message {
+    public static class TextMessage {
         @JsonProperty("chat_id")
         private String chatId;
         @JsonProperty("text")
         private String text;
+        @Nullable
+        @JsonProperty("reply_to_message_id")
+        private Long replyToMessageId;
         @JsonProperty("parse_mode")
         private String parseMode = "html";
     }
@@ -55,9 +59,7 @@ public class NotificatorTelegram extends Notificator {
         @JsonProperty("horizontal_accuracy")
         private double accuracy;
         @JsonProperty("bearing")
-        private double bearing;
-        @JsonProperty("parse_mode")
-        private String parseMode = "html";
+        private int bearing;
     }
 
     public NotificatorTelegram() {
@@ -68,30 +70,6 @@ public class NotificatorTelegram extends Notificator {
                 "https://api.telegram.org/bot%s/sendLocation",
                 Context.getConfig().getString(Keys.NOTIFICATOR_TELEGRAM_KEY));
         chatId = Context.getConfig().getString(Keys.NOTIFICATOR_TELEGRAM_CHAT_ID);
-    }
-
-    @Override
-    public void sendSync(long userId, Event event, Position position) {
-
-        Message message = new Message();
-        message.chatId = chatId;
-        message.text = NotificationFormatter.formatFullMessage(userId, event, position);
-        executeRequest(urlSendText, message);
-
-        if (position != null) {
-            LocationMessage locationMessage = new LocationMessage();
-            locationMessage.chatId = chatId;
-            locationMessage.latitude = position.getLatitude();
-            locationMessage.longitude = position.getLongitude();
-            locationMessage.bearing = position.getCourse();
-            locationMessage.accuracy = position.getAccuracy();
-            executeRequest(urlSendLocation, locationMessage);
-        }
-    }
-
-    @Override
-    public void sendAsync(long userId, Event event, Position position) {
-        sendSync(userId, event, position);
     }
 
     private void executeRequest(String url, Object message) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -70,9 +108,4 @@ public void failed(Throwable throwable) {
         });
     }
 
-    @Override
-    public void sendAsync(long userId, Event event, Position position) {
```
tananaev, line 74: What happened to this method?

src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -55,8 +75,26 @@ public void sendSync(long userId, Event event, Position position) {
...
+        executeRequest(urlSendText, message);
+
+        if (position != null) {
+            LocationMessage locationMessage = new LocationMessage();
+            locationMessage.chatId = chatId;
+            locationMessage.latitude = position.getLatitude();
+            locationMessage.longitude = position.getLongitude();
+            locationMessage.bearing = position.getCourse();
+            locationMessage.accuracy = position.getAccuracy();
+            executeRequest(urlSendLocation, locationMessage);
```
tananaev, line 88: Is there a way to send location message together with text?

soshial: It would be great, but according to [their API](https://core.telegram.org/bots/api#sendlocation) it is not really possible.

I tried adding `text` parameter, but to no avail.

src/main/java/org/traccar/notificators/NotificatorTelegram.java
soshial: This method is safe and sound, you may [check the class file in my repo](https://github.com/soshial/traccar/blob/patch-1/src/main/java/org/traccar/notificators/NotificatorTelegram.java).
Github diff algorithms are not ideal.

src/main/java/org/traccar/notificators/NotificatorTelegram.java
tananaev: Should we at least use `reply_to_message_id` to indicate which message this location is related to?

src/main/java/org/traccar/notificators/NotificatorTelegram.java
tananaev: Can we move it around, so GitHub doesn't show it as a change?

soshial:
I wish we could support showing _Live location_ in Telegram:
* live location is activated when device comes online
* tracked device location inside initial Telegram message is updated with every location update on server
* live location stops when device comes offline

But I guess, it would need more substantial changes, than the one I've submitted.

soshial:
Do you mind adding Telegram to [your documentation](https://www.traccar.org/documentation/notifications/)? You may just copy most of it from #4274.

tananaev:
Can update docs once we merge this.

soshial:
Very well. What do we still have left to finish the merge?

tananaev:
You need to resolve all checks and comments.

soshial:
Could you please help me to parse JSON result with the most suitable JSON library that you have in gradle? I tried this, but can't figure this out. Not that good with server-side Java.

    ObjectMapper mapper = new ObjectMapper();
    JsonNode root = mapper.readTree(o);
    int h = root.path("result").path("message_id").getValueAsInt();

I will fix the rest.

soshial:
By the way, why don't we include exact latitude and longitude in [short messages](https://github.com/traccar/traccar/tree/master/templates/short)? I think, that this information is crucial even in short messages, especially in SMS (when people have no internet connection).

tananaev:
Localized SMS has a limit of 70 characters, so we want to keep it short to fit within that limit.

soshial:
Could you, please, merge and add a commit with parsing of telegram massage ID in response? I have no idea how to do this.

tananaev:
Sorry, but I don't think I have time to work on this. We need to resolve it before merging.

soshial:
I have resolved all issues, and the code compiles and fully works: location is sent as a native Telegram message.
So let's merge this pull request, since I implemented my intentions fully.
If anyone would want to improve how it looks, it should be in a separate PR.

src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -16,6 +16,7 @@
 package org.traccar.notificators;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+
```
tananaev, line 19: Undo

src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -24,39 +25,54 @@
 import org.traccar.model.Position;
 import org.traccar.notification.NotificationFormatter;
 
+import javax.annotation.Nullable;
```
tananaev, line 28: Remove

src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -24,39 +25,54 @@
...
 
-    public static class Message {
+    public static class TextMessage {
         @JsonProperty("chat_id")
         private String chatId;
         @JsonProperty("text")
         private String text;
+        @Nullable
+        @JsonProperty("reply_to_message_id")
+        private Long replyToMessageId;
```
tananaev, line 47: Remove if we don't use yet

src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -70,6 +86,24 @@ public void failed(Throwable throwable) {
...
+    @Override
+    public void sendSync(long userId, Event event, Position position) {
+        if (position != null) {
+            LocationMessage locationMessage = new LocationMessage();
+            locationMessage.chatId = chatId;
+            locationMessage.latitude = position.getLatitude();
+            locationMessage.longitude = position.getLongitude();
+            locationMessage.bearing = (int) Math.ceil(position.getCourse());
+            locationMessage.accuracy = position.getAccuracy();
+            executeRequest(urlSendLocation, locationMessage);
```
tananaev, line 98: Please extract into a separate method.

src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -70,6 +81,28 @@ public void failed(Throwable throwable) {
         });
     }
 
+    private LocationMessage createMessage(Position position) {
```
tananaev, line 84: This should probable be called `createLocationMessage`.

src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -70,6 +81,28 @@ public void failed(Throwable throwable) {
...
+        locationMessage.bearing = (int) Math.ceil(position.getCourse());
+        locationMessage.accuracy = position.getAccuracy();
+        return locationMessage;
+    }
+
+    @Override
+    public void sendSync(long userId, Event event, Position position) {
+        if (position != null) {
+            LocationMessage locationMessage = createMessage(position);
+            executeRequest(urlSendLocation, locationMessage);
```
tananaev, line 98:
```suggestion
            executeRequest(urlSendLocation, createMessage(position));
```

tananaev:
Merged, thanks.

PR #4668, opened 2021-04-23
KEY_DRIVER_UNIQUE_ID is hardcoded on copyAttributes

jcardus:
Why are we forcing KEY_DRIVER_UNIQUE_ID on copyAttributes?
Also, if exists why should it be duplicated?


tananaev:
Merged.

PR #4684, opened 2021-05-24
Fix poweroff alarm and extend megastek protocol

Elsensee:
Fixed that the poweroff alert would trigger a poweron event. Also added more events for possible alert types.

Additionally the external accessories status is now parsed if it exists.

While I found a value of "DD" for the external accessories field for the [MT200x](https://drive.google.com/file/d/1mSwMTGcAmvExJpBDKXzVIUDYc0HE0FWu/view) and [MT60x](https://rrrorwxhqjmklm5m-static.micyjz.com/GPRS-Uploading-Data2015-10-26_11-aidliBpiKkrljSRoikqniiqio.pdf?dp=GvUApKfKKUAU) protocol sheets, I also found [this post](https://www.traccar.org/forums/topic/megastek-gvt-800-unidentified/) where the field seems to be filled with "D-D" while the [GVT800 protocol](https://rrrorwxhqjmklm5m-static.micyjz.com/GVT800-GPRS-Uploading-Data-aidlnBpiKkrljSRoikqpqlmiq.pdf?dp=GvUApKfKKUAU) states "reserved". So to be compliant with any values, I kept the old mechanism and made the new one optional.

Also, the MT200x states some additional values after the last alert event type field, while the MT60x doesn't. I didn't find the semicolon delimiting the message in the decoder so I added it, so that it doesn't get consumed by the alert type.


tananaev:
Can you please add unit tests for all the cases.

Elsensee:
I assume you meant like this? Testing if text yields out some alert event type would probably be unneccessary. But testing the provided examples and testing whether some input would set the right fields to the right values is what I could do. :)

src/test/java/org/traccar/protocol/MegastekProtocolDecoderTest.java
```diff
@@ -95,6 +101,9 @@ public void testDecode() throws Exception {
         verifyPosition(decoder, text(
                 "LOGSTX,123456789012345,$GPRMC,230739.000,A,3841.81895,N,09494.12409,W,0.00,0.00,270914,,,A*70,L,,imei:123456789012345,0/7,269.7,Battery=100%,,0,,,5856,78A3;78"));
         
+        Position testPosition = (Position) decoder.decode(null, null, "$MGV002,860719020193193,DeviceName,R,240214,104742,A,2238.20471,N,11401.97967,E,00,03,00,1.20,0.462,356.23,137.9,1.5,460,07,262C,0F54,25,0000,0000,0,0,0,28.5,28.3,,10,100,Timer;!");
```
tananaev, line 104: We have a special method `verifyAttribute`. See other tests for examples.

tananaev (approved):

src/test/java/org/traccar/protocol/MegastekProtocolDecoderTest.java
Elsensee: Ah right, I knew there had to be a better way! Fixed that. :)

src/test/java/org/traccar/protocol/MegastekProtocolDecoderTest.java
```diff
@@ -94,7 +106,6 @@ public void testDecode() throws Exception {
 
         verifyPosition(decoder, text(
                 "LOGSTX,123456789012345,$GPRMC,230739.000,A,3841.81895,N,09494.12409,W,0.00,0.00,270914,,,A*70,L,,imei:123456789012345,0/7,269.7,Battery=100%,,0,,,5856,78A3;78"));
-
```
tananaev, line 97: Please undo this.

src/main/java/org/traccar/protocol/MegastekProtocolDecoder.java
```diff
@@ -354,6 +358,13 @@ private Position decodeNew(Channel channel, SocketAddress remoteAddress, String
         }
 
         position.set(Position.KEY_DRIVER_UNIQUE_ID, parser.next());
+        
+        if (parser.hasNext()) {
+            position.set(Position.KEY_CHARGE, parser.nextInt() == 1);
```
tananaev, line 363:
```suggestion
            position.set(Position.KEY_CHARGE, parser.nextInt() > 0);
```

src/main/java/org/traccar/protocol/MegastekProtocolDecoder.java
```diff
@@ -275,9 +275,13 @@ private Position decodeOld(Channel channel, SocketAddress remoteAddress, String
             .or().text(" ")
             .groupEnd("?").text(",")
             .number("(d+)?,")                    // rfid
+            .groupBegin()                        // ext accessories
+            .number("([01])")                    // charging
+            .number("(d)?")                      // belt status
+            .groupEnd("?")
```
tananaev, line 281: I would replace it with something like this because it really looks like a single field:

```suggestion
            .number("([01])(d)?,").optional() // charge and belt status
```

src/test/java/org/traccar/protocol/MegastekProtocolDecoderTest.java
```diff
@@ -50,9 +51,20 @@ public void testDecode() throws Exception {
                 "0125$MGV002,860719020193193,DeviceName,R,240214,104742,A,2238.20471,N,11401.97967,E,00,03,00,1.20,0.462,356.23,137.9,1.5,460,07,262C,0F54,25,0000,0000,0,0,0,28.5,28.3,,,100,Timer;"));
 
         verifyPosition(decoder, text(
-                "$MGV002,860719020193193,DeviceName,R,240214,104742,A,2238.20471,N,11401.97967,E,00,03,00,1.20,0.462,356.23,137.9,1.5,460,07,262C,0F54,25,0000,0000,0,0,0,28.5,28.3,,,100,Timer;!"),
```
tananaev, line 53: Was this sample removed?

Elsensee: That was mostly a mistake by me. I restored the old line.

Elsensee:
Should I squash the commits or is it ok like this?

tananaev:
No need to squash, but please fix the CI issue.

Elsensee:
Ok. Fixed it :)

tananaev:
Merged, thank you.

Thiritin:
Thank you for doing this PR. Does that also fix the issue that the property belt is not showing for the MT200X?

Elsensee:
@Thiritin Yes, but it will show as a numerical value. You can look at the specification to see what value means what.

PR #4689, opened 2021-06-03
Support for Type - MicTrack

jaimzj:
Support Type value for MicTrack protocol Helps identify the current mode of data received.


tananaev:
I don't mind adding it, but why do we need to add it in 3 different places? Also, please make sure to fix CI errors.

tananaev:
Fixed and merged.

PR #4697, opened 2021-06-11
Modify NotificatorTelegram.java

RafaelMiquelino:
Add chatId based on "notificationTelegramChatId" user attributes, if available.

According to https://github.com/traccar/traccar/pull/4309

### DIFF
--- a/src/main/java/org/traccar/notificators/NotificatorTelegram.java
+++ b/src/main/java/org/traccar/notificators/NotificatorTelegram.java
@@ -1,5 +1,6 @@
 /*
  * Copyright 2019 - 2021 Anton Tananaev (anton@traccar.org)
+ * Copyright 2021 Rafael Miquelino (rafaelmiquelino@gmail.com)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -93,12 +95,17 @@ private LocationMessage createLocationMessage(Position position) {
 
     @Override
     public void sendSync(long userId, Event event, Position position) {
+        User user = Context.getPermissionsManager().getUser(userId);
         TextMessage message = new TextMessage();
-        message.chatId = chatId;
+        if (user.getAttributes().containsKey("notificationTelegramChatId")) {
```
tananaev, line 100: The key seems to be too long. We can probably cut it down to just `telegramChatId`.

RafaelMiquelino: Would that be the only comment?

tananaev: I think so. Otherwise looks fine to me. One other minor thing is that I would probably just call
```
user.getString("notificationTelegramChatId")
```
instead of checking `containsKey` first.

RafaelMiquelino: Done.

tananaev:
Merged, thank you.

PR #4699, opened 2021-06-17
Take "steps" value from LK frames in Watch protocol

l-e-o-n-c-e:
"Steps" counter value is currently parsed from a complete GPS frame only, but according to the protocol it is also transmitted in all "LK" keepalive frames, as well as the battery level.

Example :
`[3G*123456789*000D*LK,12345,0,98]
`means 12345 steps counted and 98% battery level.
Note : for this value to be incremented, the device should receive a 'PEDO,1' command (and 'PEDO,0' to stop the counter)


tananaev:
Please fix CI errors.

tananaev:
Merged, thank you.

PR #4700, opened 2021-06-17
Make access to permissions cache thread-safe

tobyfoo:
Race condition during API requests concerning devices (e.g. positions, trips) occasionally causes the check if a user has access to a device to fail, requiring admin or manager access, the request fails with a 400 "manager access required".

Tested by adding a Thread.sleep() between the cache clear() and re-population from the database, in refreshDeviceAndGroupPermissions(), and then running a script to exercise the API. First an update to a device, causing refreshDeviceAndGroupPermissions() to be called and then numerous requests reading device, positions and trips. All these call checkDevice(). During the first thread’s sleep the requests failed with 400 manager access required.

Fix was to make access to the device/user permissions cache thread safe. Test results confirm the read requests wait until the refresh task completes, then continue with the request and return a 200 and the expected content.

### DIFF
--- a/src/main/java/org/traccar/database/PermissionsManager.java
+++ b/src/main/java/org/traccar/database/PermissionsManager.java
@@ -38,6 +38,8 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


tananaev:
Merged, thank you.

PR #4706, opened 2021-06-24
Improve compatibility with megastek devices

Elsensee:
Megastek devices send some data which either is not fully compliant with the megastek protocol (by omitting some data) or which is (for our purposes) garbage data, which is only delimited with a single LF (0x0A).

Not recognising the garbage data will lead to giving the wrong data to the ProtocolDecoder.


src/main/java/org/traccar/protocol/MegastekProtocolDecoder.java
```diff
@@ -276,7 +279,7 @@ private Position decodeOld(Channel channel, SocketAddress remoteAddress, String
             .groupEnd("?").text(",")
             .number("(d+)?,")                    // rfid
             .number("([01])(d)?").optional()     // charge and belt status
-            .expression("[^,]*,").optional()
+            .expression("[^,]*,")
```
Elsensee, line 282: I've changed this back, as the comma is necessary and the `*` allows no text to be present as well. However, the comma is necessary still.

src/main/java/org/traccar/protocol/MegastekProtocolDecoder.java
```diff
@@ -141,6 +141,9 @@ private Position decodeOld(Channel channel, SocketAddress remoteAddress, String
 
             beginIndex = endIndex + 2;
             endIndex = sentence.indexOf('*', beginIndex) + 3;
+            if (beginIndex > endIndex) {
+                return null;
+            }
```
tananaev, line 146: Can you please explain this part.

Elsensee: If the `*` is not found in the text, endIndex will be `-1`, resulting in `beginIndex > endIndex`. If that's true, `sentence.substring` will throw an exception. I'd totally be fine to remove this, as traccar continues to run, but it's just not worth to log that error imho.

This happens when either garbage data is sent (as can be seen in the FrameDecoderTest, starting with 0x63...) or if the configuration data is sent, which seems to be happening at least twice a day, starting with `$CMV001`. (It's nowhere documented that this happens but it evidently happens twice a day and whenever the configuration is changed)

(We could totally introduce the configuration data into traccar if there is some place for that, I have no idea. But it seems also possible, that maybe traccar can send some commands via the TCP connection in the specific format, I have yet to test that)

tananaev: OK, let's change it to `endIndex < 0` to make it more clear that what it's for.

Elsensee: Oh yeah, that makes more sense. Fixed.

tananaev:
Merged, thank you.

PR #4712, opened 2021-07-06
Persist Session in Database

jcardus:
This allows some decoupling of the traccar instance, a first step to a scalable arquitecture.

My personal goal here is to have auto scalable traccar instances just for reports, this way a very big report won't take down the main instance.

Another big advantage with this is that when you restart your server your clients won't even realize it. The socket reconnects automatically and gets the same session, so it resumes normally. Before clients would need to login again.


src/main/java/org/traccar/web/WebServer.java
```diff
@@ -172,6 +178,19 @@ private void initApi(Config config, ServletContextHandler servletHandler) {
     }
 
     private void initSessionConfig(Config config, ServletContextHandler servletHandler) {
+        boolean sessionPersisted = config.getBoolean(Keys.WEB_SESSION_PERSISTED);
```
tananaev, line 181: This can probably be inlined.

src/main/java/org/traccar/web/WebServer.java
```diff
@@ -172,6 +178,19 @@ private void initApi(Config config, ServletContextHandler servletHandler) {
...
+        if (sessionPersisted) {
+            DatabaseAdaptor databaseAdaptor = new DatabaseAdaptor();
+            databaseAdaptor.setDatasource(Context.getDataManager().getDataSource());
+            JDBCSessionDataStoreFactory jdbcSessionDataStoreFactory = new JDBCSessionDataStoreFactory();
+            jdbcSessionDataStoreFactory.setDatabaseAdaptor(databaseAdaptor);
+            SessionHandler sessionHandler = new SessionHandler();
+            SessionCache sessionCache = new DefaultSessionCache(sessionHandler);
+            sessionCache.setSessionDataStore(jdbcSessionDataStoreFactory.getSessionDataStore(sessionHandler));
+            sessionHandler.setSessionCache(sessionCache);
+            servletHandler.setSessionHandler(sessionHandler);
```
tananaev, line 191: What are the consequences of creating new `SessionHandler` and `SessionCache` instead of using default ones?

jcardus: the default SessionCache is null

src/main/java/org/traccar/config/Keys.java
```diff
@@ -1198,6 +1198,13 @@ private Keys() {
             "web.sameSiteCookie",
             Collections.singletonList(KeyType.GLOBAL));
 
+    /**
+     * Enables persisting Jetty session to the database
+     */
+    public static final ConfigKey<Boolean> WEB_SESSION_PERSISTED = new ConfigKey<>(
```
tananaev, line 1204: One other minor thing. Can we please rename it to be "persisted session" or "persist session" instead of "session persisted".

tananaev:
Merged, thank you.

marcobag:
Hi @jcardus thank you for your modification, but i've noticed that if enabled expired sessions (logout) are not removed from the database. Is that behaviour normal? I think that high load server will fill up DB in few days.
Thank you.

PR #4717, opened 2021-07-08
bulk permission update

jcardus:


src/main/java/org/traccar/api/resource/PermissionsResource.java
```diff
@@ -17,7 +17,13 @@
 package org.traccar.api.resource;
 
 import java.sql.SQLException;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
```
tananaev, line 26: Let's not use stream API. We are not using it yet and it's probably not really needed here.

src/main/java/org/traccar/api/resource/PermissionsResource.java
```diff
@@ -55,30 +61,61 @@ private void checkPermission(Permission permission, boolean link) {
...
     }
 
+    private void refreshPermissions(List<LinkedHashMap<String, Long>> entities) {
+        Set<String> ownerProperty = new HashSet<>();
+        entities.stream()
+            .map(e -> {
+                try {
+                    return new Permission(e);
+                } catch (ClassNotFoundException classNotFoundException) {
+                    classNotFoundException.printStackTrace();
```
tananaev, line 71: Please use proper logging if needed.

src/main/java/org/traccar/api/resource/PermissionsResource.java
```diff
@@ -55,30 +61,61 @@ private void checkPermission(Permission permission, boolean link) {
...
+                    return null;
+                }
+            })
+            .filter(Objects::nonNull)
+            .filter(p -> ownerProperty.add(p.getOwnerClass().getName() + p.getPropertyClass().getName()))
```
tananaev, line 76: There has to be a better way of doing this based on the original `LinkedHashMap`.

src/main/java/org/traccar/api/resource/PermissionsResource.java
```diff
@@ -55,30 +61,61 @@ private void checkPermission(Permission permission, boolean link) {
...
+            checkPermission(permission, false);
+            Context.getDataManager().linkObject(permission.getOwnerClass(), permission.getOwnerId(),
+                    permission.getPropertyClass(), permission.getPropertyId(), false);
+            LogAction.unlink(getUserId(), permission.getOwnerClass(), permission.getOwnerId(),
+                    permission.getPropertyClass(), permission.getPropertyId());
+        }
+        refreshPermissions(entities);
         return Response.noContent().build();
     }
-
```
tananaev, line 83: revert

jcardus:
would be easier to implement if we reload everything after the bulk update request

tananaev:
Another option is to assume/enforce that all permissions are of the same type.

jcardus:
you mean all with the same owner / property combination?

src/main/java/org/traccar/api/resource/PermissionsResource.java
jcardus: revert what?

jcardus:
Can you check?

src/main/java/org/traccar/api/resource/PermissionsResource.java
```diff
@@ -57,27 +59,48 @@ private void checkPermission(Permission permission, boolean link) {
...
-        Context.getPermissionsManager().refreshPermissions(permission);
+        for (LinkedHashMap<String, Long> entity: entities) {
+            Permission permission = new Permission(entity);
+            checkPermission(permission, true);
+            Context.getDataManager().linkObject(permission.getOwnerClass(), permission.getOwnerId(),
+                    permission.getPropertyClass(), permission.getPropertyId(), true);
+            LogAction.link(getUserId(), permission.getOwnerClass(), permission.getOwnerId(),
+                    permission.getPropertyClass(), permission.getPropertyId());
+        }
+        // we assume all permissions are of same type so we use the first one for refreshing
```
tananaev, line 77: I think we should probably throw an exception if permissions are not the same type.

src/main/java/org/traccar/api/resource/PermissionsResource.java
```diff
@@ -74,13 +76,25 @@ public Response add(List<LinkedHashMap<String, Long>> entities) throws SQLExcept
...
+        if (!entities.isEmpty()) {
+            Permission first = new Permission(entities.get(0));
+            for (LinkedHashMap<String, Long> entity: entities) {
+                Permission permission = new Permission(entity);
+                if (!first.getOwnerClass().equals(permission.getOwnerClass())
+                    || !first.getPropertyClass().equals(permission.getPropertyClass())) {
+                    throw new WebApplicationException(Response.status(Response.Status.BAD_REQUEST).build());
+                }
+            }
+        }
```
tananaev, line 95: Can we do something more elegant like this:

```suggestion
        Set<String> keys = null;
        for (LinkedHashMap<String, Long> entity: entities) {
            if (keys != null & !entity.keySet().equals(keys)) {
                throw new WebApplicationException(Response.status(Response.Status.BAD_REQUEST).build());
            }
            keys = entity.keySet();
        }
```

jcardus: ok

jcardus:
can you check?

tananaev (commented):
One last thing - can we order methods in a consistent manner. One way:

1. checkPermissionTypes
2. bulk add
3. add
4. bulk remove
5. remove

tananaev:
By the way, let me know when it's ready for review. Just pushing updated code doesn't tell me that it's ready.

jcardus:
please check

tananaev:
Merged, thank you.

PR #4725, opened 2021-08-19
Add NT20 and EV02 FIX for GT06 Protocol

riccefarias:
Add NT20 and EV02 FIX for GT06 Protocol that allow to run its devices on isolated ports.

It will fix:
- Battery Level 4000% for Ev02
- Show correctly BATTERY and POWER voltage.
- Allows NT20 to run on 0x22

### DIFF
--- a/src/main/java/org/traccar/protocol/Gt06EvFrameDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06EvFrameDecoder.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2014 - 2018 Anton Tananaev (anton@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


tananaev:
It looks like you duplicated GT06 decoder. Unfortunately we won't be able to accept it like this. It's unmaintainable. We need to find a way to reuse the code.

riccefarias:
In my traccar i was add a flag on GT06 to differ these fixes... But all users prefer to use on isolated ports.

tananaev:
Why would they prefer to use different ports?

riccefarias:
I have no idea, but I made another commit with GT06Fix, this one has everything built in, separated by flags, then you can adjust the original code as you prefer.

tananaev:
Just to confirm, you are not interested in fixing it yourself, right?

riccefarias:
i was fixed it for my use... and renamed files to prevent from conflict with original updates.

tananaev:
OK, thanks. I'll leave it open in case someone is interested in finishing the work on it.

victorgaborges:
works for me.

Issue #4735, opened 2021-08-30
Filter duplicates logic is not ideal


soshial:
**Describe the bug**
Imagine that a tracking device didn't have network available and after network reappears, it starts to re-upload locations in batches. Some of these location previously already had been uploaded, some of them had not. The device can't be 100% sure, since it uses UDP.

Current logic only takes the last location from DB and compares it with the new one.  But this logic would NOT discard a location that was already in the DB a month ago, for example. Can we change the logic here?
https://github.com/traccar/traccar/blob/8ecc4af154ec67a8818a21b96a9f10a74bfc919c/src/main/java/org/traccar/handler/FilterHandler.java#L149
https://github.com/traccar/traccar/blob/8ecc4af154ec67a8818a21b96a9f10a74bfc919c/src/main/java/org/traccar/handler/FilterHandler.java#L71-L81

tananaev:
Filtering only works correctly if your device uploads locations in the correct order.

soshial:
Yes, I understand that. But this scenario (uploads locations in the correct order) doesn't sound quite realistic to me.
Can we change the logic? It shouldn't be too hard to change that.

tananaev:
Why is it not realistic?

soshial:
I explained it in the main message: sometimes there is no coverage and no internet and these parts do not get uploaded at the time they were captured by GPS. How do we upload them to avoid dupicates?

tananaev:
It's doable, but it would mean inconsistency between filtering implementations.

soshial:
I wouldn't say so. Instead of _ultimately last_ position, we should load _relatively last_ position, i.e. the last before `fixTime` (of the position being added).

tananaev:
It will have a pretty big performance impact.

soshial:
Can we create a configurable setting for filtering to switch between ultimately and relatively last position?

tananaev:
Sure. Would you be interested in contributing the change?

PR #4740, opened 2021-09-10
Add support for relative filtering

soshial:
This PR tries to address problem #4735.

### DIFF
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -22,7 +22,6 @@
 
     <entry key='filter.enable'>true</entry>
     <entry key='filter.future'>86400</entry>
-    <entry key='filter.relative'>false</entry>
 
     <entry key='event.ignoreDuplicateAlerts'>true</entry>
     <entry key='processing.computedAttributes.enable'>true</entry>
@@ -53,12 +52,8 @@
         SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime = :time LIMIT 1
     </entry>
 
-    <entry key='database.selectPrevPosition'>
+    <entry key='database.selectPrecedingPosition'>
         SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime lte :time ORDER BY fixTime DESC LIMIT 1
-    </entry>
-
-    <entry key='database.selectNextPosition'>
-        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime gt :time ORDER BY fixTime ASC LIMIT 1
     </entry>
 
     <entry key='database.selectLatestPositions'>
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -54,11 +54,11 @@
     </entry>
 
     <entry key='database.selectPrevPosition'>
-        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND NOT fixTime lt :time LIMIT 1 ORDER BY fixTime DESC
+        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime lte :time ORDER BY fixTime DESC LIMIT 1
     </entry>
 
     <entry key='database.selectNextPosition'>
-        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime gt :time LIMIT 1 ORDER BY fixTime ASC
+        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime gt :time ORDER BY fixTime ASC LIMIT 1
     </entry>
 
     <entry key='database.selectLatestPositions'>
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -49,7 +49,7 @@
     </entry>
 
     <entry key='database.selectPrecedingPosition'>
-        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime lte :time ORDER BY fixTime DESC LIMIT 1
+        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime &lt;= :time ORDER BY fixTime DESC LIMIT 1
     </entry>
 
     <entry key='database.selectLatestPositions'>
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -22,6 +22,7 @@
 
     <entry key='filter.enable'>true</entry>
     <entry key='filter.future'>86400</entry>
+    <entry key='filter.relative'>false</entry>
 
     <entry key='event.ignoreDuplicateAlerts'>true</entry>
     <entry key='processing.computedAttributes.enable'>true</entry>
@@ -46,6 +47,18 @@
 
     <entry key='database.selectPositions'>
         SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime BETWEEN :from AND :to ORDER BY fixTime
+    </entry>
+
+    <entry key='database.selectPositionByTime'>
+        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime = :time LIMIT 1
+    </entry>
+
+    <entry key='database.selectPrevPosition'>
+        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND NOT fixTime lt :time LIMIT 1 ORDER BY fixTime DESC
+    </entry>
+
+    <entry key='database.selectNextPosition'>
+        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime gt :time LIMIT 1 ORDER BY fixTime ASC
     </entry>
 
     <entry key='database.selectLatestPositions'>
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -48,10 +48,6 @@
         SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime BETWEEN :from AND :to ORDER BY fixTime
     </entry>
 
-    <entry key='database.selectPositionByTime'>
-        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime = :time LIMIT 1
-    </entry>
-
     <entry key='database.selectPrecedingPosition'>
         SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime lte :time ORDER BY fixTime DESC LIMIT 1
     </entry>
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -872,8 +872,11 @@ public final class Keys {
             Collections.singletonList(KeyType.GLOBAL));
 
     /**
-     * Filter records by Maximum Speed value in knots. Can be used to filter jumps to far locations even if they're
-     * marked as valid. Shouldn't be too low. Start testing with values at about 25000.
+     * Filter records by Maximum Speed value in knots. Can be used to filter jumps to far locations even if Position
+     * appears valid or if Position `speed` field reported by the device is also within limits. Calculates speed from
+     * the distance to the previous position and the elapsed time.
+     *
+     * Tip: Shouldn't be too low. Start testing with values at about 25000.
      */
     public static final ConfigKey<Integer> FILTER_MAX_SPEED = new ConfigKey<>(
             "filter.maxSpeed",
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -890,11 +890,12 @@ public final class Keys {
             Collections.singletonList(KeyType.GLOBAL));
 
     /**
-     * If FALSE, then filter compares duplicates, distance or period only with the last received location.
-     * Server expects all locations to come sequentially.
+     * If false, the server expects all locations to come sequentially (for each device). Filter checks for duplicates,
+     * distance, speed, or time period only against the location that was last received by server.
      *
-     * If TRUE, then filter compares duplicates, distance or period with Position's fixTime strictly before and after.
-     * Server expects locations to come at random order, since tracking device might go offline.
+     * If true, the server expects locations to come at random order (since tracking device might go offline).
+     * Filter checks for duplicates, distance, speed, or time period against the preceding Position's.
+     * Important: setting to true can cause potential performance issues.
      */
     public static final ConfigKey<Boolean> FILTER_RELATIVE = new ConfigKey<>(
             "filter.relative",
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -887,8 +887,19 @@ public final class Keys {
             Collections.singletonList(KeyType.GLOBAL));
 
     /**
-     * Time limit for the filtering in seconds. If the time difference between last position and a new one is more than
-     * this limit, the new position will not be filtered out.
+     * If FALSE, then filter compares duplicates, distance or period only with the last received location.
+     * Server expects all locations to come sequentially.
+     *
+     * If TRUE, then filter compares duplicates, distance or period with Position's fixTime strictly before and after.
+     * Server expects locations to come at random order, since tracking device might go offline.
+     */
+    public static final ConfigKey<Boolean> FILTER_RELATIVE = new ConfigKey<>(
+            "filter.relative",
+            Collections.singletonList(KeyType.GLOBAL));
+
+    /**
+     * Time limit for the filtering in seconds. If the time difference between the last position was received by server
+     * and a new position is received by server is more than this limit, the new position will not be filtered out.
      */
     public static final ConfigKey<Long> FILTER_SKIP_LIMIT = new ConfigKey<>(
             "filter.skipLimit",
@@ -896,6 +907,7 @@ public final class Keys {
 
     /**
      * Enable attributes skipping. Attribute skipping can be enabled in the config or device attributes.
+     * If position contains any attribute mentioned in "filter.skipAttributes" config key, position is not filtered out.
      */
     public static final ConfigKey<Boolean> FILTER_SKIP_ATTRIBUTES_ENABLE = new ConfigKey<>(
             "filter.skipAttributes.enable",
--- a/src/main/java/org/traccar/database/DataManager.java
+++ b/src/main/java/org/traccar/database/DataManager.java
@@ -329,15 +329,10 @@ public class DataManager {
     }
 
     public Position getPrecedingPosition(long deviceId, Date date) throws SQLException {
-        Collection<Position> positions = QueryBuilder.create(dataSource, getQuery("database.selectPrecedingPosition"))
+        return QueryBuilder.create(dataSource, getQuery("database.selectPrecedingPosition"))
                 .setLong("deviceId", deviceId)
                 .setDate("time", date)
-                .executeQuery(Position.class);
-        if (positions.isEmpty()) {
-            return null;
-        } else {
-            return positions.iterator().next();
-        }
+                .executeQuerySingle(Position.class);
     }
 
     public void updateLatestPosition(Position position) throws SQLException {
--- a/src/main/java/org/traccar/database/DataManager.java
+++ b/src/main/java/org/traccar/database/DataManager.java
@@ -340,20 +340,8 @@ public class DataManager {
         }
     }
 
-    public Position getPrevPosition(long deviceId, Date date) throws SQLException {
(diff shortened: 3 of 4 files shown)


src/main/java/org/traccar/handler/FilterHandler.java
```diff
@@ -167,19 +165,56 @@ private boolean filter(Position position) {
...
 
+                Date newfixTime = position.getFixTime();
+                Position duplicate = Context.getDataManager().getPositionByTime(position.getDeviceId(), newfixTime);
+                if (filterDuplicate(position, duplicate) && !skipLimit(position, duplicate)
+                        && !skipAttributes(position)) {
+                    filterType.append("Duplicate ");
+                }
+
+                Position previous = Context.getDataManager().getPrevPosition(position.getDeviceId(), newfixTime);
+                Position next = Context.getDataManager().getNextPosition(position.getDeviceId(), newfixTime);
```
tananaev, line 182: Can you please provide details on why we need "next" position.

soshial: The backend has configured MinPeriod filter set to 2 seconds. Imagine that these  locations were received by the server:
* Location A: `fixTime = 18:18:01`
* Location C: `fixTime = 18:18:05`

Then after 10 seconds, server receives location B with `fixTime = 18:18:04`. As it turns out, it was not delivered immediately because device was offline for 3 seconds, because it was switching between networks.

Location B is far enough from location A, but it's too close to location C time-wise. Hence, we need to take into account _next_ position as well.

tananaev: It makes sense in theory, but I'm not sure if it makes sense to actually implement it. You basically make 3 expensive SQL requests for every location report. It will cripple pretty much any server that runs any significant number of devices.

soshial: I agree that with thousands of tracking devices this will have deteriorating impact. But for private use it will not be a problem. No doubt, that we should make this option disabled by default.

What do you say, if I introduce this optimization: let's load _next_ and _prev_ positions only in case any of these filters are turned active:
- filterDistance
- filterMaxSpeed
- filterMinPeriod

So if we only duplicate filter is active, than we have only 1 request instead of 3.

tananaev: I'm mostly worried about value of this feature vs cost of maintaining it going forward.

soshial: I thought [you were Okay](https://github.com/traccar/traccar/issues/4735#issuecomment-908474813) with me adding this logic turned OFF by default.

tananaev: Yeah, the discussion was about loading one last position, if I understood it correctly.

soshial: I can rewrite the code so that it uses only 1 sql query:

`SELECT * FROM tc_positions WHERE deviceId = :deviceId AND NOT fixTime <= :time ORDER BY fixTime DESC  LIMIT 1`

Would it be satisfactory?

tananaev: Yes, that makes sense to me. In that case I think the logic would basically stay the same and you just change the way you get "last" position.

soshial: Why do you use `last` to calculate speed, when you may take speed value from Position (if available)?

tananaev: It it used for filtering big jumps, like when device suddenly moves to a 0,0 location. Speed reported by device is not useful in those cases.

soshial: Can we call `skipLimit` and `skipAttributes` only once in handleFilter()?

```
if (skipLimit(position, previous) || skipAttributes(position)) {
    return true;
}
```

soshial: We can also make lazy loading of last / previous position. What do you say? Only in the case when at least one of the corresponding filters is activated.

tananaev: Makes sense to me.

soshial: Check this out:
image

line 149: moved skipAttr to the top
line 176: lazy instantiation of _previous_

tananaev: Looks fine, but please update PR so I can review it more thoroughly.

soshial: OK, I have pushed a new commit.

src/main/java/org/traccar/handler/FilterHandler.java
```diff
@@ -167,21 +168,43 @@ private boolean filter(Position position) {
...
-            filterType.append("Distance ");
-        }
-        if (filterMaxSpeed(position, last)) {
-            filterType.append("MaxSpeed ");
-        }
-        if (filterMinPeriod(position, last)) {
-            filterType.append("MinPeriod ");
+
+        // relative filtering
+        if (filterDuplicate || filterDistance != 0 || filterMaxSpeed != 0 || filterMinPeriod != 0) {
```
tananaev, line 176:
```suggestion
        if (filterDuplicate || filterDistance > 0 || filterMaxSpeed > 0 || filterMinPeriod > 0) {
```

src/main/java/org/traccar/handler/FilterHandler.java
```diff
@@ -167,21 +168,43 @@ private boolean filter(Position position) {
...
+            Position previous = null;
+            if (filterRelative) {
+                try {
+                    Date newFixTime = position.getFixTime();
+                    previous = Context.getDataManager().getPrevPosition(position.getDeviceId(), newFixTime);
+                } catch (SQLException e) {
+                    LOGGER.warn("Error filtering position", e);
```
tananaev, line 183: This should probably be some more relevant message, like "Error retrieving previous position" or something along those lines.

src/main/java/org/traccar/handler/FilterHandler.java
```diff
@@ -167,21 +168,43 @@ private boolean filter(Position position) {
...
+            if (filterMaxSpeed(position, previous)) {
+                filterType.append("MaxSpeed ");
+            }
+            if (filterMinPeriod(position, previous)) {
+                filterType.append("MinPeriod ");
+            }
         }
 
         if (filterType.length() > 0) {
-
```
tananaev, line 184: undo

src/main/java/org/traccar/handler/FilterHandler.java
```diff
@@ -141,23 +146,19 @@ private boolean skipAttributes(Position position) {
     }
 
     private boolean filter(Position position) {
+        if (skipAttributes(position)) {
```
tananaev, line 149:
```suggestion

        if (skipAttributes(position)) {
```

src/main/java/org/traccar/handler/FilterHandler.java
tananaev: We can also fallback to `getLastPosition`.

src/main/java/org/traccar/config/Keys.java
```diff
@@ -887,15 +890,27 @@ private Keys() {
             Collections.singletonList(KeyType.GLOBAL));
 
     /**
-     * Time limit for the filtering in seconds. If the time difference between last position and a new one is more than
-     * this limit, the new position will not be filtered out.
+     * If FALSE, then filter compares duplicates, distance or period only with the last received location.
+     * Server expects all locations to come sequentially.
+     *
+     * If TRUE, then filter compares duplicates, distance or period with Position's fixTime strictly before and after.
+     * Server expects locations to come at random order, since tracking device might go offline.
```
tananaev, line 897: For consistency
```suggestion
     * If false, then filter compares duplicates, distance or period only with the last received location.
     * Server expects all locations to come sequentially.
     *
     * If true, then filter compares duplicates, distance or period with Position's fixTime strictly before and after.
     * Server expects locations to come at random order, since tracking device might go offline.
```

src/main/java/org/traccar/database/DataManager.java
```diff
@@ -328,6 +328,42 @@ public void updateDeviceStatus(Device device) throws SQLException {
...
+                .setDate("time", date)
+                .executeQuery(Position.class);
+        if (positions.isEmpty()) {
+            return null;
+        } else {
+            return positions.iterator().next();
+        }
+    }
+
+    public Position getNextPosition(long deviceId, Date date) throws SQLException {
```
tananaev, line 355: We only need one of these now, right? Also, please use full words, like `previous` instead of `prev`.

src/main/java/org/traccar/config/Keys.java
tananaev: Can you also add a quick note that setting to true can cause potential performance issues.

setup/default.xml
```diff
@@ -22,6 +22,7 @@
 
     <entry key='filter.enable'>true</entry>
     <entry key='filter.future'>86400</entry>
+    <entry key='filter.relative'>false</entry>
```
tananaev, line 25: Don't need to add this if it's already the default.

tananaev (commented):
Looks good. Add a few minor comments, mostly about formatting.

src/main/java/org/traccar/database/DataManager.java
soshial: Can I rename it everywhere to `preceeding`? It's much clearer title

tananaev: Sounds good

setup/default.xml
```diff
@@ -48,6 +48,14 @@
         SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime BETWEEN :from AND :to ORDER BY fixTime
     </entry>
 
+    <entry key='database.selectPositionByTime'>
```
tananaev, line 51: This one should be removed as well, right?

src/main/java/org/traccar/handler/FilterHandler.java
```diff
@@ -142,22 +147,19 @@ private boolean skipAttributes(Position position) {
...
-        Position last = null;
-        if (Context.getIdentityManager() != null) {
-            last = Context.getIdentityManager().getLastPosition(position.getDeviceId());
+        if (skipAttributes(position)) {
+            return false;
         }
 
+        StringBuilder filterType = new StringBuilder();
+
+        // general filtering
```
tananaev, line 156: Can you also please remove this comment.

src/main/java/org/traccar/handler/FilterHandler.java
```diff
@@ -167,17 +169,40 @@ private boolean filter(Position position) {
...
-        if (filterDistance(position, last) && !skipLimit(position, last) && !skipAttributes(position)) {
-            filterType.append("Distance ");
-        }
-        if (filterMaxSpeed(position, last)) {
-            filterType.append("MaxSpeed ");
-        }
-        if (filterMinPeriod(position, last)) {
-            filterType.append("MinPeriod ");
+
+        // relative filtering
```
tananaev, line 176: And this.

setup/default.xml
tananaev: This is not resolved yet, as far as I can tell.

soshial:
It is. I force pushed the commit.

setup/default.xml
```diff
@@ -48,6 +48,10 @@
         SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime BETWEEN :from AND :to ORDER BY fixTime
     </entry>
 
+    <entry key='database.selectPrecedingPosition'>
+        SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime lte :time ORDER BY fixTime DESC LIMIT 1
```
tananaev, line 52: One last thing. I noticed that you are using `lte`. Is it a standard SQL syntax that will work in every SQL server? Can you please provide a reference because I couldn't find any information about it.

soshial: Honestly, I thought that you were using some kind of library to parse SQL stored in XML. Did you figure out correctly, that you wrote it yourself. I couldn't use `<` sign since it's forbidden inside XML tag.

We may use this:
`SELECT * FROM tc_positions WHERE deviceId = :deviceId AND fixTime &lt;= :time ORDER BY fixTime DESC LIMIT 1`
or
`SELECT * FROM tc_positions WHERE deviceId = :deviceId AND NOT (fixTime > :time) ORDER BY fixTime DESC LIMIT 1`

What do you think?

src/main/java/org/traccar/database/DataManager.java
```diff
@@ -328,6 +328,18 @@ public void updateDeviceStatus(Device device) throws SQLException {
...
+    public Position getPrecedingPosition(long deviceId, Date date) throws SQLException {
+        Collection<Position> positions = QueryBuilder.create(dataSource, getQuery("database.selectPrecedingPosition"))
+                .setLong("deviceId", deviceId)
+                .setDate("time", date)
+                .executeQuery(Position.class);
+        if (positions.isEmpty()) {
+            return null;
+        } else {
+            return positions.iterator().next();
+        }
```
tananaev, line 340: Sorry, just noticed another thing. Can we use `executeQuerySingle` here instead?

tananaev (commented):
Also, just to confirm. Have you tested the latest to make sure everything works?

soshial:
I have tested the code and I found a serious problem (as it seems to me), that have existed in traccar even before my pull request. As it turns out, traccar doesn't save fractional TIMESTAMP values, it saves only seconds (not milliseconds). Which means that duplicate comparison didn't work for protocols that send milliseconds:

`position.getFixTime().getTime() == last.getFixTime().getTime()` returns false for seemingly duplicate position (that I send repeatedly). When comparing to last received position (as was implemented before) this problem did not appear. But ifwe take values from DB, it becomes a problem.

Compare:
* the new received position had value `1626692627800`
* and the position from DB has value `1626692627000`

tananaev:
What database are you using?

soshial:
MariaDB.

tananaev:
Is there a universal solution?

soshial:
What do you mean?

tananaev:
I mean any ideas how to fix it via liquibase, so it's supported for add databases?

soshial:
Looks like this should do the thing:

    <modifyDataType tableName="tc_positions" columnName="fixtime" newDataType="TIMESTAMP(6)" />
    <modifyDataType tableName="tc_positions" columnName="servertime" newDataType="TIMESTAMP(6)" />
    <modifyDataType tableName="tc_positions" columnName="devicetime" newDataType="TIMESTAMP(6)" />
    <modifyDataType tableName="tc_events" columnName="servertime" newDataType="TIMESTAMP(6)" />
    <modifyDataType tableName="tc_devices" columnName="lastupdate" newDataType="TIMESTAMP(6)" />

How do check that it works?

tananaev:
You have to check if it works with different database types. At least H2, MySQL, MS SQL and PostgreSQL.

soshial:
Let's switch to another PR? (since it's a different bug fix)

tananaev:
Yep, makes sense to do a separate PR.

soshial:
Besides this separate time issue, I would say everything works for me.

src/main/java/org/traccar/handler/FilterHandler.java
```diff
@@ -54,22 +58,23 @@ public FilterHandler(Config config) {
...
+                || position.getLatitude() < -90 || position.getLongitude() < -180);
     }
 
     private boolean filterZero(Position position) {
         return filterZero && position.getLatitude() == 0.0 && position.getLongitude() == 0.0;
     }
 
     private boolean filterDuplicate(Position position, Position last) {
-        if (filterDuplicate && last != null && position.getFixTime().equals(last.getFixTime())) {
+        if (filterDuplicate && last != null && position.getFixTime().getTime() == last.getFixTime().getTime()) {
```
tananaev, line 77: Why is this needed?

soshial: If we're going to store milli/micro/nanoseconds in the TIMESTAMP field in the DB, then it means we would in fact compare float numbers. So maybe to filter out duplicates we should compare `getTime()` with a specified precision.

tananaev: Sorry, where did you get floating numbers and nanoseconds? As far as I can tell, `equals` does exactly the same thing as you do manually:

https://docs.oracle.com/javase/8/docs/api/java/util/Date.html#equals-java.lang.Object-

soshial: Yes, currently it does the same thing. But, if we make this change https://github.com/traccar/traccar/pull/4746, then database would store time with milliseconds/nanosecond precision (as float). Hence, comparing new position (converted from text) and preceding (from DB) might not be equal in theory. That is why I suggest to do comparison with a certain precision. I can do additional commit for that. Do you agree?

tananaev: I thought you planned to fix database accuracy?

tsmgeek:
Note with moving to fractional seconds you may need to account for `BETWEEN` queries (from UI etc) that may be off by 0.999999 seconds.
Ie between 00:01-00:02 when actually you may have meant `00:01.000000-00:02.999999`.

In some apps at work we just use https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html#MAX in code to always push max microseconds to account for this.

src/main/java/org/traccar/handler/FilterHandler.java
soshial: Let's leave it as it is now. And deal with it in the separate PR.

tananaev: Agree but this change should be reverted because it doesn't really do anything.

soshial: Done.

src/main/java/org/traccar/handler/FilterHandler.java
```diff
@@ -142,22 +147,18 @@ private boolean skipAttributes(Position position) {
 
     private boolean filter(Position position) {
 
-        StringBuilder filterType = new StringBuilder();
-
-        Position last = null;
-        if (Context.getIdentityManager() != null) {
-            last = Context.getIdentityManager().getLastPosition(position.getDeviceId());
+        if (skipAttributes(position)) {
```
tananaev, line 150: Looks like this actually changes the filtering logic. We only want to apply `skipLimit` and `skipAttributes` for certain types of filters.

soshial: Okay. Could you please elaborate? Why should those 2 _skips_ work only for some filters?

tananaev: There was a forum discussion about it if you want more context. Basically there are two categories of filtering. The ones that filter definitely invalid data (e.g. zero coordinates) and the ones that filter excessive data (e.g. duplicates). We always want to filter former, but want to keep latter if it has some important information (e.g. alarms).

soshial: Done. I reverted to the original logic and also preserved almost the same order of filter checks (as it is in current master branch).

src/main/java/org/traccar/handler/FilterHandler.java
```diff
@@ -167,17 +165,37 @@ private boolean filter(Position position) {
...
-        }
-        if (filterMaxSpeed(position, last)) {
-            filterType.append("MaxSpeed ");
-        }
-        if (filterMinPeriod(position, last)) {
-            filterType.append("MinPeriod ");
+
+        // filter out excessive data
+        long deviceId = position.getDeviceId();
+        if (filterDuplicate || filterDistance > 0 || filterMaxSpeed > 0 || filterMinPeriod > 0) {
```
tananaev, line 171: What about `filterStatic`?

soshial: Fixed.

tananaev:
Merged, thank you.

soshial:
Could you please add the new `filter.relative` key to [the documentation](https://www.traccar.org/configuration-file/)?

tananaev:
The plan is to replace that documentation with something that's generated automatically.

PR #4746, opened 2021-09-19
Fix SQL dropping millisecond time precision

soshial:
I have tested the code and I found a serious problem (as it seems to me), that have existed in traccar even before my pull request. As it turns out, traccar doesn't save fractional TIMESTAMP values, it saves only seconds (not milliseconds). Which means that duplicate comparison didn't work for protocols that send milliseconds:

position.getFixTime().getTime() == last.getFixTime().getTime() returns false for seemingly duplicate position (that I send repeatedly). When comparing to last received position (as was implemented before) this problem did not appear (because positions were kept inside IdentityManager in RAM). But if we take values from DB that cuts off milliseconds, it becomes a problem.

Compare:

    the new received position had value 1626692627800
    and the position from DB has value 1626692627000

Honestly, I really think that traccar should save milliseconds in the database. Some usecases demand location data more frequent than once per second.

Something like this: ALTER TABLE tc_positions CHANGE COLUMN fixtime TIMESTAMP(6) and other time-related columns.

I would like to ask other users to test whether their databases also drop millisecond precision.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,6 +44,7 @@ dependencies {
     implementation "commons-codec:commons-codec:1.15"
     implementation "com.h2database:h2:1.4.200"
     implementation "mysql:mysql-connector-java:8.0.26"
+    implementation "org.mariadb.jdbc:mariadb-java-client:2.7.4"
     implementation "org.postgresql:postgresql:42.2.23"
     implementation "com.microsoft.sqlserver:mssql-jdbc:9.4.0.jre11"
     implementation "com.zaxxer:HikariCP:5.0.0"
--- a/schema/changelog-4.0-clean.xml
+++ b/schema/changelog-4.0-clean.xml
@@ -129,7 +129,7 @@
       <column name="uniqueid" type="VARCHAR(128)">
         <constraints nullable="false" unique="true" />
       </column>
-      <column name="lastupdate" type="timestamp" />
+      <column name="lastupdate" type="TIMESTAMP(6)" />
       <column name="positionid" type="INT" />
       <column name="groupid" type="INT" />
       <column name="attributes" type="VARCHAR(4000)" />
@@ -162,7 +162,7 @@
       <column name="type" type="VARCHAR(128)">
         <constraints nullable="false" />
       </column>
-      <column name="servertime" type="TIMESTAMP">
+      <column name="servertime" type="TIMESTAMP(6)">
         <constraints nullable="false" />
       </column>
       <column name="deviceid" type="INT" />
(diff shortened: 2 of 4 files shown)


soshial:
I tried altering a column and all data was gracefully converted into the new format:

`ALTER TABLE `traccar`.`tc_positions` MODIFY COLUMN fixtime TIMESTAMP(6) NOT NULL;`

But new locations are still not saved with millisecond precision. Could you @tananaev  please point in the code where Position is saved to the DB?

tananaev:
Check `DefaultDataHandler`.

soshial:
Looks like Liquibase supports such migrations: https://liquibase.jira.com/browse/CORE-3349

1. What kind of database do you have?
2. Does it currently save milliseconds / microseconds / nanoseconds for you?

tananaev:
Looks like MySQL precision is seconds by default. That's what we use on most of our servers. Looks like default H2 database has a default precision of 6.

tananaev:
I'm also worried about migrations. This will cause a lot of issues on large databases.

tshelver:
Postgresql timestamp default precision is 1 microsecond, which is what I assume Traccar generates.  The positions.device_time is timestamp with no time zone in our DB, so we should not have any impact. 

Our other tracking system generates data in milliseconds, so our downstream impacts (have a consolidation DB for functionality across both systems) are zero.

soshial:
Yes, that would fix the problem. Here I logged PreparedStatement in MariaDB vs MySQL:
* MariaDB Connector/J: (executeUpdate) HikariProxyPreparedStatement@1539062650 wrapping sql : `'UPDATE tc_devices SET lastUpdate = ? WHERE id = ?', parameters : ['2021-10-17 17:33:03.159',7]`
* MySQL Connector/J: (executeUpdate) HikariProxyPreparedStatement@618416190 wrapping com.mysql.cj.jdbc.ClientPreparedStatement: `UPDATE tc_devices SET lastUpdate = '2021-10-17 17:54:51' WHERE id = 7`

soshial:
OK.

Adding `implementation "org.mariadb.jdbc:mariadb-java-client:2.7.4"` dependency and JAR to the lib/ directory was enough to make it work. Do I additionally need to change something in the run/install script?

soshial:
Here's the MariaDB config params:

    <entry key='database.driver'>org.mariadb.jdbc.Driver</entry>
    <entry key='database.url'>jdbc:mariadb://localhost/[YOUR_DB_NAME]?serverTimezone=UTC&amp;useSSL=false&amp;allowMultiQueries=true&amp;autoReconnect=true&amp;useUnicode=yes&amp;characterEncoding=UTF-8&amp;sessionVariables=sql_mode=''</entry>

tananaev:
This thread only existed less than a month. Do you mean the whole project? There's an "unfollow" button at the top of the page.

tananaev:
Test clean installation with master changelog. We need to test at least H2, MySQL, MariaDB and MS SQL. Ideally also Oracle.

gaborgsomogyi:
Even if not listed in your last DB list I've tested it with Postgresql and the PR works fine.

PR #4750, opened 2021-09-30
Update MobilogixProtocolDecoder.java

jcardus:
We are losing data from the events with two digits like T21 and T22 (ignition on and off). Also included the event code

### DIFF
--- a/src/main/java/org/traccar/protocol/MobilogixProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/MobilogixProtocolDecoder.java
@@ -39,7 +39,7 @@ public class MobilogixProtocolDecoder extends BaseProtocolDecoder {
             .text("[")
             .number("(dddd)-(dd)-(dd) ")         // date (yyyymmdd)
             .number("(dd):(dd):(dd),")           // time (hhmmss)
-            .number("Td,")                       // type
+            .number("Td+,")                      // type
             .number("d+,")                       // device type
             .expression("[^,]+,")                // protocol version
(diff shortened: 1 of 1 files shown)


tananaev:
Please provide protocol documentation and add a test case.

jcardus:
https://drive.google.com/file/d/1FFiluchsXko0xypbR-cQBh7zo_ajrKWX/view?usp=sharing

I'll work on the test case

jcardus:
Please tell me when you get the pdf so I can remove it. I don't want to leave it here.

tananaev:
If you don't want to share it here, please send it to me via email. I want to keep a record of it somewhere so I can refer back to check protocol documentation if some issue comes up later.

jcardus:
did you receive the email? Added an example position in the test case.

tananaev:
No, haven't received any emails with a reference to this PR.

tananaev:
Can you please send it to the email address [here](https://github.com/traccar/traccar) instead of my personal address.

jcardus:
I sent an email to the email associated with the user you are using to comment this PR. The link you just wrote is not an email address. I just sent another email (the third one) to anton@traccar.org.

Next time to save us both time please just write the e-mail you want to receive the pdf.

jcardus:
I'm trying to get the type using the parser, just realized we are hardcoding it with 2 characters. But from DateTime why does it move on to serialnumber? I'm not getting that...

tananaev:
Merged, thank you.

PR #4751, opened 2021-10-01
Improve OpenCellId error logging

soshial:
Add error message received from OpenCellId geolocation provider

### DIFF
--- a/src/main/java/org/traccar/geolocation/OpenCellIdGeolocationProvider.java
+++ b/src/main/java/org/traccar/geolocation/OpenCellIdGeolocationProvider.java
@@ -48,14 +48,16 @@ public class OpenCellIdGeolocationProvider implements GeolocationProvider {
                         callback.onSuccess(
                                 json.getJsonNumber("lat").doubleValue(),
                                 json.getJsonNumber("lon").doubleValue(), 0);
-                    } else if (json.containsKey("error")) {
-                        String errorMessage = json.getJsonString("error").getString();
-                        if (json.containsKey("code")) {
-                            errorMessage += " Error code: " + json.getInt("code");
-                        }
-                        callback.onFailure(new GeolocationException(errorMessage));
                     } else {
-                        callback.onFailure(new GeolocationException("Coordinates are missing."));
+                        if (json.containsKey("error")) {
+                            String errorMessage = json.getJsonString("error").getString();
+                            if (json.containsKey("code")) {
+                                errorMessage += " (" + json.getInt("code") + ")";
+                            }
+                            callback.onFailure(new GeolocationException(errorMessage));
+                        } else {
+                            callback.onFailure(new GeolocationException("Coordinates are missing"));
+                        }
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


tananaev:
Can you please provide API response sample.

soshial:
`{"error":"Cell not found","code":1}`

src/main/java/org/traccar/geolocation/OpenCellIdGeolocationProvider.java
```diff
@@ -48,8 +48,14 @@ public void completed(JsonObject json) {
...
+                    } else if (json.containsKey("error")) {
+                        String errorMessage = json.getJsonString("error").getString();
+                        if (json.containsKey("code")) {
+                            errorMessage += " Error code: " + json.getInt("code");
+                        }
+                        callback.onFailure(new GeolocationException(errorMessage));
                     } else {
-                        callback.onFailure(new GeolocationException("Coordinates are missing"));
+                        callback.onFailure(new GeolocationException("Coordinates are missing."));
                     }
```
tananaev, line 59: I would reorganize it into one else statement and then check "error" field inside and fallback to default if missing.

src/main/java/org/traccar/geolocation/OpenCellIdGeolocationProvider.java
```diff
@@ -48,8 +48,14 @@ public void completed(JsonObject json) {
                         callback.onSuccess(
                                 json.getJsonNumber("lat").doubleValue(),
                                 json.getJsonNumber("lon").doubleValue(), 0);
+                    } else if (json.containsKey("error")) {
+                        String errorMessage = json.getJsonString("error").getString();
+                        if (json.containsKey("code")) {
+                            errorMessage += " Error code: " + json.getInt("code");
```
tananaev, line 54: I would recommend including error just in brackets without a label. Otherwise we are risking localization mismatch between original error and this label. I mean something like this:
```
Cell not found (1)
```

src/main/java/org/traccar/geolocation/OpenCellIdGeolocationProvider.java
```diff
@@ -48,8 +48,14 @@ public void completed(JsonObject json) {
...
+                        }
+                        callback.onFailure(new GeolocationException(errorMessage));
                     } else {
-                        callback.onFailure(new GeolocationException("Coordinates are missing"));
+                        callback.onFailure(new GeolocationException("Coordinates are missing."));
```
tananaev, line 58: We don't include full stops usually. Please revert.

src/main/java/org/traccar/geolocation/OpenCellIdGeolocationProvider.java
```diff
@@ -49,7 +49,15 @@ public void completed(JsonObject json) {
                                 json.getJsonNumber("lat").doubleValue(),
                                 json.getJsonNumber("lon").doubleValue(), 0);
                     } else {
-                        callback.onFailure(new GeolocationException("Coordinates are missing"));
+                        if (json.containsKey("error")) {
+                            String errorMessage = json.getJsonString("error").getString();
```
tananaev, line 53:
```suggestion
                            String errorMessage = json.getString("error");
```

tananaev:
Merged, thank you.

PR #4759, opened 2021-10-15
Totem - additional Alarm and Vehicle Status decoding

jinzo:
I added some Alarms and Vehicle Status decoding from the documentation. Cross referenced the AT05, AT07 and AT09 documentations and the Alarms are the same in all of them. The Vehicle Status in decode4 is a bit more complicated - The first few bits (up until bit bit 11) are the same, but then bit positions and meanings change.

I would like to improve this - but not sure how we could differentiate different device types and their Vehicle Status bit mapping. Is there a preferred way of handling this in Traccar? I guess we leave out/remove all the non-static ones? Or we make them "io"s so that the user can configure them how they wish?


tananaev:
Can you please provide documentation.

jinzo:
AT09 link was in the ticket, I can send that via email too. I think I have
some other devices too, will check and send you everything I have.

PR #4766, opened 2021-10-30
Update build.gradle

nixilin:
update mysql connector due to CVE-2021-2471

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,7 +43,7 @@ enforce {
 dependencies {
     implementation "commons-codec:commons-codec:1.15"
     implementation "com.h2database:h2:1.4.200"
-    implementation "mysql:mysql-connector-java:8.0.26"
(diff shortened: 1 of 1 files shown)


tananaev:
Can you please provide details about the vulnerability and how it affects Traccar.

nixilin:
Attackers can make malicious http GET, hex encode it and send to port 5055, which is default port for osmand.port.
This vulnerability directly caused by mysql-connector-java-8.0.26.jar, and it was disclosed half a month ago.
Maybe you can get more info about it by search **CVE-2021-2471 site:github.com**  on google.

tananaev:
> Attackers can make malicious http GET, hex encode it and send to port 5055, which is default port for osmand.port.

Can you elaborate what kind of malicious request that would be? Have you actually done your research to make sure it's applicable to Traccar use case? From what I can tell, the vulnerability is related to XML functions, which we don't use.

PR #4771, opened 2021-11-11
new event according to AVL ID obtained

samueloviedo04:
I have a teltonika brand gps that sends events 247 and 252 and is not reflected in the decoder. Which behaves in the same way as event 253 where it extracts its value and overwrites the event.

image

This information was obtained from:
https://wiki.teltonika-mobility.com/view/Full_AVL_ID_List

### DIFF
--- a/src/main/java/org/traccar/model/Position.java
+++ b/src/main/java/org/traccar/model/Position.java
@@ -21,6 +21,15 @@ import org.traccar.database.QueryIgnore;
 
 public class Position extends Message {
 
+
+    public static final String crash = "crash";
+    public static final String  limited_crash_trace_nc = "limitedCrashTraceNotCalibrate";
+    public static final String limited_crash_trace_c = "limitedCrashTraceCalibrate";
+    public static final String full_crash_trace_cn = "fullCrashTraceNotCalibrate";
+    public static final String full_crash_trace_n = "fullCrashTraceCalibrate";
+    public static final String crash_detected  = "crashDetected";
+    public static final String battery_present = "batteryPresent";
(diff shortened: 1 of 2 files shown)


tananaev:
Have you verified that it's the same for all devices, including old models?

samueloviedo04:
Yes, I have read the wiki and verify that the event code does not conflict with different modes.

tananaev:
Perfect. Please include links or documentation for other models.

samueloviedo04:
Here you have a link of the codes, with the different devices and specification of event codes for each of them
https://wiki.teltonika-mobility.com/view/Full_AVL_ID_List

tananaev:
I don't see any links or information about old Teltonika devices there. For example FM1100, FM2200 and FM3200.

samueloviedo04:
Here you will find the models with the EOL policy (old) and its equivalent.
https://teltonika-gps.com/es/policies-certificates/eol-policy/

For example, in the FM1100 model, its equivalent FMB120, you will get the event codes equal to those set out.
https://wiki.teltonika-gps.com/view/FMB120_Teltonika_Data_Sending_Parameters_ID

Anyway, here I leave you a PDF with the FM1100 manual where it shows the model codes and does not present a problem.
https://www.manualslib.com/manual/1666123/Teltonika-Fm1100.html?page=10

tananaev:
The fact that one model replaces old model, doesn't mean codes are the same. In fact I actually know that they are different. It tells me that you haven't done proper due diligence here.

Linking one document for FM1100 doesn't help either. What about dozens of other models?

PR #4777, opened 2021-12-02
Fix gl200 devices reporting incorrect ignition status

jbschooley:
Config parameter to reverse ignition status in IGL messages from gl200 devices. Some devices report incorrect ignition status; this fixes the reporting of those events.

```xml
<entry key="gl200.ReverseIgnitionIGL">true</entry>
```


tananaev (commented):
I think you should contact the device vendor first to figure out why it's happening.

jbschooley:
I have contacted them, but I’m expecting to wait a while for a response, if I get one at all.

PR #4778, opened 2021-12-03
Added device attribute allowForward

mivtdole:
With this change a user can select per device if the data should be forwarded or not.
In our situation we have a lot of trackers but only the data for a few needs to be forwarded.


src/main/java/org/traccar/model/Device.java
```diff
@@ -149,4 +151,13 @@ public void setDisabled(boolean disabled) {
         this.disabled = disabled;
     }
 
+    @QueryIgnore
+    public boolean getAllowForward() {
```
tananaev, line 155: Please don't add this to the device model.

src/main/java/org/traccar/model/Device.java
```diff
@@ -17,13 +17,15 @@
...
 import java.util.List;
+import java.util.Map;
 
 import org.traccar.database.QueryExtended;
 import org.traccar.database.QueryIgnore;
 
 public class Device extends GroupedModel {
 
     private String name;
+    private static final String ATTRIBUTE_DEVICE_ALLOW_FORWARD = "deviceAllowForward";
```
tananaev, line 28: This shouldn't be here either.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -287,6 +287,11 @@ public void run(Timeout timeout) {
     @Override
     protected Position handlePosition(Position position) {
 
+        Device device = identityManager.getById(position.getDeviceId());
```
tananaev, line 290: You should use attribute lookup method instead.

tananaev:
And other thing. You should probably inverse your attribute because no value should be equivalent to "false". In your case it's "true".

mivtdole:
> 

I wanted to keep the current behaviour for backwards compatibility. Current position updates for every device is always forwarded when you make no choice/add attribute. I can change to return false when attribute does not exist.
Let know what you want?

tananaev:
That's why I'm thinking that the attribute logic should be inversed. You add it to disable forwarding.

src/main/java/org/traccar/model/Device.java
mivtdole: Removed it.

src/main/java/org/traccar/model/Device.java
mivtdole: Removed it

src/main/java/org/traccar/WebDataHandler.java
mivtdole: Added attribute lookup here.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -59,6 +59,7 @@
 
     private static final String KEY_POSITION = "position";
     private static final String KEY_DEVICE = "device";
+    private static final String ATTRIBUTE_DEVICE_ALLOW_FORWARDING = "deviceAllowForwarding";
```
tananaev, line 62: If it's a device attribute, you don't need to add a prefix "device".

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -287,6 +288,15 @@ public void run(Timeout timeout) {
     @Override
     protected Position handlePosition(Position position) {
 
+        Device device = identityManager.getById(position.getDeviceId());
+
+        Map<String, Object> attributes = device.getAttributes();
+        if (attributes.containsKey(ATTRIBUTE_DEVICE_ALLOW_FORWARDING)) {
+            if (!(Boolean) attributes.get(ATTRIBUTE_DEVICE_ALLOW_FORWARDING)) {
```
tananaev, line 295: I don't think you should do a lookup twice here.

tananaev (commented):
Another thing I just thought about... what about even forwarding? It would make sense to filter it as well.

src/main/java/org/traccar/WebDataHandler.java
mivtdole: I did the same as some other attributes for the device

```bash
    data: [{
        key: 'web.reportColor',
        name: Strings.attributeWebReportColor,
        valueType: 'color'
    }, {
        key: 'devicePassword',
        name: Strings.attributeDevicePassword,
        valueType: 'string'
    }, {
        key: 'processing.copyAttributes',
        name: Strings.attributeProcessingCopyAttributes,
        valueType: 'string'
    }, {
        key: 'decoder.timezone',
        name: Strings.sharedTimezone,
        valueType: 'string',
        dataType: 'timezone'
    }, {
        key: 'deviceInactivityStart',
        name: Strings.attributeDeviceInactivityStart,
        valueType: 'number',
        allowDecimals: false,
        minValue: 1
    }, {
        key: 'deviceInactivityPeriod',
        name: Strings.attributeDeviceInactivityPeriod,
        valueType: 'number',
        allowDecimals: false,
        minValue: 1
    }, {
        key: 'deviceAllowForwarding',
        name: Strings.attributeDeviceAllowForwarding,
        valueType: 'boolean'
    }]
```
I can remove the prefix. Let me know which you want?

src/main/java/org/traccar/WebDataHandler.java
mivtdole: Fixed. Doing only one get for attribute and then check if it is null or false

src/main/java/org/traccar/WebDataHandler.java
tananaev: We only use prefix when it's not very obvious what attribute does without prefix. With forwarding it seems pretty straightforward, so I think we should remove prefix.

mivtdole: Removed the prefix.

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -287,6 +288,13 @@ public void run(Timeout timeout) {
     @Override
     protected Position handlePosition(Position position) {
 
+        Device device = identityManager.getById(position.getDeviceId());
+
+        Object allowForwarding = device.getAttributes().get(ATTRIBUTE_DEVICE_ALLOW_FORWARDING);
```
tananaev, line 293: Why not `getString`?

src/main/java/org/traccar/WebDataHandler.java
```diff
@@ -59,6 +59,7 @@
 
     private static final String KEY_POSITION = "position";
     private static final String KEY_DEVICE = "device";
+    private static final String ATTRIBUTE_DEVICE_ALLOW_FORWARDING = "allowForwarding";
```
tananaev, line 62: Let's add spacing

```suggestion

    private static final String ATTRIBUTE_DEVICE_ALLOW_FORWARDING = "allowForwarding";
```

tananaev:
Wait.. it looks like you still haven't reverted the logic.

PR #4779, opened 2021-12-08
Add new IO's

telubato:


src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -110,22 +116,19 @@ private void decodeParameter(Position position, int id, ByteBuf buf, int length)
             case 80:
                 position.set(Position.PREFIX_TEMP + (id - 78), readValue(buf, length, true) * 0.1);
                 break;
-            case 198:
```
tananaev, line 113: Why did you remove some of the IO values?

telubato: IO 198, 199 and 200 are classified with "not in use" according from IO list.

tananaev: OK, so sounds like there's a conflict with old devices. Did you check your new IO values to see if there are any conflicts?

telubato: I checked with the vendor and these IOs are no longer in use on old devices either.

tananaev: How is that possible?

telubato: IO 198, 199, 200 and 201 were used in special firmware for version 3 of Ruptela devices.
At this moment the atual list of IO and all firmwares of Ruptela devices point to:

IO ID - Name
198 - Not in use
199 - Not in use
200 - Not in use
201 - CAN AdBlue level

Thank you.

tananaev: Sounds like they confirmed that it's used in some older devices. Also, what about the ones that you've added?

telubato: The ones I've added now are transversal, that is, they haven't changed in the old devices even with special firmware.

tananaev: OK, can you please provide documentation for both, so we have a reference to check.

telubato: [FMIO list.xlsx](https://github.com/traccar/traccar/files/7678171/FMIO.list.xlsx)
I attached the file of IO list.

tananaev: What about old devices?

telubato: These IOs are transversal for all models like you can see on the file, also for old models (3 gen).

tananaev: What about previous generations?

telubato: There is none difference for previous generations, because they use the same IOs like new models.

tananaev: I just need a document to prove it and we can move forward with the merge.

telubato: The document that I attached before, you can confirm which IO are present in each type of device.
Per example, IOs:

IO29 - Power supply voltage
IO30 - Battery voltage

are common to all devices.
Others IOs corresponds to other IOs that only some models can reported (Ex. RPM).

tananaev:
Merged, thank you.

telubato:
Thank you

PR #4785, opened 2021-12-28
Update test-generator.py

NeftaliYagua:
Upgrade for Python3 compatibility


tananaev:
I think Python 2 is still more common by default.

PR #4787, opened 2021-12-31
Totem - add some basic commands

jinzo:
Added some basic commands that I needed for Totem - willing to extend it to full available list down the line. 
But wanted to open this PR and ask if there is a way to handle SMS and GPRS commands? And which Protocol implements that - as I checked some popular ones and did not find any. Would like to implement either sending via SMS or via GPRS, but need some help/example on how this is handled elsewhere. Thanks!

PR #4790, opened 2022-01-04
Totem - basic support for SMS and GPRS commands

jinzo:
As mentioned in PR #4787 (also contains protocol documentation) I tried my hand at implementing SMS and GPRS commands for the Totem protocol. 

It's not yet finished, still have to test it on real devices, but basic tests look OK. I opened this pull request as I'm not proficient Java developer and am seeking feedback on code/organization. As the base format between modes is quite similar, I organized it so that the actual command get build only in one place, and then formatted for SMS or GPRS sending.

I know you are probably very busy, so I would like to thank you for taking time to review this.

### DIFF
--- a/src/main/java/org/traccar/protocol/TotemProtocol.java
+++ b/src/main/java/org/traccar/protocol/TotemProtocol.java
@@ -25,7 +25,7 @@ import org.traccar.model.Command;
 public class TotemProtocol extends BaseProtocol {
 
     public TotemProtocol() {
-        setSupportedDataCommands(
+        String[] supportedCommands = new String[]{
                 Command.TYPE_CUSTOM,
                 Command.TYPE_REBOOT_DEVICE,
                 Command.TYPE_FACTORY_RESET,
@@ -33,7 +33,12 @@ public class TotemProtocol extends BaseProtocol {
                 Command.TYPE_POSITION_SINGLE,
                 Command.TYPE_ENGINE_RESUME,
                 Command.TYPE_ENGINE_STOP
-        );
+        };
+        setSupportedDataCommands(supportedCommands);
+
+        setTextCommandEncoder(new TotemProtocolSmsEncoder(this));
+        setSupportedTextCommands(supportedCommands);
+
         addServer(new TrackerServer(false, getName()) {
             @Override
             protected void addProtocolHandlers(PipelineBuilder pipeline) {
--- a/src/main/java/org/traccar/protocol/TotemProtocol.java
+++ b/src/main/java/org/traccar/protocol/TotemProtocol.java
@@ -25,7 +25,7 @@ import org.traccar.model.Command;
 public class TotemProtocol extends BaseProtocol {
 
     public TotemProtocol() {
-        String[] supportedCommands = new String[]{
+        setSupportedDataCommands(
                 Command.TYPE_CUSTOM,
                 Command.TYPE_REBOOT_DEVICE,
                 Command.TYPE_FACTORY_RESET,
@@ -33,11 +33,18 @@ public class TotemProtocol extends BaseProtocol {
                 Command.TYPE_POSITION_SINGLE,
                 Command.TYPE_ENGINE_RESUME,
                 Command.TYPE_ENGINE_STOP
-        };
-        setSupportedDataCommands(supportedCommands);
+        );
 
         setTextCommandEncoder(new TotemProtocolSmsEncoder(this));
-        setSupportedTextCommands(supportedCommands);
+        setSupportedTextCommands(
+                Command.TYPE_CUSTOM,
+                Command.TYPE_REBOOT_DEVICE,
+                Command.TYPE_FACTORY_RESET,
+                Command.TYPE_GET_VERSION,
+                Command.TYPE_POSITION_SINGLE,
+                Command.TYPE_ENGINE_RESUME,
+                Command.TYPE_ENGINE_STOP
+        );
 
         addServer(new TrackerServer(false, getName()) {
             @Override
--- a/src/main/java/org/traccar/protocol/TotemProtocolEncoder.java
+++ b/src/main/java/org/traccar/protocol/TotemProtocolEncoder.java
@@ -1,12 +1,11 @@
 /*
- * Copyright 2015 Irving Gonzalez
- * Copyright 2015 - 2019 Anton Tananaev (anton@traccar.org)
+ * Copyright 2016 - 2019 Anton Tananaev (anton@traccar.org)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -16,40 +15,27 @@
  */
 package org.traccar.protocol;
 
-import org.traccar.StringProtocolEncoder;
 import org.traccar.model.Command;
 import org.traccar.Protocol;
+import org.traccar.helper.Checksum;
 
-public class TotemProtocolEncoder extends StringProtocolEncoder {
+public class TotemProtocolEncoder extends TotemProtocolSmsEncoder {
 
     public TotemProtocolEncoder(Protocol protocol) {
         super(protocol);
     }
 
+    private String encodeCommand(String commandString) {
+        String builtCommand = String.format("$$%04dCF%s", 10 + commandString.getBytes().length, commandString);
+        return String.format("%s%02X", builtCommand, Checksum.xor(builtCommand));
+    }
+
     @Override
-    protected Object encodeCommand(Command command) {
+    protected String encodeCommand(Command command) {
 
         initDevicePassword(command, "000000");
 
-        switch (command.getType()) {
-            case Command.TYPE_CUSTOM:
-                return formatCommand(command, "*%s,%s#", Command.KEY_DEVICE_PASSWORD, Command.KEY_DATA);
-            case Command.TYPE_REBOOT_DEVICE:
-                return formatCommand(command, "*%s,006#", Command.KEY_DEVICE_PASSWORD);
-            case Command.TYPE_FACTORY_RESET:
-                return formatCommand(command, "*%s,007#", Command.KEY_DEVICE_PASSWORD);
-            case Command.TYPE_GET_VERSION:
-                return formatCommand(command, "*%s,056#", Command.KEY_DEVICE_PASSWORD);
-            case Command.TYPE_POSITION_SINGLE:
-                return formatCommand(command, "*%s,012#", Command.KEY_DEVICE_PASSWORD);
-            // Assuming PIN 8 (Output C) is the power wire, like manual says but it can be PIN 5,7,8
-            case Command.TYPE_ENGINE_STOP:
-                return formatCommand(command, "*%s,025,C,1#", Command.KEY_DEVICE_PASSWORD);
-            case Command.TYPE_ENGINE_RESUME:
-                return formatCommand(command, "*%s,025,C,0#", Command.KEY_DEVICE_PASSWORD);
-            default:
-                return null;
-        }
+        return encodeCommand(super.getCommandString(command));
     }
 
 }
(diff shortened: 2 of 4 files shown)


src/main/java/org/traccar/protocol/TotemProtocol.java
```diff
@@ -25,15 +25,20 @@
 public class TotemProtocol extends BaseProtocol {
 
     public TotemProtocol() {
-        setSupportedDataCommands(
+        String[] supportedCommands = new String[]{
```
tananaev, line 28: Please just duplicate this in each call. List changes pretty often and most of the time it doesn't match, so makes sense to have a little bit of duplication.

src/main/java/org/traccar/protocol/TotemProtocolEncoder.java
```diff
@@ -1,12 +1,12 @@
...
  * Copyright 2015 Irving Gonzalez
- * Copyright 2015 - 2019 Anton Tananaev (anton@traccar.org)
+ * Copyright 2016 - 2019 Anton Tananaev (anton@traccar.org)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
```
tananaev, line 9: Why are there changes to the header?

src/main/java/org/traccar/protocol/TotemProtocolEncoder.java
```diff
@@ -16,40 +16,27 @@
  */
 package org.traccar.protocol;
 
-import org.traccar.StringProtocolEncoder;
 import org.traccar.model.Command;
 import org.traccar.Protocol;
+import org.traccar.helper.Checksum;
 
-public class TotemProtocolEncoder extends StringProtocolEncoder {
+public class TotemProtocolEncoder extends TotemProtocolSmsEncoder {
```
tananaev, line 23: I don't think it makes sense to extend it. One possible option is to define a static function here that can be reused. I also would recommend defining it here in `TotemProtocolEncoder`.

src/main/java/org/traccar/protocol/TotemProtocolEncoder.java
```diff
@@ -16,40 +16,27 @@
...
     }
 
+    private String encodeCommand(String commandString) {
+        String builtCommand = String.format("$$%04dCF%s", 10 + commandString.getBytes().length, commandString);
+        return String.format("%s%02X", builtCommand, Checksum.xor(builtCommand));
+    }
+
     @Override
-    protected Object encodeCommand(Command command) {
+    protected String encodeCommand(Command command) {
```
tananaev, line 35: Please don't change the return type. It might break some refactoring later on if signature doesn't match.

src/main/java/org/traccar/protocol/TotemProtocolEncoder.java
```diff
@@ -16,40 +16,27 @@
...
+import org.traccar.helper.Checksum;
 
-public class TotemProtocolEncoder extends StringProtocolEncoder {
+public class TotemProtocolEncoder extends TotemProtocolSmsEncoder {
 
     public TotemProtocolEncoder(Protocol protocol) {
         super(protocol);
     }
 
+    private String encodeCommand(String commandString) {
```
tananaev, line 29: This can probably be inlined.

src/test/java/org/traccar/protocol/TotemProtocolEncoderTest.java
```diff
@@ -15,10 +15,24 @@ public void testEncode() throws Exception {
 
         Command command = new Command();
         command.setDeviceId(2);
-        command.setType(Command.TYPE_ENGINE_STOP);
+        command.setType(Command.TYPE_REBOOT_DEVICE);
         command.set(Command.KEY_DEVICE_PASSWORD, "000000");
 
-        assertEquals("*000000,025,C,1#", encoder.encodeCommand(command));
+        assertEquals("$$0020CF000000,0061D", encoder.encodeCommand(command));
```
tananaev, line 21: Please don't change existing unit tests.

src/main/java/org/traccar/protocol/TotemProtocolEncoder.java
jinzo: I based this off of another Encoder at the start - and forgot about the changes to the header/how it will look. I will fix it back like it was in this file at the start :)

src/test/java/org/traccar/protocol/TotemProtocolEncoderTest.java
jinzo: I mean I sorta have to, the current unit test with the refactoring won't work. As the TotemProtocolEncoder is used in GPRS communication (and the existing test is wrong for that) and the TotemProtocolSmsEncoder is used for SMS communication  - same style as the other two protocols that I found that have SMS and GPRS Encoders/commands. So that the *ProtocolSmsEncoder class handles SMS, and *ProtocolEncoder handles GPRS.

tananaev: Are you saying that existing encoder was wrong?

src/main/java/org/traccar/protocol/TotemProtocolEncoder.java
jinzo: The problem with static function is that I cannot use formatCommand, as it's non-static and I would like to avoid duplicating it. Any ideas?

tananaev: That's why I said you can define (a new) static function and reuse it in both places.

src/test/java/org/traccar/protocol/TotemProtocolEncoderTest.java
jinzo: if I understand how this works, the existing encoder worked for SMS, but for sure not for GPRS.

tananaev: It won't be used for SMS, so sounds like what you are saying is that it was incorrect all along.

src/main/java/org/traccar/protocol/TotemProtocolEncoder.java
jinzo: Sorry I'm not sure I follow, if I move the `getCommandString` to TotemProtocolEncoder and declare it as a static function (to reuse it in TotemProtocolSmsEncoder) I get `error: non-static method formatCommand(Command,String,String...) cannot be referenced from a static context`.

Sorry if this is basic Java, I'm more home in non strongly typed languages :D

jinzo:
I implemented most of the feedback apart from the new static function that I could reuse. Looking into that (but tips in the right direction are always welcome)

src/main/java/org/traccar/protocol/TotemProtocolEncoder.java
tananaev: The idea is to do something like this:
```java
class TotemProtocolEncoder {
    static String formatContent(...) { ... }
    Object formatCommand(...) { ... formatContent(...) ... }
}

class TotemProtocolSmsEncoder {
    Object formatCommand(...) { ... TotemProtocolEncoder.formatContent(...) ... }
}
```

src/test/java/org/traccar/protocol/TotemProtocolEncoderTest.java
tananaev: Why was this marked as resolved? This sounds like a big issue.

jinzo: It was wrong all along. The old test (and command implementation) was using the SMS syntax for GPRS communication. Syntax is different between GPRS and SMS commands (not much, but it is) and I did not even notice that with my changes before (previous PR) - as I did not fully understand how this commands work. 

But current tests and setup is correct for GPRS communication (the TotemProtocolEncoder) and for SMS commands (in TotemProtocolSmsEncoder) with the test example/assertion taken from the documentation.

tananaev: The original implementation is here https://github.com/traccar/traccar/pull/1317. @al3x1s I know it's been a long time, but maybe you can provide some input here. Have you tested commands with SMS or GPRS when you implemented it originally?

jinzo: I will be able to do some tests on real devices (AT05 and AT07) soon-ish, probably early next week. But I'm quite sure that it will work, docs are quite clear and it makes sense that the SMS and GPRS format are different (and that GPRS format for sending commands is close to the format for receiving data, there's also special flag/message for GPRS command response that I need to look into).

jinzo: I have finally found some time to test this, I connected a Totem AT05 and Totem AT07 to the PC, launched Totemtek configuration and log utility and send the command. And it works, log example from the device:
`[2022.02.04, 15:20:36]GPRS REC: $$0020CF000000,0061D

[2022.02.04, 15:20:36]CMD restart
[2022.02.04, 15:20:40]Successful
[2022.02.04, 15:20:40]CMD_CODE_RESTART`
So I can confirm that the original setup/code was wrong and that the current GPRS/SMS command implementation works as tested in this unittest file.

src/main/java/org/traccar/protocol/TotemProtocolEncoder.java
```diff
@@ -19,37 +19,49 @@
...
         switch (command.getType()) {
             case Command.TYPE_CUSTOM:
-                return formatCommand(command, "*%s,%s#", Command.KEY_DEVICE_PASSWORD, Command.KEY_DATA);
+                return String.format("%s,%s",
+                                     command.getAttributes().get(Command.KEY_DEVICE_PASSWORD),
+                                     command.getAttributes().get(Command.KEY_DATA)
+                                    );
             case Command.TYPE_REBOOT_DEVICE:
-                return formatCommand(command, "*%s,006#", Command.KEY_DEVICE_PASSWORD);
+                return String.format("%s,006", command.getAttributes().get(Command.KEY_DEVICE_PASSWORD));
```
tananaev, line 38: Why do we need to change these things?

jinzo: Because if we make formatContent function static, then we cannot use formatCommand unfortunately.
That's the issue discussed here: https://github.com/traccar/traccar/pull/4790#discussion_r778360037
So this was the only way I knew how to implement the proposed change/solution.

tananaev: Ah I remember now. It's been a while.

jinzo: Yeah, life moves fast and I did not have time to setup a test instance to test this until today.

src/main/java/org/traccar/protocol/TotemProtocolSmsEncoder.java
```diff
@@ -0,0 +1,37 @@
...
+public class TotemProtocolSmsEncoder extends StringProtocolEncoder {
+
+    public TotemProtocolSmsEncoder(Protocol protocol) {
+        super(protocol);
+    }
+
+    @Override
+    protected Object encodeCommand(Command command) {
+
+        initDevicePassword(command, "000000");
```
tananaev, line 32: Why do we need this?

jinzo: If you mean the initDevicePassword - it sets the password, "000000" is the default password for this device if none is specified. If you mean the whole SmsEncoder, it's because the format of SMS Commands and GPRS commands is a bit different (the "core" is the same, but start and end are different - example checksum is used with GPRS commands but not with GPS ones)

tananaev: Where do you use this password?

jinzo: In formatCommand we get this password from the command for both protocols

tananaev: OK, this is not good. We need to pass the pass word to the static method as an argument.

tananaev: Wait. Ignore what I just said. We are passing it.

tananaev:
Merged, thank you.

PR #4792, opened 2022-01-11
Immobilizer

DemonExposer:
Added support for the immobilizer on the H02 protocol


DemonExposer:
I did not mean to create this pull request here, oops.

PR #4796, opened 2022-01-14
Integration with Public Transport Feeds

engalissongomes:
hello, I would like to suggest implementing a .pb generator for integration with GTFS Realtime public transport feeds. In underdeveloped countries, generic gps would be an excellent solution for tracking public transport, as the investment value is low, I thought that maybe it would be possible to generate this data directly from Traccar, since the .proto file already exists within Traccar and the information required are already generated by the program as shown below:
Google Reference: https://developers.google.com/transit/gtfs-realtime/gtfs-realtime-proto
Example feed: https://data.texas.gov/download/eiei-9rpf/application%2Foctet-stream
Reference material: https://gtfs.org/reference/realtime/v2/


tananaev:
What is this? Why are you sending this a pull request?

PR #4797, opened 2022-01-16
Implement polling protocol support

tananaev:


soshial:
What does this polling do? Does this have to be supported by the device/protocol?

tananaev:
It's when server requests data. Specifically this implementation is to poll an API for data.

PR #4801, opened 2022-01-26
Add an alarm code on Watch protocol

l-e-o-n-c-e:
Some 2G devices send a 00004000 code when the device is removed from its charger. This is very useful as many cheap devices have a strange magnet connector which is very easy to unplug by accident. If not working, the feature might have to be activated by sending AON,1# to the device. Does NOT work on 4G devices which send a "sos" alert in that situation (00010000).

### DIFF
--- a/src/main/java/org/traccar/protocol/WatchProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/WatchProtocolDecoder.java
@@ -89,6 +89,8 @@ public class WatchProtocolDecoder extends BaseProtocolDecoder {
             return Position.ALARM_GEOFENCE_EXIT;
         } else if (BitUtil.check(status, 2)) {
             return Position.ALARM_GEOFENCE_ENTER;
+        } else if (BitUtil.check(status, 14)) {
+            return Position.ALARM_POWER_CUT; // some 2G devices raise this alarm when the device is unplugged from its charger
         } else if (BitUtil.check(status, 16)) {
             return Position.ALARM_SOS;
         } else if (BitUtil.check(status, 17)) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/protocol/WatchProtocolDecoder.java
```diff
@@ -89,6 +89,8 @@ private String decodeAlarm(int status) {
             return Position.ALARM_GEOFENCE_EXIT;
         } else if (BitUtil.check(status, 2)) {
             return Position.ALARM_GEOFENCE_ENTER;
+        } else if (BitUtil.check(status, 14)) {
+            return Position.ALARM_POWER_CUT; // some 2G devices raise this alarm when the device is unplugged from its charger
```
tananaev, line 93: Comment is unnecessary here. Please remove.

tananaev (commented):
Can you please provide protocol documentation.

src/main/java/org/traccar/protocol/WatchProtocolDecoder.java
l-e-o-n-c-e: ok

l-e-o-n-c-e:
Sorry for the 2nd submit, did not see your request to remove the comment.
This value is not documented but I have seen it working on several 2G devices.

tananaev:
Can you please provide protocol documentation.

l-e-o-n-c-e:
> Can you please provide protocol documentation.

As I told you earlier, this value is undocumented in the different docs I have for watch protocol. However I have seen it working in many chinese devices using this protocol.

tananaev:
Are you sure there are no conflicts between devices?

l-e-o-n-c-e:
> Are you sure there are no conflicts between devices?

Yes, I have tested many different devices from different manufacturers and this value is never used, except on some 2G devices.

tananaev:
Merged, thank you.

PR #4804, opened 2022-01-29
support MariaDB

soshial:
Otherwise crashes with `ERROR: Main method error - org.mariadb.jdbc.Driver - ClassNotFoundException (... < DataManager:113 < *:90 < Context:298`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,6 +44,7 @@ dependencies {
     implementation "commons-codec:commons-codec:1.15"
     implementation "com.h2database:h2:2.0.206"
     implementation "mysql:mysql-connector-java:8.0.27"
+    implementation "org.mariadb.jdbc:mariadb-java-client:2.7.4"
     implementation "org.postgresql:postgresql:42.3.1"
     implementation "com.microsoft.sqlserver:mssql-jdbc:9.4.1.jre11"
     implementation "com.zaxxer:HikariCP:5.0.1"


tananaev:
I thought MySQL driver is compatible. Can't you use it?

soshial:
It never worked for me. Had to add this MariaDB JAR manually.

tananaev:
It worked for me before. What's the error?

soshial:
Look at the first message, please.

tananaev:
That's the error when you are trying to use MariaDB, but what's the error when you try MySQL driver?

soshial:
Where do you set this up?

tananaev:
https://www.traccar.org/mysql/

tananaev:
I don't think it's a good argument. If there are problems, please let me know what the problems are and I'll be happy to merge it. Originally MariaDB was supposed to be fully compatible with MySQL tools. If that's still the case, it makes no sense to include two identical drivers. If it's not the case anymore, I'm happy to reassess.

soshial:
I checked it out and traccar works OK with MariaDB via MySQL connector/J.

But I will stick with `mariadb-java-client` anyway. It might very well be more performant than the MySQL counterpart.

tananaev:
OK, I'm going to close it for now, but if you find some evidence that it's better, I'm happy to merge it later.

PR #4811, opened 2022-02-17
Bugfix/postgres orders columns lowercase

grmt:


schema/changelog-4.17.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd"
+  logicalFilePath="changelog-4.17">
```
tananaev, line 7: Why is this 4.17? We don't have version 4.16 yet.

schema/changelog-4.17.xml
```diff
@@ -0,0 +1,16 @@
...
+
+  <changeSet author="author" id="changelog-4.17">
+
+    <renameColumn tableName="tc_orders" oldColumnName="toAddress" newColumnName="toaddress" />
+    <renameColumn tableName="tc_orders" oldColumnName="fromAddress" newColumnName="fromaddress" />
```
tananaev, line 12: I think it might be better to remove and add columns back because depending on the database it might not recognize this as a change.

tananaev (commented):
Also please remove unrelated changes.

schema/changelog-4.17.xml
grmt: Are you sure? 
My master has 4.16, adds disabledreports columns

schema/changelog-4.17.xml
grmt: Wouldn't remove columns also remove data?

tananaev: Yes, but it's not used yet.

tananaev: Another option is to rename to some temporary names. Yes, that's probably even better.

grmt: ah, you can comment faster than I can squash commits.

grmt: done

schema/changelog-4.17.xml
```diff
@@ -0,0 +1,24 @@
...
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd"
+  logicalFilePath="changelog-4.17">
+
+  <changeSet author="author" id="changelog-4.17">
+
+    <renameColumn tableName="tc_orders" oldColumnName="toAddress" newColumnName="toAddressTmp" />
+    <renameColumn tableName="tc_orders" oldColumnName="fromAddress" newColumnName="fromAddressTmp" />
+
+    <addColumn tableName="tc_orders">
```
tananaev, line 14: I meant that we should rename existing columns back to "toaddress" and "fromaddress". Instead of adding new ones.

schema/changelog-4.17.xml
tananaev: Yeah, you should add to that file.

schema/changelog-4.17.xml
grmt: Sorry, I don't understand things now, even after sleeping over the matter. I did the rename to lowercase, then you suggested to create temporary tables, then you suggest to rename again.
With the latest changes in master, at least for postgres, you have to have the lowercase columns. If the db engine does not make a distinction between lower and upper case columns, I guess that adding columns with the lowercase names could be problematic. I think the db engines and liquibase will figure out nicely how to do the rename and as you mentioned the orders are probably not used yet. So I suggest to use  7618d84

tananaev: I suggested renaming to temporary names and then renaming to all lower case. That was my second suggestion all along.

> I think the db engines and liquibase will figure out nicely how to do the rename and as you mentioned the orders are probably not used yet.

I'm pretty it doesn't work in some cases. And that's why I suggested not to do direct renaming like you had in the first place.

tananaev: Yes, double rename.

grmt: check

grmt:
I'm not aware I changed it, just did an interactive rebase. Maybe hitting a github bug.!?
Started from scratch again, same problem.

tananaev:
Very very unlikely there's a bug. Rebasing won't revert it back.

grmt:
Ah, now I understand. I did not realize that 4.16 is the target overall release version rather than just an arbitrary number for database versions. Did not realize that master is actually the new release branch ....

tananaev:
Merged, thank you for the pull request.

PR #4814, opened 2022-02-20
Bugfix/ldap user login

grmt:
Logging in with an ldap user failed with latest master

### DIFF
--- a/src/main/java/org/traccar/database/DataManager.java
+++ b/src/main/java/org/traccar/database/DataManager.java
@@ -153,7 +153,7 @@ public class DataManager {
         } else {
             if (ldapProvider != null && ldapProvider.login(email, password)) {
                 user = ldapProvider.getUser(email);
-                Context.getUsersManager().addItem(user);
+                addLdapUser(user);
                 return user;
             }
         }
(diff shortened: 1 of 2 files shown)


tananaev:
Can you please include only the changes related to this PR. Also provide some context on why it's not working.

grmt:
tc_users table is empty
ldap properly configured
login with valid ldap_user

This is the response you receive in the ui:
ERROR: column "password" of relation "tc_users" does not exist
  Position: 22 - PSQLException (... < QueryBuilder:460 < DatabaseStorage:57 < DataManager:259 < ...)

tracker-engine will try to create user in db, but fails because the Request has an extra "password" column.

With respect to the unrelated changes, you could merge #4811 or otherwise I'll try to remove them tonight.

grmt:
"Now" rather than "tonight".  Time well spent on mastering the git command line and learning about git interactive rebase and forced push

tananaev:
https://github.com/traccar/traccar/pull/4811 is not ready to be merged yet. Not all my comments are addressed and there are also some unrelated changes that need to be removed.

As for the fix here, I don't think it's right. Password field should be excluded using `@QueryIgnore`, not explicitly. Please investigate why it's not excluded.

tananaev:
Please try this fix https://github.com/traccar/traccar/commit/23bbb48eaf69db1c858a7f027d9f7a2fba8796ea.

grmt:
tested it and works fine
so will close this issue

grmt:
thank you

PR #4827, opened 2022-03-22
Decode additional Navtelecom info

yuriy-piskarev:
+ navigation valid state
+ navigation satellites count
+ navigation course
+ device odometer (by navigation)
+ main power
+ internal battery
+ adcs 
+ inputs
+ outs
+ counter inputs
+ frequency inputs
+ engine hours (by device sensors)
+ fuel sensors (rs485 / ble)
+ fuel sensor (rs232)
+ temperature (1-wire / ble)

### DIFF
--- a/src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
@@ -195,6 +195,9 @@ public class NavtelecomProtocolDecoder extends BaseProtocolDecoder {
 
                     for (int j = 0; j < bits.length(); j++) {
                         if (bits.get(j)) {
+
+                            int tmp = 0;                                                                                    // To parse a value by multiple conditions
+
                             switch (j + 1) {
                                 case 1:
                                     position.set(Position.KEY_INDEX, buf.readUnsignedIntLE());
@@ -205,8 +208,12 @@ public class NavtelecomProtocolDecoder extends BaseProtocolDecoder {
                                 case 3:
                                     position.setDeviceTime(new Date(buf.readUnsignedIntLE() * 1000));
                                     break;
+                                case 8:
+                                    tmp = buf.readUnsignedByte();
+                                    position.setValid((tmp & 0b00000010) == 0b00000010);
+                                    position.set(Position.KEY_SATELLITES, ((tmp & 0b11111100) >> 2));
+                                    break;
                                 case 9:
-                                    position.setValid(true);
                                     position.setFixTime(new Date(buf.readUnsignedIntLE() * 1000));
                                     break;
                                 case 10:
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -205,8 +208,12 @@ protected Object decode(
                                 case 3:
                                     position.setDeviceTime(new Date(buf.readUnsignedIntLE() * 1000));
                                     break;
+                                case 8:
+                                    tmp = buf.readUnsignedByte();
+                                    position.setValid((tmp & 0b00000010) == 0b00000010);
```
tananaev, line 213: Use `BitUtil` for checking bits.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -221,6 +228,159 @@ protected Object decode(
...
+                                    position.set(Position.PREFIX_ADC + 5,buf.readUnsignedShortLE() * 0.001);
+                                    break;
+                                case 26:
+                                    position.set(Position.PREFIX_ADC + 6,buf.readUnsignedShortLE() * 0.001);
+                                    break;
+                                case 29:
+                                    tmp = buf.readUnsignedByte();
+                                    position.set(Position.PREFIX_IN + 1, (tmp & 0b00000001) == 0b00000001);
+                                    position.set(Position.PREFIX_IN + 2, (tmp & 0b00000010) == 0b00000010);
+                                    position.set(Position.PREFIX_IN + 3, (tmp & 0b00000100) == 0b00000100);
```
tananaev, line 265: Also use `BitUtil` everywhere here.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -221,6 +228,159 @@ protected Object decode(
...
+                                    position.set("freq" + 2,buf.readUnsignedShortLE());
+                                    break;
+                                case 37:
+                                    position.set("engine_hours",buf.readUnsignedIntLE());
+                                    break;
+                                case 38:
+                                    tmp = buf.readUnsignedShortLE();
+                                    if (tmp < 65500) {                                                                      // Do not write error codes
+                                        position.set("rs485fuel_level" + 1,tmp);
+                                    }
```
tananaev, line 298: I think you can extract the check into a separate function, so it returns null if check fails. That way you can avoid checks every time.

Second thing here is we need a better attribute naming. Currently it doesn't match Traccar naming convention. Why not call it just `fuelN`?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -221,6 +228,159 @@ protected Object decode(
...
+                                    break;
+                                case 44:
+                                    tmp = buf.readUnsignedShortLE();
+                                    if (tmp < 65500) {                                                                      // Do not write error codes
+                                        position.set("rs232fuel_level",tmp);
+                                    }
+                                    break;
+                                case 45:
+                                    tmp = buf.readByte();
+                                    if (tmp != 0x80) {                                                                      // Do not write error codes
```
tananaev, line 338: Same here. Extract checks into a separate function. Also remove the comments.

tananaev (commented):
Also please fix formatting. I see you are missing spacing all over the place. Run checkstyle before pushing your changes.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -195,6 +195,9 @@ protected Object decode(
 
                     for (int j = 0; j < bits.length(); j++) {
                         if (bits.get(j)) {
+
+                            int tmp = 0;                                                                                    // To parse a value by multiple conditions
```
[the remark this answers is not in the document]
yuriy-piskarev: Done

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: Done

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: In the Navtelecom protocol, in the future there will be many more parameters with checks for various values. Basically it is less than or not equal to something. I tried to put it in a function that takes into account all these options, but it turned out to be more cumbersome and in the end I decided to simplify the expression. See if this is a good option.

As for fuel parameters, there are many different parameters in Navtelecom devices that are associated with certain sources of fuel data:
sensor via RS-232 (1 piece)
sensor via RS485 or BLE (up to 16 pieces)
sensor by CAN-bus (2 pieces)
If all this data is simply numbered, then it will be extremely difficult for the end user to understand exactly where the data he needs is.
We have a table describing the "tags" that are visible in the Wialon monitoring system. Basically we work with it. For parameters where the user cannot get confused, I try to choose the names of the traccar system. For parameters where you can get confused, I wanted to use the names given in our google-sheet with a description of the parameters of the exchange protocol (column "I").
https://docs.google.com/spreadsheets/d/1ufU7l2c6_rVm9V4AjNYnKqCbGEwe7gNeYczVO5v85i8/edit#gid=284746210

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: See, maybe this is a good option.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
tananaev: If you want to be more specific with the name, then I would recommend something like `rs485FuelN`.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -221,6 +228,123 @@ protected Object decode(
...
+                                    position.set(Position.PREFIX_TEMP + 6, (value != 0x80) ? value : null);
+                                    break;
+                                case 51:
+                                    value = buf.readByte();
+                                    position.set(Position.PREFIX_TEMP + 7, (value != 0x80) ? value : null);
+                                    break;
+                                case 52:
+                                    value = buf.readByte();
+                                    position.set(Position.PREFIX_TEMP + 8, (value != 0x80) ? value : null);
+                                    break;
```
tananaev, line 347: I just realized something. These can be collapsed into something like this:
```suggestion
                                case 45:
                                ...
                                case 52:
                                    value = buf.readByte();
                                    position.set(Position.PREFIX_TEMP + (j - 45), (value != 0x80) ? value : null);
                                    break;
```
And same with fuel.

yuriy-piskarev: Yes, it looks much more compact. But we need to correct the calculation of the key index. When we get into `case 45`, then `j = 44` and the key index will be `-1`. In order for everything to be calculated correctly, we need to do something like this `(j + 2 - 45)`.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -193,8 +193,11 @@ protected Object decode(
                     Position position = new Position(getProtocolName());
                     position.setDeviceId(deviceSession.getDeviceId());
 
-                    for (int j = 0; j < bits.length(); j++) {
+                    for (int j = 0, l = 2; j < bits.length(); j++, l++) {
```
tananaev, line 196: Why do we need a second variable? Can't we just add 2 to `j`?

yuriy-piskarev: Yes, of course, we can do this:

> In order for everything to be calculated correctly, we need to do something like this `(j + 2 - 45)`.

I just compared both options and it seemed to me that with one more variable, the option looks easier.
And in the rest of the protocol, there are still many parameters that need to be added in this way.

I can remove the variable and replace `l` with `j + 2`.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -205,8 +208,12 @@ protected Object decode(
                                 case 3:
                                     position.setDeviceTime(new Date(buf.readUnsignedIntLE() * 1000));
                                     break;
+                                case 8:
+                                    value = buf.readUnsignedByte();
+                                    position.setValid(BitUtil.check(value, 1));
+                                    position.set(Position.KEY_SATELLITES, ((value & 0b11111100) >> 2));
```
tananaev, line 214: Probably the last thing. It looks like you haven't updated this to use `BitUtil`.

tananaev (commented):
Everything else looks good.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: Oh right, I'll fix it now.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -221,6 +228,73 @@ protected Object decode(
...
+                                case 33:
+                                case 34:
+                                    position.set(Position.PREFIX_COUNT + (j + 2 - 33), buf.readUnsignedIntLE());
+                                    break;
+                                case 35:
+                                case 36:
+                                    position.set("freq" + (j + 2 - 35), buf.readUnsignedShortLE());
+                                    break;
+                                case 37:
+                                    position.set("engine_hours", buf.readUnsignedIntLE());
```
tananaev, line 272: Sorry, just noticed this one. You should use `Position.KEY_HOURS` here.

yuriy-piskarev: Yes, I saw this key, but let's explain why I chose this option and then finally decide which is better...

The devices have 3 engine hours parameters:
1. engine hours from the device according to the ignition sensor (main parameter)
2. engine hours from the device under load (some other mechanism is working)
3. engine hours from CAN-bus

And as in the case of fuel, I thought to use different names.

Maybe for options 1 and 2:
`position.set(Position.KEY_HOURS + xxx, buf.readUnsignedIntLE());`

And for CAN do:
`position.set("canHours", buf.readUnsignedIntLE());`

tananaev: How about `obdHours`? It's going to be consistent with `obdSpeed` etc that we already have.

yuriy-piskarev: In total, we get...

Now committing KEY_HOURS. And in the future, when adding CAN parameters, I will add the "obd" prefix to them, if I do not find an unambiguous key from the existing ones.

tananaev: ok

tananaev:
Merged, thank you.

PR #4828, opened 2022-03-24
pass language param to OpenCage Geocoder

freyfogle:
follows the same logic as Nominatim which also accepts a language param to allow user to see response in their language

### DIFF
--- a/src/main/java/org/traccar/geocoder/OpenCageGeocoder.java
+++ b/src/main/java/org/traccar/geocoder/OpenCageGeocoder.java
@@ -26,9 +26,7 @@ public class OpenCageGeocoder extends JsonGeocoder {
             url = "https://api.opencagedata.com/geocode/v1";
         }
         url += "/json?q=%f,%f&no_annotations=1&key=" + key;
-        if (language == null) {
-            url += "&language=en";
-        } else {
+        if (language != null) {
             url += "&language=" + language;
         }
         return url;
--- a/src/main/java/org/traccar/geocoder/OpenCageGeocoder.java
+++ b/src/main/java/org/traccar/geocoder/OpenCageGeocoder.java
@@ -21,16 +21,21 @@ import javax.json.JsonObject;
 
 public class OpenCageGeocoder extends JsonGeocoder {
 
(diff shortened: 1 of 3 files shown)


freyfogle:
As a follow up question to this PR, how can I submit a PR for this page: https://www.traccar.org/reverse-geocoding/ so I could also show an example of setting the language? Thank you!

src/main/java/org/traccar/geocoder/OpenCageGeocoder.java
```diff
@@ -21,16 +21,21 @@
...
 public class OpenCageGeocoder extends JsonGeocoder {
 
-    private static String formatUrl(String url, String key) {
+    private static String formatUrl(String url, String key, String language) {
         if (url == null) {
             url = "https://api.opencagedata.com/geocode/v1";
         }
         url += "/json?q=%f,%f&no_annotations=1&key=" + key;
+        if (language == null) {
+            url += "&language=en";
```
tananaev, line 30: What if we don't pass with parameter?

src/test/java/org/traccar/geocoder/GeocoderTest.java
```diff
@@ -41,7 +41,7 @@ public void testGisgraphy() {
     @Test
     public void testOpenCage() {
         Geocoder geocoder = new OpenCageGeocoder(
-                "http://api.opencagedata.com/geocode/v1", "SECRET", 0, new AddressFormat());
+                "http://api.opencagedata.com/geocode/v1", "SECRET", "en", 0, new AddressFormat());
```
tananaev, line 44: I think we should test with `null`, like other geocoders.

tananaev:
Website source code is private, so you can't submit a PR to it.

src/test/java/org/traccar/geocoder/GeocoderTest.java
freyfogle: well ideally you would have many tests with all possible values, but sure, whatever you want

freyfogle:
> Website source code is private, so you can't submit a PR to it.

ok, you should probably document geocoder.language then

PR #4829, opened 2022-03-29
update FlexApiProtocolDecoder  

jackymlr:
some info is not required, and 4g network "rsrp" has higher priority than "rssi".


src/main/java/org/traccar/protocol/FlexApiProtocolDecoder.java
```diff
@@ -56,42 +58,59 @@ protected Object decode(
...
+                    position.setTime(new Date(payload.getInt("time") * 1000L));
+                    position.setLatitude(payload.getJsonNumber("lat").doubleValue());
+                    position.setLongitude(payload.getJsonNumber("log").doubleValue());
+                } else {
+                    position.setTime(new Date(payload.getInt("gnss.ts") * 1000L));
+                    position.setLatitude(payload.getJsonNumber("gnss.latitude").doubleValue());
+                    position.setLongitude(payload.getJsonNumber("gnss.longitude").doubleValue());
+                }
+                Optional.ofNullable(payload.getJsonNumber("gnss.altitude"))
+                        .map(JsonNumber::doubleValue).ifPresent(position::setAltitude);
```
tananaev, line 73: Personally I think this format is hard to read. If you want to do the check, I would recommend doing it explicitly.

Are you sure all the parameters you check for null are optional?

src/test/java/org/traccar/protocol/FlexApiProtocolDecoderTest.java
```diff
@@ -33,12 +33,10 @@ public void testDecode() throws Exception {
 
         verifyAttributes(decoder, text(
                 "${\"topic\":\"v1/VF3102021111601/sysinfo/info\",\"payload\":{\"sysinfo.ts\":1637224740,\"sysinfo.model_name\":\"310\",\"sysinfo.oem_name\":\"inhand\",\"sysinfo.serial_number\":\"VF3102021111601\",\"sysinfo.firmware_version\":\"VT3_V1.1.32\",\"sysinfo.product_number\":\"FQ58\",\"sysinfo.description\":\"www.inhand.com.cn\"}}xx"));
-
-        verifyAttributes(decoder, text(
-                "${\"topic\":\"v1/VF3102021111601/io/info\",\"payload\":{\"io.ts\":1637227722,\"io.AI1\":0,\"io.DI1\":1,\"io.DI2\":0,\"io.DI3\":0,\"io.DI4\":0,\"io.DI1_pullup\":0,\"io.DI2_pullup\":0,\"io.DI3_pullup\":0,\"io.DI4_pullup\":0,\"io.DO1\":0,\"io.DO2\":0,\"io.DO3\":0,\"io.IGT\":1}}xx"));
```
tananaev, line 38: Why are you removing a test?

src/main/java/org/traccar/protocol/FlexApiProtocolDecoder.java
jackymlr: I'll go confirm with firmware developer again.

PR #4834, opened 2022-04-11
Change devicelimit userlimit permission to match with LDAP

rushdimohamed09:
When you are linking traccar to an external system when LDAP is enabled, automatically the devicelimit and userlimit is set to 0, which leads the user cannot create any user and well as add any devices. You will need to log to traccar via the UI and manually change the limits in order to create a user or add a new device. 
So I have made the changes the default value as -1 for devicelimit and userlimit on User.java.

### DIFF
--- a/src/main/java/org/traccar/model/User.java
+++ b/src/main/java/org/traccar/model/User.java
@@ -167,7 +167,7 @@ public class User extends ExtendedModel {
         this.expirationTime = expirationTime;
     }
 
-    private int deviceLimit;
+    private int deviceLimit = -1;
 
     public int getDeviceLimit() {
         return deviceLimit;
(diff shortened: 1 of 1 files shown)


rushdimohamed09:
The issue I encountered is when LDAP is enable and integrating Traccar to another system. We are not able to add any new devices using the APIs. Thus devicelimit is autoset to 0. So to add a new device we have to login to traccar UI and then change the devicelimit. That is a violation of integration, it makes pointless of the API. As the system is not having user/group search option APIs. If this option was there, then changing the user permissions will not be required. 
So its like if we can have two new APIs such as filter user by username and filter group by groupName would be more advisable.

tananaev:
For the device limit should use logic like this instead:

https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/api/resource/UserResource.java#L78

As for other concerns, they don't seem to be related to this PR. Happy to discuss separately.

PR #4836, opened 2022-04-12
add id to navtelecom protocol

salamlam:
Found some id added to navtelecom protocol


tananaev:
There are too many issues with this change. Pretty much on every line of code it doesn't follow our standards.

Issue #44, opened 2012-08-26
CT58i Supported @Tracker-Protokoll


urbandigital:
Tracker: SANAV CT58i - The Tracker submits a String via http which looks like this:
imei=XXXXXXXX&rmc=$GPRMC,141109.000,A,4738.1663,N,00748.7344,E,2.23,30.88,260812,,*35,439.6,2.6,AUTO

It could also submit via tcp/udp, but which Port i have to use? You could mention one?

PR #4839, opened 2022-04-20
add 53-69 fields for Navtelecom protocol

yuriy-piskarev:
- add 53-69 fields for Navtelecom protocol;
- fix 37 field (transform seconds to hours).


src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -295,6 +295,84 @@ protected Object decode(
                                     value = buf.readByte();
                                     position.set(Position.PREFIX_TEMP + (j + 2 - 45), (value != 0x80) ? value : null);
                                     break;
+                                case 53:
+                                    value = buf.readUnsignedShortLE();
+                                    if (value == 0x7FFF) {
+                                        position.set(Position.KEY_FUEL_LEVEL + "Level", (Boolean) null);
+                                        position.set(Position.KEY_FUEL_LEVEL, (Boolean) null);
```
tananaev, line 302: What is this for?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -295,6 +295,84 @@ protected Object decode(
...
+                                    } else if (BitUtil.check(value, 7)) {
+                                        position.set(Position.KEY_FUEL_LEVEL + "Level", BitUtil.to(value, 6));
```
tananaev, line 304: Also what's `Position.KEY_FUEL_LEVEL + "Level"`?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -295,6 +295,84 @@ protected Object decode(
...
+                                    break;
+                                case 58:
+                                case 59:
+                                case 60:
+                                case 61:
+                                case 62:
+                                    value = buf.readUnsignedShortLE();
+                                    position.set(
+                                            Position.KEY_AXLE_WEIGHT + (j + 2 - 58), (value != 65535) ? value : null
+                                    );
```
tananaev, line 333: Please keep closing bracket on the previous line... same for other similar places.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: If the value is 0x7FFF, this is an error code, i.e.  the device does not receive information from the CAN bus on fuel.  It is necessary that the server does not process this value.  Neither as a parameter with a percentage value, nor as a parameter with liters.

tananaev: I understand that, but why set it to null?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: In the CAN bus, fuel data can be transmitted in liters or percentages.  For transmission to the server, we use one field.  If the most significant bit is 1, then the remaining bits are the fuel percentage value.  Otherwise, the remaining bits are the fuel value in liters * 0.1.  
Therefore, two separate parameters should be displayed for the user (as battery voltage - in volts and in percent).

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: ok, i'll fix it

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: Maybe I didn't understand how to do it right.  The plan was such that the system would not display fuel values ​​from this message... put null purely by analogy with previous error handling, for example, field #44

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
tananaev: You should just call it "fuelLevel".

yuriy-piskarev: ok, got it

tananaev:
Please also provide the protocol documentation.

IgrekLg:
> Please also provide the protocol documentation.
https://wiki.navtelecom.ru/doc/protocol/ru_protocol_ntcb_last.pdf
https://disk.yandex.ru/i/TWpinu1YPFfn4g

Protocol emulator software (device): 
https://navtelecom.ru/images/software/archive/emulators/NTC_FLEX_Emulator_v1.1.3.zip

yuriy-piskarev:
> Please also provide the protocol documentation.

Hi!

Actual protocol description is in wiki, as Igor said:
https://wiki.navtelecom.ru/doc/protocol/en_protocol_ntcb_last.pdf

And russian version:
https://wiki.navtelecom.ru/doc/protocol/ru_protocol_ntcb_last.pdf
(There may be translation errors in English, the Russian version is the original)

If need some explanations, you can ask support@navtelecom.ru
We will try to answer all questions!

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: Ok, got it, thanks for the clarification.  I'll fix it as I have free time and maybe I'll add the most requested parameters.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -269,7 +269,7 @@ protected Object decode(
                                     position.set("freq" + (j + 2 - 35), buf.readUnsignedShortLE());
                                     break;
                                 case 37:
-                                    position.set(Position.KEY_HOURS, buf.readUnsignedIntLE());
+                                    position.set(Position.KEY_HOURS, buf.readUnsignedIntLE() / 60);
```
tananaev, line 272: Also this is not correct. Hours parameter is stored in milliseconds.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -295,6 +295,81 @@ protected Object decode(
...
+                                    } else if (BitUtil.check(value, 7)) {
+                                        position.set("obdAdBlueLevel", BitUtil.to(value, 6));
+                                        position.set("obdAdBlue", (Boolean) null);
+                                    } else {
+                                        position.set("obdAdBlueLevel", (Boolean) null);
+                                        position.set("obdAdBlue", BitUtil.to(value, 6) / 10);
+                                    }
+                                    break;
+                                case 67:
+                                    position.set("obdHours", buf.readUnsignedIntLE() / 60);
```
tananaev, line 360: This should probably be just the same parameter.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -295,6 +295,81 @@ protected Object decode(
...
+                                    position.set(
+                                            Position.KEY_AXLE_WEIGHT + (j + 2 - 58), (value != 65535) ? value : null);
+                                    break;
+                                case 63:
+                                    value = buf.readUnsignedByte();
+                                    position.set("obdAccelPos", (value != 0xFF) ? value : null);
+                                    break;
+                                case 64:
+                                    value = buf.readUnsignedByte();
+                                    position.set("obdBrakePos", (value != 0xFF) ? value : null);
```
tananaev, line 340: We don't use shortened words for attributes, with some exceptions.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: got it, i'll fix it

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: Replace it to `obdBrakePosition`?

tananaev: Ideally just `brakePosition` and same for acceleration.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
tananaev: Well, we don't have two parameters. I guess we can keep it as `obdHours` if you think the other one is more important. But the value should probably be millis.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -278,11 +291,15 @@ protected Object decode(
                                 case 42:
                                 case 43:
                                     value = buf.readUnsignedShortLE();
-                                    position.set("fuel" + (j + 2 - 38), (value < 65500) ? value : null);
+                                    if (value < 65500) {
+                                        position.set("fuel" + (j + 2 - 38), value);
+                                    }
```
tananaev, line 296: Why are we changing all these?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -293,9 +310,90 @@ protected Object decode(
...
+                                        position.set(Position.PREFIX_TEMP + (j + 2 - 45), value);
+                                    }
+                                    break;
+                                case 53:
+                                    value = buf.readUnsignedShortLE();
+                                    if (value != 0x7FFF) {
+                                        if (BitUtil.check(value, 7)) {
+                                            position.set("fuelLevel", BitUtil.to(value, 6));
+                                        } else {
+                                            position.set(Position.KEY_FUEL_LEVEL, BitUtil.to(value, 6) / 10);
```
tananaev, line 323: Please explain

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -305,6 +403,86 @@ protected Object decode(
...
                                     position.set("fuelTemp" + (j + 2 - 78), (int) buf.readByte());
                                     break;
+                                case 163:
+                                case 164:
+                                case 165:
+                                case 166:
+                                    value = buf.readShortLE();
+                                    if (value != (short) 0x8000) {
+                                        position.set(
+                                                Position.PREFIX_TEMP + "HighPrecision" + (j + 2 - 163),
```
tananaev, line 413: We don't use prefixes like this.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -305,6 +403,86 @@ protected Object decode(
...
+                                case 250:
+                                case 251:
+                                case 252:
+                                    position.set("user4Byte" + (j + 2 - 238), buf.readUnsignedIntLE());
+                                    break;
+                                case 253:
+                                case 254:
+                                case 255:
+                                    position.set("user8Byte" + (j + 2 - 253), buf.readLongLE());
+                                    break;
```
tananaev, line 485: These should probably be handled as generic IO values, similar to what we do with Teltonika for example.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -208,6 +208,19 @@ protected Object decode(
                                 case 3:
                                     position.setDeviceTime(new Date(buf.readUnsignedIntLE() * 1000));
                                     break;
+                                case 4:
+                                    value = buf.readUnsignedByte();
+                                    position.set(Position.KEY_ALARM, BitUtil.check(value, 2));
+                                    int guardMode = BitUtil.between(value, 3, 4);
+                                    position.set(
+                                            Position.KEY_ARMED,
+                                            (0 < guardMode) && (guardMode < 3));
```
tananaev, line 217: The logic here seems overcomplicated for just checking 2 bits.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: It seemed to me so clearer and removed null. You wrote earlier that you shouldn't set parameters to null if the check fails.

tananaev: When you have a lot of parameters, one liners are easier to read. My previous comment was about just assigning it to null. That just doesn't make any sense. What we had here made sense because it's a conditional check.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: The FLEX protocol has 8 temperature sensors with an accuracy of one degree aim. And 4 more pieces are high-precision, with an accuracy of 0.05 degrees. How should I name them to match the style?

tananaev: If you want to use the standard prefix, they should be `Position.PREFIX_TEMP + index`.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: We just have 2 bits allocated for security modes and the description is as follows:
0 - no protection
1 - security
2 - additional security
3 - service

What's the best way to remake?

tananaev: OK, if that's how it works, I guess it makes sense to keep as is. Does it not fit on a single line?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: Understood, return the previous version?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: _Does it not fit on a single line?_
No, it fits, I don’t know why I did it =) I’ll correct it.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: Are you talking about this design?
`position.set(Position.PREFIX_IO + id, readValue(buf, length));`

If yes, then it is more convenient for users who work with our configurator, documentation and who come to you from Wialon to look not at the field number, but at the size and ordinal number of the user parameter.

It's just that the field number is a lesser known value to the user.

Our version is more similar to the design for Galileo:
`position.set("can32BitR" + (tag - 0xdb), buf.readUnsignedIntLE());`
They also have these descriptions in the configurator and documentation.

I would suggest using the current option, what do you think?

tananaev: That should be fine, but I would still move these to default case, so we don't have these massive columns of cases.

yuriy-piskarev:
@tananaev 
Hi!
Try to fix your comments.
And move user parameters to `default` - is it okay?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -208,6 +208,17 @@ protected Object decode(
                                 case 3:
                                     position.setDeviceTime(new Date(buf.readUnsignedIntLE() * 1000));
                                     break;
+                                case 4:
+                                    value = buf.readUnsignedByte();
+                                    position.set(Position.KEY_ALARM, BitUtil.check(value, 2));
```
tananaev, line 213: I just realized that this is incorrect. What alarm is this?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -305,9 +385,42 @@ protected Object decode(
...
+                                        break;
+                                    } else if ((223 <= j) && (j <= 237)) {
+                                        position.set("user2Byte" + (j + 2 - 223), buf.readUnsignedShortLE());
+                                        break;
+                                    } else if ((238 <= j) && (j <= 252)) {
+                                        position.set("user4Byte" + (j + 2 - 238), buf.readUnsignedIntLE());
+                                        break;
+                                    } else if ((253 <= j) && (j <= 255)) {
+                                        position.set("user8Byte" + (j + 2 - 253), buf.readLongLE());
+                                        break;
```
tananaev, line 419: It looks good, but please remove the extra brackets in if statements.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
tananaev: This is still not addressed. What's the difference between fuelLevel and KEY_FUEL_LEVEL?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -208,6 +208,17 @@ protected Object decode(
...
+                                case 4:
+                                    value = buf.readUnsignedByte();
+                                    position.set(Position.KEY_ALARM, BitUtil.check(value, 2));
+                                    int guardMode = BitUtil.between(value, 3, 4);
+                                    position.set(Position.KEY_ARMED, (0 < guardMode) && (guardMode < 3));
+                                    break;
+
+                                case 5:
+                                    value = buf.readUnsignedByte();
+                                    position.set(Position.KEY_ROAMING, BitUtil.check(value, 6));
```
tananaev, line 220: This should be set only if true. We don't want every message to include this flag.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -294,8 +305,77 @@ protected Object decode(
...
+                                    position.set(Position.KEY_OBD_ODOMETER, buf.readFloatLE());
+                                    break;
+                                case 58:
+                                case 59:
+                                case 60:
+                                case 61:
+                                case 62:
+                                    value = buf.readUnsignedShortLE();
+                                    position.set(
+                                            Position.KEY_AXLE_WEIGHT + (j + 2 - 58), (value != 0xFFFF) ? value : null);
```
tananaev, line 341: This cannot be used as a prefix. Use string prefix instead.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: This is a global alert status.  When the device is armed, the triggering of security sensors:
 1. Generates an Alarm event.
 2. Set the alarm flag (this bit).

 Further, all messages will be with the ALARM flag set.

 Reset bit only when disarming.

tananaev: Yeah, but the value cannot be boolean. It has to be one of the valid alarm types:

https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/model/Position.java#L114-L151

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: Ok, I'll fix

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: Mmm..

This flag is more like the state the device has entered.  They can be lengthy.  
And the ALARM keys from your list are more like EVENT, or more precisely, the cause of the event.

Or I'm wrong?

tananaev: If it's in alarm state, you should send one of those for Traccar to handle it correctly.

yuriy-piskarev: Ok, fix it

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: Ok, fix it

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: This string ok?

`obdAxleWeight`

tananaev: You can just use "axleWeight".

yuriy-piskarev: Ok

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -282,7 +295,7 @@ protected Object decode(
                                     break;
                                 case 44:
                                     value = buf.readUnsignedShortLE();
-                                    position.set(Position.KEY_FUEL_LEVEL, (value < 65500) ? value : null);
+                                    position.set("fuel", (value < 65500) ? value : null);
```
tananaev, line 298: Why are we changing this?

yuriy-piskarev:
@tananaev 
fix by comments

And change the name of the fuel level from the CAN bus. It's been mixed with conventional fuel gauges before...

obdFuel - 0.1 Liters
obdFuelLevel - 1%

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: It's been mixed with conventional fuel gauges before...

name like field 44

yuriy-piskarev: It is more correct to consider it just a fuel sensor without a number - it is only one. After it, fuel1 ... fuel16, we can have sensors from RS-485.
And separately it is necessary to select fuel from the CAN bus, I called it obdFuel - it is also the same, but with a variant in liters and percentages.

tananaev: But `Position.KEY_FUEL_LEVEL` is the same key, so this makes no sense.

yuriy-piskarev: Yes, I saw it, but I thought that just "fuel" looks a little shorter.  Change to key?

tananaev: Yes. And can we please avoid any more changes in this PR, so we can finally merge it.

yuriy-piskarev: Ok!

yuriy-piskarev:
@tananaev
Hi! last fix

tananaev:
Merged, thanks.

PR #4860, opened 2022-06-06
Allow websocket connection trough "Basic Authorization" headers

TalissonJunior:
I´m currently using angular as the client app with an implemented HTTP Interceptor to put the "Authorization" headers on each request, 
it would be great if we could connect to the WebSocket using the "Authorization" headers, 

besides.... my session is always getting null value for the "userId", which throws an parse error  on line
```
long userId = (Long) ((HttpSession) req.getSession()).getAttribute(SessionResource.USER_ID_KEY);
```
Allowing the connection trough the header would solve the problem.

### DIFF
--- a/src/main/java/org/traccar/api/security/SecurityRequestFilter.java
+++ b/src/main/java/org/traccar/api/security/SecurityRequestFilter.java
@@ -26,6 +26,7 @@ import org.traccar.model.User;
 import org.traccar.storage.StorageException;
 
 import javax.annotation.security.PermitAll;
+import javax.servlet.*;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.container.ContainerRequestContext;
@@ -33,10 +34,11 @@ import javax.ws.rs.container.ContainerRequestFilter;
 import javax.ws.rs.container.ResourceInfo;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.SecurityContext;
+import java.io.IOException;
 import java.lang.reflect.Method;
 import java.nio.charset.StandardCharsets;
 
-public class SecurityRequestFilter implements ContainerRequestFilter {
+public class SecurityRequestFilter implements ContainerRequestFilter, Filter {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SecurityRequestFilter.class);
 
@@ -73,19 +75,18 @@ public class SecurityRequestFilter implements ContainerRequestFilter {
         try {
 
             String authHeader = requestContext.getHeaderString(AUTHORIZATION_HEADER);
-            if (authHeader != null) {
 
+            if (authHeader != null) {
                 try {
-                    String[] auth = decodeBasicAuth(authHeader);
-                    User user = Context.getPermissionsManager().login(auth[0], auth[1]);
-                    if (user != null) {
-                        Main.getInjector().getInstance(StatisticsManager.class).registerRequest(user.getId());
-                        securityContext = new UserSecurityContext(new UserPrincipal(user.getId()));
+                    Long userId = GetUserIdFromBasicAuth(authHeader);
+
+                    if(userId != null) {
+                        Main.getInjector().getInstance(StatisticsManager.class).registerRequest(userId);
+                        securityContext = new UserSecurityContext(new UserPrincipal(userId));
                     }
                 } catch (StorageException e) {
                     throw new WebApplicationException(e);
                 }
-
             } else if (request.getSession() != null) {
 
                 Long userId = (Long) request.getSession().getAttribute(SessionResource.USER_ID_KEY);
@@ -116,4 +117,38 @@ public class SecurityRequestFilter implements ContainerRequestFilter {
 
     }
 
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+        HttpServletRequest httpRequest = (HttpServletRequest) request;
+        String authHeaders = httpRequest.getQueryString();
+        String queryAuthName = AUTHORIZATION_HEADER + "=";
+
+        if (authHeaders != null && authHeaders.contains(queryAuthName)) {
+            try {
+                Long userId = GetUserIdFromBasicAuth(authHeaders.replace(queryAuthName, ""));
+
+                if (userId != null) {
+                    httpRequest.getSession().setAttribute(SessionResource.USER_ID_KEY, userId);
+                }
+            }
+            catch (StorageException e) {}
+        }
+
+        chain.doFilter(httpRequest, response);
+    }
+
+    private Long GetUserIdFromBasicAuth(String authHeaders) throws StorageException {
+        if(authHeaders == null) {
+            return null;
+        }
+
+        String[] auth = decodeBasicAuth(authHeaders);
+        User user = Context.getPermissionsManager().login(auth[0], auth[1]);
+
+        if (user != null) {
+            return user.getId();
+        }
+
+        return null;
+    }
 }
--- a/src/main/java/org/traccar/api/AsyncSocketServlet.java
+++ b/src/main/java/org/traccar/api/AsyncSocketServlet.java
@@ -20,9 +20,15 @@ import org.eclipse.jetty.websocket.server.JettyWebSocketServletFactory;
 import org.traccar.Context;
 import org.traccar.api.resource.SessionResource;
 import org.traccar.config.Keys;
+import org.traccar.model.User;
+import org.traccar.storage.StorageException;
 
 import javax.servlet.http.HttpSession;
+import javax.ws.rs.WebApplicationException;
 import java.time.Duration;
+
+import static org.traccar.api.security.SecurityRequestFilter.AUTHORIZATION_HEADER;
+import static org.traccar.api.security.SecurityRequestFilter.decodeBasicAuth;
 
 public class AsyncSocketServlet extends JettyWebSocketServlet {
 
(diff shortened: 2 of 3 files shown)


TalissonJunior:
I tried to make the websocket call ```SecurityRequestFilter.filter```but it doesn´t seems to work, and this function requires a parameter of ```ContainerRequestContext``` which is different from the websocket ```JettyServerUpgradeRequest```, even a cast would be a problem...

But an alternative would be moving the repeat code into a static function ```GetUserIdFromHeaders``,
so the full code would be like: 

SecurityRequestFilter
```
 @Override
    public void filter(ContainerRequestContext requestContext) {

        if (requestContext.getMethod().equals("OPTIONS")) {
            return;
        }

        SecurityContext securityContext = null;

        try {

            String authHeader = requestContext.getHeaderString(AUTHORIZATION_HEADER);

            if (authHeader != null) {
                try {
                    Long userId = GetUserIdFromHeaders(authHeader);

                    if(userId != null) {
                        Main.getInjector().getInstance(StatisticsManager.class).registerRequest(userId);
                        securityContext = new UserSecurityContext(new UserPrincipal(userId));
                    }
                } catch (StorageException e) {
                    throw new WebApplicationException(e);
                }
            } else if (request.getSession() != null) {

                Long userId = (Long) request.getSession().getAttribute(SessionResource.USER_ID_KEY);
                if (userId != null) {
                    Context.getPermissionsManager().checkUserEnabled(userId);
                    Main.getInjector().getInstance(StatisticsManager.class).registerRequest(userId);
                    securityContext = new UserSecurityContext(new UserPrincipal(userId));
                }

            }

        } catch (SecurityException e) {
            LOGGER.warn("Authentication error", e);
        }

        if (securityContext != null) {
            requestContext.setSecurityContext(securityContext);
        } else {
            Method method = resourceInfo.getResourceMethod();
            if (!method.isAnnotationPresent(PermitAll.class)) {
                Response.ResponseBuilder responseBuilder = Response.status(Response.Status.UNAUTHORIZED);
                if (!XML_HTTP_REQUEST.equals(request.getHeader(X_REQUESTED_WITH))) {
                    responseBuilder.header(WWW_AUTHENTICATE, BASIC_REALM);
                }
                throw new WebApplicationException(responseBuilder.build());
            }
        }

    }

    public static Long GetUserIdFromHeaders(String authHeaders) throws StorageException {
        if(authHeaders == null) {
            return null;
        }

        String[] auth = decodeBasicAuth(authHeaders);
        User user = Context.getPermissionsManager().login(auth[0], auth[1]);

        if (user != null) {
            return user.getId();
        }

        return null;
    }
```

AsyncSocketServlet
```
@Override
    public void configure(JettyWebSocketServletFactory factory) {
        factory.setIdleTimeout(Duration.ofMillis(Context.getConfig().getLong(Keys.WEB_TIMEOUT)));

        factory.setCreator((req, resp) -> {
            String authHeader = req.getHeader(AUTHORIZATION_HEADER);

            if (authHeader != null) {
                try {
                    Long userId = SecurityRequestFilter.GetUserIdFromHeaders(authHeader);

                    if(userId == null) {
                        return null;
                    }
                    return new AsyncSocket(userId);
                } catch (StorageException e) {
                    throw new WebApplicationException(e);
                }
            } else {
                if (req.getSession() != null) {
                    long userId = (Long) ((HttpSession) req.getSession()).getAttribute(SessionResource.USER_ID_KEY);
                    return new AsyncSocket(userId);
                } else {
                    return null;
                }
            }
        });
    }
```

I have tested, and it works the same way as the before, i can make a pull request for a better preview...

TalissonJunior:
i see, well we could use a filter with a little different approach, since i coudn´t reach the "Authorization Headers" trough it

so basically on the client app we would connect to the socket using a query parameter 
ex: url "../api/socket?Authorization=MyBasicAuth",


Then the file ```FilterSocket```(New File) which contains 
```
 @Override
    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {

        HttpServletRequest httpRequest = (HttpServletRequest) request;
        String authHeaders = httpRequest.getQueryString();
        String queryAuthName = AUTHORIZATION_HEADER + "=";

        if (authHeaders != null && authHeaders.contains(queryAuthName)) {
            String[] auth = decodeBasicAuth(authHeaders.replace(queryAuthName, ""));
            if(auth != null && auth.length == 2) {

                try {
                    User user = Context.getPermissionsManager().login(auth[0], auth[1]);

                    if (user != null) {
                        httpRequest.getSession().setAttribute(SessionResource.USER_ID_KEY, user.getId());
                    }
                }
                catch (StorageException e) {}
            }
        }

        chain.doFilter(httpRequest, response);
    }
```
would get the id of the user based on the authorization and set it to the session, before calling the websocket  ```AsyncSocketServlet``` which would be the same as it was before

AsyncSocketServlet
```
 @Override
    public void configure(JettyWebSocketServletFactory factory) {
        factory.setIdleTimeout(Duration.ofMillis(Context.getConfig().getLong(Keys.WEB_TIMEOUT)));
        factory.setCreator((req, resp) -> {
            if (req.getSession() != null) {
                long userId = (Long) ((HttpSession) req.getSession()).getAttribute(SessionResource.USER_ID_KEY);
                return new AsyncSocket(userId);
            } else {
                return null;
            }
        });
    }
```

What do you think?
note: the query parameter is not required

tananaev:
> since i coudn´t reach the "Authorization Headers" trough it

So, do you get a filter call for the websocket connection?

tananaev:
I think you misunderstood me. I want to reuse existing filter.

TalissonJunior:
i see, when with existing filter you mean "SecurityRequestFilter" right ?

tananaev:
Correct.

TalissonJunior:
ok i got it, then yeah, i removed the file that i created and implement the filter on the ```SecurityRequestFilter``` as the folowing code shows: 

```
 @Override
    public void filter(ContainerRequestContext requestContext) {

        if (requestContext.getMethod().equals("OPTIONS")) {
            return;
        }

        SecurityContext securityContext = null;

        try {

            String authHeader = requestContext.getHeaderString(AUTHORIZATION_HEADER);

            if (authHeader != null) {
                try {
                    Long userId = GetUserIdFromBasicAuth(authHeader);

                    if(userId != null) {
                        Main.getInjector().getInstance(StatisticsManager.class).registerRequest(userId);
                        securityContext = new UserSecurityContext(new UserPrincipal(userId));
                    }
                } catch (StorageException e) {
                    throw new WebApplicationException(e);
                }
            } else if (request.getSession() != null) {

                Long userId = (Long) request.getSession().getAttribute(SessionResource.USER_ID_KEY);
                if (userId != null) {
                    Context.getPermissionsManager().checkUserEnabled(userId);
                    Main.getInjector().getInstance(StatisticsManager.class).registerRequest(userId);
                    securityContext = new UserSecurityContext(new UserPrincipal(userId));
                }

            }

        } catch (SecurityException e) {
            LOGGER.warn("Authentication error", e);
        }

        if (securityContext != null) {
            requestContext.setSecurityContext(securityContext);
        } else {
            Method method = resourceInfo.getResourceMethod();
            if (!method.isAnnotationPresent(PermitAll.class)) {
                Response.ResponseBuilder responseBuilder = Response.status(Response.Status.UNAUTHORIZED);
                if (!XML_HTTP_REQUEST.equals(request.getHeader(X_REQUESTED_WITH))) {
                    responseBuilder.header(WWW_AUTHENTICATE, BASIC_REALM);
                }
                throw new WebApplicationException(responseBuilder.build());
            }
        }

    }

    @Override
    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
        HttpServletRequest httpRequest = (HttpServletRequest) request;
        String authHeaders = httpRequest.getQueryString();
        String queryAuthName = AUTHORIZATION_HEADER + "=";

        if (authHeaders != null && authHeaders.contains(queryAuthName)) {
            try {
                Long userId = GetUserIdFromBasicAuth(authHeaders.replace(queryAuthName, ""));

                if (userId != null) {
                    httpRequest.getSession().setAttribute(SessionResource.USER_ID_KEY, userId);
                }
            }
            catch (StorageException e) {}
        }

        chain.doFilter(httpRequest, response);
    }

    private Long GetUserIdFromBasicAuth(String authHeaders) throws StorageException {
        if(authHeaders == null) {
            return null;
        }

        String[] auth = decodeBasicAuth(authHeaders);
        User user = Context.getPermissionsManager().login(auth[0], auth[1]);

        if (user != null) {
            return user.getId();
        }

        return null;
    }
```

tananaev:
You need to send a PR because I don't know what the difference is.

src/main/java/org/traccar/api/security/SecurityRequestFilter.java
```diff
@@ -116,4 +117,38 @@ public void filter(ContainerRequestContext requestContext) {
 
     }
 
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+        HttpServletRequest httpRequest = (HttpServletRequest) request;
+        String authHeaders = httpRequest.getQueryString();
```
tananaev, line 123: Why can't we use headers?

TalissonJunior: it seems that browsers don´t support sending headers with websocket, you can only send "protocols" or attach a cookie before calling the websocket
```
var ws = new WebSocket("ws://example.com/service", "soap");
```

tananaev:
What about your original solution? It didn't work? I'm reading the title and description of the PR and it doesn't make sense to me if it's not possible in the first place.

TalissonJunior:
i changed to get the authorization from headers as you suggest, and it wasn´t working , until i type the url : "api/devices" on the browser and enter my credentials, now i´m getting the authorization from headers...

tananaev:
I'm confused. Are you saying your original PR didn't use headers?

TalissonJunior:
I´m saying that my original PR did use headers, but it shouldn´t since browsers doesn´t allow it, so in my opinion we should close this PR, or allow Basic authorization trough queryParameter when working with socket.

I tried to use "the basic authorization" because the logic to get the userId from the current session was not working on my case, i was getting "WebSocket connection to 'ws://localhost:8082/api/socket' failed"  (UserId from session was always null)

Allowing basic authorization trought queryParameters would make this problem go away.

tananaev:
So, your original code didn't work?

Let's close it for now.

PR #4874, opened 2022-06-27
Fix for watch protocol with zero base stations and no TA

wkhaksar:
The issue discussion started at Traccar forum [here](https://www.traccar.org/forums/topic/watch-protocol-numberformatexception/#post-72333). 

The documentation is attached [here](https://github.com/traccar/traccar/files/8994431/Communication.protocol-V3.0_20210423.pdf).

The problem was because of timing advance, as it was still retrieved even if base station information was not attached or base stations count was zero.

If there is no base station information attached or base station number is zero then it means there is no TA as well, as TA is a part of base stations information. 

So, TA moved to inner block.

To prevent failing other test cases, a check for being empty is applied for mac address and signal strength of WiFi stations.


src/main/java/org/traccar/protocol/WatchProtocolDecoder.java
```diff
@@ -164,8 +164,11 @@ private Position decodePosition(DeviceSession deviceSession, String data) {
 
             for (int i = 0; i < wifiCount; i++) {
                 index += 1; // wifi name
-                network.addWifiAccessPoint(WifiAccessPoint.from(
-                        values[index++], Integer.parseInt(values[index++])));
+                String macAddress = values[index++];
+                String rssi = values[index++];
+               if (!macAddress.isEmpty() && !rssi.isEmpty()) {
+                   network.addWifiAccessPoint(WifiAccessPoint.from(macAddress, Integer.parseInt(rssi)));
+               }
```
tananaev, line 171: This part doesn't make sense to me. I also can't find anything about it in the documentation you provided. Can you please point to the page in the doc. Also, please attache the documentation to this ticket.

wkhaksar: The documentation is now attached.

The reason for this conditional check is not something to be mentioned from new documentation, but I added this because some tests are failing when we move timing advance into the inner block.

For example, bellow test at line 33 in WatchProtocolDecoderTest.java fails:
````
                "[SG*9059056143*0053*UD,251021,223408,A,41.46500,N,081.53128,W,0.926,000,0,00,70,70,0,50,00000000,0,1,,,,00]"));
````
I see in this piece of code, there is no base stations information and no WiFi information reported, therefore parsing empty string to integer (signal strength) making the problem and producing the error.

Comment (wkhaksar) [reply src/main/java/org/traccar/protocol/WatchProtocolDecoder.java:171]: It is according to the documentations for watch protocol. I read the documentations which are uploaded in Traccar website [here](https://www.traccar.org/protocols/) and the documentation which I uploaded to this ticket. 

The documentation are stating that, right after the status field:
1. Directly after status, there is data for base stations starting with base stations count, count will be greater than zero if base station information is available. But it will be zero if not available. 
2. And in two other documentations (documentation attached to this ticket at page 4 starting line 6 and documentation on Traccar [比电协议V1.8（20160926）.doc](https://www.traccar.org/protocol/5093-watch/%E6%AF%94%E7%94%B5%E5%8D%8F%E8%AE%AEV1.8%EF%BC%8820160926%EF%BC%89.doc) , there is WiFi data directly after base station information. The first integer will state the count for WiFi stations which are reported along with. 

So this way, there is a possible case that the WiFi count field has a value greater than zero but there is no WiFi stations attached or device has failed to transmit the WiFi station data (mac address and rssi). 

So in bellow case, I see the same as above, there is 0 base stations, but the WiFi count field is 1 and there is no WiFi stations attached.

```
                "[SG*9059056143*0053*UD,251021,223408,A,41.46500,N,081.53128,W,0.926,000,0,00,70,70,0,50,00000000,0,1,,,,00]"));
```

Comment (tananaev) [reply src/main/java/org/traccar/protocol/WatchProtocolDecoder.java:171]: What about this test case:
```
[SG*9051004074*0058*AL,120117,145602,V,40.058413,N,76.336618,W,11.519,188,99,00,01,80,0,50,00000000,0,1,0,0,,10]
```

Comment (wkhaksar) [reply src/main/java/org/traccar/protocol/WatchProtocolDecoder.java:171]: The test also pass with it, but a zero mac address withe zero rssi will be added to the position. We can avoid this by a conditional check of mac address with zero. I mean if there is a value of zero for mac address then we do not need to add it.

But the last double in these messages, the documentation says it is accuracy, but as I know, with current implementation of watch protocol and Traccar, the accuracy comes from geolocation providers. If this is true, then this last double can also be ignored.

Comment (tananaev) [reply src/main/java/org/traccar/protocol/WatchProtocolDecoder.java:171]: I think at the very least we need to check for those invalid values.

Comment (wkhaksar) [reply src/main/java/org/traccar/protocol/WatchProtocolDecoder.java]: With a1bf85f I added a check to exclude mac address if it is zero.

File: src/main/java/org/traccar/protocol/WatchProtocolDecoder.java [hunk 2]
```diff
@@ -164,8 +164,11 @@ private Position decodePosition(DeviceSession deviceSession, String data) {
 
             for (int i = 0; i < wifiCount; i++) {
                 index += 1; // wifi name
-                network.addWifiAccessPoint(WifiAccessPoint.from(
-                        values[index++], Integer.parseInt(values[index++])));
+                String macAddress = values[index++];
+                String rssi = values[index++];
+               if (!macAddress.isEmpty() && macAddress != "0" && !rssi.isEmpty()) {
```
tananaev, line 169: Please fix indentation. Also that's not how you compare strings in java.

src/test/java/org/traccar/protocol/WatchProtocolDecoderTest.java
```diff
@@ -144,7 +144,7 @@ public void testDecode() throws Exception {
                 "[ZJ*014111001350304*0038*008a*UD,070318,021027,V,00.000000,N,000.000000,E,0,0,0,0,100,18,1000,50,00000000,4,255,460,0,9346,5223,42,9346,5214,20,9784,4083,11,9346,5221,5]"));
 
         verifyPosition(decoder, buffer(
-                "[3G*8800000015*00DD*UD,010120,025946,V,0.0,N,0.0,E,22.0,0,-1,0,100,98,0,0,00000000,0,5,eduroam,f4:db:e6:d2:a8:00,-53,eduroam,f4:db:e6:da:d0:80,-79,eduroam,78:0c:f0:24:f9:80,-82,Lions,b0:be:76:0a:05:9a,-82,tubs-guest,f4:db:e6:d2:a8:01,-53,0.0]"));
+                "[3G*8800000015*00DD*UD,010120,025946,V,0.0,N,0.0,E,22.0,0,-1,0,100,98,0,0,00000000,0,5,eduroam,f4:db:e6:d2:a8:00,-53,eduroam,f4:db:e6:da:d0:80,-79,eduroam,78:0c:f0:24:f9:80,-82,Lions,b0:be:76:0a:05:9a,-82,tubs-guest,0,-53,0.0]"));
```
tananaev, line 147: Please don't change the original data from devices.

tananaev:
I think you need to rebase your branch.

wkhaksar:
Added a validation for zero mac address in protocol test.

tananaev:
It looks like something is wrong with the PR. It shows all the master changes now.

wkhaksar:
I see. Would you agree with me if I close this PR without merging and create a new one?

tananaev:
Yep, feel free to create a new PR if that's easier.

PR #4875, opened 2022-06-28
Fix for watch protocol with zero base stations

wkhaksar:
The issue discussion started at Traccar forum [here](https://www.traccar.org/forums/topic/watch-protocol-numberformatexception/#post-72333).

The documentation is attached [here](https://github.com/traccar/traccar/files/8994431/Communication.protocol-V3.0_20210423.pdf).

The problem was because of timing advance, as it was still retrieved even if base station information was not attached or base stations count was zero.

If there is no base station information attached or base station number is zero then it means there is no TA as well, as TA is a part of base stations information.
 So, TA moved to inner block.

A check for invalid mac address with value of 0 has been applied. A verification for this has been created in protocol test.

The rest of the conversation can be seen in PR #4874

### DIFF
--- a/src/main/java/org/traccar/protocol/WatchProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/WatchProtocolDecoder.java
@@ -142,8 +142,8 @@ public class WatchProtocolDecoder extends BaseProtocolDecoder {
         Network network = new Network();
 
         int cellCount = Integer.parseInt(values[index++]);
-        index += 1; // timing advance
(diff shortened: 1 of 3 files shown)


tananaev:
Why is there a change to `traccar-web`? Please remove that part.

wkhaksar:
I don't know what is happenning wrong with my PRs these days. I tried alot to revert them but I was unable to do so. 

To avoid confusions, therefore I created a clean PR #4877. 

Sorry for any inconvenience.

PR #4886, opened 2022-07-08
change database driver and config files

emilhm:


src/main/java/org/traccar/database/DataManager.java
```diff
@@ -59,6 +59,17 @@ public class DataManager {
...
     private DataSource dataSource;
 
+
+    private final String dbHost = System.getenv("DB_HOST");
+
+    private final String dbName = System.getenv("DB_NAME");
+
+    private final String dbUser = System.getenv("DB_USER");
+
+    private final String dbPassword = System.getenv("DB_PASSWORD");
```
tananaev, line 69: Just FIY we already support environment variables. You don't need to reinvent the wheel.

https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/config/Config.java#L57

PR #4905, opened 2022-07-21
Fix passwordReset.vm template

lzkill:
Adds the *reset-password* endpoint to the reset link.


tananaev:
Does this actually work? Don't we need `modern` in the path?

lzkill:
This PR if flawed, I'll create another one.

PR #4907, opened 2022-07-21
Fixes passwordReset.vm template

lzkill:
Adds the reset-password endpoint to the reset link.

### DIFF
--- a/templates/full/passwordReset.vm
+++ b/templates/full/passwordReset.vm
@@ -3,6 +3,6 @@
 <html>
 <body>
 To reset password please click on the following link:<br>
-<a href="$webUrl?passwordReset=$token">$webUrl?passwordReset=$token</a><br>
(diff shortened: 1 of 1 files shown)


tananaev:
Merged, thank you.

lzkill:
Maybe injecting the *modern* word from the config (`web.url`) might be useful in the future when the legacy frontend is removed. What do you think?

tananaev:
No. Once legacy app is removed, the modern app will be in the root folder.

lzkill:
Exactly... At this moment one could keep `web.url=https:/xxx.yyy/modern`. When the legacy is removed it the only required would be make  `web.url=https:/xxx.yyy`. But for this to work the template needs not to contain `modern` in the link, only the `reset-password endpoint`.

tananaev:
We would just need to change the template later.

tananaev:
Good catch. Fixed. I guess it doesn't matter in terms of functionality.

PR #4908, opened 2022-07-21
Adds Pushover user attributes

lzkill:
The Pushover integration doesn't allow messages to be delivered per-user,
since the `NOTIFICATOR_PUSHOVER_USER` config is single for the whole instance.

### DIFF
--- a/src/main/java/org/traccar/notificators/NotificatorPushover.java
+++ b/src/main/java/org/traccar/notificators/NotificatorPushover.java
@@ -56,25 +56,24 @@ public class NotificatorPushover implements Notificator {
         url = "https://api.pushover.net/1/messages.json";
         token = config.getString(Keys.NOTIFICATOR_PUSHOVER_TOKEN);
         user = config.getString(Keys.NOTIFICATOR_PUSHOVER_USER);
-        if (token == null || user == null) {
-            throw new RuntimeException("Pushover token or user missing");
-        }
     }
 
     @Override
     public void send(User user, Event event, Position position) {
-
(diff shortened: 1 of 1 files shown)


tananaev:
You need to fix the lint errors.

src/main/java/org/traccar/notificators/NotificatorPushover.java
```diff
@@ -56,25 +56,24 @@ public NotificatorPushover(Config config, NotificationFormatter notificationForm
...
-        }
-
         var shortMessage = notificationFormatter.formatMessage(user, event, position, "short");
 
         Message message = new Message();
         message.token = token;
-        message.user = this.user;
-        message.device = device;
+
+        message.user = user.getString("pushoverUserKey");
```
tananaev, line 68: I think we should reuse `NOTIFICATOR_PUSHOVER_USER` here.

lzkill: I've used the [NotificatorTelegram.java ](https://github.com/lzkill/traccar/blob/ea68646e6572945db82bfdf8e732957492e19b03/src/main/java/org/traccar/notificators/NotificatorTelegram.java)as a guide, where the config provides a wide value that can be overwritten by an user attribute. (`NOTIFICATOR_TELEGRAM_CHAT_ID` / `telegramChatId`)

tananaev: We can't really change what's already there, but we can enforce this for new attributes.

PR #4910, opened 2022-07-21
Update Xexun2ProtocolDecoder.java

stefanclark:
[New detailed positioning data (bit 7).pdf](https://github.com/traccar/traccar/files/9162567/New.detailed.positioning.data.bit.7.pdf)


tananaev:
Where is this documentation from? Can we see the full original document?

stefanclark:
https://www.traccar.org/protocol/5233-xexun2/XEXUN%20Devices%20and%20Server%20Data%20Protocol%20for%20GPS%20tracker%2020210520.pdf

tananaev:
I think you misunderstood me. I need documentation that you used.

By the way, in our document I see that bit 7 is extended data that can contain different information. How does your PR handle this?

stefanclark:
Extra documentation info for bit 7 of position data came directly from xexun. PDF attached is unaltered from what I received directly. This would be added as 3.2.3.7 of main documentation.

bit 7 of mask (read line 132) is extended data

bit 7 of positionMask (read line 138) is new detailed positioning data

Extended data does not need to be parsed (it would go in around line 205. Line 211 `buf.readerIndex(endIndex);` adjusts readerIndex for unparsed extended data.

src/main/java/org/traccar/protocol/Xexun2ProtocolDecoder.java
```diff
@@ -176,7 +176,19 @@ protected Object decode(
                         position.set(Position.KEY_SATELLITES, buf.readUnsignedByte());
                         position.setLongitude(convertCoordinate(buf.readDouble()));
                         position.setLatitude(convertCoordinate(buf.readDouble()));
-
+                    }
+                    if (BitUtil.check(positionMask, 7)) {
+                        buf.skipBytes(2); // length = 30
+                        buf.skipBytes(1); // marker = 'G'
+                        buf.skipBytes(2); // length = 27
```
tananaev, line 183: I think these needs to be handled properly. You need to honor the length fields. And also probably we need to check for the marker. They clearly intend this to be extendable, even if for now it's fixed.

src/main/java/org/traccar/protocol/Xexun2ProtocolDecoder.java
```diff
@@ -176,7 +176,19 @@ protected Object decode(
...
+                        position.setLongitude(convertCoordinate(buf.readDouble()));
+                        position.setLatitude(convertCoordinate(buf.readDouble()));
+                        position.setValid(buf.readUnsignedByte() > 0); // 0=invalid,1=normal,2=sub-meter,3=differential
```
tananaev, line 186: Remove comment.

src/main/java/org/traccar/protocol/Xexun2ProtocolDecoder.java
```diff
@@ -176,7 +176,19 @@ protected Object decode(
...
+                        position.set(Position.KEY_SATELLITES, buf.readUnsignedByte());
+                        buf.skipBytes(1); // SNR
```
tananaev, line 188: This should be `readUnsignedByte` and either remove the comment or change it to full `satellite signal to noise ratio`.

tananaev:
Looks like you haven't even tried compiling your code after the changes.

Please also add a test case.

stefanclark:
Sorry I was rushing. Added sample data for test and updated code.

I also ran ```./gradlew test --tests org.traccar.protocol.Xexun2*``` which didn't appear to show errors. (Assuming that's right?)

I will add encoder later as a separate pull request to enable sending commands to xexun2 devices. I have working for OpenGTS, but need to port code.

tananaev:
Merged, thank you.

PR #4912, opened 2022-07-25
Xexun2 Encoder

stefanclark:
Code to validate checksum could be added to Decoder.


src/main/java/org/traccar/protocol/Xexun2ProtocolDecoder.java
```diff
@@ -99,12 +100,12 @@ protected Object decode(
             return null;
         }
 
-        sendResponse(channel, type, index, imei);
-
         buf.readUnsignedShort(); // attributes
         buf.readUnsignedShort(); // checksum
 
         if (type == MSG_POSITION) {
+            sendResponse(channel, type, index, imei);
```
tananaev, line 107: Why is this moved?

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright 2016 - 2022 Anton Tananaev (anton@traccar.org)
```
tananaev, line 2: You need to fix the notice.

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
```diff
@@ -0,0 +1,121 @@
...
+            } else {
+                result.writeByte(b);
+            }
+        }
+        result.writeBytes(buf.readBytes(2));
+
+        return result;
+    }
+
+    private static int checksum(byte[] data)
```
tananaev, line 65: Are you sure it's not one of the standard checksums we already support?

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
```diff
@@ -0,0 +1,121 @@
...
+
+
+    private static ByteBuf encodeContent(String uniqueId, String content) {
+        ByteBuf buf = Unpooled.buffer();
+
+        byte[] message = content.getBytes();
+
+        buf.writeShort(0xFAAF);
+        buf.writeShort(0x0007);
+        buf.writeShort(0x0001);
```
tananaev, line 93: What are these?

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
```diff
@@ -0,0 +1,121 @@
...
+        buf.writeBytes(DataConverter.parseHex(uniqueId + "0"),0,8);
```
tananaev, line 94: Why is there a range?

tananaev (commented):
Also make sure you run gradle check before sending your pull requests.

src/main/java/org/traccar/protocol/Xexun2ProtocolDecoder.java
stefanclark: When sending a command (type = 0x07) the device will acknowledge the command with a message type = 0x07.

If decoder responded to the command ack with another ack, the device would interpret that as another command and create a loop.

Decoder should only ack (`sendResponse()`) when type = 0x14

tananaev: Shouldn't we send to `type != 0x07`?

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
stefanclark: To ensure that only 8 bytes are written. Can we assume uniqueId is only ever 15 characters?

tananaev: Well, the id will be a valid unique id.

src/main/java/org/traccar/protocol/Xexun2ProtocolDecoder.java
stefanclark: I don't think it makes a difference. As far as I know only 0x07 and 0x14 are used for message id. None other are mentioned in documentation.

I'm not sure the server should blindly acknowledge other messages? Although I realise it might be a breaking change if devices do send other message id, so perhaps for backwards compatibility sending a response when type != 0x07 is safer?

Your call.

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
stefanclark: I've tried all the CRC16 algorithms and none generate the same checksum.

src/main/java/org/traccar/protocol/Xexun2ProtocolDecoder.java
tananaev: Let's do `type != 0x07`. This is more in line with other protocols, so I think that's probably be right if they extend the protocol.

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
tananaev: Can you please include the documentation.

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
stefanclark: 0x0001 is Serial Number for message. With messages coming from device it increases by 1 each message sent to server.

For messages sent to device you could increase each time a message is sent so that you can directly match ack to message sent. Would this be possible with traccar? Is there an enumerated property for Command passed to encodeCommand()?

I have had no issues using 0x0001 for every command sent to X06 device.

tananaev: Then write it like this:
```
buf.writeShort(1); // index
```

What about the other one?

stefanclark: message id 0x07 = remote control command

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
stefanclark: https://www.traccar.org/protocol/5233-xexun2/XEXUN%20Devices%20and%20Server%20Data%20Protocol%20for%20GPS%20tracker%2020210520.pdf

1.4.4 ECC on page 3&4 - no description on algorithm, just java & C code provided.

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
tananaev: You should reuse the type from `Xexun2ProtocolDecoder`.

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
tananaev: Looks like it's a standard TCP/UDP checksum. We should add it to our standard checksum class.

http://minirighi.sourceforge.net/html/udp_8c-source.html

tananaev: Also you should use buffer, not array.

stefanclark:
This should be good to merge?

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
```diff
@@ -0,0 +1,102 @@
...
+import org.traccar.Protocol;
+
+public class Xexun2ProtocolEncoder extends BaseProtocolEncoder {
+
+    public Xexun2ProtocolEncoder(Protocol protocol) {
+        super(protocol);
+    }
+
+    public static final int FLAG = 0xfaaf;
+    public static final int MSG_COMMAND = 0x07;
```
tananaev, line 33: Both constants should be defined in the decoder.

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private static ByteBuf encodeFrame(ByteBuf buf) {
+        int bufLength = buf.readableBytes();
+        if (bufLength < 5) {
+            return null;
+        }
```
tananaev, line 39: What is this for?

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
```diff
@@ -0,0 +1,102 @@
...
+public class Xexun2ProtocolEncoder extends BaseProtocolEncoder {
+
+    public Xexun2ProtocolEncoder(Protocol protocol) {
+        super(protocol);
+    }
+
+    public static final int FLAG = 0xfaaf;
+    public static final int MSG_COMMAND = 0x07;
+
+    private static ByteBuf encodeFrame(ByteBuf buf) {
```
tananaev, line 35: I'm pretty sure I had a comment about this, but this needs to be moved out into a frame encoder.

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
```diff
@@ -0,0 +1,102 @@
...
+    private static ByteBuf encodeContent(String uniqueId, String content) {
+        ByteBuf buf = Unpooled.buffer();
+
+        ByteBuf message = Unpooled.copiedBuffer(content.getBytes());
+
+        buf.writeShort(FLAG);
+        buf.writeShort(MSG_COMMAND);
+        buf.writeShort(1); // index
+        buf.writeBytes(DataConverter.parseHex(uniqueId + "0"));
+        buf.writeShort(message.capacity());
```
tananaev, line 75: You should not use capacity. Why not just use content length?

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+        result.writeBytes(buf.readBytes(2));
+
+        return result;
+    }
+
+    private static ByteBuf encodeContent(String uniqueId, String content) {
+        ByteBuf buf = Unpooled.buffer();
+
+        ByteBuf message = Unpooled.copiedBuffer(content.getBytes());
```
tananaev, line 69: Don't you need encoding here for the string? ASCII?

src/main/java/org/traccar/helper/Checksum.java
```diff
@@ -200,4 +200,23 @@ public static long luhn(long imei) {
         return (10 - (checksum % 10)) % 10;
     }
 
+    public static int udp(ByteBuffer data) {
```
tananaev, line 203: I think this checksum is also used for TCP. Maybe call it `ip`?

src/main/java/org/traccar/helper/Checksum.java
```diff
@@ -200,4 +200,23 @@ public static long luhn(long imei) {
...
+        int sum = 0;
+        int len = data.capacity();
```
tananaev, line 205: Again, don't use capacity.

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
stefanclark: ok

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
stefanclark: the method would fail if the buffer to encode was < 5 bytes. it should never be the case so could remove these lines.

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
stefanclark: not sure it was mentioned, if it was I missed it. moved now and added a test case (although the test case is not a real packet that would be sent/received from device).

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
stefanclark: ByteBuff doesn't have a length() method. A number of other decoder/encoders use capacity() e.g. T800xProtocolDecoder.java

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
stefanclark: now set explicitly

src/main/java/org/traccar/helper/Checksum.java
stefanclark: I couldn't find other examples of this checksum. The "Internet Checksum" from RFC 791 / 1071 appears to be slightly different. Renamed ip().

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
tananaev: Let's remove if it never happens.

src/main/java/org/traccar/helper/Checksum.java
stefanclark: ByteBuffer does not appear to have a length property/method. I have rewritten the for loop to while ByteBuffer.hasRemaining() however not sure what difference it makes.

src/main/java/org/traccar/protocol/Xexun2ProtocolEncoder.java
tananaev: You should use `readableBytes`. That's the real length of the data. Other places should probably be updated as well, but not part of this pr.

src/main/java/org/traccar/helper/Checksum.java
tananaev: See comment above.

stefanclark:
Is this okay now?

tananaev:
Looks good, thank you. Merged.

PR #4916, opened 2022-08-05
CAN bus data support for huabao

K1DV5:
According to the [JT-T808 standard](https://www.roadragon.com/uploads/JT808-protocol.pdf), a device on a vehicle can periodically report [CAN bus data](https://www.csselectronics.com/pages/can-bus-simple-intro-tutorial) to the server using the `0x0705` message ID. The format of the body of the data is like:
```
[2 byte integer count of the data items]
[5 byte BCD encoded time of the data reception]
[can frame]
[can frame]
...
```
Where `can frame` has a format like
```
[4 byte can ID][8 byte data]
```
What the CAN ID and data mean is beyond the scope of the mentioned standard (and I think, by extension, Traccar.) And it can be vehicle manufacturer specific. I only implemented decoding it so far.


tananaev:
1. Provide description and protocol documentation
2. Revert the submodule change

PR #4921, opened 2022-08-14
Add gotify notificator

TheZoker:
This add a notificator for gotify:
https://gotify.net/

I saw a lot of people having issues setting up gotify via event forward, so I looked at the code and saw, that the notificator implementation is quite simple. To ease the process of those people (and myself) I added a simple integration. 
Since traccar is selfhosted and gotify is selfhosted as well, in my opinion those are a great match! 

Docs and frontend (or is this handled via translation platform?) PR will follow, if this one is alright. 

I tested it on my local installation and it works without issue.


tananaev:
Can you please provide some references for this statement:
> I saw a lot of people having issues setting up gotify via event forward

TheZoker:
Sure, here are some links:

- https://www.traccar.org/forums/topic/forward-events-to-php-application/
- https://github.com/neo7530/traccar_pusher/issues/1 (comment at the bottom)
- Myself (I get the same issue as in the first topic and did not manage to fix this)

Maybe I should not used the phrase "a lot" but "some" instead.

tananaev:
First two refer to the same thing. Also it's about forwarding, not notifications, from what I can tell.

Also, couldn't you use SMS API option to basically achieve the same thing?

TheZoker:
Yes right, but different people having the same issue (and myself as well). 

II want to use the SMS function for different notifications using twilio in the future (so I want to use both gotify and twilio at the same time) and I think that is not possible if I already use the SMS Gateway for gotify notification. 

Is there a reason that is against integration gotify into traccar?

tananaev:
I'm kind of against it because it seems like pretty niche service and it's not going to be used a lot, but having it in the codebase means more maintenance burden.

TheZoker:
Hmm I think that gotify is kind of popular in the self hosting community (just like traccar for its purpose) it has about 7.6k stars. 
And gotify did not change the very small message API call in years now (the project only has 530 commits and most of them are UI related or updating dependencies), so the endpoint should not change in the near future (and even if it does, it is like one line of code change). 
The `message.go` file only has like 25 changes within the last 4 years, but I understand your point.

tananaev:
Maintenance needed not only when API changes, but also, for example, if we decide to refactor our internal stuff.

Let's leave it as open for now and see how much interest we get.

tamcore:
Another vote for integrating this from me. Gotify is not a niche service, as /r/selfhosted on Reddit would happily confirm. 

Using a SMS gateway would mean added cost. Other integrations like Firebase and Telegram route your data over 3rd party servers. Gotify can be hosted on your own infrastructure.

So, yeah, please include this :)

poltrian:
Same for me, i have a ntfy Server, but i have no idea how it works with Traccar. Gotfy is very wellcome

tamcore:
> Same for me, i have a ntfy Server, but i have no idea how it works with Traccar. Gotfy is very welcome. Synapse Matrix is also very welcome as an alternative to Telegram.

With ntfy it's easy, same boat for me now :) See my config bellow. Just point `sms.http.url` to your instance. Adjust `sms.http.authorization*` if you have ACL stuff configured. And update the `topic`in the message template. Then traccar will send the SMS-Style messages to ntfy. And to get ntfy messages into Matrix, check https://gitlab.com/999eagle/maubot-ntfy.

```
<entry key='notificator.types'>web,sms</entry>
<entry key='sms.http.url'>http://ntfy.ntfy.svc.cluster.local:2586</entry>
<entry key='sms.http.authorizationHeader'>Authorization</entry>
<entry key='sms.http.authorization'>Bearer tk_YOURTOKEN</entry>
<entry key='sms.http.template'>
    {
        "topic": "YavTZnJjSsX9ltep",
        "message": "{message}"
    }
</entry>
```

poltrian:
I use https and thats  not working with your configuration. Do you have any idea?

tamcore:
Depends :)
- When you use the `Test Channels` function when configuring a new notification, do you get any error message at the bottom of the screen?
- How did you configure it? Did you hardcode the topic in `sms.http.template`, or did you use `{phone}` there?
  - If you hardcoded it, you still need to set a dummy phone number in your profile, as otherwise it'll not trigger the processing (https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/notificators/NotificatorSms.java#L48)
  - If you used `{phone}` there, did you set the proper topic as your account's phone number? (Whitespaces or something?)
- Are you using your own ntfy instance, or upstream ntfy.sh?
  - If it's your own, 
    - is access-control enabled?
      - If access-control is enabled, did you verify, that the token you provided can POST to your topic? (curl)
    - are you using a self-signed SSL certificate?
- How did you deploy traccar? Which version / container image?

I just double-checked it against ntfy.sh based on my example configuration which i submitted for PR there (https://github.com/binwiederhier/ntfy/pull/631) and that seems to work :)

poltrian:
Please excuse me for the time you took. I forgot to enter the dummy phone number to make it work. Thanks alot!

Dasinfomedia2:
+1 for gotify

TheZoker:
@tananaev Since other users have also expressed their interest in this, would you now consider this change?

DIO85-git:
I agree with PR. I'm a self-hosted and open source enthusiast. Someday Firebase may be charged for usage, it's good to have alternatives like this. Congratulations @TheZoker! Please approve this.

SecT0uch:
I would also love seing this merged!

> it seems like pretty niche service
This is kind of the same niche as those self-hosting a tracking service TBH.

Many of us here are having a self-hosted infrastructure with Gotify  without Google apps.

SecT0uch:
After a few trial and error, I managed to configure SMS API to work with Gotify.

First **make sure you have a dummy phone number  in Settings > Account > Preferences** otherwise messages won't be sent.
Then in the configuration
```xml
<entry key='notificator.types'>sms,web</entry>
<entry key='sms.http.url'>https://<GOTIFY_SERVER>/message?token=<APP_TOKEN></entry>
<entry key='sms.http.template'>
{
        "message": "{message}",
        "priority":  3
}
    </entry>
```

dlm21:
I spent sooo long trying to set up Gotify only to finally find this PR that mentions setting up a dummy phone number. Thank yall that added this crucial detail. Considering Gotify is twice as popular as Traccar (if using dumb metrics like github followers and Stars) I'd say this would be a welcome addition.
Would also be nice to see more template variables other than just {phone} and {message} , but that's a topic for another PR~
At least making a note on https://www.traccar.org/http-sms-api/ about the need for a dummy/placeholder phone number (if we are mandated to use the sms http gateway) would be really helpful to those walking this path in the future.

PR #4941, opened 2022-09-07
Support for Piligrim M-5000/M-10000 protocol

anton2920:
This branch adds basic support for protocol used by Piligrim M-5000/M-10000 and probably some other devices.

### DIFF
--- a/src/main/java/org/traccar/helper/NMEA.java
+++ b/src/main/java/org/traccar/helper/NMEA.java
@@ -1,0 +1,126 @@
+package org.traccar.helper;
+
+import java.util.HashMap;
+import java.util.Map;
+
+
+public class NMEA {
+
+    // fucking java interfaces
+    interface SentenceParser {
+        public boolean parse(String[] tokens, GPSPosition position);
+    }
+
+    // utils
+    static float Latitude2Decimal(String lat, String NS) {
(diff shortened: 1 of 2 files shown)


tananaev:
This looks like work in progress. Let me know when it's ready for review.

anton2920:
Example:

```
POST /push.do HTTP/1.0
Host: wayker.com
Content-Type: application/x-www-form-urlencoded
Content-Length: 198
Connection: close
user-agent: Piligrim-M7 (#76)

&phoneNumber=%2B+7800XXXXXXX&message=ALARM KEY; $GPRMC,180752.000,A,XXXX.0857,N,XXXXX.8173,E,0.00,104.74,220722,,,A,V* 29,05; GSM: XXX-XX XXXX-XXXX,XXXX,XXXX,XXXX,XXXX 25;  S; Batt: 405,M
```

I've replaced private information with `X`, but overall structure should be clear. Also, instead of `ALARM KEY; `, it's possible that message has nothing or some other words, which depend on event. All of those are currently ignored.

tananaev:
A quick question here. Is there a reason you didn't use our standard way of decoding text data? We have many examples of NMEA decoding.

tananaev:
Refactor if you want to merge to your local version - https://github.com/traccar/traccar/commit/4de8efe1ef0810af492c161bfc1d3200958d75d8.

PR #4943, opened 2022-09-07
Support for G1RUS protocol (hex-encoding)

anton2920:
This branch adds basic support for G1RUS protocol used by PromaSat 1000 Next (and other GoSafe devices). Probably _G1rusProtocol*.java_ files could be merged with _GosafeProtocol*.java_ files, if there's evidence that these protocols are in fact 100% identical, just the encoding is different.


tananaev:
All of your pull requests are not really following our style. Are you interested in fixing it or do you want us to fix it?

By the way, can you provide details on which devices are supported for each of the protocols. And also provide protocol documentation if you can for each.

anton2920:
Unfortunately, I'm not really interested in fixing style and/or writing tests for these, as these branches were just modifications for internal needs and not originally intended to be shared as pull requests.

The spec. for this protocol was provided by you, so it's easy to verify the changes against it and/or extend/modify them. Let me know if you need an example. I'll then remove all private data from HEX and share it.

tananaev:
Yes, please provide examples.

tananaev:
Refactored this one as well - https://github.com/traccar/traccar/commit/5612d29b74a1c5ded13955565cf1b48f2307813c. Please try and provide examples for tests if possible.

PR #4960, opened 2022-10-12
Fixed invalid header, version and IMEI parsing

anton2920:


tananaev:
Here's a better fix I think: https://github.com/traccar/traccar/commit/c801d33f44ae346d01e19e47a3b97c8405f4d5b4.

PR #4970, opened 2022-10-26
2 Bugs on maintenance

jcardus:
According to the documentation:

> Traccar generates event every time totalDistance attribute gets over maintenance.start + maintenance.interval * N value where N is a natural number.
> 
> Example:
> maintenance.start=6000000, maintenance.interval=8000000
> 
> Events will be generated when totalDistance goes over 6000 km, 14000 km, 22000 km and so on.

The formula isn't coherent with the example because 0 is not a natural number. But the formula is correctly implemented, so, the event **won't be fired** when it goes over 6000 km. This the first bug.

If maintenance.interval < maintenance.start  and maintenance.start is multiple of maintenance.interval then the event will be fired before maintenance.start because the comparison will be on negative numbers. 
Another solution would be to use absolute values on the comparison.  This is the second bug.

### Example:
maintenance.start = 10 000 000
maintenance.interval = 2 000 000

the event **will be fired** when it goes over 2 000 000 
the event **wont' be fired** when it goes over 10 000 000

### DIFF
--- a/src/main/java/org/traccar/handler/events/MaintenanceEventHandler.java
+++ b/src/main/java/org/traccar/handler/events/MaintenanceEventHandler.java
@@ -51,13 +51,15 @@ public class MaintenanceEventHandler extends BaseEventHandler {
             if (maintenance.getPeriod() != 0) {
                 double oldValue = lastPosition.getDouble(maintenance.getType());
                 double newValue = position.getDouble(maintenance.getType());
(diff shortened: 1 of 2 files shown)


tananaev:
Can we add some tests to validate?

jcardus:
ok

jcardus:
please check

tananaev:
Merged, thank you.

PR #4975, opened 2022-11-01
Added Can Properties and Value to Hex Decimal

kenanilgun:
CAN elements a standart on below documentations. If we use with decodeParameter can come just long value on position. I will have plan to change this situation with mask system. It's first step.

That IDs just using with 1 protocol and using with can spesifcly/unique differents(Manual CAN Data)

https://wiki.teltonika-mobility.com/view/TFT100_AVL_ID_List#Manual_CAN_I.2FO_elements

### DIFF
--- a/src/main/java/org/traccar/model/Position.java
+++ b/src/main/java/org/traccar/model/Position.java
@@ -145,6 +145,24 @@ public class Position extends Message {
     public static final String ALARM_TAMPERING = "tampering";
     public static final String ALARM_REMOVING = "removing";
 
+    public static final String KEY_CAN0 = "can0";
+    public static final String KEY_CAN1 = "can1";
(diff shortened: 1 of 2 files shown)


tananaev:
1. Please provide documentation
2. Confirm that these IDs are the same for all models
3. If you want to add new common position attributes, it needs to be used at the very least by 2 protocols
4. Provide details on what those parameters mean

kenanilgun:
Hi,

CAN elements a standart on below documentations. If we use with decodeParameter can come just long value on position. I will have plan to change this situation with mask system. It's first step.

https://wiki.teltonika-mobility.com/view/TFT100_AVL_ID_List#Manual_CAN_I.2FO_elements

PR #4994, opened 2022-12-08
MQTT Notifier added

sveba:
A rebase of #4558, so it can be merged easy

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,6 +85,7 @@ dependencies {
     implementation "com.amazonaws:aws-java-sdk-sns:1.12.349"
     implementation "org.apache.kafka:kafka-clients:3.3.1"
     implementation "com.google.firebase:firebase-admin:9.1.1"
+    implementation 'org.eclipse.paho:org.eclipse.paho.client.mqttv3:1.2.5'
     testImplementation "junit:junit:4.13.2"
     testImplementation "org.mockito:mockito-core:4.+"
 }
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -85,6 +85,7 @@ dependencies {
     implementation "com.amazonaws:aws-java-sdk-sns:1.12.349"
     implementation "org.apache.kafka:kafka-clients:3.3.1"
     implementation "com.google.firebase:firebase-admin:9.1.1"
+    implementation 'org.eclipse.paho:org.eclipse.paho.client.mqttv3:1.2.5'
```
tananaev, line 88: First of all, looks like this library is basically abandoned. Secondly, I think it's better to use existing Netty library to communicate via MQTT.

sveba:
@tananaev well, it is not my code. The only thing I did was to rebase, so it can be merged easily. If I understand you right, you don't like the implementation at all, because it is using the wrong library, pattern and interface. So it is completely wrong :)

tananaev:
I like the idea of supporting MQTT, but yeah, there are too many issues in this implementation. Should we close it for now?

sveba:
sure

sveba:
@tananaev I would try to Implement this as a forwarder, but I can not understand how netty can be used as Mqtt-Client. So I would prefer to use https://hivemq.github.io/hivemq-mqtt-client/ 

If you have a problem with that, pls tell me before I start coding :)

tananaev:
Maybe it's ok because it seems like they use Netty under the hood.

PR #5014, opened 2023-01-06
Lbs for tlt2h

wkhaksar:
This branch adds LBS support for tlt2h protocol.

According to [this protocol](https://www.mictrack.com/downloads/protocols/Mictrack_Communication_Protocol_For_MT700_V1.0.pdf) there is also LBS data available in the packet if this is enabled in the device.

The docs mentions, this LBS data is an optional part of data packet, if it is enabled then it would be provided both in GPS and WiFi packets.


tananaev:
Merged with a small fix. Even if GPS is marked as invalid, you have to decode it. You can't ignore it.

PR #5026, opened 2023-02-10
Update SuntechProtocolDecoder.java

fernandohallberg:
Hi,

I added battery level position attribute to suntech protocol (ST940).

test OK:
{"versionFw":"556",**"battery":4.0**,"distance":2184.45,"totalDistance":128182.02,"motion":true}


tananaev:
Please provide the protocol documentation and add a test case.

PR #5031, opened 2023-02-15
Update Gt06FrameDecoder.java

cylyl:
Return frame based on protocol length.

### DIFF
--- a/src/main/java/org/traccar/protocol/Gt06FrameDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06FrameDecoder.java
@@ -38,7 +38,9 @@ public class Gt06FrameDecoder extends BaseFrameDecoder {
             length += 2 + buf.getUnsignedShort(buf.readerIndex() + 2);
         }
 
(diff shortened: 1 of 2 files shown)


tananaev:
Please include a unit test for you change.

cylyl:
Test added Update Gt06FrameDecoderTest.java #5032

src/main/java/org/traccar/protocol/Gt06FrameDecoder.java
```diff
@@ -38,7 +38,9 @@ protected Object decode(
             length += 2 + buf.getUnsignedShort(buf.readerIndex() + 2);
         }
 
-        if (buf.readableBytes() >= length && buf.getUnsignedShort(buf.readerIndex() + length - 2) == 0x0d0a) {
+        if (buf.readableBytes() >= (length + 2) && buf.getUnsignedShort(buf.readerIndex() + length) == 0x0d0a) {
```
tananaev, line 41: Why are we adding 2 to the length? Do you have documentation explaining it?

Issue #5028, opened 2023-02-14
Cannot declare variables in Computed attributes


memesaregood1:
**Describe the bug**
Local JEXL variables cannot be accessed or declared anymore. Worked in 5.3, as far as I remember.

**To Reproduce**
1. Go to 'Computed Attributes'
2. Click on '+'
3. Enter this expression: `var theAnswer = 42; theAnswer += 2; theAnswer`, type: number
4. See error: `org.traccar.handler.ComputedAttributesHandler.computeAttribute:105@1:1 parsing error in 'var' - Parsing (... < ComputedAttributesHandler:105 < AttributeResource:63 < ...)`

**Expected behavior**
44

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser; Chrome
 - Version: 109

**Smartphone (please complete the following information):**
 - Device: Xiaomi Mi 9
 - OS: Android 11
 - Browser: Chrome
 - Version 107

**Additional context**
Traccar version: 5.6, upgraded from 5.5, previously upgraded from 5.3. Java 17.
Reference: https://commons.apache.org/proper/commons-jexl/reference/syntax.html

tananaev:
I think it's just a limitation of the new version of JEXL. We run it as an expression using [createExpression](https://commons.apache.org/proper/commons-jexl/apidocs/org/apache/commons/jexl3/JexlEngine.html#createExpression(java.lang.String)) method.

memesaregood1:
It does not make sense to me. If this is mentioned in Syntax documentation, as well as function declaration, why isn't it supported in _your_ JEXL?

tananaev:
I don't know, but feel free to investigate and let us know if you find anything.

memesaregood1:
I've indeed done some digging. It appears that some functions are indeed disabled, but they can be enabled using ``JexlFeatures``. If used correctly, it may be used to enable missing funcitonality and not expose any security vulnerabilities.
I'll be preparing a PR in a short while.

PR #5036, opened 2023-02-21
Implement JEXL parser settings

memesaregood1:
Following https://github.com/traccar/traccar/issues/5028, this PR adds various preferences options to extend or disable JEXL functionality on-demand. The parser now runs in the JEXL Sandbox, which should stop most, if not any, RCE attacks. The following configuration lines were added:

- `processing.computedAttributes.parser.annotation` - if disabled, using `@annotation` constructs will result in an exception
- `processing.computedAttributes.parser.arrayReferences` - if disabled, using 'obj[ ref ]' where ref is not a string or integer literal will result in a parsing exception;
- `processing.computedAttributes.parser.lambda` - enable lambda declaration
- `processing.computedAttributes.parser.lexical` - with this on, redefining local variables will result in an exception
- `processing.computedAttributes.parser.lexicalShade` - with this on, local variables will shade global ones outside their scope
- `processing.computedAttributes.parser.localVariables` - with this on, local variables declaration will be allowed
- `processing.computedAttributes.parser.loops` - enable loop execution
- `processing.computedAttributes.parser.method_calls` - if this is disabled, using `obj.samplemethod()` would result in an exception
- `processing.computedAttributes.parser.newInstanceCreation` - if this is disabled, using `new()` will result in an exception
- `processing.computedAttributes.parser.pragma` - enable pragma expressions
- `processing.computedAttributes.parser.scriptConstructs` - enable script constructs
- `processing.computedAttributes.parser.sideEffect` - modifying variables or members using constructs will result in an exception
- `processing.computedAttributes.parser.sideEffectGlobal` - same as `sideEffect`, but in global scope
- `processing.computedAttributes.parser.structuredLiteral` - enable array/map/set literal expressions processing
I'd like @tananaev to review this PR and check if there any flaws or mistakes in my solution. 
I also want to point out the `scriptConstructs` option - it appears dangerous to me. I tried this JEXL expression:
`"''.getClass().forName('java.lang.Runtime').getRuntime().exec('bash')"`
to execute Bash on the machine, which works, if JEXL is not running under the Sandbox. If it does, this does not work. However, as I'm not experienced enough for this, I can't figure out if this expression can be altered to be successfully used in the Sandbox environment, so, Anton, please check this out.

References:
[JexlFeatures](https://svn.apache.org/repos/infra/websites/production/commons/content/proper/commons-jexl/apidocs/org/apache/commons/jexl3/JexlFeatures.html)
[JexlOptions](https://commons.apache.org/proper/commons-jexl/apidocs/org/apache/commons/jexl3/JexlOptions.html)
[JexlSandbox](https://commons.apache.org/proper/commons-jexl/apidocs/org/apache/commons/jexl3/introspection/JexlSandbox.html)
[Jexl Syntax](https://commons.apache.org/proper/commons-jexl/reference/syntax.html)

### DIFF
--- a/src/main/java/org/traccar/handler/ComputedAttributesHandler.java
+++ b/src/main/java/org/traccar/handler/ComputedAttributesHandler.java
@@ -57,39 +57,16 @@ public class ComputedAttributesHandler extends BaseDataHandler {
     @Inject
     public ComputedAttributesHandler(Config config, CacheManager cacheManager) {
         this.cacheManager = cacheManager;
-        boolean enableJEXLAnnotation = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_ANNOTATION);
-        boolean enableJEXLArrayReferences = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_ARRAY_REFERENCES);
-        boolean enableJEXLLambdas = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_LAMBDA);
-        boolean enableJEXLLexical = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_LEXICAL);
-        boolean enableJEXLLexicalShade = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_LEXICAL_SHADE);
         boolean enableJEXLLocalVariables = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_LOCAL_VARIABLES);
         boolean enableJEXLLoops = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_LOOPS);
-        boolean enableJEXLMethodCalls = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_METHOD_CALLS);
         boolean enableJEXLNewInstanceCreation = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_NEW_INSTANCE_CREATION);
-        boolean enableJEXLPragma = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_PRAGMA);
-        boolean enableJEXLScriptConstructs = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_SCRIPT_CONSTRUCTS);
-        boolean enableJEXLSideEffect = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_SIDE_EFFECT);
-        boolean enableJEXLSideEffectGlobal = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_SIDE_EFFECT_GLOBAL);
-        boolean enableJEXLStructuredLiteral = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_STRUCTURED_LITERAL);
         JexlSandbox sandbox = new JexlSandbox(false);
         sandbox.allow("com.safe.Functions");
-        boolean enableJEXLRegister = true;
         JexlFeatures features = new JexlFeatures()
-                .annotation(enableJEXLAnnotation)
-                .arrayReferenceExpr(enableJEXLArrayReferences)
-                .lambda(enableJEXLLambdas)
-                .lexical(enableJEXLLexical)
-                .lexicalShade(enableJEXLLexicalShade)
                 .localVar(enableJEXLLocalVariables)
                 .loops(enableJEXLLoops)
-                .methodCall(enableJEXLMethodCalls)
                 .newInstance(enableJEXLNewInstanceCreation)
-                .pragma(enableJEXLPragma)
-                .register(enableJEXLRegister)
-                .script(enableJEXLScriptConstructs)
-                .sideEffect(enableJEXLSideEffect)
-                .sideEffectGlobal(enableJEXLSideEffectGlobal)
-                .structuredLiteral(enableJEXLStructuredLiteral);
+                .structuredLiteral(true);
         engine = new JexlBuilder()
                 .strict(true)
                 .namespaces(Collections.singletonMap("math", Math.class))
--- a/src/main/java/org/traccar/handler/ComputedAttributesHandler.java
+++ b/src/main/java/org/traccar/handler/ComputedAttributesHandler.java
@@ -115,7 +115,9 @@ public class ComputedAttributesHandler extends BaseDataHandler {
      */
     @Deprecated
     public Object computeAttribute(Attribute attribute, Position position) throws JexlException {
-        return engine.createScript(features, engine.createInfo(), attribute.getExpression()).execute(prepareContext(position));
+        return engine.createScript(features,
+                engine.createInfo(),
+                attribute.getExpression()).execute(prepareContext(position));
     }
 
     @Override
--- a/src/main/java/org/traccar/handler/ComputedAttributesHandler.java
+++ b/src/main/java/org/traccar/handler/ComputedAttributesHandler.java
@@ -26,10 +26,8 @@ import java.util.Map;
 import java.util.Set;
 
 import io.netty.channel.ChannelHandler;
-import org.apache.commons.jexl3.JexlBuilder;
-import org.apache.commons.jexl3.JexlEngine;
-import org.apache.commons.jexl3.JexlException;
-import org.apache.commons.jexl3.MapContext;
+import org.apache.commons.jexl3.*;
+import org.apache.commons.jexl3.introspection.JexlSandbox;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.traccar.BaseDataHandler;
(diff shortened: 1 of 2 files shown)


tananaev:
I feel like this is too many options. Which parameters do you personally need?

I think it makes sense to have one config parameter to enable/disable variable. A lot of other stuff should probably be never enabled. We want to limit expressions only to very simple things.

memesaregood1:
When writing this, I assumed users would want some more flexibility. If we're talking minimal, then I think local vars, loops and, maybe, `new` instance creations would be sufficient?

tananaev:
Most users are confused even with basic stuff. Yeah, I think variables, loops and new instance makes sense to me. Can you please update the PR with those 3 options?

memesaregood1:
No arrays?

tananaev:
I think arrays can probably be enabled by default?

memesaregood1:
Alright. 3 options with 1 hardcoded, coming right up.

tananaev:
You can also probably remove `.parser.` prefix for the attributes. Make them shorter.

memesaregood1:
Better?

src/main/java/org/traccar/handler/ComputedAttributesHandler.java
```diff
@@ -55,14 +57,28 @@ public class ComputedAttributesHandler extends BaseDataHandler {
 
     private final boolean includeDeviceAttributes;
 
+
     @Inject
     public ComputedAttributesHandler(Config config, CacheManager cacheManager) {
         this.cacheManager = cacheManager;
+        boolean enableJEXLLocalVariables = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_LOCAL_VARIABLES);
+        boolean enableJEXLLoops = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_LOOPS);
+        boolean enableJEXLNewInstanceCreation = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_NEW_INSTANCE_CREATION);
```
tananaev, line 66: I would inline these.

src/main/java/org/traccar/handler/ComputedAttributesHandler.java
```diff
@@ -55,14 +57,28 @@ public class ComputedAttributesHandler extends BaseDataHandler {
 
     private final boolean includeDeviceAttributes;
 
+
```
tananaev, line 60: Remove extra space.

src/main/java/org/traccar/handler/ComputedAttributesHandler.java
```diff
@@ -55,14 +57,28 @@ public class ComputedAttributesHandler extends BaseDataHandler {
...
+                .newInstance(enableJEXLNewInstanceCreation)
+                .structuredLiteral(true);
         engine = new JexlBuilder()
                 .strict(true)
                 .namespaces(Collections.singletonMap("math", Math.class))
+                .sandbox(sandbox)
+                .features(features)
                 .create();
         includeDeviceAttributes = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_DEVICE_ATTRIBUTES);
+
```
tananaev, line 81: Remove space.

tananaev (commented):
A couple of minor comments.

Does this mean we can re-enable attributes for regular users? Or is it still too risky?

memesaregood1:
It's probably safe to let non-admins edit or create their own attributes. The code is now running in a sandbox, so no apparent damage to the system, but there's a but: I've just discovered that JEXL has some other functionality enabled by default, such as method calls from the objects. If we were to have minimal functionality, we would need to disable these features explicitly.

tananaev:
Please fix the lint issue. Also can you please add empty lines between config key definitions, similar to how we have them between other keys.

memesaregood1:
Can you please explain on lint issues? Everything looks good to me, so please point at what I should be looking for.

Will add empty lines.

tananaev:
You can check the CI output:

> Error: eckstyle] [ERROR] /home/runner/work/traccar/traccar/src/main/java/org/traccar/config/Keys.java:1316: Line is longer than 120 characters (found 134). [LineLength]

memesaregood1:
Fixed, please review.

tananaev:
It's still failing. Please make sure you run the gradle check locally first.

memesaregood1:
It runs just fine locally, that's the problem. The tasks run almost flawlessly. I was hoping it's related to Github Actions and your tests.

tananaev:
Now it looks like one of the unit tests is failing:
```
org.traccar.handler.ComputedAttributesTest > testComputedAttributes FAILED
    org.apache.commons.jexl3.JexlException$Method at ComputedAttributesTest.java:[54](https://github.com/traccar/traccar/actions/runs/4245922144/jobs/7382243428#step:4:55)
```

memesaregood1:
Try this one. Made features apply during expression execution, not during engine initialization.

tananaev:
What do you mean try? Have you tested it? I'm confused.

memesaregood1:
Wait, nevermind, I think I just finally got the tests failing on my machine now. Sorry for the confusion, will be back with the new commit in a minute.

Also, yes, I test everything I commit before pushing.

memesaregood1:
Fixed namespace. The test regarding the Math namespace was failing, and it's a mystery to me how it was passing before. You now, apart from including the namespace during engine initialization, provide it to the sandbox object in order for the engine to solve functions.

src/main/java/org/traccar/handler/ComputedAttributesHandler.java
```diff
@@ -102,7 +115,7 @@ private MapContext prepareContext(Position position) {
      */
     @Deprecated
     public Object computeAttribute(Attribute attribute, Position position) throws JexlException {
-        return engine.createExpression(attribute.getExpression()).evaluate(prepareContext(position));
+        return engine.createScript(features, engine.createInfo(), attribute.getExpression()).execute(prepareContext(position));
```
tananaev, line 118: What is the difference between `createExpression().evaluate()` and `createScript().execute()`?

memesaregood1: Let's start with the difference between expressions and scripts: [docs say](https://commons.apache.org/proper/commons-jexl/apidocs/org/apache/commons/jexl3/JexlExpression.html) that expression is a single, simple expression, and no declarations, loops or blocks are allowed there. Scripts, however, are a different story - they do exactly the same things `JexlExpression` does, but also allows for local variables, loops and so on to be defined and used. And, after all, I had to use `createScript()` to plug the `JexlFeatures` object during the expression execution.

`.execute()` and `.evaluate()` do the same thing (execute the given expression), they're just named different for some reason.

tananaev: A follow up question. With variables, are they persisted between executions or they are scoped to a single execution and cleared after?

memesaregood1: They're in their own scope. When the execution is compete, the variables are released.

tananaev:
Unfortunately we're back to lint issues.

memesaregood1:
The checkstyle gradle task doesn't let make a one-line return statement like it was before. Got to store data into variables for that to pass.

tananaev:
As far I can tell, the errors have nothing to do with that.

memesaregood1:
`ComputedAttributesHandler.java:118: Line is longer than 120 characters (found 127). [LineLength]`
119 characters of return line + 8 spaces.

tananaev:
Yeah, just add some line breaks. Why do you need variabled?

memesaregood1:
Not my usual code style, but that's one way to do it, sure. Will do tomorrow.

memesaregood1:
Patched lint issues. Review?

tananaev:
Merged, thank you.

memesaregood1:
Any way for me to update the documentation?

tananaev:
Do you mean add the configuration parameters to the website? Or write a separate page about it?

memesaregood1:
Yeah, I meant adding new configuration parameters. Plus, we have a page for computed attributes already, I can just add some examples regarding new syntax features there, if that's possible.

tananaev:
Configuration page is automatically generated based on the `Keys.java`. As for the computed attributes page, feel free to send me any updates or additions you want to make there and I'll apply them.

tananaev:
Thanks. Let's update the documentation once the release is out.

memesaregood1:
Bump? The release is out.

tananaev:
Updated the documentation here:

https://www.traccar.org/computed-attributes/

victor-butler:
@memesaregood1, may I ask for advice please?

In traccar I need to use `getTime()`  which used to work in previous versions. I believe the `method_calls` should be enabled for this and borrowing your code, I added the relevant configuration. However, the `getTime()` method still returns `null`. 

What am I missing? Your advice will be appreciated:

In `src/main/java/org/traccar/config/Keys.java`, I added this:

```
public static final ConfigKey<Boolean> PROCESSING_COMPUTED_ATTRIBUTES_METHOD_CALLS = new BooleanConfigKey(
            "processing.computedAttributes.method_calls",
            List.of(KeyType.CONFIG));
```
In `src/main/java/org/traccar/handler/ComputedAttributesHandler.java`, I edited this:

```
features = new JexlFeatures()
                .localVar(config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_LOCAL_VARIABLES))
                .loops(config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_LOOPS))
                .methodCall(config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_METHOD_CALLS)) //<-----The new line
                .newInstance(config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_NEW_INSTANCE_CREATION))
                .structuredLiteral(true);
```
Note: the `method_calls` are enabled it in the traccar.xml and server is restarted.

memesaregood1:
@victor-butler try the following and report back the results:

- try to disable the method_calls option and see if the result changes. it shall throw an error
- I believe the root of the issue is the JEXL queries being executed in a sandbox. in this case, try to bypass the sandbox for the sake of an experiment and see if the method returns the correct value.

> used to work in previous versions
can you tell what versions were those?

P.S. is there a reason you can't use the time parameter from the messages the device sends?

victor-butler:
@memesaregood1, thanks for the response!

I tired to set the `'processing.computedAttributes.method_calls` to `false`. Running the same command I am now getting an error:

Command: `serverTime.getTime()`
Error: 
`@1:11 method call error in '' - Feature (... < ComputedAttributesHandler:140 < AttributeResource:63 < ... < OverrideFilter:49 < ...)`

To bypass the sandbox, I assume I have to remove the below code. Is there anything else that will be needed?
```
JexlSandbox sandbox = new JexlSandbox(false);
sandbox.allow("com.safe.Functions");
```

I am not quite sure which was the last version this used to work. Maybe 5.6 or 5.8.

I need to compare dates, i.e. `serverTime.getTime() - deviceTime.getTime() < 10000`. Using the date object won't work.

memesaregood1:
If memory serves right, a JEXL engine without a sandbox would look like this:

```java
    @Inject
    public ComputedAttributesHandler(Config config, CacheManager cacheManager) {
        this.cacheManager = cacheManager;
        features = new JexlFeatures()
                .localVar(config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_LOCAL_VARIABLES))
                .loops(config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_LOOPS))
                .methodCall(config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_METHOD_CALLS)) //<-----The new line
                .newInstance(config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_NEW_INSTANCE_CREATION))
                .structuredLiteral(true);
        engine = new JexlBuilder()
                .strict(true)
                .namespaces(Collections.singletonMap("math", Math.class))
                .create();
        includeDeviceAttributes = config.getBoolean(Keys.PROCESSING_COMPUTED_ATTRIBUTES_DEVICE_ATTRIBUTES);
    }
```

victor-butler:
I confirm the method calls now run fine without sandbox enabled.

I am reading there are allowed and blocked classes by default. So in theory if we allow method calls it should work with sandbox, like:

`sandbox.allow(Foo.class.methodCall())`

What should we use instead of `Foo` in our particular case?

memesaregood1:
Since `getTime()` is a method of a `Date` class, objects of which are the serverTime and deviceTime parameters, we may as well just allow the Date class as a whole, since [the docs](https://commons.apache.org/proper/commons-jexl/apidocs/org/apache/commons/jexl3/introspection/JexlSandbox.html#allow-java.lang.String-) say that the `allow` method may only allow a **class**, not a particular method:

```java
    List.of(
        Double.class, Float.class, Integer.class, Long.class, Short.class,
        Character.class, Boolean.class, String.class, Byte.class, Date.class)
            .forEach((type) -> sandbox.allow(type.getName()));
```

tananaev:
This makes sense to me. Do you want to send a pr?

victor-butler:
I get an error on code compile:

```
/src/main/java/org/traccar/handler/ComputedAttributesHandler.java:69: error: cannot find symbol
                Character.class, Boolean.class, String.class, Byte.class, Date.class)
                                                                          ^
  symbol:   class Date
  location: class ComputedAttributesHandler
```
Some class declaration missing?

tananaev:
You're probably missing the import declaration.

victor-butler:
Yes, pretty much sure that is, however I am not sure which is the required declaration for the Date class.

memesaregood1:
```java
import java.util.Date;
```

victor-butler:
Ahh, thanks. I thought it was a JEXL related declaration. Testing and will confirm.

victor-butler:
Tested and it works. No need for additional `method_calls` computed attribute. Thanks @memesaregood1 !!!

PR #5037, opened 2023-02-22
Better handling of attributes (boolean attributes + setAccuracy special case) for Wialon protocol

truppelito:
Pull request as discussed [here](https://www.traccar.org/forums/topic/motion-attribute-using-wialon-protocol/#post-81534).

### DIFF
--- a/src/main/java/org/traccar/protocol/WialonProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/WialonProtocolDecoder.java
@@ -148,7 +148,16 @@ public class WialonProtocolDecoder extends BaseProtocolDecoder {
                     // (true, yes, etc), in which case we convert into a boolean
 
                     try {
-                        position.set(key, Double.parseDouble(value));
+                        double double_value = Double.parseDouble(value);
+
+                        // Since accuracy is not part of the general parameter list,
+                        // we need to handle it separately by calling setAccuracy directly
+                        if (key.equals("accuracy")) {
+                            position.setAccuracy(double_value);
+                        }
+                        else {
+                            position.set(key, double_value);
+                        }
                     } catch (NumberFormatException e) {
                         if (value.equalsIgnoreCase("true") || value.equalsIgnoreCase("t")
                             || value.equalsIgnoreCase("yes") || value.equalsIgnoreCase("y")
--- a/src/main/java/org/traccar/protocol/WialonProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/WialonProtocolDecoder.java
@@ -39,31 +39,31 @@ public class WialonProtocolDecoder extends BaseProtocolDecoder {
     }
 
     private static final Pattern PATTERN_ANY = new PatternBuilder()
-            .expression("([^#]+)?")              // imei
-            .text("#")                           // start byte
-            .expression("([^#]+)")               // type
-            .text("#")                           // separator
-            .expression("(.*)")                  // message
+            .expression("([^#]+)?") // imei
+            .text("#") // start byte
+            .expression("([^#]+)") // type
+            .text("#") // separator
+            .expression("(.*)") // message
             .compile();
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/WialonProtocolDecoder.java
```diff
@@ -39,31 +39,31 @@ public WialonProtocolDecoder(Protocol protocol) {
     }
 
     private static final Pattern PATTERN_ANY = new PatternBuilder()
-            .expression("([^#]+)?")              // imei
-            .text("#")                           // start byte
```
tananaev, line 43: Please revert all these formatting changes.

src/main/java/org/traccar/protocol/WialonProtocolDecoder.java
```diff
@@ -135,10 +135,41 @@ private Position decodePosition(Channel channel, SocketAddress remoteAddress, St
...
+                            position.setAccuracy(double_value);
+                        }
+                        else {
+                            position.set(key, double_value);
+                        }
                     } catch (NumberFormatException e) {
-                        position.set(paramParser.group(1).toLowerCase(), paramParser.group(2));
+                        if (value.equalsIgnoreCase("true") || value.equalsIgnoreCase("t")
+                            || value.equalsIgnoreCase("yes") || value.equalsIgnoreCase("y")
+                            || value.equalsIgnoreCase("on")) {
```
tananaev, line 163: Why do we need to check for all these values?

src/test/java/org/traccar/protocol/WialonProtocolDecoderTest.java
```diff
@@ -75,6 +78,15 @@ public void testDecode() throws Exception {
...
 
+        verifyAttributes(decoder, text(
+                "#D#120319;112003;NA;NA;NA;NA;0.000;NA;NA;0;NA;NA;NA;NA;NA;motion:3:true"));
+
+        verifyAttributes(decoder, text(
+                "#D#120319;112003;NA;NA;NA;NA;0.000;NA;NA;0;NA;NA;NA;NA;NA;motion:3:false"));
+
+        Position p = (Position)decoder.decode(null, null, text(
+                "#D#120319;112003;NA;NA;NA;NA;0.000;NA;NA;0;NA;NA;NA;NA;NA;accuracy:2:12.3"));
+        assertEquals(p.getAccuracy(), 12.3, 0.001);
```
tananaev, line 89: Looks a bit messy. Maybe we can just skip this test.

src/test/java/org/traccar/protocol/WialonProtocolDecoderTest.java
```diff
@@ -75,6 +78,15 @@ public void testDecode() throws Exception {
         verifyPositions(decoder, text(
                 "#B#110315;045857;5364.0167;N;06127.8262;E;0;155;965;7;2.40;4;0;14.77,0.02,3.6;AB45DF01145;"));
 
+        verifyAttributes(decoder, text(
+                "#D#120319;112003;NA;NA;NA;NA;0.000;NA;NA;0;NA;NA;NA;NA;NA;motion:3:true"));
+
+        verifyAttributes(decoder, text(
+                "#D#120319;112003;NA;NA;NA;NA;0.000;NA;NA;0;NA;NA;NA;NA;NA;motion:3:false"));
```
tananaev, line 85: One is probably enough. And use `verifyAttribute` so you can validate the value.

src/main/java/org/traccar/protocol/WialonProtocolDecoder.java
truppelito: We don't. It's up to you to decide which values you'd like to check for.

tananaev: Maybe let's start with true/false for now.

tananaev:
Merged, thank you.

truppelito:
Thank you!

Issue #5023, opened 2023-01-31
RegisterUnknown filter only allow number as identifier


bernamaxim:
Hi Anton,

I let Register Unknown be true.

``` <entry key='database.ignoreUnknown'>true</entry> ```
``` <entry key='database.registerUnknown'>true</entry> ```

But the problem is lots of strange data as new and unknown unit is entered, even unit without Identifier can be entered and register as new unit. Most are entered with an identifier in the form of a hexadecimal number.

Is it possible to implement the Traccar feature to filter out units like this (example: only allow numbers, contain certain words, etc.)

<img width="1273" alt="Screen Shot 2023-01-31 at 2 25 14 PM" src="https://user-images.githubusercontent.com/2969479/215695785-e0999f10-b206-4365-99c3-46be8100281f.png">

tananaev:
Seems like a good idea.

PR #5044, opened 2023-03-10
RegisterUnknown filter

bernamaxim:
RegisterUnknown filter only allow number and not empty uniqueId #5023


src/main/java/org/traccar/session/ConnectionManager.java
```diff
@@ -123,7 +123,9 @@ public DeviceSession getDeviceSession(
         Device device = deviceLookupService.lookup(uniqueIds);
 
         if (device == null && config.getBoolean(Keys.DATABASE_REGISTER_UNKNOWN)) {
-            device = addUnknownDevice(uniqueIds[0]);
+            if (uniqueIds[0].matches("[0-9]+") && !uniqueIds[0].isBlank()) {
```
tananaev, line 126: I think checking for `isBlank` is redundant here.

As for the regex, I think it's better to make it configurable. And for the default I would make it something like `\w{3,15}`. It would make it restrictive enough to filter the noise and generic enough to handle most of the identifiers.

bernamaxim: \w stands for "word character". Looks like it would match better with \d
\d{3,15} = matches any digit captures from 3 up to 15 digits
CMIIW

tananaev: Identifiers are not always all digits, so we need to make it more generic by default.

bernamaxim: "I think checking for isBlank is redundant here."

Please check this image
215695785-e0999f10-b206-4365-99c3-46be8100281f

There are one or more blank uniqueIds registered here

bernamaxim:
```
if (device == null && config.getBoolean(Keys.DATABASE_REGISTER_UNKNOWN)) {
    if (uniqueIds[0].matches("\\w{3,15}")) {
        if (uniqueIds[0].matches("[0-9]+") && config.getBoolean(Keys.DATABASE_REGISTER_UNKNOWN_NUMBER_ONLY)) {
            device = addUnknownDevice(uniqueIds[0]);
        } else {
            device = addUnknownDevice(uniqueIds[0]);
        }
    }
}
```

Like this?

tananaev: A blank string doesn't match the regex, so how does it get added? Am I missing something?

Your code snippet doesn't look correct. I think the regex itself should be configurable. It shouldn't be a boolean config.

bernamaxim:
```
if (device == null && config.getBoolean(Keys.DATABASE_REGISTER_UNKNOWN)) {
    if (uniqueIds[0].matches("\\w{3,15}")) {
        if (uniqueIds[0].matches(config.getString(Keys.DATABASE_REGISTER_UNKNOWN_REGEX))) {
            device = addUnknownDevice(uniqueIds[0]);
        } else {
            device = addUnknownDevice(uniqueIds[0]);
        }
    }
}
```

How it does like?

tananaev: No. Why do you have two regular expressions?

bernamaxim: > No. Why do you have two regular expressions?

```And for the default I would make it something like \w{3,15}```

I think it will be for default value, then if there are other can be added on config.

Comment (bernamaxim) [reply src/main/java/org/traccar/session/ConnectionManager.java]: Just like this?:
```
if (device == null && config.getBoolean(Keys.DATABASE_REGISTER_UNKNOWN)) {
    if (uniqueIds[0].matches(config.getString(Keys.DATABASE_REGISTER_UNKNOWN_REGEX))) {
        device = addUnknownDevice(uniqueIds[0]);
    } else {
        device = addUnknownDevice(uniqueIds[0]);
    }
}
```

tananaev: Why is there else statement?

bernamaxim: if (device == null && config.getBoolean(Keys.DATABASE_REGISTER_UNKNOWN)) {
    if (uniqueIds[0].matches(config.getString(Keys.DATABASE_REGISTER_UNKNOWN_REGEX))) {
        device = addUnknownDevice(uniqueIds[0]);
    }
}

tananaev: Yes, something like that.

PR #5047, opened 2023-03-13
RegisterUnknown filter with regex

bernamaxim:
RegisterUnknown filter with regex #5023

### DIFF
--- a/setup/default.xml
+++ b/setup/default.xml
@@ -39,6 +39,8 @@
     <entry key='commands.queueing'>true</entry>
 
     <entry key='database.ignoreUnknown'>true</entry>
+    <entry key='database.registerUnknown.regex'>\w{3,15}</entry>
+
     <entry key='database.changelog'>./schema/changelog-master.xml</entry>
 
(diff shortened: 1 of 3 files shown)


setup/default.xml
```diff
@@ -39,6 +39,8 @@
     <entry key='commands.queueing'>true</entry>
 
     <entry key='database.ignoreUnknown'>true</entry>
+    <entry key='database.registerUnknown.regex'>\w{3,15}</entry>
```
tananaev, line 42: Can you please move this default into the key definition instead.

bernamaxim: Do you mean move key to traccar.xml?

tananaev: No, I mean more it into the code.

bernamaxim: I'm sorry, Anton I cannot catch your point. 🙏
Can you give an example to make this is clear?

tananaev:
The CI check has failed. Make sure you run the test locally first.

tananaev:
Merged, thank you.

PR #5054, opened 2023-03-31
a fix to permission check

wkhaksar:
A quick fix to permission check. 

Currently, if a manager has access to a single user and a single device, but he/she can assign all other devices to this user, even those device to which the manager doesn't have acces, because the check was only applied to the owner but not property.

With this small change it will fix the issue. The check will be applied to both owner and property to ensure the manager has access to both of them.


tananaev:
Good catch, thank you.

wkhaksar:
You are welcome. 

I will also suggest a change to the API reference [here](https://www.traccar.org/api-reference/#tag/Permissions/paths/~1permissions/post), because adding a permission will return 204 if succeeded (not 200 as currently), same as removing.

Issue #4207, opened 2019-01-30
Feature Request: SSO Support


victorferreira:
It would be great if Traccar had support for Single Sign-On. This feature would allow Traccar to be easily integrated with others services. I've been looking into [Keycloak](https://www.keycloak.org) and seams like a great open source option for a self-hosted SSO and a good fit to be integrated into Traccar. I wonder how much work would take implement support for something like this. 

@tananaev do you have any recommendation on how to approach an integration like this?

tananaev:
It would be nice to have something that can be directly integrated into Traccar, not a standalone server.

victorferreira:
Just like there's third-party integrations for geocoding and map layer, it's possible to have more than one option. An implementation like this would be optional and additional to the system in place. For example, I could add the config bellow and point the authentication process to another service.

```
<entry key='auth.enabled'>true</entry>
<entry key='auth.service'>Keycloak</entry>
<entry key='auth.url'>https://auth.example.com</entry>
<entry key='auth.key'>h7vVHhjvlJVY8</entry>
```

In this case the end user could use the login and password they already. Also, having a standalone server would help build a micro-service architecture, instead of having everything inside Traccar. Complementary services like billing, support, etc... could easily integrated sharing the same credentials for the user.

tananaev:
I don't agree with your opinion. If it's a relatively small module and can be integrated directly, I don't see why we would want third party service.

As for micro-service architecture, it doesn't make sense for our case.

tsmgeek:
Keycloak is a good product that I use, but its just a oauth/saml provider so you would not intregate it directly into Traccar but instead build support for oauth and this can be done using the Traccar java libs, then you are free to use any other service you wish.
Basic idea but the actual implementation may be more complicated as you may want to accept additional roles and attributes to be passed ie admin/read only/TZ etc.

mxplusb:
I would love to have OAuth2/OIDC support built into Traccar, but I agree with @tananaev, it should be an integration, not something bundled.

mrabbah:
Hello, having the possibility to interface with a external OAuth2/OpenID Connect solution for SSO is interesting, because most company now have already their own Identity Access Management System, like Keycloak, Okta..., and all authorization and authentication are managed in those systems. The embedded SSO solution will be a nice to have, but does not answer to the need of most company.

Jafner:
[...]

PR #5055, opened 2023-04-01
Implement header authentication for reverse proxies

dan-r:
I host Traccar behind a reverse proxy that handles authentication. Because of this, having Traccar automatically authenticate users based on a header provided by the proxy would be useful.

This of course puts a lot of trust in the reverse proxy, so ideally an IP range check would be added too, but I wanted to share a 'prototype' solution to get the general consensus on this feature before continuing with it.

In my environment, to test this I set web.headerAuth to Remote-Email, which my haproxy instance fills with the users email. I tested and plan to use this with LDAP - it may make sense to lock it down to only LDAP, but again I wanted to get the views of the community/maintainers.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -764,15 +764,9 @@
           "required": true
         },
         "responses": {
-          "200": {
-            "description": "OK",
-            "content": {
-              "application/json": {
-                "schema": {
-                  "$ref": "#/components/schemas/Permission"
-                }
-              }
-            }
+          "204": {
+            "description": "No Content",
+            "content": {}
           },
           "400": {
             "description": "No permission",
@@ -786,16 +780,6 @@
         "tags": [
           "Permissions"
         ],
-        "requestBody": {
-          "content": {
-            "application/json": {
-              "schema": {
-                "$ref": "#/components/schemas/Permission"
-              }
-            }
-          },
-          "required": true
-        },
         "responses": {
           "204": {
             "description": "No Content",


tananaev:
Seems like a very specific use case. Not sure if it makes sense unless we can generify this.

I'm also not sure I understand the LDAP role in this. If your proxy is doing the authentication, then why do you need LDAP?

dan-r:
A more useful implementation may be using OIDC to implement proper SSO instead of this reverse proxy approach (there's some interest in #4207 by the looks of it).

With my specific implementation, the LDAP server just serves to assign the admin role and trigger the initial account creation.

tananaev:
Proper SSO with generic implementation will definitely be useful.

dan-r:
Okay I'll close this PR and start working on an SSO implementation.
 
@tananaev do you have any issue with an SSO implementation using OpenID Connect? 

From my point of view it seems like the current de facto standard and would allow authenticating against private directories like Keyclock, Authelia and Auth0 and also against third-parties like Google. 

My current implementation idea is a new endpoint (maybe under /api/session) to initiate the SSO login, and to provide a configuration option to either allow local login alongside SSO or to skip the login page and go straight to the SSO provider.

tananaev:
OpenID Connect makes sense. I would recommend drawing a simple diagram or something like that before actually starting coding, so we get on the same page about the implementation details.

dan-r:
So focusing on the OIDC code flow, I'd propose 2 new API endpoints:
- /api/session/openid/auth - Checks for an existing session and redirects to identity provider if necessary 
- /api/session/openid/callback - Callback URL for the identity provider

I'd also add the following configuration variables:
- oidc.enable - boolean state
- oidc.force - true = don't show Traccar login page, redirect straight to identity provider
- oidc.clientId - OIDC client ID
- oidc.authURL - OIDC authorization URL
- oidc.tokenURL - OIDC token URL
- oidc.userinfoURL - OIDC user info URL
- oidc.adminGroup - OIDC group to grant admin access

The proposed implementation steps would be:
- Add config keys to Config.java
- Add new API endpoints to SessionResource.java
- Create a new module in org.traccar.api.security to handle everything SSO
  - Potentially refactor some of LoginService.java if a lot of it is ending up copied to the new module
- Add a redirect for when oidc.force is true to either traccar-web logic or WebServer.java
- Update tests, Swagger spec and documentation

 Diagram

tananaev:
Sounds reasonable to me. I think we can go ahead with the implementation.

dan-r:
Just moved across to the issue at #5056, apologies, messed up this PR with my fork branch.

Issue #5056, opened 2023-04-01
Implement OpenID Connect SSO


dan-r:
Initially discussed in #5055

So focusing on the OIDC code flow, I'd propose 2 new API endpoints:
- /api/session/openid/auth - Checks for an existing session and redirects to identity provider if necessary 
- /api/session/openid/callback - Callback URL for the identity provider

I'd also add the following configuration variables:
- oidc.enable - boolean state
- oidc.force - true = don't show Traccar login page, redirect straight to identity provider
- oidc.clientId - OIDC client ID
- oidc.authURL - OIDC authorization URL
- oidc.tokenURL - OIDC token URL
- oidc.userinfoURL - OIDC user info URL
- oidc.adminGroup - OIDC group to grant admin access

The proposed implementation steps would be:
- Add config keys to Config.java
- Add new API endpoints to SessionResource.java
- Create a new module in org.traccar.api.security to handle everything SSO
  - Potentially refactor some of LoginService.java if a lot of it is ending up copied to the new module
- Add a redirect for when oidc.force is true to either traccar-web logic or WebServer.java
- Update tests, Swagger spec and documentation

![Untitled Diagram](https://user-images.githubusercontent.com/1384852/229298644-0f534013-d5ba-4a78-8f9a-884ec7bfc426.jpg)

tananaev:
Would you be interesting it writing a short documentation on how to set everything up? Maybe with one example provider, ideally something popular, like Google. I want to publish it in the documentation section on the website.

tananaev:
Yes, something like that.

tananaev:
Either one. I think generic like LDAP works. It would also be nice to have some real example.

dan-r:
Sure, I'll give it a go and put it in here

tananaev:
You can just prepare a Google doc or something like that. I'll transfer it to the website after. The website source code is not public.

tananaev:
Thanks for such a detailed documentation. I've published it on the website:

https://www.traccar.org/openid-sso/

I haven't linked to it from the main documentation page yet to not confuse people. With the next official release I'll link to it from some prominent place there and of course mention it in the release blog post.

PR #5057, opened 2023-04-02
Implement SSO with OpenID Connect

dan-r:
Implementing OIDC as discussed in #5056. I've also opened a PR on traccar-web (PR) for the frontend changes to show a 'Login with SSO' button and to redirect the user to the identity provider straight away if oidc.force is set.

I'm by no means an expert on OpenID Connect, but it works for me and seems to catch malformed and fabricated callbacks from my testing.

This change does require web.url to be set in order to form the redirect URL sent to the identity provider, not sure if this is a sane requirement or we should try and use the logic to 'guess' the URL here.

The dependencies are Apache 2.0 licensed, so shouldn't be a problem as far as I can see.

Mainly raising this now to get feedback and more testing, my Java is a bit rusty!

So far I've only this with my local instance of Authelia, configured as followed:
```
      - id: traccar
        description: Traccar
        secret: <REDACTED>
        public: false
        redirect_uris:
          - <REDACTED>
        scopes:
          - openid
          - profile
          - groups
          - email
```
with Traccar configured accordingly:
```
<entry key='oidc.enable'>true</entry>
<entry key='oidc.force'>true</entry>
<entry key='oidc.clientId'>traccar</entry>
<entry key='oidc.clientSecret'>REDACTED</entry>
<entry key='oidc.authUrl'>https://authelia/api/oidc/authorize</entry>
<entry key='oidc.tokenUrl'>https://authelia/api/oidc/token</entry>
<entry key='oidc.userInfoUrl'>https://authelia/api/oidc/userinfo</entry>
<entry key='oidc.adminGroup'>admins</entry>
<entry key='web.url'>https://traccar</entry>
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,6 +86,8 @@ dependencies {
     implementation "com.hivemq:hivemq-mqtt-client:1.3.0"
     implementation "redis.clients:jedis:4.3.1"
     implementation "com.google.firebase:firebase-admin:9.1.1"
+    implementation "com.nimbusds:oauth2-oidc-sdk:10.7.1"
+    implementation "net.minidev:json-smart:2.4.10"
     testImplementation "org.junit.jupiter:junit-jupiter-api:$junitVersion"
     testImplementation "org.junit.jupiter:junit-jupiter-engine:$junitVersion"
     testImplementation "org.mockito:mockito-core:5.1.1"
--- a/src/main/java/org/traccar/api/security/OpenIDProvider.java
+++ b/src/main/java/org/traccar/api/security/OpenIDProvider.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2017 - 2023 Anton Tananaev (anton@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.api.security;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.traccar.config.Config;
+import org.traccar.config.Keys;
+import org.traccar.model.User;
+
+public class OpenIDProvider {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(OpenIDProvider.class);
+
+    private final Boolean force;
+    private final String clientId;
+    private final String authUrl;
+    private final String tokenUrl;
+    private final String userInfoUrl;
+    private final String adminGroup;
+
+    public OpenIDProvider(Config config) {
+        force = config.getBoolean(Keys.OIDC_FORCE);
+        clientId = config.getString(Keys.OIDC_CLIENTID);
+        authUrl = config.getString(Keys.OIDC_AUTHURL);
+        tokenUrl = config.getString(Keys.OIDC_TOKENURL);
+        userInfoUrl = config.getString(Keys.OIDC_USERINFOURL);
+        adminGroup = config.getString(Keys.OIDC_ADMINGROUP);
+    }
+
+}
--- a/src/main/java/org/traccar/api/security/OpenIDProvider.java
+++ b/src/main/java/org/traccar/api/security/OpenIDProvider.java
@@ -32,6 +32,7 @@ import java.util.Date;
 import java.util.List;
 import java.io.IOException;
 import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import com.google.inject.Inject;
 
@@ -43,6 +44,7 @@ import com.nimbusds.oauth2.sdk.AuthorizationGrant;
 import com.nimbusds.oauth2.sdk.TokenRequest;
 import com.nimbusds.oauth2.sdk.TokenResponse;
 import com.nimbusds.oauth2.sdk.AuthorizationCodeGrant;
+import com.nimbusds.oauth2.sdk.AuthorizationErrorResponse;
 import com.nimbusds.oauth2.sdk.ParseException;
 import com.nimbusds.oauth2.sdk.AuthorizationResponse;
 import com.nimbusds.oauth2.sdk.auth.Secret;
@@ -134,7 +136,7 @@ public class OpenIDProvider {
         }
     }
 
-    private AuthorizationCode parseCallback(URI requri) {
+    private AuthorizationCode parseCallback(URI requri) throws WebApplicationException {
         AuthorizationResponse response;
 
         try {
@@ -144,7 +146,8 @@ public class OpenIDProvider {
         }
 
         if (!response.indicatesSuccess()) {
-            return null;
+            AuthorizationErrorResponse error = response.toErrorResponse();
+            throw new WebApplicationException(Response.status(403).entity(error.getErrorObject().getDescription()).build());
         }
 
         return response.toSuccessResponse().getAuthorizationCode();
@@ -196,19 +199,19 @@ public class OpenIDProvider {
         return user;
     }
 
-    public Response handleCallback(URI requri, HttpServletRequest request) throws StorageException {
+    public Response handleCallback(URI requri, HttpServletRequest request) throws StorageException, WebApplicationException {
         // Parse callback
         AuthorizationCode authCode = this.parseCallback(requri);
 
         if (authCode == null) {
-            return Response.ok().entity("Callback parse fail").build();
+            return Response.status(403).entity( "Invalid OpenID Connect callback.").build();
         }
 
         // Get token from IDP
         OIDCTokenResponse tokens = this.getToken(authCode);
 
         if (tokens == null) {
-            return Response.ok().entity("Token request failed").build();
+            return Response.status(403).entity("Unable to authenticate with the OpenID Connect provider. Please try again.").build();
         }
 
         BearerAccessToken bearerToken = tokens.getOIDCTokens().getBearerAccessToken();
@@ -217,7 +220,7 @@ public class OpenIDProvider {
         UserInfo idpUser = this.getUserInfo(bearerToken);
 
         if (idpUser == null) {
-            return Response.ok().entity("User info request failed").build();
+            return Response.status(500).entity("Failed to access OpenID Connect user info endpoint. Please contact your administrator.").build();
         }
 
         String email = idpUser.getEmailAddress();
--- a/src/main/java/org/traccar/api/security/OpenIDProvider.java
+++ b/src/main/java/org/traccar/api/security/OpenIDProvider.java
@@ -61,7 +61,7 @@ import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
 import com.nimbusds.openid.connect.sdk.claims.UserInfo;
 
 public class OpenIDProvider {
-    private final Boolean force;
+    public final Boolean force;
     private final ClientID clientId;
     private final Secret clientSecret;
     private URI callbackUrl;
--- a/src/main/java/org/traccar/api/security/OpenIDProvider.java
+++ b/src/main/java/org/traccar/api/security/OpenIDProvider.java
@@ -15,30 +15,228 @@
  */
 package org.traccar.api.security;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.traccar.config.Config;
 import org.traccar.config.Keys;
+import org.traccar.api.resource.SessionResource;
 import org.traccar.model.User;
+import org.traccar.storage.Storage;
+import org.traccar.storage.StorageException;
+import org.traccar.storage.query.Request;
+import org.traccar.storage.query.Columns;
+import org.traccar.helper.LogAction;
+import org.traccar.helper.ServletHelper;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Date;
+import java.util.List;
+import java.io.IOException;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.core.Response;
+import com.google.inject.Inject;
+
+import com.nimbusds.oauth2.sdk.http.HTTPResponse;
+import com.nimbusds.oauth2.sdk.AuthorizationCode;
+import com.nimbusds.oauth2.sdk.ResponseType;
+import com.nimbusds.oauth2.sdk.Scope;
+import com.nimbusds.oauth2.sdk.AuthorizationGrant;
+import com.nimbusds.oauth2.sdk.TokenRequest;
+import com.nimbusds.oauth2.sdk.TokenResponse;
+import com.nimbusds.oauth2.sdk.AuthorizationCodeGrant;
+import com.nimbusds.oauth2.sdk.ParseException;
+import com.nimbusds.oauth2.sdk.AuthorizationResponse;
+import com.nimbusds.oauth2.sdk.auth.Secret;
+import com.nimbusds.oauth2.sdk.auth.ClientSecretBasic;
+import com.nimbusds.oauth2.sdk.auth.ClientAuthentication;
+import com.nimbusds.oauth2.sdk.token.BearerAccessToken;
+import com.nimbusds.oauth2.sdk.id.State;
+import com.nimbusds.oauth2.sdk.id.ClientID;
+import com.nimbusds.openid.connect.sdk.OIDCTokenResponse;
+import com.nimbusds.openid.connect.sdk.Nonce;
+import com.nimbusds.openid.connect.sdk.OIDCTokenResponseParser;
+import com.nimbusds.openid.connect.sdk.UserInfoResponse;
+import com.nimbusds.openid.connect.sdk.UserInfoRequest;
+import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
+
+import com.nimbusds.openid.connect.sdk.claims.UserInfo;
 
 public class OpenIDProvider {
-
-    private static final Logger LOGGER = LoggerFactory.getLogger(OpenIDProvider.class);
-
     private final Boolean force;
-    private final String clientId;
-    private final String authUrl;
-    private final String tokenUrl;
-    private final String userInfoUrl;
+    private final ClientID clientId;
+    private final Secret clientSecret;
+    private URI callbackUrl;
+    private URI authUrl;
+    private URI tokenUrl;
+    private URI userInfoUrl;
+    private URI baseUrl;
     private final String adminGroup;
 
-    public OpenIDProvider(Config config) {
+    private Config config;
+    private LoginService loginService;
+    private Storage storage;
+
+    @Inject
+    public OpenIDProvider(Config config, LoginService loginService, Storage storage) {
         force = config.getBoolean(Keys.OIDC_FORCE);
-        clientId = config.getString(Keys.OIDC_CLIENTID);
-        authUrl = config.getString(Keys.OIDC_AUTHURL);
-        tokenUrl = config.getString(Keys.OIDC_TOKENURL);
-        userInfoUrl = config.getString(Keys.OIDC_USERINFOURL);
+        clientId = new ClientID(config.getString(Keys.OIDC_CLIENTID));
+        clientSecret = new Secret(config.getString(Keys.OIDC_CLIENTSECRET));
+
+        this.config = config;
+        this.storage = storage;
+        this.loginService = loginService;
+
+        try {
+            callbackUrl = new URI(config.getString(Keys.WEB_URL, "") + "/api/session/openid/callback");
+            authUrl = new URI(config.getString(Keys.OIDC_AUTHURL, ""));
+            tokenUrl = new URI(config.getString(Keys.OIDC_TOKENURL, ""));
+            userInfoUrl = new URI(config.getString(Keys.OIDC_USERINFOURL, ""));
+            baseUrl = new URI(config.getString(Keys.WEB_URL, ""));
+        } catch (URISyntaxException e) {
+        }
+
         adminGroup = config.getString(Keys.OIDC_ADMINGROUP);
     }
 
+    public URI createAuthRequest() {
+        AuthenticationRequest request = new AuthenticationRequest.Builder(
+                new ResponseType("code"),
+                new Scope("openid", "profile", "email", "groups"),
+                this.clientId,
+                this.callbackUrl)
+                .endpointURI(this.authUrl)
+                .state(new State())
+                .nonce(new Nonce())
+                .build();
+
+        return request.toURI();
+    }
+
+    private OIDCTokenResponse getToken(AuthorizationCode code) {
+        // Credentials to authenticate us to the token endpoint
+        ClientAuthentication clientAuth = new ClientSecretBasic(this.clientId, this.clientSecret);
+        AuthorizationGrant codeGrant = new AuthorizationCodeGrant(code, this.callbackUrl);
+
+        TokenRequest request = new TokenRequest(this.tokenUrl, clientAuth, codeGrant);
+        TokenResponse tokenResponse;
+
+        try {
+            HTTPResponse tokenReq = request.toHTTPRequest().send();
+            tokenResponse = OIDCTokenResponseParser.parse(tokenReq);
+            if (!tokenResponse.indicatesSuccess()) {
+                return null;
+            }
+
+            return (OIDCTokenResponse) tokenResponse.toSuccessResponse();
+        } catch (IOException e) {
+            return null;
+        } catch (ParseException e) {
+            return null;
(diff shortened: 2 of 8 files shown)


src/main/java/org/traccar/api/security/OpenIDProvider.java
```diff
@@ -0,0 +1,245 @@
...
+        user.setName(name);
+        user.setEmail(email);
+        user.setFixedEmail(true);
+        user.setDeviceLimit(this.config.getInteger(Keys.USERS_DEFAULT_DEVICE_LIMIT));
+
+        int expirationDays = this.config.getInteger(Keys.USERS_DEFAULT_EXPIRATION_DAYS);
+
+        if (expirationDays > 0) {
+            user.setExpirationTime(new Date(System.currentTimeMillis() + expirationDays * 86400000L));
+        }
```
tananaev, line 191: We should probably extract this logic somewhere to avoid duplication. Maybe `UserUtil`?

src/main/java/org/traccar/api/security/OpenIDProvider.java
```diff
@@ -0,0 +1,245 @@
...
+
+        return request.toURI();
+    }
+
+    private OIDCTokenResponse getToken(AuthorizationCode code) {
+        // Credentials to authenticate us to the token endpoint
+        ClientAuthentication clientAuth = new ClientSecretBasic(this.clientId, this.clientSecret);
+        AuthorizationGrant codeGrant = new AuthorizationCodeGrant(code, this.callbackUrl);
+
+        TokenRequest request = new TokenRequest(this.tokenUrl, clientAuth, codeGrant);
```
tananaev, line 121: Can this be reused? That way we don't need to keep so many instance variables.

src/main/java/org/traccar/api/security/OpenIDProvider.java
```diff
@@ -0,0 +1,245 @@
...
+            if (!tokenResponse.indicatesSuccess()) {
+                return null;
+            }
+
+            return (OIDCTokenResponse) tokenResponse.toSuccessResponse();
+        } catch (IOException e) {
+            return null;
+        } catch (ParseException e) {
+            return null;
+        }
```
tananaev, line 136: I would probably at least log these.

src/main/java/org/traccar/api/security/OpenIDProvider.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    private AuthorizationCode parseCallback(URI requri) throws WebApplicationException {
+        AuthorizationResponse response;
+
+        try {
+            response = AuthorizationResponse.parse(requri);
+        } catch (ParseException e) {
+            return null;
```
tananaev, line 145: Same here.

src/main/java/org/traccar/api/security/OpenIDProvider.java
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        if (!response.indicatesSuccess()) {
+            AuthorizationErrorResponse error = response.toErrorResponse();
+            throw new WebApplicationException(Response.status(403).entity(error.getErrorObject().getDescription()).build());
```
tananaev, line 150: This class should not use `WebApplicationException` directly.

src/main/java/org/traccar/api/security/OpenIDProvider.java
```diff
@@ -0,0 +1,245 @@
...
+
+            return (OIDCTokenResponse) tokenResponse.toSuccessResponse();
+        } catch (IOException e) {
+            return null;
+        } catch (ParseException e) {
+            return null;
+        }
+    }
+
+    private AuthorizationCode parseCallback(URI requri) throws WebApplicationException {
```
tananaev, line 139: I would probably inline this method.

src/main/java/org/traccar/api/security/OpenIDProvider.java
```diff
@@ -0,0 +1,245 @@
...
+
+            userInfoResponse = UserInfoResponse.parse(httpResponse);
+        } catch (IOException e) {
+            return null;
+        } catch (ParseException e) {
+            return null;
+        }
+
+        if (!userInfoResponse.indicatesSuccess()) {
+            // User info request failed - usually from expiring
```
tananaev, line 172: Can we log some useful information about the error?

src/main/java/org/traccar/api/security/OpenIDProvider.java
```diff
@@ -0,0 +1,245 @@
...
+        try {
+            HTTPResponse httpResponse = new UserInfoRequest(this.userInfoUrl, token)
+                    .toHTTPRequest()
+                    .send();
+
+            userInfoResponse = UserInfoResponse.parse(httpResponse);
+        } catch (IOException e) {
+            return null;
+        } catch (ParseException e) {
+            return null;
```
tananaev, line 168: I see you have a lot of try catches. I would move these higher up the chain.

tananaev (commented):
Thanks for working on this. It's one of the frequently requested features.

I did a first review pass. Overall seems like a pretty good implementation. Most of the comments are just minor formatting / styling / naming etc, so should be easy to fix.

src/main/java/org/traccar/api/security/OpenIDProvider.java
dan-r: Is this needed for our user creation? Doesn't look like we do it for LDAP at the moment, but we do when manually creating a user.

src/main/java/org/traccar/config/Keys.java
```diff
@@ -610,6 +610,62 @@ private Keys() {
             "ldap.adminGroup",
             List.of(KeyType.CONFIG));
 
+
+    /**
+     * OIDC enable.
```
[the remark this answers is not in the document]
dan-r: Agreed

build.gradle
```diff
@@ -86,6 +86,8 @@ dependencies {
     implementation "com.hivemq:hivemq-mqtt-client:1.3.0"
     implementation "redis.clients:jedis:4.3.1"
     implementation "com.google.firebase:firebase-admin:9.1.1"
+    implementation "com.nimbusds:oauth2-oidc-sdk:10.7.1"
+    implementation "net.minidev:json-smart:2.4.10"
```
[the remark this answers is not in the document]
dan-r: `net.minidev:json-smart` is a dependency of `com.nimbusds:oauth2-oidc-sdk`

src/main/java/org/traccar/api/security/OpenIDProvider.java
dan-r: clientAuth can be and thats the only place we use clientSecret, so I've moved that to a class property.

dan-r:
Thanks for the fast review! I've gone through and resolved all the issues as far as I can see, with the exception of:

> We should probably extract this logic somewhere to avoid duplication. Maybe `UserUtil`?

which I've left a comment on.

src/main/java/org/traccar/api/security/OpenIDProvider.java
tananaev: This is probably needed for LDAP as well. I think it was just never implemented there.

swagger.json
```diff
@@ -2769,6 +2821,12 @@
           "coordinateFormat": {
             "type": "string"
           },
+          "oidcEnabled": {
+            "type": "boolean"
+          },
+          "oidcForce": {
+            "type": "boolean"
+          },
```
tananaev, line 2829: Don't forget to rename these as well.

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -0,0 +1,170 @@
...
+
+    public URI createAuthUri() {
+        AuthenticationRequest request = new AuthenticationRequest.Builder(
+                new ResponseType("code"),
+                new Scope("openid", "profile", "email", "groups"),
+                clientId,
+                callbackUrl)
+                .endpointURI(authUrl)
+                .state(new State())
+                .nonce(new Nonce())
```
tananaev, line 105: Are these two needed?

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -0,0 +1,170 @@
...
+        adminGroup = config.getString(Keys.OPENID_ADMINGROUP);
+    }
+
+    public URI createAuthUri() {
+        AuthenticationRequest request = new AuthenticationRequest.Builder(
+                new ResponseType("code"),
+                new Scope("openid", "profile", "email", "groups"),
+                clientId,
+                callbackUrl)
+                .endpointURI(authUrl)
```
tananaev, line 103: This is kind of hard to read. Maybe split the builder creation from the rest?

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -0,0 +1,170 @@
...
+                .build();
+
+        return request.toURI();
+    }
+
+    private OIDCTokenResponse getToken(AuthorizationCode code) throws IOException, ParseException, GeneralSecurityException {
+        AuthorizationGrant codeGrant = new AuthorizationCodeGrant(code, callbackUrl);
+        TokenRequest tokenRequest = new TokenRequest(tokenUrl, clientAuth, codeGrant);
+
+        HTTPResponse tokenReq = tokenRequest.toHTTPRequest().send();
```
tananaev, line 115: This should be called `tokenResponse` or something like that, right?

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -0,0 +1,170 @@
...
+        TokenResponse tokenResponse = OIDCTokenResponseParser.parse(tokenReq);
+        if (!tokenResponse.indicatesSuccess()) {
+            LOGGER.warn("Invalid authorization code provided to OpenID callback");
```
tananaev, line 118: We probably don't need to log if we also throwing an exception, especially if it doesn't add any specific useful information.

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -0,0 +1,170 @@
...
+        UserInfoResponse userInfoResponse;
+
+        HTTPResponse httpResponse = new UserInfoRequest(userInfoUrl, token)
+                .toHTTPRequest()
+                .send();
+
+        userInfoResponse = UserInfoResponse.parse(httpResponse);
+
+        if (!userInfoResponse.indicatesSuccess()) {
+            LOGGER.error("Failed to access OpenID user info endpoint");
```
tananaev, line 135: Same here.

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -0,0 +1,170 @@
...
+        if (!tokenResponse.indicatesSuccess()) {
+            LOGGER.warn("Invalid authorization code provided to OpenID callback");
+            throw new GeneralSecurityException("Unable to authenticate with the OpenID Connect provider.");
+        }
+
+        return (OIDCTokenResponse) tokenResponse.toSuccessResponse();
+    }
+
+    private UserInfo getUserInfo(BearerAccessToken token) throws IOException, ParseException, GeneralSecurityException {
+        UserInfoResponse userInfoResponse;
```
tananaev, line 126: Why is this defined separately here? You can move it to the same line as the assignment.

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        return userInfoResponse.toSuccessResponse().getUserInfo();
+    }
+
+    public URI handleCallback(URI requestUri, HttpServletRequest request) throws StorageException, ParseException, IOException, GeneralSecurityException {
+        AuthorizationResponse response = AuthorizationResponse.parse(requestUri);
+
+        if (!response.indicatesSuccess()) {
+            LOGGER.warn("Callback received error response from OpenID identity provider");
```
tananaev, line 146: Same here

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -0,0 +1,170 @@
...
+            throw new GeneralSecurityException(response.toErrorResponse().getErrorObject().getDescription());
+        }
+
+        AuthorizationCode authCode = response.toSuccessResponse().getAuthorizationCode();
+
+        if (authCode == null) {
+            LOGGER.warn("Malformed OpenID callback");
```
tananaev, line 153: And here.

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -0,0 +1,170 @@
...
+                .state(new State())
+                .nonce(new Nonce())
+                .build();
+
+        return request.toURI();
+    }
+
+    private OIDCTokenResponse getToken(AuthorizationCode code) throws IOException, ParseException, GeneralSecurityException {
+        AuthorizationGrant codeGrant = new AuthorizationCodeGrant(code, callbackUrl);
+        TokenRequest tokenRequest = new TokenRequest(tokenUrl, clientAuth, codeGrant);
```
tananaev, line 113: I think I asked this original, but can this be stored as an instance? That way we don't need so many instance variables.

tananaev (commented):
Added some more comments.

tananaev:
Looks like there are also a bunch of lint issues. Make sure to run `gradle check` locally.

src/main/java/org/traccar/database/OpenIdProvider.java
dan-r: state is, nonce isnt, I'll take nonce out.

dan-r:
Just finished addressing the comments, will just sort out the linting now.

tananaev:
It looks like you updated the submodule for some reason. Can you please undo that.

dan-r:
Right, that should all be sorted

tananaev:
I still see the submodule change. Everything else looks great.

tananaev:
Merged. Thank you for contributing and also for fixing all the feedback quickly.

maltegrosse:
great work @dan-r - looking forward to integrate it with keycloak.
@tananaev is there any ETA for a next (docker/container) release?

tananaev:
No ETA, but probably within a few weeks.

dan-r:
@maltegrosse if you get a chance to build it yourself and try it out with Keycloak, it'd be great to get some extra feedback/testing!
I've tested Authelia (local) and Google so far.

maltegrosse:
@tananaev there arent any dev/nightly builds triggered by PRs somewhere on dockerhub, right? (looong time ago I worked with java :)

@dan-r I will give it a try the next days

dan-r:
@maltegrosse if you wanted to build your own Docker images from the preview, something like this should work:

https://github.com/dan-r/traccar-docker/tree/preview/preview

(Untested! 😀)

maltegrosse:
thank you both @dan-r  and @tananaev , I built my own docker based on dan-r docker files. 

I can not find anything regarding openid on the webui. I am using the same settings as [above](https://github.com/traccar/traccar/pull/5057#issue-1650941861) 

I checked the backend api endpoint `api/session/openid/auth` and it exists (post returns a 405 not 404)

I checked the frontend source code (good that is not minified at this point :) , and the openid changes are included, that means the latest preview includes the PR.

Behaviour:
Opening the default webpath `/` redirects me to the `/register`  endpoint.
I also can see in the frontend webdev console on `/api/session` throws a 404 on first visit (takes quite long), after reload the 404 dissapears. I guess it checks if an session exists for this visitor.

Do I miss some step enabling openid?

tananaev:
Try registering the admin account first.

maltegrosse:
Okay I got it running, sorry my fault, the config parameters are outdated in the post above:
`oidc.force -> openid.force` etc... as written in the [sso documentation ](https://www.traccar.org/openid-sso/)
I figured it out that somehow the config was not applied as the `/api/server` call returned `... ,"openIdEnabled":false," ...`

perhaps it should also be added to the [config page](https://www.traccar.org/configuration-file/) too (once its a release)

**So it works ! great work!**

Some notes:

1. A general openid role/group should be considered, e.g. anyone who not belongs to the group `traccar_user` is blocked. `openid.defaultGroup` (string or wildcard). The admin group works perfect. But restricting normal users on openid client side is somehow tricky...

2. Shouldnt the config just point to the url where the `.../.well-known/openid-configuration` is located? I think its an [openid spec](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest) and would reduce confusion :)

tananaev:
Both seem like good ideas, especially the second point.

maltegrosse:
just attach my setup if someone wants to reproduce with Keycloak:

My setup: treafik(running in some other docker compose file),postgres, docker-compose

```
version: "3.5"

services:
  postgres-traccar:
    image: postgres:14.6
    environment:
      POSTGRES_USER: traccar
      POSTGRES_PASSWORD: somepw
      POSTGRES_DB: traccar
      PGDATA: /var/lib/postgresql/data/pgdata
    hostname: postgres-traccar
    volumes:
      - /home/demo/gps/pgdata:/var/lib/postgresql/data
    restart: always
    networks:
      - backend

  traccar:
    image: local/traccar:preview
    hostname: traccar
    depends_on:
    - postgres-traccar
    restart: always
    ports:
    - "5000-5150:5000-5150" #some traefik tcp router should be considered
  #  - "8082:8082" traefik handles http port
    volumes:
    - ./conf/traccar.xml:/opt/traccar/conf/traccar.xml

    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.traccar.rule=Host(`traccar.example.com`)"
      - "traefik.http.routers.traccar.entrypoints=websecure"
      - "traefik.http.routers.traccar.tls.certResolver=le"
      - "traefik.docker.network=traefik_default"
      - "traefik.http.routers.traccar.priority=1"
      - "traefik.http.services.traccar.loadbalancer.server.port=8082"
    networks:
      - web
      - backend 
networks:
  web:
    external:
      name: traefik_default 
  backend:
    external: true
```

traccar.xml
```
<?xml version='1.0' encoding='UTF-8'?>
<!DOCTYPE properties SYSTEM 'http://java.sun.com/dtd/properties.dtd'>
<properties>

    <entry key="config.default">./conf/default.xml</entry>

    <entry key='database.driver'>org.postgresql.Driver</entry>
    <entry key='database.url'>jdbc:postgresql://postgres-traccar/traccar</entry>
    <entry key='database.user'>traccar</entry>
    <entry key='database.password'>somepw</entry>
    <entry key='openid.enable'>true</entry>
    <entry key='openid.force'>true</entry>
    <entry key='openid.clientId'>traccar</entry>
    <entry key='openid.clientSecret'>somesec</entry>
    <entry key='openid.authUrl'>https://auth.example.com/realms/REALMNAME/protocol/openid-connect/auth</entry>
    <entry key='openid.tokenUrl'>https://auth.example.com/realms/REALMNAME/protocol/openid-connect/token</entry>
    <entry key='openid.userInfoUrl'>https://auth.example.com/realms/REALMNAME/protocol/openid-connect/userinfo</entry>
    <entry key='openid.adminGroup'>admin</entry>
    <entry key='web.url'>https://traccar.example.com</entry>


</properties>

```

Keycloak:
1. Add new client
2. set to confidential
3. note the secret
4. add redirected url to `https://traccar.example.com/api/session/openid/callback`

Hints: 
1. Since few versions Keycloak does not include the "auth" in the `/auth/realms/REALMNAME/` path anymore, so its just `/realms/REALMNAME/`
2. By default Keycloak adds a `/` in front of every group. So either consider it in the traccar config or change it (somewhere - I forgot :-) in Keycloak

dan-r:
Good to hear, thanks for testing it @maltegrosse 😁 

I'll add those changes (might be a day or two before I have a chance). 

So:
- openid.defaultGroup - if set, restrict access to this group. Is it worth passing OpenID errors to the UI, or shall I just throw an exception for unauthorised?
- openid.issuerUrl - if set, the OpenIdProvider constructor pulls the configuration endpoint and sets auth/token/userInfoUrl automatically. 

My plan is to keep the manual configuration options too, unless there are any objections.

maltegrosse:
@dan-r  sounds great.  
If you willing to do some additional changes I am totally happy with this. :)
1) unauthorised would be fine for me
2) yes, perhaps some provider do not provide that endpoint
- some applications want them provided manually, eg [jupyterhub](https://oauthenticator.readthedocs.io/en/latest/tutorials/provider-specific-setup/providers/generic.html)
- [some application](https://github.com/skooner-k8s/skooner#oidc) just need clientid, secret and url

In addition, what do you think about a sign out url? Now signout just redirects back to the dashboard...

dan-r:
I'll wait on @tananaev opinion on the configuration options. Thanks for providing a couple of examples from other apps.

I'm torn on the idea of a sign out URL - when openid.force is enabled it does just redirect straight back to the sign in, but this may be sane behaviour. Our other options would be:
1. Log out of Traccar then redirect to identity provider logout URL (Which I believe the standard insists is /logout)
2. Log out and redirect somewhere arbitrary (config option for sign out URL)

Option 1 could cause frustrating behaviour for the user - why when I sign out of Traccar do I get signed out of my Google account too?
Option 2 could work but may also be a bit inconsistent.
Maybe its worth making this configurable, but not sure how necessary it is.

maltegrosse:
@dan-r yeah, thats true. There is the openid well known configuration about `end_session_endpoint` - but I could only find it in the keycloak config - not the google config, so better skip it for now :)

tananaev:
I think redirecting back to the login page makes sense to me as a default behavior. Adding logout URL would require us adding some special page in the web app just for that. I feel like it's a bit too much.

As for URL configuration, I think it's ok to support single URL and also individual URLs if it's required for some providers.

dan-r:
@maltegrosse The auto discovery using `openid.issuerUrl` and access restrictions with `openid.allowGroup` are now in (under #5064). I've tested with Google and Authelia, might be useful to check with your Keyclock setup if you get a minute!

maltegrosse:
@dan-r thank you, I tried to test the issuer url, but it fails, I guess traccar can not find the right urls in the well known file.

1. Requesting https://traccar.example.com/api/session/openid/auth, returns 303, 
2. then redirects to a HTTP ??  (the `s` is missing)

http://traccar.example.com/api/?response_type=code&redirect_uri=https%3A%2F%2Ftraccar.example.com%2F%2Fapi%2Fsession%2Fopenid%2Fcallback&state=something&client_id=traccar&scope=openid+profile+email+groups

not sure if the latest changes are in the [latest preview](https://traccar.nyc3.digitaloceanspaces.com/builds/traccar-other-preview.zip) ?
I rebuild your preview container with --no-cache flag....

comparing well-known of google vs keycloak, its somehow weird.... I perhaps the issue is regarding the content type:

- opening https://accounts.google.com/.well-known/openid-configuration in chrome shows a well formatted json, but in dev network tools you can not find any content-type definition in the response.

- opening https://sso.redhat.com/auth/realms/redhat-external/.well-known/openid-configuration (as a keycloak example), chrome show a unformatted json, but as content-type: application/json

so why does google work on your site, even you added json only on your request:... see https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/database/OpenIdProvider.java#L95

that confuses me :)

dan-r:
Looks like the preview build hasn't been updated since those changes went in. 

`curl -I https://traccar.nyc3.digitaloceanspaces.com/builds/traccar-other-preview.zip
HTTP/2 200 
content-length: 138393111
accept-ranges: bytes
last-modified: Tue, 04 Apr 2023 13:52:53 GMT`

maltegrosse:
thanks for the hin, @tananaev can you please trigger a new preview build?

tananaev:
https://github.com/traccar/traccar/actions/runs/4638714676

maltegrosse:
thanks @tananaev  @dan-r - the openid changes work well. great job!
so the new traccar config for keycloak looks like this:
```
<?xml version='1.0' encoding='UTF-8'?>
<!DOCTYPE properties SYSTEM 'http://java.sun.com/dtd/properties.dtd'>
<properties>
    <entry key="config.default">./conf/default.xml</entry>
    <entry key='database.driver'>org.postgresql.Driver</entry>
    <entry key='database.url'>jdbc:postgresql://postgres-traccar/traccar</entry>
    <entry key='database.user'>traccar</entry>
    <entry key='database.password'>somepw</entry>
    <entry key='openid.enable'>true</entry>
    <entry key='openid.force'>true</entry>
    <entry key='openid.clientId'>traccar</entry>
    <entry key='openid.clientSecret'>somesec</entry>
    <entry key='openid.issuerUrl'>https://auth.example.com/realms/REALMNAME</entry>
    <entry key='openid.adminGroup'>admin</entry>
    <entry key='web.url'>https://traccar.example.com</entry>
</properties>
```

dan-r:
Looks good but you don't need openid.enable, that was taken out as we just enable based on the other options being set.

masterwishx:
Sorry cant find info about authelia , i can put traccar behind nginx + Authela but then cant see gps clients .
using in Authelia config : 

  - domain: "*.mydomain.work"
        resources:
          - "^/api([/?].*)?$"
          - "^/rest([/?].*)?$" 
          - "^/identity.*$"
          - "^/triggers.*$"
          - "^/meshagents.*$"
          - "^/meshsettings.*$"
          - "^/agent.*$"
          - "^/control.*$"
          - "^/meshrelay.*$"
          - "^/wl.*$"
          .
          .
        policy: bypass

for clients can communicate ... with api

masterwishx:
@dan-r if I use also Nginx proxy manager, do I need some settings in it?

PR #5060, opened 2023-04-03
Minor tweaks to OpenID Connect integration

dan-r:
Including anything I come across whilst preparing documentation (work in progress).

### 1. Make adminGroup parameter optional/don't evaluate groups if it isnt provided
I think its sensible to require the openid, profile and email scopes as:
- openid - required
- email - we're mapping to accounts based on email! Aside: We should maybe look into using the 'sub' token in the future, and perhaps tie that in with storing the LDAP DN (I think they're analogous),
- profile - its useful to know the users name

Although my local SSO service (Authelia) supports groups and it seems like enterprise directories like Okta do too, public providers like Google seem not to, which does makes sense.

### DIFF
--- a/src/main/java/org/traccar/database/OpenIdProvider.java
+++ b/src/main/java/org/traccar/database/OpenIdProvider.java
@@ -94,9 +94,15 @@ public class OpenIdProvider {
     }
 
     public URI createAuthUri() {
+        Scope scope = new Scope("openid", "profile", "email");
+
+        if (adminGroup != null) {
+            scope.add("groups");
+        }
+
         AuthenticationRequest.Builder request = new AuthenticationRequest.Builder(
                 new ResponseType("code"),
-                new Scope("openid", "profile", "email", "groups"),
+                scope,
                 clientId,
                 callbackUrl);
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -94,9 +94,15 @@ public OpenIdProvider(Config config, LoginService loginService) {
     }
 
     public URI createAuthUri() {
+        Scope scope = new Scope("openid", "profile", "email");
+
+        if (adminGroup != null) {
+            scope.add("groups");
```
tananaev, line 100: So does it fail if you request it where it's not supported?

dan-r: It does, with Google you get a list of valid scopes and a list of invalid ones (just groups in this case)

tananaev: Then the PR looks good to me. Are you still working on it?

dan-r: Yeah I'll leave it drafted until I finish off the docs if thats good with you, incase I need to add anything else in

dan-r:
The only other thing I could see for now was that we don't actually stop the user authenticating with a username and password with the API when openID is forced. Its not 100% consistent with the LDAP implementation so of course open to changing how that second commit is done.

tananaev:
Merged, thanks.

PR #5061, opened 2023-04-04
Add DELETE api endpoint for Positions

### LINKED ISSUE #4387: [Feature request] delete position API endpoint

rovertzxd:
Add DELETE endpoint for Positions based on time span.
closes #4387

### DIFF
--- a/src/main/java/org/traccar/api/resource/PositionResource.java
+++ b/src/main/java/org/traccar/api/resource/PositionResource.java
@@ -31,6 +31,7 @@ import org.traccar.storage.query.Request;
 import javax.inject.Inject;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
@@ -43,6 +44,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
 import java.util.List;
+import java.util.LinkedList;
 
 @Path("positions")
 @Produces(MediaType.APPLICATION_JSON)
@@ -84,6 +86,21 @@ public class PositionResource extends BaseResource {
         } else {
             return PositionUtil.getLatestPositions(storage, getUserId());
         }
+    }
+
+    @DELETE
+    public Response remove(
+            @QueryParam("deviceId") long deviceId,
+            @QueryParam("from") Date from, @QueryParam("to") Date to) throws StorageException {
+        permissionsService.checkPermission(Device.class, getUserId(), deviceId);
+        permissionsService.checkRestriction(getUserId(), UserRestrictions::getReadonly);
+
+        var conditions = new LinkedList<Condition>();
+        conditions.add(new Condition.Equals("deviceId", deviceId));
+        conditions.add(new Condition.Between("fixTime", "from", from, "to", to));
+        storage.removeObject(Position.class, new Request(Condition.merge(conditions)));
+
+        return Response.status(Response.Status.NO_CONTENT).build();
     }
 
     @Path("kml")
--- a/swagger.json
+++ b/swagger.json
@@ -873,6 +873,54 @@
             }
           }
         }
+      },
+      "delete": {
+        "summary": "Deletes all the Positions of a device in the time span specified",
+        "description": "",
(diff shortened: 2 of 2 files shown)


src/main/java/org/traccar/api/resource/PositionResource.java
```diff
@@ -86,6 +88,25 @@ public Collection<Position> getJson(
...
+        if (from == null || to == null || deviceId <= 0) {
+            return Response.status(Response.Status.BAD_REQUEST).build();
+        }
+        permissionsService.checkPermission(Device.class, getUserId(), deviceId);
+        permissionsService.checkRestriction(getUserId(), UserRestrictions::getReadonly);
+        permissionsService.checkRestriction(getUserId(), UserRestrictions::getDeviceReadonly);
+
+        var conditions = new LinkedList<Condition>();
+        conditions.add(new Condition.Equals("deviceId", deviceId));
+        conditions.add(new Condition.Between("serverTime", "from", from, "to", to));
```
tananaev, line 104: This should be `fixTime` instead. We always use fix time by default.

src/main/java/org/traccar/api/resource/PositionResource.java
```diff
@@ -86,6 +88,25 @@ public Collection<Position> getJson(
         }
     }
 
+    @DELETE
+    public Response remove(
+            @QueryParam("deviceId") long deviceId, @QueryParam("from") Date from,
+            @QueryParam("to") Date to) throws StorageException {
```
tananaev, line 94:
```suggestion
            @QueryParam("deviceId") long deviceId,
            @QueryParam("from") Date from, @QueryParam("to") Date to) throws StorageException {
```

src/main/java/org/traccar/api/resource/PositionResource.java
```diff
@@ -86,6 +88,25 @@ public Collection<Position> getJson(
         }
     }
 
+    @DELETE
+    public Response remove(
+            @QueryParam("deviceId") long deviceId,
+            @QueryParam("from") Date from, @QueryParam("to") Date to) throws StorageException {
+        if (from == null || to == null || deviceId <= 0) {
+            return Response.status(Response.Status.BAD_REQUEST).build();
+        }
```
tananaev, line 97: We usually don't validate input parameters like this. It will fail later or if the parameters are wrong/missing.

src/main/java/org/traccar/api/resource/PositionResource.java
```diff
@@ -86,6 +88,25 @@ public Collection<Position> getJson(
...
+        permissionsService.checkPermission(Device.class, getUserId(), deviceId);
+        permissionsService.checkRestriction(getUserId(), UserRestrictions::getReadonly);
+        permissionsService.checkRestriction(getUserId(), UserRestrictions::getDeviceReadonly);
```
tananaev, line 100: Not sure about this one. I think "device readonly" user should still be able to remove positions.

tananaev (commented):
A couple of more small comments.

src/main/java/org/traccar/api/resource/PositionResource.java
rovertzxd: After reviewing other endpoints I think the following aligns more with the current code standards:
```suggestion
        if (deviceId > 0) {
            permissionsService.checkPermission(Device.class, getUserId(), deviceId);
            if (from != null && to != null) {
                permissionsService.checkRestriction(getUserId(), UserRestrictions::getReadonly);
                ....
            }
        }
```
What do you think?

tananaev: I think we should remove the checks completely.

src/main/java/org/traccar/api/resource/PositionResource.java
rovertzxd: I haven't explored the source enough to know what "readonly" and "device readonly" actually mean. Will remove the check for "device readonly"

src/main/java/org/traccar/api/resource/PositionResource.java
rovertzxd: Okay, sounds good.

tananaev:
Merged, thanks.

PR #5064, opened 2023-04-05
Updates to OpenID Implementation

dan-r:
Implements auto endpoint discovery and the ability to restrict access to groups.

Discussion in #5057

### DIFF
--- a/src/main/java/org/traccar/MainModule.java
+++ b/src/main/java/org/traccar/MainModule.java
@@ -75,6 +75,7 @@ import org.traccar.handler.GeolocationHandler;
 import org.traccar.handler.SpeedLimitHandler;
 import org.traccar.helper.ObjectMapperContextResolver;
 import org.traccar.helper.SanitizerModule;
+import org.traccar.helper.WebHelper;
 import org.traccar.mail.LogMailManager;
 import org.traccar.mail.MailManager;
 import org.traccar.mail.SmtpMailManager;
@@ -390,19 +391,7 @@ public class MainModule extends AbstractModule {
     public static VelocityEngine provideVelocityEngine(Config config) {
         Properties properties = new Properties();
         properties.setProperty("resource.loader.file.path", config.getString(Keys.TEMPLATES_ROOT) + "/");
-
-        if (config.hasKey(Keys.WEB_URL)) {
-            properties.setProperty("web.url", config.getString(Keys.WEB_URL).replaceAll("/$", ""));
-        } else {
-            String address;
-            try {
-                address = config.getString(Keys.WEB_ADDRESS, InetAddress.getLocalHost().getHostAddress());
-            } catch (UnknownHostException e) {
-                address = "localhost";
-            }
-            String url = URIUtil.newURI("http", address, config.getInteger(Keys.WEB_PORT), "", "");
-            properties.setProperty("web.url", url);
-        }
+        properties.setProperty("web.url", WebHelper.retrieveWebUrl(config));
 
         VelocityEngine velocityEngine = new VelocityEngine();
         velocityEngine.init(properties);
--- a/src/main/java/org/traccar/MainModule.java
+++ b/src/main/java/org/traccar/MainModule.java
@@ -26,7 +26,6 @@ import com.google.inject.name.Names;
 import io.netty.util.HashedWheelTimer;
 import io.netty.util.Timer;
 import org.apache.velocity.app.VelocityEngine;
-import org.eclipse.jetty.util.URIUtil;
 import org.traccar.broadcast.BroadcastService;
 import org.traccar.broadcast.MulticastBroadcastService;
 import org.traccar.broadcast.NullBroadcastService;
@@ -96,9 +95,7 @@ import javax.inject.Singleton;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientBuilder;
 import java.io.IOException;
-import java.net.InetAddress;
 import java.net.URISyntaxException;
-import java.net.UnknownHostException;
 import java.net.http.HttpClient;
 import java.util.Properties;
 
--- a/src/main/java/org/traccar/MainModule.java
+++ b/src/main/java/org/traccar/MainModule.java
@@ -97,6 +97,7 @@ import javax.ws.rs.client.ClientBuilder;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.net.http.HttpClient;
 import java.util.Properties;
 
 public class MainModule extends AbstractModule {
@@ -174,11 +175,12 @@ public class MainModule extends AbstractModule {
 
     @Singleton
     @Provides
-    public static OpenIdProvider provideOpenIDProvider(Config config, LoginService loginService) {
-        if (config.hasKey(Keys.OPENID_CLIENTID)) {
-            return new OpenIdProvider(config, loginService);
-        }
-        return null;
+    public static OpenIdProvider provideOpenIDProvider(
+        Config config, LoginService loginService, ObjectMapper objectMapper) throws InterruptedException, IOException {
+            if (config.hasKey(Keys.OPENID_CLIENT_ID)) {
+                return new OpenIdProvider(config, loginService, HttpClient.newHttpClient(), objectMapper);
+            }
+            return null;
     }
 
     @Provides
--- a/src/main/java/org/traccar/MainModule.java
+++ b/src/main/java/org/traccar/MainModule.java
@@ -96,6 +96,7 @@ import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientBuilder;
 import java.io.IOException;
 import java.net.InetAddress;
+import java.net.URISyntaxException;
 import java.net.UnknownHostException;
 import java.net.http.HttpClient;
 import java.util.Properties;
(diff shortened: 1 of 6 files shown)


src/main/java/org/traccar/config/Keys.java
```diff
@@ -636,19 +636,28 @@ private Keys() {
             "openid.clientSecret",
             List.of(KeyType.CONFIG));
 
+    /**
+     * OpenID Connect Issuer (Base) URL.
+     * This is used to automatically configure the authorization, token and user info URLs if
+     * they are not provided.
+     */
+    public static final ConfigKey<String> OPENID_ISSUERURL = new StringConfigKey(
```
tananaev, line 644: Actually can we update the keys to separate works. Haven't noticed it in the original PR. It should be something like `OPENID_ISSUER_URL` for this one.

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -82,15 +91,47 @@ public OpenIdProvider(Config config, LoginService loginService) {
...
+            } else {
+                authUrl = new URI(config.getString(Keys.OPENID_AUTHURL));
+                tokenUrl = new URI(config.getString(Keys.OPENID_TOKENURL));
+                userInfoUrl = new URI(config.getString(Keys.OPENID_USERINFOURL));
+            }
         } catch (URISyntaxException error) {
             LOGGER.error("Invalid URIs provided in OpenID configuration");
+        } catch (InterruptedException | IOException error) {
+            LOGGER.error("OpenID Connect auto discovery failed");
         }
```
tananaev, line 131: Do we need to catch these? I would just let it crash.

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -82,15 +91,47 @@ public OpenIdProvider(Config config, LoginService loginService) {
 
         try {
             callbackUrl = new URI(config.getString(Keys.WEB_URL, "") + "/api/session/openid/callback");
-            authUrl = new URI(config.getString(Keys.OPENID_AUTHURL, ""));
-            tokenUrl = new URI(config.getString(Keys.OPENID_TOKENURL, ""));
-            userInfoUrl = new URI(config.getString(Keys.OPENID_USERINFOURL, ""));
             baseUrl = new URI(config.getString(Keys.WEB_URL, ""));
+
+            if (
```
tananaev, line 96: I would simplify this to just take `OPENID_ISSUERURL` if it's set. Otherwise assume that separate URLs are provided.

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -82,15 +91,47 @@ public OpenIdProvider(Config config, LoginService loginService) {
...
+                    || !config.hasKey(Keys.OPENID_TOKENURL)
+                    || !config.hasKey(Keys.OPENID_USERINFOURL))
+                ) {
+                    HttpClient httpClient = HttpClient.newHttpClient();
+
+                    HttpRequest httpRequest = HttpRequest.newBuilder(
+                        URI.create(
+                            config.getString(Keys.OPENID_ISSUERURL) + "/.well-known/openid-configuration")
+                            )
+                        .header("accept", "application/json")
```
tananaev, line 109:
```suggestion
                        .header("Accept", "application/json")
```

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -82,15 +91,47 @@ public OpenIdProvider(Config config, LoginService loginService) {
...
+                    !config.hasKey(Keys.OPENID_AUTHURL)
+                    || !config.hasKey(Keys.OPENID_TOKENURL)
+                    || !config.hasKey(Keys.OPENID_USERINFOURL))
+                ) {
+                    HttpClient httpClient = HttpClient.newHttpClient();
+
+                    HttpRequest httpRequest = HttpRequest.newBuilder(
+                        URI.create(
+                            config.getString(Keys.OPENID_ISSUERURL) + "/.well-known/openid-configuration")
+                            )
```
tananaev, line 108:
```suggestion
                        URI.create(config.getString(Keys.OPENID_ISSUERURL) + "/.well-known/openid-configuration"))
```

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -82,15 +91,47 @@ public OpenIdProvider(Config config, LoginService loginService) {
...
             baseUrl = new URI(config.getString(Keys.WEB_URL, ""));
+
+            if (
+                config.hasKey(Keys.OPENID_ISSUERURL)
+                && (
+                    !config.hasKey(Keys.OPENID_AUTHURL)
+                    || !config.hasKey(Keys.OPENID_TOKENURL)
+                    || !config.hasKey(Keys.OPENID_USERINFOURL))
+                ) {
+                    HttpClient httpClient = HttpClient.newHttpClient();
```
tananaev, line 103: Can we instead inject the mapper and the client?

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -82,15 +91,47 @@ public OpenIdProvider(Config config, LoginService loginService) {
...
+                            )
+                        .header("accept", "application/json")
+                        .build();
+
+                    String httpResponse = httpClient.send(httpRequest, BodyHandlers.ofString()).body();
+
+                    Map<String, Object> discoveryMap = new ObjectMapper().readValue(
+                        httpResponse, new TypeReference<Map<String, Object>>() { });
+
+                    authUrl = new URI(discoveryMap.get("authorization_endpoint").toString());
```
tananaev, line 117: Can we do a type cast instead of `toString()`?

src/main/java/org/traccar/config/Keys.java
dan-r: To confirm, the config file property names are fine, just the internal need changing?

tananaev: Yes, just the Java keys.

dan-r:
That should be everything. I'll go through and update the docs once merged

src/main/java/org/traccar/MainModule.java
```diff
@@ -174,11 +175,12 @@ public static LdapProvider provideLdapProvider(Config config) {
...
-        if (config.hasKey(Keys.OPENID_CLIENTID)) {
-            return new OpenIdProvider(config, loginService);
-        }
-        return null;
+    public static OpenIdProvider provideOpenIDProvider(
+        Config config, LoginService loginService, ObjectMapper objectMapper) throws InterruptedException, IOException {
+            if (config.hasKey(Keys.OPENID_CLIENT_ID)) {
+                return new OpenIdProvider(config, loginService, HttpClient.newHttpClient(), objectMapper);
+            }
+            return null;
```
tananaev, line 183: The indentation should 4 instead of 8 spaces here.

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -69,28 +77,51 @@ public class OpenIdProvider {
...
+                    userInfoUrl = new URI((String) discoveryMap.get("userinfo_endpoint"));
+
+                    LOGGER.info("OpenID Connect auto discovery successful");
+                } else {
+                    authUrl = new URI(config.getString(Keys.OPENID_AUTH_URL));
+                    tokenUrl = new URI(config.getString(Keys.OPENID_TOKEN_URL));
+                    userInfoUrl = new URI(config.getString(Keys.OPENID_USERINFO_URL));
+                }
         } catch (URISyntaxException error) {
             LOGGER.error("Invalid URIs provided in OpenID configuration");
```
tananaev, line 120: I would probably keep this exception as well. An exception will actually be more useful because it will include line numbers.

tananaev: And we can probably remove try catch

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -69,28 +77,51 @@ public class OpenIdProvider {
...
+        ) throws InterruptedException, IOException {
+            this.loginService = loginService;
+
+            force = config.getBoolean(Keys.OPENID_FORCE);
+            clientId = new ClientID(config.getString(Keys.OPENID_CLIENT_ID));
+            clientAuth = new ClientSecretBasic(clientId, new Secret(config.getString(Keys.OPENID_CLIENT_SECRET)));
+
+            try {
+                callbackUrl = new URI(config.getString(Keys.WEB_URL, "") + "/api/session/openid/callback");
+                baseUrl = new URI(config.getString(Keys.WEB_URL, ""));
```
tananaev, line 96: We have this logic here to constructing default URL:

https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/MainModule.java#L390-L401

Maybe we can extract it into some common place and reuse here as well?

dan-r: Did you have anywhere specific in mind? None of the helper files seem to be an obvious choice to me

tananaev: Maybe add it to the `ServletHelper`? We can also rename it to something more generic, like `WebHelper`.

dan-r: Something like that?

tananaev:
Perfect, thank you again.

dan-r:
Any time. I've updated the documentation doc that should still be shared with you

tananaev:
Updated, thanks.

PR #5069, opened 2023-04-14
Extended Gator Protocol (M558FS)

abish7643:
## Functionalities Added
1. RFID Packet Decoding (Tested)
2. Alarm Packet Decoding (Power Off, SOS Tested)
3. Engine Start Encoding (Tested)
4. Engine Stop Decoding (Tested)

Attaching Logs for anyone to do Test Cases
```
24248000253c0d0733230414195650008307570765682900000000c0470100000c0f0000473000ff710d
2424210005718000d50d
24248000253c0d0733230414195720008307570765682900000000c0470100000c0f0000473600ff060d
2424210005068000a20d
24248000253c0d0733230414195750008307570765682900000000c0470100000c0f0000473a00ff7a0d
24242100057a8000de0d
24248000253c0d0733230414195820008307570765683000000000c0470100000c0f0000473d00ff1b0d
24242100051b8000bf0d
24248000253c0d0733230414195850008307570765682900000000c0470100000c0f0000474000ff0f0d
24242100050f8000ab0d
24248000253c0d0733230414195920008307580765682900000000c0470100000c0f0000474200ff730d
2424210005738000d70d
24248000253c0d0733230414195950008307580765683100000000c0470100000c0f0000474800ff110d
24248000253c0d0733230414200020008307590765683200000000c0470100000c0f0000474b00ff000d
2424210005008000a40d
24248000253c0d0733230414200050008307590765683200000000c0470100000c0f0000474d00ff760d
24248000253c0d0733230414200120008307590765683200000000c0470100000c0f0000475100ff1b0d
24242100051b8000bf0d
24248000253c0d0733230414200135008307590765683200000000c0470000000c0d0000475300ff0f0d
24242100050f8000ab0d
24248200273c0d0733230414200145008307590765683200000000c04700000000000000475300ff0008760d
2424210005768200d00d
24248000253c0d0733230414200149008307590765683200000000c04701000000000000475300ff730d
2424210005738000d70d
24247200203c0d07330242313631394143430d2a4123041420015600830758076568329f0d
24242100059f7200c90d
24247200203c0d07330231394143324630430d2a412304142002010083075807656831c90d
2424210005c972009f0d
24248000253c0d0733230414200219008307580765683200000000c0470100000c0d0000475500ff260d
2424210005268000820d
24248200273c0d0733230414200220008307580765683200000000c0470100000c0d0000475500ff00011e0d
24242100051e8200b80d
24248000253c0d0733230414200249008307580765683200000000c0470100000c0f0000475700ff760d
2424210005768000d20d
```


tananaev (commented):
Let's start with some big themes first:
1. Please add unit tests for both decoding and encoding
2. Fix formatting. For example, variable use our variable name convention. No underscores. Lower camel case names for functions and variables. No shortened words.
3. Run gradle check to make sure all checks pass.

abish7643:
> Let's start with some big themes first:
> 
>     1. Please add unit tests for both decoding and encoding
> 
>     2. Fix formatting. For example, variable use our variable name convention. No underscores. Lower camel case names for functions and variables. No shortened words.
> 
>     3. Run gradle check to make sure all checks pass.

I'll do all of this in the upcoming days.
Haven't worked on Java, may need some time.

abish7643:
Added new commit.

Gradle Build ran successfully without any StyleCode errors.

@abish7643
[Formatting Fixed, Naming Patterns Fixed, Gradle Build Successful](https://github.com/traccar/traccar/pull/5069/commits/c08d5bb26edc5b4dc1f8d795d952d4276e5aff55)

tananaev:
OK, can you please add unit tests and provide the protocol documentation.

abish7643:
> OK, can you please add unit tests and provide the protocol documentation.

Attaching Protocol Document. The protocol is exactly same as the one found on Protocol page.
[M508A_M508G_M528S_M508S_M558_M558T_M588F_M588FS_tracker_data_protocol.pdf](https://github.com/traccar/traccar/files/11248183/M508A_M508G_M528S_M508S_M558_M558T_M588F_M588FS_tracker_data_protocol.pdf)

debug.xml
```diff
@@ -27,4 +27,15 @@
...
+
+    <entry key='commands.queueing'>true</entry>
+
+    <entry key="filter.enable">true</entry>
+    <entry key="filter.zero">true</entry>
+
+    <entry key="filter.skipAttributes.enable">true</entry>
+    <entry key="filter.skipAttributes">command,event,result</entry>
+
+
```
tananaev, line 40: Please revert this.

src/main/java/org/traccar/protocol/GatorProtocolEncoder.java
```diff
@@ -0,0 +1,192 @@
...
+import org.traccar.helper.Checksum;
+import org.traccar.model.Command;
+
+public class GatorProtocolEncoder extends BaseProtocolEncoder {
+
+    public GatorProtocolEncoder(Protocol protocol) {
+        super(protocol);
+    }
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(Main.class);
```
tananaev, line 34: You shouldn't do logging here. We already log commands.

src/main/java/org/traccar/protocol/GatorProtocolEncoder.java
```diff
@@ -0,0 +1,192 @@
...
+        // Add Checksum
+        buf.writeByte(Checksum.xor(buf.nioBuffer(2, buf.writerIndex())));
+
+        // End of Packet -> 0D
+        buf.writeByte(0x0D);
+
+        return buf;
+    }
+
+    public int[] numToIp(String sim) {
```
tananaev, line 103: This function doesn't look right. First of all, the naming of the function and variables is not per our standards. Also, it looks too complicated. It just doesn't make sense. What are you trying to do here?

src/main/java/org/traccar/protocol/GatorProtocolEncoder.java
```diff
@@ -0,0 +1,192 @@
...
+
+    public static final int GET_POSITION = 0x30;
+    public static final int SET_ENGINE_START = 0x38;
+    public static final int SET_ENGINE_STOP = 0x39;
+
+
+    private ByteBuf engineExecute(String pseudoAddress, int engineState) {
+        ByteBuf buf = Unpooled.buffer(256);
+
+        // Add Header - 24 24
```
tananaev, line 44: Please remove most of these comments. Only include comment when absolutely necessary. Don't include comments for something that's immediately obvious from the code.

src/main/java/org/traccar/protocol/GatorProtocolEncoder.java
```diff
@@ -0,0 +1,192 @@
...
+        buf.writeByte(0x24);
+        buf.writeByte(0x24);
+
+        // Add Main Order - 0x38 for Engine Start, 0x39 for Engine Stop
+        buf.writeByte(engineState);
+
+        // Add Packet Length - 00 06
+        buf.writeByte(0x00);
+        buf.writeByte(0x06);
```
tananaev, line 53: Things like this should be added as "short", not two bytes.

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
```diff
@@ -120,17 +175,102 @@ protected Object decode(
...
+            if (deviceSession == null) {
+                return null;
+            }
+
+            position.setDeviceId(deviceSession.getDeviceId());
+            position.setTime(new Date());
+
+            getLastLocation(position, null);
+
+            // Main Order
```
tananaev, line 242: Remove all these comments as well

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
```diff
@@ -120,17 +175,102 @@ protected Object decode(
             position.set(Position.KEY_SATELLITES, flags & 0x0f);
 
             position.set(Position.KEY_STATUS, buf.readUnsignedByte());
-            position.set("key", buf.readUnsignedByte());
+            position.set(Position.KEY_IGNITION, buf.readUnsignedByte() == 0x01);
```
tananaev, line 178: Why are we changing this?

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
```diff
@@ -90,7 +101,51 @@ protected Object decode(
...
+            }
+
+            StringBuilder rfidData = new StringBuilder();
+            while (buf.readableBytes() > 0) {
+                int rfidDataByte = buf.readUnsignedByte();
+                if (rfidDataByte == 0x0D) {
+                    break;
+                }
+                rfidData.append((char) rfidDataByte);
+            }
```
tananaev, line 126: Use readCharSequence?

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
```diff
@@ -90,7 +101,51 @@ protected Object decode(
...
+        if (type == MSG_RFID_DATA) {
+            Position position = new Position(getProtocolName());
+
+            DeviceSession deviceSession = getDeviceSession(channel, remoteAddress, "1" + id, id);
+            if (deviceSession == null) {
+                return null;
+            }
+            position.setDeviceId(deviceSession.getDeviceId());
+
+            int subSignal = buf.readUnsignedByte();
```
tananaev, line 113: What is sub signal?

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
abish7643: It's clearly said in the protocol documents that this byte is regarding `ignition status`.
0x01 -> ON
0x00 -> OFF

The devices we have is also connected to ignition. I wonder why this protocol didn't use the already existing key.

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
tananaev: You can just use the search, so it shouldn't be a problem.

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
tananaev: Wait what? If it's a type, it should definitely not be called `isRFIDData`. And that name is not even consistent with our standards, as we discussed before already.

abish7643: Apologies. I'll use **subType**.

abish7643:
I have made the required changes as you mentioned, I need to test the device tomorrow since I have changed some parts which you suggested.

If there are any modification required, do mention.

tananaev:
If you use `getLastLocation`, you shouldn't set any GPS related fields, including validity.

abish7643:
Removed That Line. I'll test all the functionalities tomorrow and will update.

abish7643:
@tananaev 

Works perfect. Tested everything.
You could check, only a line was removed (setValid).

src/main/java/org/traccar/protocol/GatorProtocolEncoder.java
tananaev: There has to be a better way. This is what it looks like in the decoder:

https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/protocol/GatorProtocolDecoder.java#L51-L60

abish7643: @tananaev 

I simplified the encoding function.
Tested it with my actual devices.

https://github.com/abish7643/traccar/blob/Gator_Protocol_Improved/src/main/java/org/traccar/protocol/GatorProtocolEncoder.java#L82-L98

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
```diff
@@ -120,17 +176,99 @@ protected Object decode(
...
+
+            // Sub Type - Skip 1 Byte
+            buf.readUnsignedByte();
+
+            int commandExecuted = buf.readUnsignedByte();
+
+            // Reserved - Skip 2 Bytes
+            buf.readUnsignedShort();
+
+            position.set(Position.KEY_EVENT, Event.TYPE_COMMAND_RESULT);
```
tananaev, line 253: Do not do this. Event code should only include if device sends it.

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
```diff
@@ -120,17 +176,99 @@ protected Object decode(
...
+                if (mainType == 0x38) {
+                    position.set(Position.KEY_COMMAND, Command.TYPE_ENGINE_RESUME);
+                    position.set(Position.KEY_RESULT, "Engine Started");
+                } else if (mainType == 0x39) {
+                    position.set(Position.KEY_COMMAND, Command.TYPE_ENGINE_STOP);
+                    position.set(Position.KEY_RESULT, "Engine Stopped");
+                }
+            } else {
+                position.set(Position.KEY_RESULT, "Command Failed");
+            }
```
tananaev, line 265: These all are also incorrect. In the `KEY_RESULT` we only include the original data from the device. You should not have hardcoded text there.

src/main/java/org/traccar/protocol/GatorProtocolEncoder.java
```diff
@@ -0,0 +1,115 @@
...
+        buf.writeByte(ipAddress[3]);
+
+        buf.writeByte(Checksum.xor(buf.nioBuffer(2, buf.writerIndex())));
+
+        buf.writeByte(0x0D);
+
+        return buf;
+    }
+
+    private ByteBuf getFromDevice(String pseudoAddress, int type) {
```
tananaev, line 59: It seems like the two functions are basically identical, so I think they can be merged.

src/main/java/org/traccar/protocol/GatorProtocolEncoder.java
```diff
@@ -0,0 +1,115 @@
...
+        ByteBuf buf = Unpooled.buffer(256);
```
tananaev, line 60: You don't need to hardcode the length.

src/main/java/org/traccar/protocol/GatorProtocolEncoder.java
```diff
@@ -0,0 +1,115 @@
...
+
+        buf.writeByte(Checksum.xor(buf.nioBuffer(2, buf.writerIndex())));
+
+        buf.writeByte(0x0D);
+
+        return buf;
+    }
+
+    public int[] idToPseudoIPAddress(String deviceID) {
+        if (deviceID.length() != 11) {
```
tananaev, line 83: What is this for?

src/main/java/org/traccar/protocol/GatorProtocolEncoder.java
```diff
@@ -0,0 +1,115 @@
...
+
+public class GatorProtocolEncoder extends BaseProtocolEncoder {
+
+    public GatorProtocolEncoder(Protocol protocol) {
+        super(protocol);
+    }
+
+    public static final int GET_POSITION = 0x30;
+    public static final int SET_ENGINE_START = 0x38;
+    public static final int SET_ENGINE_STOP = 0x39;
```
tananaev, line 33: These are message types. We always define them in the decoder class. And they should be prefixed with `MSG_`. Make sure the naming also matched the protocol documentation.

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
```diff
@@ -90,7 +97,56 @@ protected Object decode(
...
+            if (rfidDataEndIndex != -1) {
+                int length = rfidDataEndIndex - buf.readerIndex();
+                rfidData.append(buf.toString(buf.readerIndex(), length, Charset.defaultCharset()));
+
+                buf.skipBytes(length + 1);
+            } else {
+                return null;
+            }
+
+            position.set(Position.KEY_DRIVER_UNIQUE_ID, rfidData.toString());
```
tananaev, line 128: I previously asked to simplify this, but it still seems overcomplicated. Why do you need a `StringBuilder`?

tananaev (commented):
I think we're almost there. I've added a few smaller comments, mostly around formatting.

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
abish7643: Then how would the user know that the command sent to the device was successfully executed?

I cannot create a notification / sound event if this is removed, how else to counter?

abish7643: But you asked not to hardcode KEY_RESULT, what received within the response from the device is just a byte.

Or should I use **TYPE_ENGINE_RESUME** in result?

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
abish7643: Are you suggesting to remove **KEY_COMMAND** and **KEY_RESULT**?

src/main/java/org/traccar/protocol/GatorProtocolDecoder.java
tananaev: Well, you should use the values from the device. You should almost never hardcode any values in the decoder. I would probably use `commandExecuted` as the `KEY_RESULT` and `mainType` as the `KEY_COMMAND`. This makes the most sense from what I can tell.

src/main/java/org/traccar/protocol/GatorProtocolEncoder.java
abish7643: I was about to mention that the gradle build was failing if I don't include that length check.

Definitely the test case failed but I'm not sure why that's the case. May be you could help.

image

tananaev: So which message is correct?

abish7643: @tananaev 

I have tracked it down, the issue lies in the substring creation block. The device ID should be at least 11 to continue with that logic.

```java
    public int[] idToPseudoIPAddress(String deviceID) {
        if (deviceID.length() != 11) {
            return new int[4];
        }

        int[] ipAddress = new int[4];
        for (int i = 0; i < 4; i++) {
            ipAddress[i] = Integer.parseInt(deviceID.substring(3 + i * 2, 5 + i * 2));
        }

        ipAddress[0] = ipAddress[0] % 0x7f;
        ipAddress[1] = ipAddress[1] % 0x7f;
        ipAddress[2] = ipAddress[2] % 0x7f;
        ipAddress[3] = ipAddress[3] % 0x7f;

        return ipAddress;
    }
```

According to the protocol document, the device id can be varied in length. The previous developer did assume that the device ID length was 11, that's the reason why the decoder function was simple while my initial encoder draft was quite long, taking care of the length. Since the decoder creates an ID with fixed length, there's no point in improving the capability of just the encoder. But in order get past Gradle build, I may have to add a conditional check on length there.

So what's your take on this, should I extend the decoder function just like how initially made the encoder function. Or should I keep it simple like this.

tananaev: If it's incorrect, then yes, ideally we should fix it.

PR #5070, opened 2023-04-15
Translate camel case names of alarms into proper readable names in notifications

gustavofcalmeida:
This fix has 2 main objectives:
1. to make the alarm names in notifications more readable to end user. E.g. "GPS Antenna Cut" instead of "gpsAntennaCut"
2. Facilitate localization of alarm names to other languages, since it's not possible using Transifex.

It's secure when new alarm entries are added. They will just have their normal camelCase name.


tananaev:
Merged, thank you.

PR #5077, opened 2023-05-05
Teltonika IO30 Fault count

K-J-Dod24:
Addition of Teltonika decode FMBXXX IO30 fault count, updated modern web pull request will be sent shortly.


tananaev:
Our policy for adding new attributes to `Position` is that it has be used at the very least by two completely different protocols. If it's not the case here, you should just add it as a custom attribute using a hardcoded string key.

K-J-Dod24:
New code changed is this now correct or do I need to perform a new pull request

tananaev:
Merged, thank you.

PR #5082, opened 2023-05-10
Teltonika io32 coolant temp/io115 engine temp

K-J-Dod24:
Coolant temperature sent from teltonika devices comes in as the io32 but is currently written in code as io115 which is engine temperature as sent from device. This is the backend code fixes to correctly separate and define the io's.


tananaev:
1. Please provide the documentation link.
2. It seems like you're making the same mistake as before. You're adding a new attribute to Position when it's only used by a single protocol.

tananaev:
I understand what you're trying to do, but we have a strict policy on when we add new parameters into `Position.java`, as I already explained before. If you refuse to follow it, we won't be able to take your change in.

tananaev:
Merged, thanks.

PR #5084, opened 2023-05-14
Disabled user login form message

iabesadze:
Hello

if the user is disabled it is advisable to tell him at the entrance that his account is disabled, now the system says that either Invalid username or password


tananaev:
What is this? Why are you creating a pull request from `jcache` branch?

PR #5087, opened 2023-05-17
Teltonika io69 GNSS status

K-J-Dod24:
Added IO69 GNSS Status to the teltonika decoder


tananaev:
Sorry, but I don't think we can take this change. We don't hardcode usually strings like that.

K-J-Dod24:
I am confused as to why this change cant be merged all its doing is assigning a human readable value to the io69 as it is in the Teltonika parameter list. there is other similarly coded parameters why is this one not able to be accepted?

K-J-Dod24:
image

        register(253, null, (p, b) -> {
            switch (b.readUnsignedByte()) {
                case 1:
                    p.set(Position.KEY_ALARM, Position.ALARM_ACCELERATION);
                    break;
                case 2:
                    p.set(Position.KEY_ALARM, Position.ALARM_BRAKING);
                    break;
                case 3:
                    p.set(Position.KEY_ALARM, Position.ALARM_CORNERING);
                    break;
                default:
                    break;
            }
        });

This bit of code already exists in the Teltonika Decoder and is written in the same way as my proposed code for decoding the values from io69 (GNSS Status) (we used this code as the example to write register 69) as I said all this code will do is assign a human readable value to the output of the sensor i.e. 

1= the value "GNSS On wtih fix"

image

K-J-Dod24:
I meant the format of the code being similar. And the values I am assigning to each case are the literal human readable outputs from the sensor as shown in the last image. Why can this not be done this way? How would we achieve this if not at this level, as there is no other output that could be assigned from this sensor. (see the previous image)

tananaev:
They are no localizable this value. Alarms are localized.

K-J-Dod24:
Sorry I didn't put this change through positions.java for localization this was because I was trying to abide by your rule for requiring multiple devices for updating positions. Is there another way I'm missing to add this into localization for integer to string translation?

tananaev:
Again, unfortunately it's not possible to take this change currently. To make it properly localized, it has to be used by at least some other protocols. You should just use computed attributes to convert the int value into whatever human readable text you need.

K-J-Dod24:
image

what about this bit of code here also in the teltonika decoder is this not the same as my proposed change ? or is this also different ?

tananaev:
I don't see any hardcoded strings as values.

K-J-Dod24:
Ok I will approach this sensor in a different way.

K-J-Dod24:
can this sensor be written in with the results shown below
image

but instead of being hardcoded in can they be written like below so that this can be put in the en.json for us and then it would be able to be translated for other languages using your method
image

PR #5093, opened 2023-05-25
Acknowledge mileage alert and decode third input status

edvalley:


tananaev:

tananaev:
Can you attach the documentation here instead?

edvalley:
[Protocol  for SF Series of GPS Tracking Product V1 16-0731.pdf](https://github.com/traccar/traccar/files/11943963/Protocol.for.SF.Series.of.GPS.Tracking.Product.V1.16-0731.pdf)
[Protocol  for S911 Series of GPS Tracking Product V1 45-1102.pdf](https://github.com/traccar/traccar/files/11943962/Protocol.for.S911.Series.of.GPS.Tracking.Product.V1.45-1102.pdf)

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -110,17 +110,26 @@ private String decodeAlarm(String event) {
...
 
-        if (event.length() == 1) {
-            char inputStatus = event.charAt(0);
-            if (inputStatus >= 'A' && inputStatus <= 'D') {
-                int inputStatusInt = inputStatus - 'A';
-                position.set(Position.PREFIX_IN + 1, inputStatusInt & 1);
-                position.set(Position.PREFIX_IN + 2, inputStatusInt & 2);
-                return null;
-            }
+        if (event.length() != 1) {
```
tananaev, line 113: Please keep the same structure as before. That way it's easier to see the difference.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -110,17 +110,26 @@ private String decodeAlarm(String event) {
...
+        } else if (inputStatus >= 'O' && inputStatus <= 'R') {
+            inputStatusInt = inputStatus - 'O' + 4;
+        } else {
+            return event;
         }
 
-        return event;
+        position.set(Position.PREFIX_IN + 1, inputStatusInt & 1);
+        position.set(Position.PREFIX_IN + 2, inputStatusInt & 2);
+        position.set(Position.PREFIX_IN + 3, inputStatusInt & 4);
```
tananaev, line 130: These should probably be using `BitUtil.check`.

edvalley:
Sorry about multiple commits, I used GitHub's web IDE. Changed code to be like before. I see your point, but IMHO, this way code is less compact. Do you still want me to change it to use BitUtil? Thanks.

tananaev:
Yes please.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -114,13 +115,21 @@ private String decodeEvent(String event, Position position) {
             char inputStatus = event.charAt(0);
             if (inputStatus >= 'A' && inputStatus <= 'D') {
                 int inputStatusInt = inputStatus - 'A';
-                position.set(Position.PREFIX_IN + 1, inputStatusInt & 1);
-                position.set(Position.PREFIX_IN + 2, inputStatusInt & 2);
+                position.set(Position.PREFIX_IN + 1, BitUtil.check(inputStatusInt, 0));
+                position.set(Position.PREFIX_IN + 2, BitUtil.check(inputStatusInt, 1));
+                position.set(Position.PREFIX_IN + 3, 0);
```
tananaev, line 120: Why is this zero?

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -114,13 +115,21 @@ private String decodeEvent(String event, Position position) {
...
+                return null;
+            } else if (inputStatus >= 'O' && inputStatus <= 'R') {
+                int inputStatusInt = inputStatus - 'O';
+                position.set(Position.PREFIX_IN + 1, BitUtil.check(inputStatusInt, 0));
+                position.set(Position.PREFIX_IN + 2, BitUtil.check(inputStatusInt, 1));
+                position.set(Position.PREFIX_IN + 3, 1);
```
tananaev, line 126: Is seems like these should be 4,5,6... Also why value 1 is hardcoded?

edvalley:
I think if you read the code in the very first commit you will get it...

edvalley:
You can also check the table on page 17 in the first document I uploaded before

edvalley:
That is:
On devices lacking input 3, will never see a report about it (so no O, P, Q or R event codes), so it will be always zero when we receive A, B, C or D event codes.
On devices having input 3, we will receive A, B, C or D event codes when input 3 is zero, and O, P, Q or R event codes when input 3 is one.

tananaev:
1. How do we handle models that don't have the third input?
2. The value should be boolean, not an integer.

edvalley:
> 1. How do we handle models that don't have the third input?

I guess there is no simple answer. We would need to condition setting (or not) input 3 according to the device model. To get the device model, we need to request it from the device by issuing a command. I think relaying on it won't be forward compatible and could be weak.

Otherwise, without taking device model into account, we can:
- If we receive O, P, Q or R event codes, we have for sure a model with input 3, and we will set it to TRUE.
- If we receive A, B, C or D event codes, we can not say if device have input 3 or not. So we can:
    - Not set input 3 value at all (incorrectly saying device lacks input 3 in case it has it)
    - Set input 3 to FALSE (incorrectly saying device have input 3 in case it don't) (The choose I already made)

Any suggestion?

> 2. The value should be boolean, not an integer.

You're right. I just kept it as it was before. Let's change it.

tananaev:
We have model name in the device settings. You can use that. If not specified, use only two inputs.

edvalley:
> We have model name in the device settings. You can use that. If not specified, use only two inputs.

So we hardcode comparisons with models known to have input 3?

tananaev:
Yes.

tananaev:
Looks like it doesn't compile.

edvalley:
> Looks like it doesn't compile.

Yeah. Sorry. Still using web editor.

edvalley:
Cloning...

edvalley:
Please, review whenever you can. Thanks...

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -209,6 +226,12 @@ protected Object decodeAvrmc(
             return null;
         }
 
+        String deviceModel = null;
+        Device device = getCacheManager().getObject(Device.class, deviceSession.getDeviceId());
+        if (device != null) {
+            deviceModel = device.getModel();
+        }
```
tananaev, line 233:
```suggestion
        String model = getCacheManager().getObject(Device.class, deviceId).getModel();
```

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -230,12 +253,17 @@ protected Object decodeAvrmc(
...
         position.set(Position.KEY_ODOMETER, parser.nextDouble() * 1000);
         position.set(Position.KEY_GPS, parser.nextInt());
         position.set(Position.PREFIX_ADC + 1, parser.nextDouble() * 0.001);
-        position.set(Position.PREFIX_ADC + 2, parser.nextDouble() * 0.001);
+
+        if ("AVL110".equals(deviceModel) || "AVL120".equals(deviceModel)) {
+            position.set(Position.PREFIX_ADC + 2, parser.nextDouble() * 0.001);
+        } else {
+            parser.next();
+        }
```
tananaev, line 266: Why are we changing this?

edvalley: > Why are we changing this?

There are models not having analog input 2 (SF-Lite for example). I don't know about other models (beside AVL110/120) that has it. Right now I don't remember if I wrote that part myself in the past, but that time, I didn't see the need to use device model to conditionally set analog input 2.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Is it possible that "device" is not still in cache? (Race condition?) I ask this because I'm having (unrelated?) errors in log (I've not debugged them yet) that reads: "WARN: Device xxx cache missing"

edvalley: Anyway, I can follow your suggestion and deffer solving this.

tananaev: There might be some very corner case condition, but in that case we want it to fail.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
tananaev: What do they report? Empty or actual value? I see some empty fields in the documentation.

edvalley: SF-Lite, always reports a zero for analog input 2, but physically, it lacks such input. Uploaded documentation is really old (I don't have recent one) but firmware is more recent. We can actually condition setting analog input 2 according to the field being empty or not, but, as I said, and as currently, we will see/store info about it even if the device lacks it. You tell me...

tananaev:
Merged, thanks.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
edvalley: Ok. Let me debug this and I'll open an issue or pull request. Thanks...

edvalley:
> Merged, thanks.

Thanks to you and your patience

PR #5095, opened 2023-05-28
fix NullPointerException when attribute key is defined and value is null

merabtenei:
Today I got a pretty annoying error that caused the server to not work properly.
The issue came after setting a user attributes field to something like this :
```
{
    ....
    attributes: {
        timezone: "Africa/Algiers",
        telegramChatId: null
    }
    ...
}
```
And got the following error on the server trying to generate a notification :
```
2023-05-28 16:25:51  INFO: [T3bc3c196] error - NullPointerException (ExtendedModel:94 < *:101 < NotificatorTelegram:93 < NotificationManager:110 < ...)
```
This caused by the line 94 in ExtendedModel:
```java
    public String getString(String key, String defaultValue) {
        if (attributes.containsKey(key)) {
            return attributes.get(key).toString(); // Here ******
        } else {
            return defaultValue;
        }
    }
```
If any attribute is explicitely set to null, the containsKey will return true but the actual value is null and calling toString will raise a NullPointerException.

### DIFF
--- a/src/main/java/org/traccar/model/ExtendedModel.java
+++ b/src/main/java/org/traccar/model/ExtendedModel.java
@@ -90,7 +90,7 @@ public class ExtendedModel extends BaseModel {
     }
 
     public String getString(String key, String defaultValue) {
-        if (attributes.containsKey(key)) {
+        if (attributes.containsKey(key) && attributes.get(key) != null) {
             return attributes.get(key).toString();
         } else {
             return defaultValue;


src/main/java/org/traccar/model/ExtendedModel.java
```diff
@@ -90,7 +90,7 @@ public void add(Map.Entry<String, Object> entry) {
     }
 
     public String getString(String key, String defaultValue) {
-        if (attributes.containsKey(key)) {
+        if (attributes.containsKey(key) && attributes.get(key) != null) {
             return attributes.get(key).toString();
```
tananaev, line 94: Null is a valid value. We can probably do something like this:

```suggestion
        if (attributes.containsKey(key)) {
            Object value = attributes.get(key);
            return value != null ? value.toString() : value;
```

merabtenei: I though that would be the same as getString(key) would call getString(key, defaultValue: null).
So the question is when calling getString(key, someValue) and the actual key is defined and its value is null, do we return null or do we return someValue (the default value) ?
Both are reasonable to me, We just need to avoid Exceptions that cause unexpected behaviors.

merabtenei:
Keeping the suggested changes fails to build due to the toString signature expecting a String in return. I don't know if changing the signature is ok. The first commit a suggested didn't cause this issue as the returned value is the passed defaultValue.

tananaev:
You can just return `null` instead to avoid the compilation error.

merabtenei:
@tananaev do you have an idea why tests are failing for this ?

src/main/java/org/traccar/model/ExtendedModel.java
```diff
@@ -91,7 +91,8 @@ public void add(Map.Entry<String, Object> entry) {
 
     public String getString(String key, String defaultValue) {
         if (attributes.containsKey(key)) {
-            return attributes.get(key).toString();
+            Object value = attributes.containsKey(key);
```
tananaev, line 94: Wait. There's a typo. It should obviously be `.get(key)` instead of `.containsKey(key)`

merabtenei: My bad. Fixed.

tananaev:
Merged, thanks.

PR #5096, opened 2023-05-28
do not add permission if user is serviceAccount

merabtenei:
This pull request allows creating objects (maintenance, geofence, notification) using the web.serviceAccountToken, this is useful in case of scripts that would for example create default maintenance, geofences and notifications for newly created users.
Resources are create without any link to a user, the script can then call api/permissions to link those resources to a given user.


tananaev:
Merged, thank you.

PR #5097, opened 2023-05-29
Viber notificator

merabtenei:
I added a ViberNotificator based on the TelegramNotificator, it works the same way.
You need a bot API Key and a user unique ID obtained from webhook sent by viber (read viber documentation for details: [https://developers.viber.com/docs/api/rest-bot-api/](https://developers.viber.com/docs/api/rest-bot-api/))
We tested this notificator and we are using it in our current project built on top of Traccar. We built an OTP system where we ask a user to send a code to our bot in order to get his/her viber_uid.
We are sharing this with the community in case you find this interesting.


src/main/java/org/traccar/config/Keys.java
```diff
@@ -1162,6 +1162,27 @@ private Keys() {
             "notificator.telegram.sendLocation",
             List.of(KeyType.CONFIG));
 
+    /**
+     * Viber notification API key.
+     */
+    public static final ConfigKey<String> NOTIFICATOR_VIBER_KEY = new StringConfigKey(
+                    "notificator.viber.key",
+                    List.of(KeyType.CONFIG));
```
tananaev, line 1170: Please fix indentation.

src/main/java/org/traccar/config/Keys.java
```diff
@@ -1162,6 +1162,27 @@ private Keys() {
...
+     * Viber notification chat id to post messages to.
+     */
+    public static final ConfigKey<String> NOTIFICATOR_VIBER_CHAT_ID = new StringConfigKey(
+                    "notificator.viber.uid",
+                    List.of(KeyType.CONFIG));
+
+    /**
+     * Viber notification send location message.
+     */
+    public static final ConfigKey<Boolean> NOTIFICATOR_VIBER_SEND_LOCATION = new BooleanConfigKey(
```
tananaev, line 1182: We should probably unify this config parameter between telegram and viber.

src/main/java/org/traccar/notificators/NotificatorViber.java
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright 2019 - 2023 Anton Tananaev (anton@traccar.org)
+ * Copyright 2021 Rafael Miquelino (rafaelmiquelino@gmail.com)
```
tananaev, line 3: Fix this. You can either put your name/email or keep mine, but the dates should be right.

src/main/java/org/traccar/notificators/NotificatorViber.java
```diff
@@ -0,0 +1,112 @@
...
+        chatId = config.getString(Keys.NOTIFICATOR_VIBER_CHAT_ID);
+        sendLocation = config.getBoolean(Keys.NOTIFICATOR_VIBER_SEND_LOCATION);
+    }
+
+    private LocationMessage createLocationMessage(String messageChatId, Position position) {
+        LocationMessage locationMessage = new LocationMessage();
+        locationMessage.chatId = messageChatId;
+        locationMessage.location = new LatLng();
+        locationMessage.location.latitude = position.getLatitude();
+        locationMessage.location.longitude = position.getLongitude();
```
tananaev, line 88:
```suggestion
        LatLng latLng = new LatLng();
        latLng.latitude = position.getLatitude();
        latLng.longitude = position.getLongitude();
        locationMessage.location = latLng;
```

src/main/java/org/traccar/notificators/NotificatorViber.java
```diff
@@ -0,0 +1,112 @@
...
+        private String chatId;
+        @JsonProperty("text")
+        private String text;
+        @JsonProperty("type")
+        private String parseMode = "text";
+        @JsonProperty("min_api_version")
+        private int minApiVersion = 1;
+    }
+
+    public static class LatLng {
```
tananaev, line 56: I would probably rename this to `Location`.

src/main/java/org/traccar/notificators/NotificatorViber.java
```diff
@@ -0,0 +1,112 @@
...
+        var shortMessage = notificationFormatter.formatMessage(user, event, position, "short");
+
+        TextMessage message = new TextMessage();
+        message.chatId = user.getString("viberUid");
+        if (message.chatId == null) {
+            message.chatId = chatId;
+        }
+        message.text = shortMessage.getBody();
+        client.target(urlSendText).request().header("X-Viber-Auth-Token",
+                config.getString(Keys.NOTIFICATOR_VIBER_KEY)).post(Entity.json(message)).close();
```
tananaev, line 103: I think you should get the key once in the constructor and save it in a variable.

Also please reformat the request to something like:
```suggestion
        client.target(urlSendText).request()
                .header("X-Viber-Auth-Token", ...)
                .post(Entity.json(message))
                .close();
```

src/main/java/org/traccar/notificators/NotificatorViber.java
```diff
@@ -0,0 +1,112 @@
...
+        if (sendLocation && position != null) {
+            client.target(urlSendLocation).request().header("X-Viber-Auth-Token",
+                    config.getString(Keys.NOTIFICATOR_VIBER_KEY)).post(
+                            Entity.json(createLocationMessage(message.chatId, position)))
```
tananaev, line 107: Same here.

src/main/java/org/traccar/config/Keys.java
merabtenei: What formatter are you using, because my Visual Studio code uses a different formatting and it changes formatting as soon as I save, I had to manually save without formatting.

src/main/java/org/traccar/config/Keys.java
merabtenei: What would we name this unified parameter ? Is NOTIFICATOR.SENDLOCATION = "notificator.sendLocation" okey ?

merabtenei: I think we may keep this as it is for now to ensure backward compatibility, changing the parameter name would require modifications when upgrading from an older version for anyone using this features. What do you think ?

src/main/java/org/traccar/config/Keys.java
tananaev: I use IntelliJ.

src/main/java/org/traccar/config/Keys.java
tananaev: `NOTIFICATOR_SEND_LOCATION = "notificator.sendLocation"` sounds good to me. I think it's a rarely used key, so it's ok to break backward compatibility. Also, it doesn't really break notifications. You just won't get the location info.

rkalman:
It is nice, but unfortunatelly a more interesting part is missing: getting viber uids from the viber webhook callbacks.

PR #5107, opened 2023-06-07
Feature/redis broadcast

reajuria:
Implementation of Redis PubSub as broadcast layer. Currently VPCs on cloud providers don't support Multicast Ex. [GCP](https://cloud.google.com/vpc/docs/vpc#specifications).

PD. tananaev, might be missing a few things as I've little experience with Java... open to corrections/suggestions, thanks.


src/main/java/org/traccar/MainModule.java
```diff
@@ -340,8 +341,16 @@ public static SpeedLimitHandler provideSpeedLimitHandler(@Nullable SpeedLimitPro
     @Provides
     public static BroadcastService provideBroadcastService(
             Config config, ObjectMapper objectMapper) throws IOException {
+        String broadcastType = config.getString(Keys.BROADCAST_TYPE);
         if (config.hasKey(Keys.BROADCAST_ADDRESS)) {
-            return new MulticastBroadcastService(config, objectMapper);
+            switch (broadcastType) {
```
tananaev, line 346: I think it makes sense to just do switch on `config.getString(Keys.BROADCAST_TYPE)` without checking `BROADCAST_ADDRESS`.

src/main/java/org/traccar/broadcast/RedisBroadcastService.java
```diff
@@ -0,0 +1,206 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(RedisBroadcastService.class);
+
+    private final ObjectMapper objectMapper;
+
+    private final ExecutorService service = Executors.newSingleThreadExecutor();
+
+    private final Set<BroadcastInterface> listeners = new HashSet<>();
+
+    private final String url;
+    private final String pubsubChannel = "traccar:cast";
```
tananaev, line 50: Maybe just call it "traccar"? I would also rename the variable to just `channel`.

src/main/java/org/traccar/broadcast/RedisBroadcastService.java
```diff
@@ -0,0 +1,206 @@
...
+
+    private final Jedis subscriberJedis;
+    private final Jedis publisherJedis;
```
tananaev, line 53: Remove Jedis suffix.

src/main/java/org/traccar/broadcast/RedisBroadcastService.java
```diff
@@ -0,0 +1,206 @@
...
+
+    private final String id;
+
+    public RedisBroadcastService(Config config, ObjectMapper objectMapper) throws IOException {
+        this.objectMapper = objectMapper;
+        url = config.getString(Keys.BROADCAST_ADDRESS);
+
+        subscriberJedis = new Jedis(url);
+        publisherJedis = new Jedis(url);
```
tananaev, line 62: Do we actually need two instances of Jedis?

src/main/java/org/traccar/broadcast/RedisBroadcastService.java
```diff
@@ -0,0 +1,206 @@
...
+
+        // id that will be used to identify this instance of the server
+        id = String.valueOf(System.currentTimeMillis());
```
tananaev, line 65: You can probably assign `id` directly in declaration. And maybe it's better to use something like UUID?

src/main/java/org/traccar/broadcast/RedisBroadcastService.java
```diff
@@ -0,0 +1,206 @@
...
+        message.setCommandDeviceId(deviceId);
+        sendMessage(message);
+    }
+
+    @Override
+    public void invalidateObject(boolean local, Class<? extends BaseModel> clazz, long id) {
+        BroadcastMessage message = new BroadcastMessage();
+        message.setChanges(Map.of(Permission.getKey(clazz), id));
+        sendMessage(message);
+    }
```
tananaev, line 112: There's a lot of duplication. Can we make a base class with all common stuff that doesn't change between implementations?

src/main/java/org/traccar/broadcast/RedisBroadcastService.java
```diff
@@ -0,0 +1,206 @@
...
+                }
+            }, pubsubChannel);
+
+            while (!service.isShutdown()) {
+                try {
+                    Thread.sleep(1000);
+                } catch (InterruptedException e) {
+                    break;
+                }
+            }
```
tananaev, line 194: Is there a better way of waiting? Do we even need this?

src/main/java/org/traccar/broadcast/RedisBroadcastService.java
```diff
@@ -0,0 +1,206 @@
...
+
+            try {
+                subscriberJedis.close();
+                publisherJedis.close();
+            } catch (Exception e) {
+                LOGGER.warn("Failed to close pubsub", e);
+                throw new RuntimeException(e);
+            }
```
tananaev, line 202: It's not a good idea to catch generic `Exception`. And you should either throw or log. Otherwise you will have duplication in logs. I prefer throw because that's what multicast does.

src/main/java/org/traccar/broadcast/RedisBroadcastService.java
```diff
@@ -0,0 +1,206 @@
...
+    private final String pubsubChannel = "traccar:cast";
+
+    private final Jedis subscriberJedis;
+    private final Jedis publisherJedis;
+
+    private final String id;
+
+    public RedisBroadcastService(Config config, ObjectMapper objectMapper) throws IOException {
+        this.objectMapper = objectMapper;
+        url = config.getString(Keys.BROADCAST_ADDRESS);
```
tananaev, line 59: Probably need to update javadoc for `BROADCAST_ADDRESS`.

src/main/java/org/traccar/broadcast/RedisBroadcastService.java
reajuria: Positive about [subscribe](https://redis.io/commands/subscribe/), need to check how Jedis works

src/main/java/org/traccar/broadcast/RedisBroadcastService.java
reajuria: Could be moved to BroadcastService?

tananaev: It's not needed for Noop implementation, so I would just make another base class like `BaseBroadcastService`.

src/main/java/org/traccar/MainModule.java
reajuria: Hi, I didn't want to introduce a breaking change in here, on Keys made BROADCAST_TYPE default to "multicast" to match current behavior, with that in mind would you either prefer to remove BROADCAST_TYPE default and BROADCAST_ADDRESS check or switch's default? thanks.

reajuria:
@tananaev made some changes as requested, please let me know if more changes are needed, thanks.

src/main/java/org/traccar/MainModule.java
tananaev: Let's do a breaking change. It's a relatively new feature, so should be fine.

src/main/java/org/traccar/broadcast/BaseBroadcastService.java
```diff
@@ -0,0 +1,118 @@
...
+                } else {
+                    listeners.forEach(listener -> listener.invalidateObject(
+                            false,
+                            Permission.getKeyClass(first.getKey()), first.getValue()));
+                }
+            }
+        }
+    }
+
+}
```
tananaev, line 118: Add line break

reajuria:
@tananaev made requested changes, please have a look again, thanks.

src/main/java/org/traccar/config/Keys.java
```diff
@@ -1755,6 +1755,15 @@ private Keys() {
...
             "time,position,speed,course,accuracy,result");
 
+    /**
+     * Broadcast method. Available options are "multicast" and "redis". By default (if the value is not
+     * specified or does not matches available options) server disables broadcast.
+     */
+    public static final ConfigKey<String> BROADCAST_TYPE = new StringConfigKey(
+            "broadcast.type",
+            List.of(KeyType.CONFIG),
+            "");
```
tananaev, line 1765: Can we remove this empty string?

reajuria: It prevents NullPointerException to be thrown by broadcast provider injector 😅

tananaev: Yeah, but I would rather handle null separately.

tananaev:
Merged, thank you.

PR #5111, opened 2023-06-09
Add ability to filter devices by attribute

### LINKED ISSUE #3776: Get devices by attribute
Hi.

I was trying to get devices by attribute but I'm having problems to do that.
I have created an attribute as "master" and I want to get all devices with "master" attribute equals "X".

Anyone know how to do this?

I just want to update all slaves devices that with master position if is an specific status.

rovertzxd:
Fixes #3776 
This simply adds a new query parameter to filter by attributes. My use case is similar to @marlomgirardi. Need to get the devices with a specified attribute and value. As @tsmgeek mentions you can just pull the complete list of devices and filter from there, but my concern that the request could take longer as you have more devices.

My intention is to request comments and feedback and maybe hear from the community any other use cases that are not covered by this implementation.

The current limitiations are:
- `JSON_CONTAINS` is specific to MySQL and MariaDB (it doesn't work on H2 databases);
- The filter looks for an exact match
- If you add multiple attribute parameters to the query it all conditions must match

**Additional Notes:**
I need to add documentation, but before I do that I want to get feedback.

### DIFF
--- a/src/main/java/org/traccar/api/resource/DeviceResource.java
+++ b/src/main/java/org/traccar/api/resource/DeviceResource.java
@@ -75,7 +75,8 @@ public class DeviceResource extends BaseObjectResource<Device> {
     public Collection<Device> get(
             @QueryParam("all") boolean all, @QueryParam("userId") long userId,
             @QueryParam("uniqueId") List<String> uniqueIds,
-            @QueryParam("id") List<Long> deviceIds) throws StorageException {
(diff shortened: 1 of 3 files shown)


tananaev:
Pulling the database and then filtering would be almost the same thing as filtering it in the database. That's because there are no indexes that would help you. So it doesn't really make sense to add this complexity.

PR #5113, opened 2023-06-11
Ruptela IO decoding for Odometer, OBD Odometer and OBD Speed

jinzo:
Hi,

Ruptela IO List: [FMIO list.xlsx](https://doc.ruptela.com/resources/Storage/FMIO%20List/FMIO%20list.xlsx)

I want to update the Ruptela Protocol decoder with more data from the official FMIO list. First commit is a basic implementation of Odometer, OBD Odometer and OBD Speed -> which put me in a bit of a dilemma. Odometer and OBD Odometer map cleanly to the existing Position KEYs, but there are several possible IO Attributes that could map for OBD Speed for example. I currently took the one from the same group as OBD Odometer, as this looked the more sane option. But IO "165  - TCO Vehicle Speed" could also map into this KEY and "176 - GPS speed" too and "210 - CAN wheel based speed" and so on. What would be the best here? Do we completely avoid setting KEY_OBD_SPEED? This is just an example, there are several such KEYs, already implemented KEY_RPM among them. 

My thinking is that KEY_OBD_SPEED stays mapped like I mapped it here, because it makes logical sense with KEY_OBD_ODOMETER and we leave all other IO Attributes that don't map "cleanly" with the default ioNUMBER mapping?

Also another question would be, the Teltonika decoder is decoding a lot of "custom" KEYs, do we want that with this protocol too? To me, it looks cleaner with the ioNUMBER mapping for the ones that don't cleanly map into existing Position.KEY_* ?

I will be going thorough other Position.KEY_*s shortly and push additional commits if I find clean mappings/get feedback on the above questions.

### DIFF
--- a/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
@@ -108,6 +108,9 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
             case 30:
                 position.set(Position.KEY_BATTERY, readValue(buf, length, false) * 0.001);
                 break;
+            case 65:
+                position.set(Position.KEY_ODOMETER, readValue(buf, length, true));
+                break;
             case 74:
                 position.set(Position.PREFIX_TEMP + 3, readValue(buf, length, true) * 0.1);
                 break;
@@ -115,6 +118,9 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
             case 79:
             case 80:
                 position.set(Position.PREFIX_TEMP + (id - 78), readValue(buf, length, true) * 0.1);
+                break;
+            case 95:
+                position.set(Position.KEY_OBD_SPEED, UnitsConverter.knotsFromKph(readValue(buf, length, true)));
                 break;
             case 134:
                 if (readValue(buf, length, false) > 0) {
@@ -128,6 +134,9 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
                 break;
             case 197:
                 position.set(Position.KEY_RPM, readValue(buf, length, false) * 0.125);
+                break;
+            case 645:
+                position.set(Position.KEY_OBD_ODOMETER, readValue(buf, length, true) * 1000);
                 break;
             default:
                 position.set(Position.PREFIX_IO + id, readValue(buf, length, false));
--- a/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
@@ -108,6 +108,9 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
             case 30:
                 position.set(Position.KEY_BATTERY, readValue(buf, length, false) * 0.001);
                 break;
+            case 32:
+                position.set(Position.KEY_DEVICE_TEMP, readValue(buf, length, true));
+                break;
             case 65:
                 position.set(Position.KEY_ODOMETER, readValue(buf, length, true));
                 break;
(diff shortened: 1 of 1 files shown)


jinzo:
Oh I also forgot a question, some IOs have Error values - what's the expected/desired behavior with that? do we just pass it along? do we not set the KEY? do we pass null? Did not find any instances of how this is handled elsewhere/in other decoders.

tananaev:
Are you sure absolutely all Ruptela devices use the same IOs?

jinzo:
Yes, all Ruptela devices use the same IOs - there is some special firmware
IOs, but they don't overlap/the IDs are not used in this 'normal' IO List.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -116,6 +122,9 @@ private void decodeParameter(Position position, int id, ByteBuf buf, int length)
             case 80:
                 position.set(Position.PREFIX_TEMP + (id - 78), readValue(buf, length, true) * 0.1);
                 break;
+            case 95:
+                position.set(Position.KEY_OBD_SPEED, UnitsConverter.knotsFromKph(readValue(buf, length, true)));
```
tananaev, line 126: OBD speed we actually store in km/h usually, from what I remember.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -126,9 +135,25 @@ private void decodeParameter(Position position, int id, ByteBuf buf, int length)
                     position.set(Position.KEY_ALARM, Position.ALARM_ACCELERATION);
                 }
                 break;
+            case 170:
+                position.set(Position.KEY_CHARGE, readValue(buf, length, false) > 0);
```
tananaev, line 139: This doesn't look right. This should we a charging current. You should always store original values, not do any calculations like this.

tananaev (commented):
If there are similar values like with speed and odometer, you can usually reuse the same keys. In practice people usually use one or the other. Not everything at the same time.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
jinzo: I went with this comment:  https://github.com/traccar/traccar/blob/7f2ded68ef015684f90bef59c055a1c980688069/src/main/java/org/traccar/model/Position.java#L98

Also this decoder does similar conversion:
https://github.com/traccar/traccar/blob/7f2ded68ef015684f90bef59c055a1c980688069/src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java#L662

On the other hand, TotemProtocolDecoder is storing raw value, which is km/h per my documentation
https://github.com/traccar/traccar/blob/7f2ded68ef015684f90bef59c055a1c980688069/src/main/java/org/traccar/protocol/TotemProtocolDecoder.java#L549

I would prefer km/h storing too - but then we need to update the comment in Position.java and fix at least the Gl200TextProtocolDecoder. Should I do it?

tananaev: hmmm.. I guess most of the places use km/h, so maybe update those inconsistent places.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
jinzo: Ok, I would prefer that too, but some of the KEY_CHARGE decoders do it similarly. Generally speaking KEY_CHARGE is boolean. So either I omit that completely (there is no specific Ruptela IO for charging) or this attribute won't be boolean in this instance (I'm not entirely sure if that is even possible). 
https://github.com/traccar/traccar/blob/7f2ded68ef015684f90bef59c055a1c980688069/src/main/java/org/traccar/protocol/ItsProtocolDecoder.java#L228
https://github.com/traccar/traccar/blob/7f2ded68ef015684f90bef59c055a1c980688069/src/main/java/org/traccar/protocol/JidoProtocolDecoder.java#L118
https://github.com/traccar/traccar/blob/7f2ded68ef015684f90bef59c055a1c980688069/src/main/java/org/traccar/protocol/Gps103ProtocolDecoder.java#L343

tananaev:
OK, I merged as is for now. It would nice to unify the data though.

jinzo:
Thank you, I will keep it in mind and probably put some effort into this data unification down the line.

PR #5114, opened 2023-06-11
Add increment value attribute on Duplicate Device Name for Route Report.

LiGuru:
[fix] Excel limitation workaround for duplicate Sheets names
[Forum discussion](https://www.traccar.org/forums/topic/report-route-if-there-are-duplicate-device-names/)
[Expected result](https://github.com/traccar/traccar/files/11714435/report-110623135938.xlsx)

Declaring uniqueSheetsTitle
Add private method getUniqueSheetName
Usage in getExcel sheetNames

### DIFF
--- a/src/main/java/org/traccar/reports/RouteReportProvider.java
+++ b/src/main/java/org/traccar/reports/RouteReportProvider.java
@@ -42,6 +42,9 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
+import java.util.Map;
+import java.util.HashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 public class RouteReportProvider {
 
@@ -66,7 +69,13 @@ public class RouteReportProvider {
         }
         return result;
     }
+    private final Map<String, Integer> uniqueSheetsTitle = new HashMap<>();
 
+    private String getUniqueSheetName(String key) {
+        AtomicInteger atomic = new AtomicInteger(uniqueSheetsTitle.getOrDefault(key, -1));
+        uniqueSheetsTitle.put(key, atomic.incrementAndGet());
+        return (uniqueSheetsTitle.get(key) > 0 ? key + "-" + uniqueSheetsTitle.get(key) : key);
+    }
     public void getExcel(OutputStream outputStream,
             long userId, Collection<Long> deviceIds, Collection<Long> groupIds,
             Date from, Date to) throws StorageException, IOException {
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/reports/RouteReportProvider.java
```diff
@@ -66,7 +69,13 @@ public Collection<Position> getObjects(long userId, Collection<Long> deviceIds,
         }
         return result;
     }
+    private final Map<String, Integer> uniqueSheetsTitle = new HashMap<>();
 
+    private String getUniqueSheetName(String key) {
+        AtomicInteger atomic = new AtomicInteger(uniqueSheetsTitle.getOrDefault(key, -1));
+        uniqueSheetsTitle.put(key, atomic.incrementAndGet());
+        return (uniqueSheetsTitle.get(key) > 0 ? key + "-" + uniqueSheetsTitle.get(key) : key);
```
tananaev, line 77: Why do you need `AtomicInteger`? Also I think the numbers should start with 1 and not -1. So basically if there's 1 device, we don't show. If more than 1, we do "Name", "Name-2", "Name-3" etc.

src/main/java/org/traccar/reports/RouteReportProvider.java
```diff
@@ -66,7 +69,13 @@ public Collection<Position> getObjects(long userId, Collection<Long> deviceIds,
         }
         return result;
     }
+    private final Map<String, Integer> uniqueSheetsTitle = new HashMap<>();
```
tananaev, line 72: Please move to the top next to other variables. And also name doesn't really make sense. It just contains the number of times a name is included. Maybe just call `namesCount` or something like that.

LiGuru:
I use AtomicInteger because of NullPointerException and adding an element with default value. 

Less code I guess.

If you suggest another implementation I will consider it.

tananaev:
Why would there be a `NullPointerException`?

As for default, you should use `computeIfAbsent`.

LiGuru:
What about putIfAbsent instead of computeIfAbsent
Is a lambda/function needed here?

`putIfAbsent(key, 0);`
`computeIfPresent(key, (k, v) -> v + 1);`

tananaev:
Actually maybe you just use compute like this:
```
map.compute(key, (k, value) -> (value == null) ? 1 : (value + 1))
```

LiGuru:
I'm thinking should it be implemented in all reports where Excel Sheets are needed?

tananaev:
Probably in all reports that use tabs.

PR #5115, opened 2023-06-12
Ruptela IO Decoding improvements

jinzo:
Hi,

another set of Ruptela IO improvements that builds upon https://github.com/traccar/traccar/pull/5113.

Main motivation was that I missed the IGNITION decoding. Previous implementation was rather simplistic, as different Ruptela devices have different "designated" ignition inputs now (it might be that in the past all of them were at DIN4) and the possibility of defining custom ignition logic (eg. which input is actual ignition & other possibilities).  I'm always vary of changing stuff that could break ignition decoding for some setups (eg. where the device does not have custom ignition defined and only relies on DIN4 being the ignition) - but on the other hand, without this change it absolutely breaks (at least some) properly configured Ruptela devices. Anyone affected could still use computed attributes for IGNITION, and everyone else can use DIN4 for other uses too.

Also added some other decoding alarms/ios.

### DIFF
--- a/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
@@ -177,7 +177,7 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
                 }
                 break;
             case 415:
-                if (readValue(buf, length, false) != 1) {
+                if (readValue(buf, length, false) == 0) {
                     position.set(Position.KEY_ALARM, Position.ALARM_GPS_ANTENNA_CUT);
                 }
                 break;
--- a/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
@@ -157,14 +157,14 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
                 position.set(Position.KEY_CHARGE, readValue(buf, length, false) > 0);
                 break;
             case 173:
-                position.set(Position.KEY_MOTION, readValue(buf, length, false) == 1);
+                position.set(Position.KEY_MOTION, readValue(buf, length, false) > 0);
                 break;
             case 197:
                 position.set(Position.KEY_RPM, readValue(buf, length, false) * 0.125);
                 break;
             case 251:
             case 409:
-                position.set(Position.KEY_IGNITION, readValue(buf, length, false) == 1);
+                position.set(Position.KEY_IGNITION, readValue(buf, length, false) > 0);
                 break;
             case 410:
                 if (readValue(buf, length, false) > 0) {
--- a/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
@@ -132,6 +132,11 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
             case 80:
                 position.set(Position.PREFIX_TEMP + (id - 78), readValue(buf, length, true) * 0.1);
                 break;
+            case 88:
+                if (readValue(buf, length, false) == 1) {
+                    position.set(Position.KEY_ALARM, Position.ALARM_JAMMING);
+                }
+                break;
             case 95:
                 position.set(Position.KEY_OBD_SPEED, UnitsConverter.knotsFromKph(readValue(buf, length, true)));
                 break;
@@ -144,6 +149,9 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
                 if (readValue(buf, length, false) > 0) {
                     position.set(Position.KEY_ALARM, Position.ALARM_ACCELERATION);
                 }
+                break;
+            case 150:
+                position.set(Position.KEY_OPERATOR, readValue(buf, length, false));
                 break;
             case 170:
                 position.set(Position.KEY_CHARGE, readValue(buf, length, false) > 0);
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -97,10 +97,20 @@ private void decodeParameter(Position position, int id, ByteBuf buf, int length)
             case 2:
             case 3:
             case 4:
+            case 5:
                 position.set("di" + (id - 1), readValue(buf, length, false));
                 break;
-            case 5:
-                position.set(Position.KEY_IGNITION, readValue(buf, length, false) == 1);
+            case 20:
+                position.set("ai3", readValue(buf, length, false));
```
tananaev, line 104: What is AI?

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -135,12 +150,22 @@ private void decodeParameter(Position position, int id, ByteBuf buf, int length)
                     position.set(Position.KEY_ALARM, Position.ALARM_ACCELERATION);
                 }
                 break;
+            case 150:
+                position.set(Position.KEY_OPERATOR, readValue(buf, length, false));
```
tananaev, line 154: What is the value here?

tananaev (commented):
Also please don't change `traccar-web`.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
jinzo: I was going with the "di" as a digital input and "ai" as an analog input. I guess this touches on the next PREFIX_IN comment to. How do we handle device that has DIN1 and AIN1 as a digital input 1 and analog input 1?

jinzo:
> Also please don't change `traccar-web`.

Oh, sorry my bad, was unintentional.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
jinzo: Unique Mobile operator code - MCC and MNC on my devices

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
tananaev: "IN" is only for digital inputs. For analog we have "ADC", which is the standard.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
tananaev: I guess it's ok for now, but ideally we only store the name here. For MCC and MNC we have Network object.

jinzo:
This should resolve all the issues. Thanks again for your guidance and all the help!

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -151,9 +176,19 @@ private void decodeParameter(Position position, int id, ByteBuf buf, int length)
...
+            case 415:
+                if (readValue(buf, length, false) != 1) {
+                    position.set(Position.KEY_ALARM, Position.ALARM_GPS_ANTENNA_CUT);
+                }
+                break;
             case 645:
                 position.set(Position.KEY_OBD_ODOMETER, readValue(buf, length, true) * 1000);
                 break;
+            case 758:
+                if (readValue(buf, length, false) == 1) {
```
tananaev, line 188: Can this also be `> 0`?

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -151,9 +176,19 @@ private void decodeParameter(Position position, int id, ByteBuf buf, int length)
                     position.set(Position.KEY_ALARM, Position.ALARM_ACCIDENT);
                 }
                 break;
+            case 415:
+                if (readValue(buf, length, false) != 1) {
```
tananaev, line 180: What about this one?

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -122,6 +132,11 @@ private void decodeParameter(Position position, int id, ByteBuf buf, int length)
             case 80:
                 position.set(Position.PREFIX_TEMP + (id - 78), readValue(buf, length, true) * 0.1);
                 break;
+            case 88:
+                if (readValue(buf, length, false) == 1) {
```
tananaev, line 136: And this?

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
jinzo: This one not, 0 -> no tampering, 1 -> tampering activated and 255 - no data from tampering switch, which is present in my tests when there is no tampering

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
tananaev: Why would unknown be an alarm?

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
tananaev: I see. Sounds good.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
jinzo: Hm, I was thinking to err on the side of caution (eg. everything that is not connected, is setting an alarm), but don't have a better reason than this. But don't have a strong opinion on this, so we can change it no problems.

tananaev: Let's only use 0 as antenna cut. Im actually not entirely sure this is the right thing to use here. Some devices report it as a separate alarm, but in this case we're kind of using the state.

jinzo: Ok, changed it and will revisit this later, when I take a look at the events and what exactly does Ruptela send when it detects that the GPS antenna is disconnected.

Issue #5116, opened 2023-06-12
Computed Attributes, JEXL and advanced usage 


jinzo:
While trying to do a little bit more advanced computed attributes I encountered some difficulties and I'm opening this bug report to track what I found out and hopefully provide a pull request. And maybe @memesaregood1 can provide some insight, as he did some changes to this lately. 

Namely - first, what I'm sure is a bug/needs documentation update is the mention that new() can be used: 
```
new("java.lang.Double", 10)
```
This needs an additional note that it needs to be enabled before with (same as the note for localVariables)
```
<entry key='processing.computedAttributes.newInstanceCreation'>true</entry>
```
And from what I can see, this computedAttributes settings are not mentioned in the auto generated config file documentation?

But even after enabling this, it does not work. I suspect it's because of the sandboxing - but need to investigate further
**Test**
```
var foo = new("java.lang.Double", 10)
return "baz" + foo
```
**Error**
```
org.traccar.handler.ComputedAttributesHandler.computeAttribute:119@1:1 unsolvable function/method 'java.lang.Double(Byte)' - Method (ComputedAttributesHandler:120 < AttributeResource:63 < ... < OverrideFilter:50 < ...)
```
As I'm a little bit pressed for time, I'm not sure if I will tackle this immediately or solve my problem in a different way for now and then investigate this further. From a quick look JEXL config/sandboxing is not so simple to understand, at least for me.

memesaregood1:
You are right.
I [threw together](https://gist.github.com/memesaregood1/18d24f79e8a3c8b5f47be312e9ddb6ed) some code to play around with the engine, and, as you suspected, Sandbox is blocking the `new()` calls for denied types, if such are not allowed. Strangely enough, when I was testing around when I made changes to Traccar, things worked fine.

jinzo:
Thanks for your effort - so I would suggest that the basic types that could be used for manipulating the attributes are added:
```
        sandbox.allow(Double.class.getName()); //types have to be allowed too!
        sandbox.allow(Float.class.getName());
        sandbox.allow(Integer.class.getName());
        sandbox.allow(Long.class.getName());
        sandbox.allow(Short.class.getName());
        sandbox.allow(Character.class.getName());
        sandbox.allow(Boolean.class.getName());
        sandbox.allow(String.class.getName());
```
And documentation update with this info should be enough for now? And then later down the line this can be extended/made configurable if the need arises?

@memesaregood1 @tananaev what do you guys think about this suggestion?

[...]

PR #5122, opened 2023-06-15
Allow primitive Java types in JEXL Engine

memesaregood1:
Following #5116, this PR allows primitive Java types to be used in JEXL scripts. This makes it possible to initialize variables with these types using the `new()` operator.

### DIFF
--- a/src/main/java/org/traccar/handler/ComputedAttributesHandler.java
+++ b/src/main/java/org/traccar/handler/ComputedAttributesHandler.java
@@ -66,9 +66,9 @@ public class ComputedAttributesHandler extends BaseDataHandler {
         JexlSandbox sandbox = new JexlSandbox(false);
         sandbox.allow("com.safe.Functions");
         sandbox.allow(Math.class.getName());
-        List.of(Double.class, Float.class, Integer.class,
-                        Long.class, Short.class, Character.class,
-                        Boolean.class, String.class, Byte.class)
+        List.of(
+            Double.class, Float.class, Integer.class, Long.class, Short.class,
+            Character.class, Boolean.class, String.class, Byte.class)
                 .forEach((type) -> sandbox.allow(type.getName()));
         features = new JexlFeatures()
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/handler/ComputedAttributesHandler.java
```diff
@@ -65,6 +65,14 @@ public ComputedAttributesHandler(Config config, CacheManager cacheManager) {
...
         sandbox.allow("com.safe.Functions");
         sandbox.allow(Math.class.getName());
+        sandbox.allow(Double.class.getName());
+        sandbox.allow(Float.class.getName());
+        sandbox.allow(Integer.class.getName());
+        sandbox.allow(Long.class.getName());
+        sandbox.allow(Short.class.getName());
+        sandbox.allow(Character.class.getName());
+        sandbox.allow(Boolean.class.getName());
+        sandbox.allow(String.class.getName());
```
tananaev, line 75: Can we do something like:
```
List.of(Double.class, Float.class, ...).forEach((type) -> sandbox.allow(type.getName()));
```

memesaregood1: Certainly.

src/main/java/org/traccar/handler/ComputedAttributesHandler.java
```diff
@@ -65,6 +66,10 @@ public ComputedAttributesHandler(Config config, CacheManager cacheManager) {
         JexlSandbox sandbox = new JexlSandbox(false);
         sandbox.allow("com.safe.Functions");
         sandbox.allow(Math.class.getName());
+        List.of(Double.class, Float.class, Integer.class,
+                        Long.class, Short.class, Character.class,
+                        Boolean.class, String.class, Byte.class)
+                .forEach((type) -> sandbox.allow(type.getName()));
```
tananaev, line 72:
```suggestion
        List.of(
            Double.class, Float.class, Integer.class, Long.class, Short.class,
            Character.class, Boolean.class, String.class, Byte.class)
                .forEach((type) -> sandbox.allow(type.getName()));
```

tananaev:
Merged, thanks.

PR #5124, opened 2023-06-20
Ruptela Decoder - implement Extended protocol extended messages merging

jinzo:
Hi,

since some time I have noticed strange behavior on some Ruptela devices, where seemingly IOs were not sent to the server. After some digging I found out that in fact - device was sending two records, and I was just seeing the last one. After some digging in the documentation I found out that this is called Extended protocol extended messages merging, as Ruptela devices are sending more records with the same header if it cannot fit all the IOs in a single one. 

Documentation: [Ruptela Protocol 1.67](https://pdfcoffee.com/ruptela-protocol-v167pdf-4-pdf-free.html) page 15, heading "2.4 Extended protocol extended records".

I have been running this in limited testing on my staging setup and it looks good. I added the (hopefully) redundant time check, just to be more sure we are not merging wrong records. Also the line is broken because of checkstyle limit and I'm not entirely sure how else should I break it to be 'in style'. I'll extend the testing to more devices tomorrow and report back if there are any unforeseen consequences, but I think it's coded so that it fails gracefully if something does not check out.

### DIFF
--- a/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
@@ -236,7 +236,18 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
                 buf.readUnsignedByte(); // timestamp extension
 
                 if (type == MSG_EXTENDED_RECORDS) {
-                    buf.readUnsignedByte(); // record extension
+                    int b = buf.readUnsignedByte(); // record extension
+                    int noRecordsToMerge = (b & 0xf0) >> 4;
+                    int currentRecord = b & 0x0f;
+
+                    if (currentRecord > 0 && noRecordsToMerge >= currentRecord) {
+                        if (!positions.isEmpty()
+                                && positions.get(positions.size() - 1).getDeviceTime()
+                                   .compareTo(position.getDeviceTime()) == 0) {
+                            position = positions.remove(positions.size() - 1);
+                        }
+                    }
+
                 }
 
                 buf.readUnsignedByte(); // priority (reserved)
--- a/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
@@ -238,7 +238,7 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
 
                 if (type == MSG_EXTENDED_RECORDS) {
                     int recordExtension = buf.readUnsignedByte();
-                    int mergeRecordCount = BitUtil.between(recordExtension, 4, 8);
+                    int mergeRecordCount = BitUtil.from(recordExtension, 4);
                     int currentRecord = BitUtil.to(recordExtension, 4);
 
                     if (currentRecord > 0 && currentRecord <= mergeRecordCount) {
--- a/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
@@ -124,7 +124,7 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
                 position.set(Position.PREFIX_ADC + 2, readValue(buf, length, false));
                 break;
             case 29:
-                position.set(Position.KEY_POWER, readValue(buf, length, false));
+                position.set(Position.KEY_POWER, readValue(buf, length, false) * 0.001);
                 break;
             case 30:
                 position.set(Position.KEY_BATTERY, readValue(buf, length, false) * 0.001);
--- a/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
@@ -22,6 +22,7 @@ import org.traccar.BaseProtocolDecoder;
 import org.traccar.session.DeviceSession;
 import org.traccar.NetworkMessage;
 import org.traccar.Protocol;
+import org.traccar.helper.BitUtil;
 import org.traccar.helper.DataConverter;
 import org.traccar.helper.UnitsConverter;
 import org.traccar.model.Position;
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -236,7 +236,18 @@ protected Object decode(
                 buf.readUnsignedByte(); // timestamp extension
 
                 if (type == MSG_EXTENDED_RECORDS) {
-                    buf.readUnsignedByte(); // record extension
+                    int b = buf.readUnsignedByte(); // record extension
```
tananaev, line 239: Needs better name

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -236,7 +236,18 @@ protected Object decode(
...
+                    int noRecordsToMerge = (b & 0xf0) >> 4;
+                    int currentRecord = b & 0x0f;
```
tananaev, line 241: Use BitUtil

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -236,7 +236,18 @@ protected Object decode(
...
+
+                    if (currentRecord > 0 && noRecordsToMerge >= currentRecord) {
+                        if (!positions.isEmpty()
+                                && positions.get(positions.size() - 1).getDeviceTime()
+                                   .compareTo(position.getDeviceTime()) == 0) {
```
tananaev, line 246: Why do we need this comparison?

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -236,7 +236,18 @@ protected Object decode(
...
+                            position = positions.remove(positions.size() - 1);
```
tananaev, line 247: We should probably merge the attributes, right? Otherwise you're actually losing the data.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
jinzo: It's an additional safety check, that we are truly merging two Positions with the same time (as all such extended Positions have the same header, so time, lat/long,...

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
jinzo: We are not loosing data, as this just checks if it needs to get the latest Position (instead of creating/using the newly created one) and then gets it. Down the line from this, this "old/previous" position, with all the attributes it already has, just gets some new attributes added to it (and then it gets re-added to the positions List). See attached screenshot for example. Previously this IOs were split in two Positions in this instance:
image

jinzo: P.S: I am aware that I need to fix the Power decoding so that we have V instead of mV stored, it's coming with some other PR soon-ish :)

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
tananaev: I would remove any unnecessary checks.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -236,7 +237,17 @@ protected Object decode(
                 buf.readUnsignedByte(); // timestamp extension
 
                 if (type == MSG_EXTENDED_RECORDS) {
-                    buf.readUnsignedByte(); // record extension
+                    int recordExtension = buf.readUnsignedByte(); // record extension
```
tananaev, line 240:
```suggestion
                    int recordExtension = buf.readUnsignedByte();
```

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -236,7 +237,17 @@ protected Object decode(
...
+                    int noRecordsToMerge = BitUtil.between(recordExtension, 4, 8);
```
tananaev, line 241: I would rename to something like `mergeRecordCount`.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -236,7 +237,17 @@ protected Object decode(
...
+                    int currentRecord = BitUtil.to(recordExtension, 4);
+
+                    if (currentRecord > 0 && noRecordsToMerge >= currentRecord) {
```
tananaev, line 244: Reads better this way

```suggestion
                    if (currentRecord > 0 && currentRecord <= noRecordsToMerge) {
```

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -236,7 +237,17 @@ protected Object decode(
...
+                        if (!positions.isEmpty()
```
tananaev, line 245: Would there be any case where the list is empty?

jinzo: If everything goes as described in the documentation then no - this falls under "remove any unnecessary checks" and I'll remove this whole check.

jinzo:
Applied suggested changes, tested it on live devices and it looks good. I also 'sneaked' in the Power scaling for mV to V and can confirm that behaves now as expected too.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -236,7 +237,13 @@ protected Object decode(
                 buf.readUnsignedByte(); // timestamp extension
 
                 if (type == MSG_EXTENDED_RECORDS) {
-                    buf.readUnsignedByte(); // record extension
+                    int recordExtension = buf.readUnsignedByte();
+                    int mergeRecordCount = BitUtil.between(recordExtension, 4, 8);
```
tananaev, line 241: One last minor thing
```suggestion
                    int mergeRecordCount = BitUtil.from(recordExtension, 4);
```

tananaev:
Merged, thank you.

PR #5134, opened 2023-07-13
Megastel Protocol - Exstended with WiFi data and Protocol Encoder

hellenthal:
Improvements on MegastekProtocolDecoder decodeNew;
- Added WifiAccessPoint (for LBS Geolocation).
- BatteryLevel (percentage) was wrongly stored as battery what is battery voltage.
- Satellite counts now all satellites (BDS, GPS and Glonass).
- Belt status set to KEY_LOCK
- A 'restart' was not covered, this will now trigger a ALARM_POWER_ON.
- A 'detect water'. was not covered, this will now trigger a ALARM_FUEL_LEAK (most similar of all Alarm options)
- Command Response messages (for new and old protocol) including for new protocol sending current timestamp back to the device on request (Q051;).

Added;
- MegastekProtocolEncoder, for new and old protocol. New protocol has more SupportedDataCommands. With Custom Commands only need to enter the actual command and not the full message including device id. This part is only tested with  the new protocol.

--
The old and the already existing new Megastek protocol in Traccar could perhaps better be divided up in two protocols? Megastek and Megastek2 like with Xexun and Xexun2 protocol?

### DIFF
--- a/src/main/java/org/traccar/protocol/MegastekProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/MegastekProtocolDecoder.java
@@ -18,16 +18,22 @@ package org.traccar.protocol;
 import io.netty.channel.Channel;
 import org.traccar.BaseProtocolDecoder;
 import org.traccar.session.DeviceSession;
+import org.traccar.NetworkMessage;
 import org.traccar.Protocol;
 import org.traccar.helper.DateBuilder;
 import org.traccar.helper.Parser;
 import org.traccar.helper.PatternBuilder;
 import org.traccar.model.CellTower;
+import org.traccar.model.WifiAccessPoint;
 import org.traccar.model.Network;
 import org.traccar.model.Position;
 
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
 import java.net.SocketAddress;
 import java.util.regex.Pattern;
+import java.util.Date;
+import java.util.TimeZone;
 
 public class MegastekProtocolDecoder extends BaseProtocolDecoder {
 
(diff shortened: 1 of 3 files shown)


hellenthal:
[Megastel Protocol v2.pdf](https://github.com/traccar/traccar/files/12040207/Megastel.Protocol.v2.pdf)
[Megastek Command List v2.pdf](https://github.com/traccar/traccar/files/12040210/Megastek.Command.List.v2.pdf)

Here are the Megastek Protocol and Command List documentation.

hellenthal:
Can I ask what you like to be cleaned up of the formatting?
Except the Satellites part is everything taken over from other protocol class files.

src/main/java/org/traccar/protocol/MegastekProtocolDecoder.java
```diff
@@ -231,21 +237,51 @@ private Position decodeOld(Channel channel, SocketAddress remoteAddress, String
         return position;
     }
 
+    private static final Pattern PATTERN_COMMAND_RESULT_OLD = new PatternBuilder()
+            .text("$")                           // start
+            .number("(d+),")                     // device id
+            .expression("(.*);")				 // command
+            .number("xx")                     	 // checksum
+            .any()               				 //
```
tananaev, line 245: What is this comment?

src/main/java/org/traccar/protocol/MegastekProtocolDecoder.java
```diff
@@ -231,21 +237,51 @@ private Position decodeOld(Channel channel, SocketAddress remoteAddress, String
...
+
     private static final Pattern PATTERN_NEW = new PatternBuilder()
             .number("dddd").optional()
             .text("$MGV")
             .number("ddd,")
             .number("(d+),")                     // imei
-            .expression("[^,]*,")                // name
-            .expression("([RS]),")
+            .expression("[^,]*,")                // device name
+            .expression("([RS]),")				 // R = Realtime data, S = Stored data
```
tananaev, line 276: Same here. You shouldn't use tabs anywhere in the code.

tananaev (commented):
Just a couple of examples. You should run the check before pushing the code though. It will catch a log of these things.

PR #5135, opened 2023-07-13
PositionForwarderUrl Added batteryLevel and charge.

hellenthal:
Data as batteryLevel and charge are also important information to forward.

Those are also supported by the OsmAnd protocol (batt and charge).


tananaev:
I don't think hardcoding all the possible attributes is the way to go. If you really want to implement separate attributes, I think the right way is to refactor the formatting to check for attributes based on the template key.

hellenthal:
I agree with that not all possible attributes are the way to go.

I use the forwarding to send tracker data from one traccar server to another traccar server (my home server to a friends home server) by the OsmAnd protocol.

And battery level and charge are also one of the only attributes that are left that is supported by the OsmAnd protocol and not in the PositionForwarderUrl.

After battery level and charge is think only hdop not supported in the PositionForwarderUrl.

tananaev:
So do you plan to refactor with my suggestion?

Issue #5138, opened 2023-07-19
Web app does not keep user logged in


jbschooley:
**Describe the bug**
In the old web app there was a remember user checkbox that would keep the user logged in. That checkbox is now gone. The cookie does not have a Max-Age, so the browser (or mobile app) clears it relatively quickly.

**To Reproduce**
1. Sign in
2. Close the site or app
3. Wait a few hours
4. User will be shown the login screen

**Expected behavior**
The user should stay logged in. This would be accomplished by setting the session cookie's Max-Age to the value of web.sessionTimeout. A checkbox for the user to choose whether to stay logged in would be helpful, but having it as default would be sufficient.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Firefox
 - Version 114.0.2

**Smartphone (please complete the following information):**
 - Device: iPhone 13
 - OS: iOS 16.5
 - Browser: Safari; Traccar Manager app

**Additional context**

PR #5139, opened 2023-07-19
Set session cookie max age if web.sessionTimeout is set

jbschooley:
Fixes https://github.com/traccar/traccar/issues/5138


tananaev:
Merged, thank you.

PR #5141, opened 2023-07-19
added engineLoad + AdblueLevel

eisenkoch:
https://wiki.teltonika-gps.com/view/FMB204_Teltonika_Data_Sending_Parameters_ID


src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -245,7 +245,9 @@ private static void register(int id, Set<String> models, BiConsumer<Position, By
         register(89, fmbXXX, (p, b) -> p.set("fuelLevelPercentage", b.readUnsignedByte()));
         register(90, null, (p, b) -> p.set(Position.KEY_DOOR, b.readUnsignedShort()));
         register(110, fmbXXX, (p, b) -> p.set(Position.KEY_FUEL_CONSUMPTION, b.readUnsignedShort() * 0.1));
+        register(111, fmbXXX, (p, b) -> p.set("AdblueLevel", b.readUnsignedByte()));
         register(113, fmbXXX, (p, b) -> p.set(Position.KEY_BATTERY_LEVEL, b.readUnsignedByte()));
+        register(114, fmbXXX, (p, b) -> p.set("engineLoad", b.readUnsignedByte()));
```
tananaev, line 250: Don't we already have a standard parameter for engine load?

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -245,7 +245,9 @@ private static void register(int id, Set<String> models, BiConsumer<Position, By
         register(89, fmbXXX, (p, b) -> p.set("fuelLevelPercentage", b.readUnsignedByte()));
         register(90, null, (p, b) -> p.set(Position.KEY_DOOR, b.readUnsignedShort()));
         register(110, fmbXXX, (p, b) -> p.set(Position.KEY_FUEL_CONSUMPTION, b.readUnsignedShort() * 0.1));
+        register(111, fmbXXX, (p, b) -> p.set("AdblueLevel", b.readUnsignedByte()));
```
tananaev, line 248: The key should be lower camel case, so something like `adBlueLevel` or `adblueLevel`.

tananaev: Also I noticed that the length is not correct. It's not a byte, according to the documentation. Make sure you include a unit test.

PR #5146, opened 2023-07-26
Watch Protocol Decoder update

KhalilHadded:
- handle blood oxygen data

### DIFF
--- a/src/main/java/org/traccar/model/Position.java
+++ b/src/main/java/org/traccar/model/Position.java
@@ -44,6 +44,9 @@ public class Position extends Message {
     public static final String KEY_ODOMETER_TRIP = "tripOdometer"; // meters
     public static final String KEY_HOURS = "hours"; // milliseconds
     public static final String KEY_STEPS = "steps";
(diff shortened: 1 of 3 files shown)


src/main/java/org/traccar/model/Position.java
```diff
@@ -44,6 +44,9 @@ public class Position extends Message {
     public static final String KEY_ODOMETER_TRIP = "tripOdometer"; // meters
     public static final String KEY_HOURS = "hours"; // milliseconds
     public static final String KEY_STEPS = "steps";
+    public static final String KEY_BLOOD_OXYGEN = "bloodOxygen";
```
tananaev, line 47: We only add new parameters here where they are used by at least 2 protocols. You would also have to update the web app with those new attributes.

KhalilHadded: I think the web app can handle this update automatically.

image

image

Do i have to change something else in the web app ?

tananaev: No, it cannot. Known attributes have to be proper translatable attributes.

Issue #1154, opened 2015-04-02
New model of TK 110


joeinaustin:
I have a newly acquired tracker that says it is a TK 110.  That setting is not working unless I am doing something else wrong.  I get no updates after adding the device.  Can someone look at my log and tell me if I am sending on the right port?
```
2015-04-02 12:51:57  WARN: An existing connection was forcibly closed by the remote host - java.io.IOException (null:-2)
2015-04-02 12:51:57  INFO: Closing connection by exception
2015-04-02 12:51:57  INFO: Closing connection by disconnect
2015-04-02 12:52:00 DEBUG: [5013 <- 172.56.16.6] - HEX: 78780d0103554880201151530003c3830d0a
2015-04-02 12:52:16 DEBUG: [5013 <- 172.56.16.6] - HEX: 78780d0103554880201151530003c3830d0a
2015-04-02 12:52:17 DEBUG: [5013 <- 172.56.16.6] - HEX: 78781f120d0b0b000000c0000000000000000000200001360426f300370b00033de30d0a
2015-04-02 12:52:48 DEBUG: [5013 <- 172.56.16.6] - HEX: 78781f120d0b0b000000c0000000000000000000200001360426f300370b0004495c0d0a
2015-04-02 12:53:18 DEBUG: [5013 <- 172.56.16.6] - HEX: 78781f120d0b0b000000c0000000000000000000200001360426f300370b000558d50d0a
2015-04-02 12:53:48 DEBUG: [5013 <- 172.56.16.6] - HEX: 78781f120f04020b360bc5034273100a80aadd003d4601360426f300370b0006b39f0d0a
2015-04-02 12:54:18 DEBUG: [5013 <- 172.56.16.6] - HEX: 78781f120f04020b3629c6034272e20a80aae5003d4b01360426f300370b0007e6310d0a
2015-04-02 12:54:19 DEBUG: [5013 <- 172.56.16.6] - HEX: 78780a134600390001000848950d0a
2015-04-02 12:54:25 DEBUG: [5013 <- 172.56.16.6] - HEX: 78780d01035548802011515300096cd90d0a
2015-04-02 12:54:48 DEBUG: [5013 <- 172.56.16.6] - HEX: 78781f120f04020b370bc6034272f80a80aaf7003d4b01360426f300370b000924b50d0a
[...]

PR #5147, opened 2023-07-28
Watch Protocol Decoder update

KhalilHadded:
- handle blood oxygen data

web app updates [#1154](https://github.com/traccar/traccar-web/pull/1154)

### DIFF
--- a/src/main/java/org/traccar/model/Position.java
+++ b/src/main/java/org/traccar/model/Position.java
@@ -44,6 +44,9 @@ public class Position extends Message {
     public static final String KEY_ODOMETER_TRIP = "tripOdometer"; // meters
     public static final String KEY_HOURS = "hours"; // milliseconds
     public static final String KEY_STEPS = "steps";
+    public static final String KEY_BLOOD_OXYGEN = "bloodOxygen";
+    public static final String KEY_PRESSURE_HIGH = "pressureHigh";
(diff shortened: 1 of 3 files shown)


src/main/java/org/traccar/model/Position.java
```diff
@@ -44,6 +44,9 @@ public class Position extends Message {
     public static final String KEY_ODOMETER_TRIP = "tripOdometer"; // meters
     public static final String KEY_HOURS = "hours"; // milliseconds
     public static final String KEY_STEPS = "steps";
+    public static final String KEY_BLOOD_OXYGEN = "bloodOxygen";
+    public static final String KEY_PRESSURE_HIGH = "pressureHigh";
+    public static final String KEY_PRESSURE_LOW = "pressureLow";
```
tananaev, line 49: This is a comment from previous PR:

> We only add new parameters here where they are used by at least 2 protocols.

KhalilHadded: Okay. I will set those parameters as custom parameters. So i ll hardcode them as they was in the first commit.
And I ll close the pull request in the web app.
is this ll be ok?

tananaev: That should be ok and then please close the other PR.

PR #5148, opened 2023-07-28
Watch Protocol Decoder update

KhalilHadded:
- handle blood oxygen data


tananaev:
Merged, thanks.

PR #5153, opened 2023-08-07
fix: AMQP auto close connection

RodolfoSilva:
Currently, the AMQP connection is being closed after the completion of block execution.
I made a mistake in my last PR #5136.

### DIFF
--- a/src/main/java/org/traccar/forward/AmqpClient.java
+++ b/src/main/java/org/traccar/forward/AmqpClient.java
@@ -47,13 +47,6 @@ public class AmqpClient {
             Connection connection = factory.newConnection();
             channel = connection.createChannel();
             channel.exchangeDeclare(exchange, BuiltinExchangeType.TOPIC, true);
-            Runtime.getRuntime().addShutdownHook(new Thread(() -> {
-                try {
-                    connection.close();
-                } catch (IOException e) {
-                    throw new RuntimeException(e);
-                }
-            }));
         } catch (IOException | TimeoutException e) {
             throw new RuntimeException("Error while creating and configuring RabbitMQ channel", e);
         }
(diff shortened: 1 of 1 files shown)


tananaev:
So where do you close connection then?

tananaev:
But you're removing it from the try-with-resources statement. I'm asking about your new code.

RodolfoSilva:
Today, this will not be closed. Is there a place where I can add a shutdown method? I haven't seen this in the EventForwarder and PositionForwarder interfaces.

tananaev:
In other places we close connection after each message, like you had before. That's why I'm asking.

RodolfoSilva:
This is not how RabbitMQ works. 

Repeatedly opening and closing connections for each message can significantly reduce. It can also put unnecessary load on the RabbitMQ server, potentially affecting its performance or stability.

It's generally recommended to establish a connection when the application starts up, and keep it open for as long as it's needed. This allows you to send many messages (potentially millions) over the same connection, reducing the per-message overhead and thus increasing the overall efficiency and performance.

In addition, RabbitMQ's protocol provides for certain optimizations like publishing confirmations and delivery tags that rely on maintaining a stable connection.

Te RabbitMQ client provide an automatic connection recovery to handle temporary network failures.

However, I could use the `Runtime.addShutdownHook()` to close the connection when the app is stopping. What do you think?

tananaev:
I don't think we need `Runtime.addShutdownHook()`. The OS will close all the network connections if needed. I was more worried about failed connections and retried. Is it handed automatically by the library? If yes, what's the policy?

RodolfoSilva:
@tananaev The official documentation from library says that library will handle this connection recovery process automatically by default:

[RabbitMQ Java Client Documentation](https://www.rabbitmq.com/api-guide.html#recovery)
`ConnectionFactory` API documentation: [ConnectionFactory (RabbitMQ Java Client API Guide)](https://rabbitmq.github.io/rabbitmq-java-client/api/current/com/rabbitmq/client/ConnectionFactory.html)

image

Should I remove the `ShutdownHook`?

tananaev:
Yeah, let's remove the `ShutdownHook` then. Please create the connection in the constructor then.

RodolfoSilva:
Done.
The connection is already created inside the AmqpClient constructor.

tananaev:
Merged, thanks.

PR #5158, opened 2023-08-14
Add support for MQTT version 3.x

jokakilla:
I had issues connecting with traccar to my aedes broker that currently only supports MQTT 3.x. It doesn't support MQTT5 yet.
It seems like the hivemq client library's Mqtt5AsyncClient doesn't fallback to MQTT 3.x when the server doesn't support version 5 features. The hivemq dev have a completely separate package structure (redudant) for v3 and v5. The classes don't have common interfaces.

I've added a setting to tell traccar which version to use. If the setting is not there or different from v3 version 5 like before will be used by default.

Discussion was here: https://www.traccar.org/forums/topic/release-56-mqtt-support-for-event-forwarding/page/3/

In my setup I could finally connect to my MQTT broker with these changes.

### DIFF
--- a/src/main/java/org/traccar/MainModule.java
+++ b/src/main/java/org/traccar/MainModule.java
@@ -35,6 +35,7 @@ import org.traccar.config.Keys;
 import org.traccar.database.LdapProvider;
 import org.traccar.database.OpenIdProvider;
 import org.traccar.database.StatisticsManager;
+import org.traccar.forward.MqttClientFactory;
 import org.traccar.forward.EventForwarder;
 import org.traccar.forward.EventForwarderJson;
 import org.traccar.forward.EventForwarderAmqp;
@@ -367,7 +368,7 @@ public class MainModule extends AbstractModule {
                 case "kafka":
                     return new EventForwarderKafka(config, objectMapper);
                 case "mqtt":
-                    return new EventForwarderMqtt(config, objectMapper);
+                    return new EventForwarderMqtt(config, objectMapper, new MqttClientFactory());
                 case "json":
                 default:
                     return new EventForwarderJson(config, client);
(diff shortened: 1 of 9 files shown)


tananaev:
I don't like this duplication. Are they not backward or forward compatible?

jokakilla:
Me too. But I didn't see a more elegant solution if you want to Stick to hivemq. Maybe I just overlooked it. I wish hivemq would detect the version supported by the server or at least offer a version setting in the client builder.

I tried to minimize the Code redundancy by separating the EventForwarderMqtt from the client creation.

Maybe there is another mqtt library for java if this ist not acceptable.

tananaev:
Is 3 forward compatible with 5? Can MQTT 3 client send data to MQTT 5 server?

jokakilla:
Not an expert for MQTT. I stumbled over this: "Most likely important: MQTT v5.0 is not backward compatible (like v3.1.1). Obviously too many new things are introduced so existing implementations have to be revisited."

ChatGPT says a MQTT 3.1 client is able to connect to a MQTT 5 server. You'll only lose the latest features. Not Sure If this ist true. That would mean we could simply switch to the MQTT3 client.

Unfortunately I don't have a v5 Server around and have no experience setting one up.

jokakilla:
Maybe ClientV5 -> Server v3.1 doesn't work but the opposite direction does?

ivanfmartinez:
> Is 3 forward compatible with 5? Can MQTT 3 client send data to MQTT 5 server?

I have used v3 clients fine with v5 servers in other applications. I have used paho in many applications connecting to v3 and v5. Hivemq have different classes for interacting with v5 and v3 servers.

mqtt 5 have some new features, but as I have seen in current traccar code when I made a refactor to create a position forwarded traccar only use basic things.

I will change my code to use the mqtt3 client to test with my v5 server, and if it works fine in next days will push a new commit for the #5202

ramonsmits:
Its better to only support MQTT3 as there as no really useful additions like headers. As long as the MQTT broker accepts v3 messages its ok. All brokers will happily send to v3 and v5 subscribers.

PR #5159, opened 2023-08-15
Throw overspeed event immediately, accounting for minimumDuration

### LINKED ISSUE #5103: Overspeed event
Hi,

I noticed that since the attribute minimalDuration was added to over speeding events, it takes 2 frames to generate an event.
This causes that when the vehicle exceeds the speed for a moment, the event is not generated even if the attribute minimalDuration is 0.

I suggest adding a condition:
(position != null && position.getSpeed() > speedLimit && minimalDuration ==0) then generate event.

jguthrie100:
Fixes #5103 

As soon as a speed limit is broken, an overspeed event is thrown. Current functionality waits until the 2nd reading of the overspeed instance before throwing the event.

### DIFF
--- a/src/main/java/org/traccar/session/state/OverspeedProcessor.java
+++ b/src/main/java/org/traccar/session/state/OverspeedProcessor.java
@@ -34,7 +34,7 @@ public final class OverspeedProcessor {
         if (oldState) {
             boolean newState = position.getSpeed() > speedLimit;
             if (newState) {
-                setEvent(state, position, speedLimit, minimalDuration);
+                checkEvent(state, position, speedLimit, minimalDuration);
             } else {
                 state.setOverspeedState(false);
                 state.setOverspeedTime(null);
@@ -45,11 +45,11 @@ public final class OverspeedProcessor {
             state.setOverspeedTime(position.getFixTime());
             state.setOverspeedGeofenceId(geofenceId);
 
-            setEvent(state, position, speedLimit, minimalDuration);
+            checkEvent(state, position, speedLimit, minimalDuration);
         }
     }
 
-    private static void setEvent(OverspeedState state, Position position, double speedLimit, long minimalDuration) {
+    private static void checkEvent(OverspeedState state, Position position, double speedLimit, long minimalDuration) {
         if (state.getOverspeedTime() != null) {
             long oldTime = state.getOverspeedTime().getTime();
             long newTime = position.getFixTime().getTime();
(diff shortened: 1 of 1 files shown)


jguthrie100:
Speed limit: 30

[Screencast from 2023-08-15 12:50:44.webm](https://github.com/traccar/traccar/assets/4670502/952e9aa2-81e5-48ab-829b-206302217953)

tananaev:
So you don't respect the duration anymore?

jguthrie100:
minimalDuration is checked in the refactored method on line 56.. the functionality hasn't changd in any way from before, from what I can tell.

I changed the minimalDuration check to be `>=`, because if you have a minimalDuration of 2 seconds, then if exactly 2 seconds has passed, then that satisfies the minimalDuration (whereas before it had to wait until 3 seconds or so). It also means that if the minimalDuration is 0, then it sends the event immediately.

tananaev:
How do you do it without a second reading?

jguthrie100:
I don't understand what you mean..

On each iteration, from what I can tell, the current time is checked against the time at which the OverspeedState is first initialised. So on the first matching check, the current time will match the initial overspeed time (they are both set in the same iteration), and unless the minimalDuration is 0, then the minimalDuration check will fail.

Then on the next iteration, the updated position time is checked against the original OverspeedState time, and this keeps happening until the minimalDuration is exceeded..

Am I understanding it wrong?

jguthrie100:
If I change the `event.overspeed.minimalDuration` config value to 5, then the event only gets triggered after 5 seconds..

[Screencast from 2023-08-15 15:53:16.webm](https://github.com/traccar/traccar/assets/4670502/fe08b82e-d85f-424c-8286-f799ae84070f)

tananaev:
So what is the difference with previous implementation? Support for zero duration?

src/main/java/org/traccar/session/state/OverspeedProcessor.java
```diff
@@ -59,7 +44,27 @@ public static void updateState(
             state.setOverspeedState(true);
             state.setOverspeedTime(position.getFixTime());
             state.setOverspeedGeofenceId(geofenceId);
+
+            setEvent(state, position, speedLimit, minimalDuration);
         }
     }
 
+    private static void setEvent(OverspeedState state, Position position, double speedLimit, long minimalDuration) {
```
tananaev, line 52: I think the method should be rename to something like `checkEvent`.

PR #5176, opened 2023-09-13
filter positions without jump but with invalid speed

jcardus:


tananaev:
That is not the point of that filter.

jcardus:
Yes, but it's still related to maxSpeed which is what I'm trying to filter. Some devices send this valid positions with invalid speeds generating false notifications.

tananaev:
It shouldn't be a part of this filter. Also please provide specific examples.

PR #5182, opened 2023-09-19
Speed is expected as knots value.

geduxas:
OBD speed showed wrong in GUI, because it's expected to be saved as knot. To fix that value should be converted to knot's from km/h as used in other code location's.


geduxas:
In web, obdspeed and speed treaded as same, so should i modify obdspeed in web?

tananaev:
Yes.

PR #5187, opened 2023-09-26
fix PR #4839:

yuriy-piskarev:
- before "case 5" delete blank line;
- field 15 conversion to meters;
- fields 53, 66 (check 15 bit, value 0-14 bits);
- field 54 check negative, conversion to liters;
- field 57 conversion to meters;
- fields 207-255 fix field num check (I didn't compare the index correctly)

### DIFF
--- a/src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
@@ -193,13 +193,12 @@ public class NavtelecomProtocolDecoder extends BaseProtocolDecoder {
                     Position position = new Position(getProtocolName());
                     position.setDeviceId(deviceSession.getDeviceId());
 
-                    for (int j = 0; j < bits.length(); j++) {
-                        if (bits.get(j)) {
+                    for (int j = 1; j <= bits.length(); j++) {
+                        if (bits.get(j - 1)) {
 
-                            int bitNum = j + 1;
                             int value;
 
-                            switch (bitNum) {
+                            switch (j) {
                                 case 1:
                                     position.set(Position.KEY_INDEX, buf.readUnsignedIntLE());
                                     break;
@@ -259,7 +258,7 @@ public class NavtelecomProtocolDecoder extends BaseProtocolDecoder {
                                 case 24:
                                 case 25:
                                 case 26:
-                                    position.set(Position.PREFIX_ADC + (j + 2 - 21), buf.readUnsignedShortLE() * 0.001);
+                                    position.set(Position.PREFIX_ADC + (j + 1 - 21), buf.readUnsignedShortLE() * 0.001);
                                     break;
                                 case 29:
                                     value = buf.readUnsignedByte();
@@ -275,11 +274,11 @@ public class NavtelecomProtocolDecoder extends BaseProtocolDecoder {
                                     break;
                                 case 33:
                                 case 34:
-                                    position.set(Position.PREFIX_COUNT + (j + 2 - 33), buf.readUnsignedIntLE());
+                                    position.set(Position.PREFIX_COUNT + (j + 1 - 33), buf.readUnsignedIntLE());
                                     break;
                                 case 35:
                                 case 36:
-                                    position.set("freq" + (j + 2 - 35), buf.readUnsignedShortLE());
+                                    position.set("freq" + (j + 1 - 35), buf.readUnsignedShortLE());
                                     break;
                                 case 37:
                                     position.set(Position.KEY_HOURS, buf.readUnsignedIntLE() * 1000);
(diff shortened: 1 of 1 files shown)


yuriy-piskarev:
@tananaev 

Hi Anton!

I haven't been able to check yet (there were some minor difficulties with server deployment). I'll check it tomorrow, I hope.
So far I have added fixes based on comments from Igor (there are obvious errors) and found several more similar bugs.

Check out the current version, have any comments on the style?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -196,9 +196,10 @@ protected Object decode(
                     for (int j = 0; j < bits.length(); j++) {
                         if (bits.get(j)) {
 
+                            int bitNum = j + 1;
```
tananaev, line 199: Can we just have `j` value to be `1` to `bits.length()`?

yuriy-piskarev: Anton, most likely I don’t understand the translation exactly, can you formulate the question differently? Sorry)

yuriy-piskarev:
Tested it and fixed a few bugs. But I would check several times before Monday.

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
tananaev: My point is that you should just have `j` start from 1 instead of 0. That way you won't need a second variable.

yuriy-piskarev: Great solution.  I'll fix it today and test it again.

yuriy-piskarev:
@tananaev 
Anton, look, are there any more comments?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -308,20 +307,21 @@ protected Object decode(
...
+                                            position.set("obdFuelLevel", BitUtil.to(value, 14));
                                         } else {
-                                            position.set("obdFuel", BitUtil.to(value, 6) / 10);
+                                            position.set("obdFuel", BitUtil.to(value, 14) / 10.0);
                                         }
                                     }
                                     break;
                                 case 54:
-                                    position.set(Position.KEY_FUEL_USED, buf.readFloatLE());
+                                    double dValue = buf.readFloatLE() * 0.5;
```
tananaev, line 323: Why are we excluding zero?

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
```diff
@@ -357,10 +357,10 @@ protected Object decode(
                                 case 66:
                                     value = buf.readUnsignedShortLE();
                                     if (value != 0x7FFF) {
-                                        if (BitUtil.check(value, 7)) {
-                                            position.set("obdAdBlueLevel", BitUtil.to(value, 6));
+                                        if (BitUtil.check(value, 15)) {
+                                            position.set("obdAdBlueLevel", BitUtil.to(value, 14));
                                         } else {
-                                            position.set("obdAdBlue", BitUtil.to(value, 6) / 10);
+                                            position.set("obdAdBlue", BitUtil.to(value, 14) / 10.0);
```
tananaev, line 363:
```suggestion
                                            position.set("obdAdBlue", BitUtil.to(value, 14) * 0.1);
```

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: For this field value is not valid if < 0

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
yuriy-piskarev: Ok, I'll fix it

src/main/java/org/traccar/protocol/NavtelecomProtocolDecoder.java
tananaev: OK, can we name the value properly then. I recommend something like `fuelUsed`.

yuriy-piskarev: Ok!

tananaev: What you did doesn't make sense. It's not even the same line of code. You shouldn't change the key. I'm asking to change `dValue` to `fuelUsed`.

yuriy-piskarev: done!

tananaev:
Merged, thanks.

PR #5190, opened 2023-09-30
Ruptela - Heartbeat support & extended records spanning multiple messages fix

jinzo:
Hi,

this pull request fixes a bug (first reported to me by Joao Santos via email, thanks!) where Ruptela Extended Messages could split the records to merge in two messages. Eg: the device buffered a lot of extended records (more than 11 that go into one message) because of network problems or some such, and then when it started sending them it could happen that first message contained some records that should be merged with the records in second message. When that second message came, the system previously errored out because it couldn't merge them, as the array of records (= positions in Traccar) was empty. So the correct solution is/was to get the last stored record/position and merge it with that one. 
I was finally able to replicate this and figure it out. I'm also preparing some more changes for a later PR, and I hope I'll be able to extend the tests too - current sample has too much PII to include in the tests.

I included a simple Heartbeat implementation as I noticed that the devices do behave in a nicer way if they get a positive response to the heartbeat message.

### DIFF
--- a/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
@@ -51,6 +51,7 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
     public static final int MSG_SMS_VIA_GPRS = 8;
     public static final int MSG_DTCS = 9;
     public static final int MSG_IDENTIFICATION = 15;
+    public static final int MSG_HEARTBEAT = 16;
     public static final int MSG_SET_IO = 17;
     public static final int MSG_FILES = 37;
     public static final int MSG_EXTENDED_RECORDS = 68;
@@ -388,7 +389,7 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
 
             return null;
 
-        } else if (type == MSG_IDENTIFICATION) {
+        } else if (type == MSG_IDENTIFICATION || type == MSG_HEARTBEAT) {
 
             ByteBuf content = Unpooled.buffer();
             content.writeByte(1);
--- a/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
@@ -227,92 +227,103 @@ public class RuptelaProtocolDecoder extends BaseProtocolDecoder {
 
             List<Position> positions = new LinkedList<>();
 
-            buf.readUnsignedByte(); // records left
-            int count = buf.readUnsignedByte();
+            try {
+                buf.readUnsignedByte(); // records left
+                int count = buf.readUnsignedByte();
 
-            for (int i = 0; i < count; i++) {
-                Position position = new Position(getProtocolName());
-                position.setDeviceId(deviceSession.getDeviceId());
+                for (int i = 0; i < count; i++) {
+                    Position position = new Position(getProtocolName());
+                    position.setDeviceId(deviceSession.getDeviceId());
 
-                position.setTime(new Date(buf.readUnsignedInt() * 1000));
-                buf.readUnsignedByte(); // timestamp extension
+                    position.setTime(new Date(buf.readUnsignedInt() * 1000));
+                    buf.readUnsignedByte(); // timestamp extension
 
-                if (type == MSG_EXTENDED_RECORDS) {
-                    int recordExtension = buf.readUnsignedByte();
-                    int mergeRecordCount = BitUtil.from(recordExtension, 4);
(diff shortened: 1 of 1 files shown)


jinzo:
I squeezed in another improvement - one thing while debugging this was that the device started burning thorough data. This was caused by the exception, device not getting any response back and retrying almost immediately. I wrapped the record parsing with an exception catcher, where we just write back the negative response and the device knows that something is wrong and introduces a retry delay. This then prevents it from burning thorough data.

src/main/java/org/traccar/protocol/RuptelaProtocolDecoder.java
```diff
@@ -226,88 +227,103 @@ protected Object decode(
...
+            } catch (Exception e) {
+                ByteBuf content = Unpooled.buffer();
+                content.writeByte(0);
+                ByteBuf response = RuptelaProtocolEncoder.encodeContent(0, content);
+                content.release();
+                if (channel != null) {
+                    channel.writeAndFlush(new NetworkMessage(response, remoteAddress));
+                }
+                throw e;
             }
```
tananaev, line 327: Are you saying there's a device bug and we're fixing it? Doesn't feel right to me. What if we just ignore it?

jinzo: Not exactly. It is documented behaviour. It's just a little bit badly designed on their end. Instead of treating anything that is not an ACK as a failed delivery, they have three states:
- ACK: everything OK, delete the sent messages from flash
- NACK: negative, increase delay period and retry
- get no response: continuously send the same message with high frequency

All of this could be avoided if they would treat everything that is not an ACK as failed delivery/NACK. But the devices do not and behave like mentioned above (and in the protocol docs). So this catch makes sure that in the event of an error the devices get appropriate NACK and don't spam the server with retries.

tananaev: Documented behavior? I'm talking about the fact that device sends broken messages.

jinzo: Ah sorry misunderstood. They are not broken per-se, it's just not mentioned anywhere in the documentation (at least that I could see) that this edge case could happen. But it makes sense once you think about it. Ruptela is just packing as much positions it can in one message and it can happen that positions that should be merged end up in different messages. 
eg. Ruptela is storing the positions that it can't send in flash, and then when it detects that it can send them (because the network is back up for example) it just takes first X positions (that fit into the bytes-that-can-be-sent-in-one-message) and sends them, does no checking to determine that if positions-that-should-be-merged end up in the same message. So it can happen that they are split in multiple messages (but are always ordered) and we need to account that when merging them. So it's a completely valid message/behaviour, I just did not expect that this can also happen when doing this. As it happens relatively rarely (that there is soo much positions that they need to be sent in multiple messages AND that position-that-needs-to-be-merged ends up split between two messages). My oversight for sure - that failed rarely but when it failed it failed preatty bad (because of the ACK/NACK stuff that I described above)

tananaev: OK, if that's expected then it shouldn't be handled as an exception. I'm still not entirely sure what's happening and why it causes and error. Maybe you can explain on an example? Thank you

jinzo: Yeah we are misunderstanding what's actually happening - let me try to comment specific lines and reasoning for them.

tananaev:
@jinzo looks like you requested another review, but you haven't provided what I asked.

jinzo:
I provided additional comments that hopefully clarify things? They are marked as Pending, not sure what that means tho? you do not see 3 new comments I made?

tananaev:
Pending means that you haven't actually sent them.

jinzo (commented):
Ok, this should now publish the review/comments

tananaev:
I still need to understand what exactly is happening. What is the root cause of the exception? You keep repeating about the ack issue, but that's a side effect not a root cause. I want to know why there's an exception on some real example.

jinzo:
I fixed the one root cause of exception in this PR (the first change, "positions-that-should-be-merged are in different messages"). I don't know of any others currently (if I would, I would fix them too). It's mostly a precaution, that if it happens, it won't eat all my data before I notice it and fix it.

tananaev:
Unfortunately I don't see it as a proper fix. I see it as a pretty nasty workaround. Waiting for a specific example with a breakdown to understand what the problem is and how we can handle it properly.

jinzo:
Ok, I removed it. I understand your reasoning, from my perspective it fit nicely into what Ruptela documented. But then the two other original changes are ready for merging I think?

tananaev:
Merged, thank you.

PR #5192, opened 2023-10-05
(Bugfix-T800X Protocol) Fix messages response for 0x2323

sherbinator:
### The problem
The devices that use header `0x2323` can't establish a stable TCP connection which can lead to unexpected behavior because the devices don't receive the correct response from the server for login/heartbeat messages.

### The solution
According to the attached document, the devices of header `0x2323` are expecting the response to have the serial bytes to be `0x00 0x01` in response messages.

steps:
1. the device sends a login message to the server ex. `232301001500030[IMEI]150208625020`
2. the server responds with `232301000f00010[IMEI]` which is the same message that the device sent. note that the serial bytes are `0x00 0x01`
3. the device sends a heartbeat message to the server ex. `232303000f00b50[IMEI]`
4. the server responds with `232303000f00010[IMEI]` which is the same message that the device sent. note that the serial bytes are `0x00 0x01`, the correct response has to be sent other wise the device will store cache of GPS/ALARM messages and send them again. (this will dramatically increase the telemetry data usage)
5. if the device send alarm message. then, the server has to respond or the device will cache the message and send it again.
6. settings/forward messages also needs response.

notes:
- we have tested the solution with real devices and it works fine. models:
    - T8808B
    - T8808+
- according to the manufacture the header `0x2323` is EOL and we are not expecting new devices to use it. so, we are not expecting any side effects from this change.

## PR Checklist

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## References
[TOPFLYTECH T880X Protocol V6.2](https://www.traccar.org/protocol/5094-t800x/TOPFLYTECH%20T880X%20Protocol%20V6.2.xlsx)

## Credits
[Jaime Duardo - Hardware Engineer](https://github.com/jduardo2704)

[Abdelrahman Mousa - Software Engineer](https://github.com/sherbinator)

### DIFF
--- a/src/main/java/org/traccar/protocol/T800xProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/T800xProtocolDecoder.java
@@ -163,12 +163,8 @@ public class T800xProtocolDecoder extends BaseProtocolDecoder {
 
         boolean positionType = type == MSG_GPS || type == MSG_GPS_2 || type == MSG_ALARM || type == MSG_ALARM_2;
         if (!positionType) {
-
-            if (header == 0x2323) {
-                sendResponse(channel, header, type, 0x0001, imei, 0);
-            } else {
-                sendResponse(channel, header, type, index, imei, 0);
-            }
+            var serial = header == 0x2323 ? 0x0001 : index;
+            sendResponse(channel, header, type, serial, imei, 0);
         }
 
         if (positionType) {
@@ -523,12 +519,8 @@ public class T800xProtocolDecoder extends BaseProtocolDecoder {
         boolean acknowledgement = AttributeUtil.lookup(
                 getCacheManager(), Keys.PROTOCOL_ACK.withPrefix(getProtocolName()), deviceSession.getDeviceId());
         if (acknowledgement || type == MSG_ALARM || type == MSG_ALARM_2) {
-
-            if (header == 0x2323) {
-                sendResponse(channel, header, type, 0x0001, imei, alarm);
-            } else {
-                sendResponse(channel, header, type, index, imei, alarm);
-            }
+            var serial = header == 0x2323 ? 0x0001 : index;
+            sendResponse(channel, header, type, serial, imei, alarm);
         }
 
         return position;


tananaev (commented):
You have to provide a description.

sherbinator:
I have updated the PR. lemme know if you need more clarification.

src/main/java/org/traccar/protocol/T800xProtocolDecoder.java
```diff
@@ -163,7 +163,12 @@ protected Object decode(
 
         boolean positionType = type == MSG_GPS || type == MSG_GPS_2 || type == MSG_ALARM || type == MSG_ALARM_2;
         if (!positionType) {
-            sendResponse(channel, header, type, index, imei, 0);
+
+            if (header == 0x2323) {
+                sendResponse(channel, header, type, 0x0001, imei, 0);
```
tananaev, line 168: Can we send 1 for other devices as well?

sherbinator: according the official documentation of the manufacturer, we believe that all devices use header 0x2323 will need to receive the serial `1`.

tananaev: I meant devices with other headers?

sherbinator: I guess no, starting from header 0x2525 the serial will be computed depending on device info

We are testing some devices and I can confirm that it will not be the case in the following headers:

0x2525 gen 1 and gen 2
0x2626
0x2727

jduardo2704: Hello Anton, it's not possible; it only applies to devices with header ```0x2323```. In other cases, you need to respond with the serial number that comes in the device's payload. The documentation regarding other headers from Topflytech can be found in the description of this [pull request](https://github.com/traccar/traccar/pull/5191).

tananaev: OK, can we do something like this then: `header == 0x2323 ? 1 : index`

sherbinator: Yeah sure, I'll add it

sherbinator: done 👍

PR #5202, opened 2023-10-25
add mqtt position forwarder and refactor mqtt event forwarder to use shared code

ivanfmartinez:
added support to send all position data  using mqtt
the Event forwarder was refactored to use same code as the new position forwarder.

this code is running fine in my personal instance and I use it to integrate with some applications that use mqtt

### DIFF
--- a/src/main/java/org/traccar/forward/PositionForwarderMqtt.java
+++ b/src/main/java/org/traccar/forward/PositionForwarderMqtt.java
@@ -16,6 +16,7 @@
 package org.traccar.forward;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -26,12 +27,11 @@ import org.traccar.model.Position;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.hivemq.client.mqtt.datatypes.MqttQos;
 import com.hivemq.client.mqtt.mqtt3.Mqtt3AsyncClient;
 import com.nimbusds.oauth2.sdk.util.StringUtils;
 
 public class PositionForwarderMqtt implements PositionForwarder {
-
+	
 	private final Mqtt3AsyncClient client;
     private final ObjectMapper objectMapper;
     private final String topic;
@@ -80,14 +80,21 @@ public class PositionForwarderMqtt implements PositionForwarder {
 
     @Override
     public void forward(PositionData positionData, ResultHandler resultHandler) {
-    	final Set<String> options = new HashSet<>(Arrays.asList(positionData.getDevice().getString(MQTT_FORWARD_OPTIONS, "").split(",")));
-        /**
+
+    	// get options and remove spaces
+    	final String optionsSt = positionData.getDevice().getString(MQTT_FORWARD_OPTIONS, "").replaceAll(" ", "");
+    	final Set<String> options = StringUtils.isBlank(optionsSt) ? Collections.emptySet() : new HashSet<>(Arrays.asList(optionsSt.split(",")));
+    	
+    	/**
          * If mqtt.alias is defined on device publish position information
          * otherwhise publish full positionData
          * this make easy to parse specific device location
          */
         final String alias = positionData.getDevice().getString(MQTT_ALIAS, "");
-        if (StringUtils.isNotBlank(alias) && (options.isEmpty() || options.contains(OPTION_ALIAS)) && !options.contains(OPTION_NOALIAS)) {
+
+		if (StringUtils.isNotBlank(alias) && 
+				(options.isEmpty() || options.contains(OPTION_ALIAS)) && 
+				!options.contains(OPTION_NOALIAS)) {
             publishPosition(topic + "/" + alias , positionData.getPosition(), resultHandler, options);
         }
 
@@ -95,7 +102,7 @@ public class PositionForwarderMqtt implements PositionForwarder {
             publishPosition(topic + "/" + positionData.getPosition().getDeviceId() , positionData.getPosition(), resultHandler, options);
         }
         
-        if (StringUtils.isBlank(OPTION_ALIAS) || options.contains(OPTION_COMBINED)) {
+        if (StringUtils.isBlank(alias) || options.contains(OPTION_COMBINED)) {
             publish(topic, positionData, resultHandler);
         }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/traccar/forward/PositionForwarderMqtt.java
```diff
@@ -0,0 +1,130 @@
...
+     * Device alias to be used when publising data
+     */
+    private static final String MQTT_ALIAS = "mqtt.alias";
+    /**
+     * Options for publishing empty/not defined - combined data, and if alias
+     * defined position information combined - combined data alias - device position
+     * deviceId - device position noalias - dont publish with alias even when alias
+     * defined
+     */
+    private static final String MQTT_FORWARD_OPTIONS = "mqtt.forward.options";
```
tananaev, line 49: We should start without any options. I'm not sure about including it.

src/main/java/org/traccar/forward/PositionForwarderMqtt.java
```diff
@@ -0,0 +1,130 @@
...
+
+        if (StringUtils.isBlank(alias) || options.contains(OPTION_COMBINED)) {
+            publish(topic, positionData, resultHandler);
+        }
+
+    }
+
+    private void publishPosition(final String topic, final Position position, final ResultHandler resultHandler,
+            final Set<String> options) {
+        // TODO option to split position attributes in specific topics
```
tananaev, line 110: There shouldn't be any TODOs.

src/main/java/org/traccar/forward/PositionForwarderMqtt.java
```diff
@@ -0,0 +1,130 @@
...
+        // for now just publish single topic with json
+        if (options.isEmpty() || options.contains(OPTION_DEVICE_JSON)) {
+            publish(topic, position, resultHandler);
+        }
+
+        // TODO option to send geofence names
```
tananaev, line 116: No

src/main/java/org/traccar/helper/MqttUtil.java
```diff
@@ -0,0 +1,101 @@
...
+import com.hivemq.client.mqtt.mqtt3.Mqtt3Client;
+import com.hivemq.client.mqtt.mqtt3.Mqtt3ClientBuilder;
+import com.hivemq.client.mqtt.mqtt3.message.auth.Mqtt3SimpleAuth;
+import com.hivemq.client.mqtt.mqtt3.message.publish.Mqtt3Publish;
+import com.nimbusds.oauth2.sdk.util.StringUtils;
+
+/**
+ * MQTT generic calls
+ *
+ */
```
tananaev, line 34: Please remove most of the comments. Only include where it's absolutely necessary and the code is not clear for some reason.

src/main/java/org/traccar/forward/PositionForwarderMqtt.java
```diff
@@ -0,0 +1,77 @@
...
+public class PositionForwarderMqtt implements PositionForwarder {
+
+    private final Mqtt3AsyncClient client;
+    private final ObjectMapper objectMapper;
+    private final String topic;
+
+    /*
+     * Device alias to be used when publishing data
+     */
+    private static final String MQTT_ALIAS = "mqtt.alias";
```
tananaev, line 37: What is this parameter?

ivanfmartinez: Instead of publishing the data only on a single topic for all devices if a device have an alias defined (in device attributes) position for this device will be published on specific topic. This make easy to filter / subscribe to specific device position information

like
/traccar/position/alias

instead of
/traccar/position



You can see on : https://github.com/traccar/traccar/pull/5202/files/34c88d44dcec94ed13b87eae516581602a0d2362#diff-a232b34e7d9106c3ae8030135f5038e844dc0f6550c8f73f78028a396e9d3195R48

tananaev: I thought we agreed on removing configuration parameters for now.

ivanfmartinez: You have requested to remove that OPTIONS parameters.

This is the minimum that I use to integrate this with my environment, split positions by an easy to use name.

tananaev: This is not something we do for forwarding now.

ivanfmartinez:
@tananaev check if this last commit are acceptable, and I will make local changes to instantiate a subclass with my needs.

src/main/java/org/traccar/helper/MqttUtil.java
```diff
@@ -0,0 +1,91 @@
...
+import com.hivemq.client.mqtt.mqtt3.message.publish.Mqtt3Publish;
+import com.nimbusds.oauth2.sdk.util.StringUtils;
+
+public final class MqttUtil {
+
+    private MqttUtil() {
+
+    }
+
+    public static Mqtt3AsyncClient createClient(final String url, final String willTopic) {
```
tananaev, line 37: What is `willTopic`? And why do we need to send "connected" and "disconnected" payload?

ivanfmartinez: This is standard a way to tell any other mqtt client that this application is connected or not.

We send "connected" when connected, and if the connection is lost the server will automatically send the "disconnected" payload to the topic. This is called last will.

An application that use this data have a way to know if this specific process is connected or not to the mqtt server.

in this case if admin configure the position forward to "/traccar/position" the code will send "connected" with retain flag to "/traccar/position/state" and when the traccar server stops and disconnect from mqtt server, the server will change this retained topic to "disconnected"

tananaev: Can you please provide a link to this standard.

ivanfmartinez: Here you can read more about this procedure : https://www.hivemq.com/blog/mqtt-essentials-part-9-last-will-and-testament/

tananaev: This doesn't look like a standard to me. Even the article implies that it's an optional thing.

ivanfmartinez: The article says "To summarize, the Last Will and Testament (LWT) feature in MQTT is crucial in ensuring efficient communication and maintaining system integrity in the event of unexpected client disconnections. By combining LWT with retained messages, developers can store and communicate client state on specific topics, providing valuable information to subscribed clients"

And in my environment I use this. This dont hurt (very low extra traffic) and gives extra context information.

Removing this will make more difficult to extend with my needs, in this case I will stay with my changes locally.

tananaev: OK. Let's leave the PR open in case anyone is interested in updating it later.

ramonsmits: Its 2 different things. Last will is used for health monitoring. I would not combine that in this PR. Last will behavior could be added later after this is merged?

ramonsmits:
What topics are used for position/event forwarding? It seems the current MQTT implementation isn't really following how most MQTT systems publish topics around isolated topics.

For example, https://github.com/ivanfmartinez/ivanfm-traccar-mqtt seems to have topics per certain values to make is easy for subscribers to subscribe at a very granular level to minimize traffic/communication between the MQTT broker and the subscriber.

src/main/java/org/traccar/helper/MqttUtil.java
supriyoMondal: @tananaev if forwarding of connected and disconnected is disabled is this pr acceptable ?

tananaev: Yes, I think most other stuff is fine.

geduxas: Truly I don't see here anything wrong with LWT topic, it's used almost everywhere in same way as @ivanfmartinez implemented it. Through this topic all subscribers will know status of slender, if traccar suddenly dies, or disconnects from MQTT server, there's no way to know that.
@tananaev please accept this pr in full, i am also waiting for it.

tananaev: This PR won't be accepted as is. It does not make sense to implement a special case just for one forwarder.

geduxas: But it is, @tananaev , maybe you are seeing something more about it, but last will testament works like that, it set's retain flag for offline message, and keeps it updated with online message.. so that's how it's possible implement disconnected state in mqtt. Sorry if i don't understand something else...

tananaev: Correct, but I'm against including LWT in this PR.

PR #5204, opened 2023-10-29
totp

e-macgregor:


drattek:
How does this work when you are setting up the system for the first time, im being asked the TOTP code but dont have one yet.

tananaev:
It shouldn't be asking. Make sure you're using the latest master.

schaurian:
I can't find any option to enable it using the current TOTP. What am i missing? Is there some docs i wasn't able to find?

PR #5207, opened 2023-11-09
request pagination

gbrvalerio:
- This draft PR creates the basis for implementing Pagination in storage requests

#### Core
- [x] Create Pagination class
- [x] Adapt Request class to include pagination
- [x] Adapt DatabaseStorage to handle the request pagination
- [x] Remove the limit prop from the Order class
- [x] Fix current usage of the Request class to use Pagination

#### Implementation
- [ ] Add skip and limit url params to the list devices endpoint
  - [ ] Update Docs
  - [x] Update query
- [ ] Add skip and limit url params to the list users endpoint
  - [ ] Update Docs
  - [x] Update query
- [ ] Add skip and limit url params to the base class SimpleObjectResource (affects Groups, Calendars, Orders)
  - [ ] Update Docs
  - [x] Update query
- [ ] Add skip and limit url params to the list drivers endpoint
  - [ ] Update Docs
  - [x] Update query
  
### Databases Tested
- [x] Postgres
- [x] H2
- [ ] MS SQL

@tananaev

### DIFF
--- a/src/main/java/org/traccar/storage/DatabaseStorage.java
+++ b/src/main/java/org/traccar/storage/DatabaseStorage.java
@@ -316,15 +316,23 @@ public class DatabaseStorage extends Storage {
     private String formatPagination(Pagination pagination) {
         StringBuilder result = new StringBuilder();
         if (pagination != null) {
+        	if (databaseType.equals("Microsoft SQL Server")) {
+        		result.append(" OFFSET ");
+        		result.append(pagination.getSkip());
+        		result.append(" ROWS FETCH FIRST ");
+                result.append(pagination.getLimit());
+                result.append(" ROWS ONLY");
+                return result.toString();
+        	}
+        	
         	if(pagination.getSkip() > 0) {
         		result.append(" OFFSET ");
                 result.append(pagination.getSkip());
         	}
+        	
             if (pagination.getLimit() > 0) {
             	if (databaseType.equals("Microsoft SQL Server")) {
-                    result.append(" OFFSET 0 ROWS FETCH FIRST ");
-                    result.append(pagination.getLimit());
-                    result.append(" ROWS ONLY");
+                    
                 } else {
                     result.append(" LIMIT ");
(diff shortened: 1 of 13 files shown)


src/main/java/org/traccar/storage/DatabaseStorage.java
```diff
@@ -307,14 +309,25 @@ private String formatOrder(Order order) {
...
+            	if (databaseType.equals("Microsoft SQL Server")) {
                     result.append(" OFFSET 0 ROWS FETCH FIRST ");
-                    result.append(order.getLimit());
+                    result.append(pagination.getLimit());
                     result.append(" ROWS ONLY");
                 } else {
                     result.append(" LIMIT ");
-                    result.append(order.getLimit());
+                    result.append(pagination.getLimit());
                 }
```
tananaev, line 331: Which databases have you tested?

gbrvalerio: PG so far

tananaev: We need to test at least H2, MySQL, Postgres and MS SQL. Is there an SQL standard for it?

gbrvalerio: code wise, the path is correct?

tananaev: I only did a brief look, but I think it makes sense. I would separate database changes and report changes though.

gbrvalerio:
@tananaev regarding docs: is it automatically updated or should I do some manual work here? If so, can you point me where should I pay attention and update?

ty

tananaev:
What docs do you need to update?

gbrvalerio:
the api ones to include the new `skip` and `limit` url params in the endpoints

tananaev:
Well, we need to split the PR first.

gbrvalerio:
closed in favor of #5208 and #5209

PR #5208, opened 2023-11-10
make DatabaseStorage support pagination

gbrvalerio:
This PR adds pagination capability on the `DatabaseStorage` and the correct handlers on the `Request` class.
Also removes the `Order.limit` in favor of the new `Pagination.limit` and fix the usage of the class throughout the app.

### DIFF
--- a/src/main/java/org/traccar/database/CommandsManager.java
+++ b/src/main/java/org/traccar/database/CommandsManager.java
@@ -33,6 +33,7 @@ import org.traccar.storage.StorageException;
 import org.traccar.storage.query.Columns;
 import org.traccar.storage.query.Condition;
 import org.traccar.storage.query.Order;
+import org.traccar.storage.query.Pagination;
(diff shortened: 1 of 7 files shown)


tananaev:
So, which databases have you tested and is the syntax an SQL standard?

gbrvalerio:
yes, the syntax is SQL standard (OFFSET/LIMIT).

I've tested on PGSQL and H2. Yet to validate on MS SQL

tananaev:
Please provide a link to the SQL standard with this.

And let me know once you test everything.

gbrvalerio:
[PGSQL](https://www.postgresql.org/docs/current/queries-limit.html#QUERIES-LIMIT) 
["For compatibility with PostgreSQL, MySQL also supports the LIMIT row_count OFFSET offset syntax."](https://dev.mysql.com/doc/refman/8.0/en/select.html) ...

the first definition appears on the SQL:2011 spec.

PR #5226, opened 2023-11-28
extract the logic for handling changes into a separate method (handle…

moussa-elAqqaoui:
…Changes) to improves the modularity of the code. Each method now has a clear responsibility


tananaev:
I don't think we can accept it. It doesn't really solve any clear problem.

PR #5230, opened 2023-12-02
Revert and fix previous conflicting commits TeltonikaProtocolDecoder.…

victor-butler:
- Adding the missing 4G devices from the FMBXXX group - FMCs and FMMs. Those are the same as FMB devices just the 4G versions.
- Reverting changes of previous edit (cd1ce7c3e89d12f59339ccd2676d7c2bfc1596cf):
1. Removing the parameters which are only available via additional CAN readers and are not reported by standalone FMBXXX devices. There are non CAN alternatives for those parameters which creates a conflict.
2. Removing the alarm association for ALVs 246, 247, 249, 252. The proposed code only records a single alarm for those AVLs and only if the AVL value is greater than 0. ALV 246, 247, 249, 252 parameters can be in fact zeros or even bear additional information with values 1-8 for the accident AVL. Since traccar doesn't have a separate alarm for the zero values of those parameters, it's better for those to stay generic and handle them with computed attributes if needed.

Reference:
https://wiki.teltonika-gps.com/view/FMB003_Teltonika_Data_Sending_Parameters_ID

Note: People should refrain from making protocol wide changes that only suit their specific use cases.


src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -254,28 +246,16 @@ private static void register(int id, Set<String> models, BiConsumer<Position, By
...
         register(205, fmbXXX, (p, b) -> p.set("cid2g", b.readUnsignedShort()));
         register(206, fmbXXX, (p, b) -> p.set("lac", b.readUnsignedShort()));
-        register(232, fmbXXX, (p, b) -> p.set("cngStatus", b.readUnsignedByte() > 0));
-        register(233, fmbXXX, (p, b) -> p.set("cngUsed", b.readUnsignedInt() * 0.1));
-        register(234, fmbXXX, (p, b) -> p.set("cngLevel", b.readUnsignedShort()));
-        register(235, fmbXXX, (p, b) -> p.set("oilLevel", b.readUnsignedByte()));
+        register(232, fmbXXX, (p, b) -> p.set("cngStatus", b.readUnsignedByte() > 0)); //only available via additional CAN reader
+        register(233, fmbXXX, (p, b) -> p.set("cngUsed", b.readUnsignedInt() * 0.1)); //only available via additional CAN reader
+        register(234, fmbXXX, (p, b) -> p.set("cngLevel", b.readUnsignedShort())); //only available via additional CAN reader
+        register(235, fmbXXX, (p, b) -> p.set("oilLevel", b.readUnsignedByte())); //only available via additional CAN reader
```
tananaev, line 252: Why do we need these comments?

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -254,28 +246,16 @@ private static void register(int id, Set<String> models, BiConsumer<Position, By
...
-        });
-        register(247, fmbXXX, (p, b) -> {
-            p.set(Position.KEY_ALARM, b.readUnsignedByte() > 0 ? Position.ALARM_ACCIDENT : null);
-        });
-        register(249, fmbXXX, (p, b) -> {
-            p.set(Position.KEY_ALARM, b.readUnsignedByte() > 0 ? Position.ALARM_JAMMING : null);
-        });
-        register(252, fmbXXX, (p, b) -> {
-            p.set(Position.KEY_ALARM, b.readUnsignedByte() > 0 ? Position.ALARM_POWER_CUT : null);
-        });
```
tananaev, line 278: Can you please explain what the problem is with all these?

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -236,15 +236,7 @@ private static void register(int id, Set<String> models, BiConsumer<Position, By
             }
         });
         register(80, fmbXXX, (p, b) -> p.set("dataMode", b.readUnsignedByte()));
-        register(81, fmbXXX, (p, b) -> p.set(Position.KEY_OBD_SPEED, b.readUnsignedByte()));
-        register(82, fmbXXX, (p, b) -> p.set(Position.KEY_THROTTLE, b.readUnsignedByte()));
```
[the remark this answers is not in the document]
victor-butler: Sure:
1. OBD Speed is AVL 37. This is natively reported by the FMBXXX devices without the need for CAN reader. Hence the conflict.
image

2. Throttle is AVL. 41. This is also natively reported by the FMBXXX devices without the need for CAN reader. Hence the conflict.
image

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
tananaev: It is absolutely expected that not all of the parameters will be reported by all devices. That's fine as long as there are no conflicts. Please remove these comments.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
victor-butler: Sure:
1. AVL 246 is towing alarm. It can have values of `0: No Towing` and `1: Towing`. Traccar only has a ALARM_TOW key so we cannot record the No Towing event:
image


2. AVL 247 is accident alarm. It can have values from 1 to 8. This code only records one `accident` value and we have no idea what the AVL actual value was.
image


3. AVL 249 is jamming alarm. It can have values of `0: Jamming Stop` and `1: Jamming Start`. Traccar only has a ALARM_JAMMING key so we cannot record the Jamming Stop event:
image


4. AVL 252 is related to the device power. It can have values of `0: battery present` and `1: battery unplugged`:
image

tananaev: Traccar does not have "no towing" event or "jamming stopped" event. I think it's ok if we also record the original IO value, but I want supported events to be decoded properly.

victor-butler: I am all for decoding properly the supported events but if we leave this code, we create the following issues:

1. If the device sends AVL 249 - jamming stopped, currently the event is entirely skipped. So we have no record of the event ever being sent, unless in the log file.
2. A workaround would be to record the original io as you suggest but then we will first unnecessarily duplicate data in the database, therefore in the in the device details too. Second, the jamming event will be decoded with the protocol and the no jamming has to be decoded manually which means inconsistent handling of the same AVL. 
The accepted best practice would be to either entirely decode the AVL or not at all and then use computed attributes to handle it in a consistent way.

tananaev: I disagree with your conclusion of the best practice. The best practice in this case would be to store both raw value and the alarm.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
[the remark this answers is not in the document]
victor-butler: No, the devices report both - that's the idea of the different AVL ids. This code forces the assignment of the Throttle and  speed keys to the CAN AVLs.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
victor-butler: Commenting removed. Still not comfortable with replacing the original AVL ids with custom names. Equally, someone might prefer to call it gasStatus or lpgStatus.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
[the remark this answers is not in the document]
victor-butler: We are talking about different things - devices can report one or the other, or both. 
The problem is if we wish to have AVL 37 for throttle, the decoder assigns the CAN AVL.

Also, another thing that I just crossed my mind, the CAN adapters are not exclusively used with FMBXXX devices only.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
victor-butler: Let's agree to disagree :)

How do you suggest to fix this because the currently important data is being entirely skipped and missing from the server.

tananaev: I already explained twice what I think it a good solution for this.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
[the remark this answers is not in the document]
victor-butler: Here's the disadvantage to your suggestion - if we want to receive both AVL 37 and AVL 82 (KEY_THROTTLE) in the parameters we will see:
`io37: io37value, throttle: io82value`

Then if, we wish to use `io37` for throttle, then we cannot keep both values using computed attributes because we will end up with this in the positions table:
`throttle: io82value`

Hence the conflict. I don't know how to explain it more clearly.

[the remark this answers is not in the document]
tananaev: This is the first time you mentioned that someone would need both. Why that would be the case? And it also seems like a very corner case that vast majority of people don't care about.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
victor-butler: Well, currently the original raw value is not stored so the code is bad.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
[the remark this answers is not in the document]
victor-butler: I cannot decide for what people might need or not need. Depending on the vehicle and reported data people might need both AVLs. 
I see a conflict (or for the very least unclean code) and I point it out. That's all.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
victor-butler: Yes I am. When I asked how do you suggest to fix this, I meant what are the suggested code change, not about the logic (keep both values).

victor-butler:
> will lv-can work with this modification? if this were not the case it would be a big problem, in fact 70-80% of the world is using these modules

LV-CAN can and worked well without those modifications long before they were made. And it's difficult to make an educated guess of the percentage of people using LV-CAN. This information is simply not available.

geduxas:
Yes i am agree, there is problem with Teltonika because it doesn't use same for all it's device range.. its hard to maintain old devices and new which split i think in 3-4 ranges.. so in any device range avl's could mean different values.. so we need to split them in code and use by configuration. I am glad that @tananaev implemented seperate fmbxxx model's and i was almost shure that for fmbxxx all devices will share same avl's id's.. so my code changes was to support lv-can device's. So i think we need different group to use for devices with lv-can and other's..

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
geduxas: > Yes I am. When I asked about how do you suggest to fix this, I meant what are the suggested code changes, not about the logic (keep both values).

Ok, so we have device which plugs directly to obd and device which uses lv_can.. shame for teltonka that for same functions it will have different avl's don't think ever it will be used in both cases.. so you use one or another..

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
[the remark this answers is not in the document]
geduxas: > I cannot decide for what people might need or not need. Depending on the vehicle and reported data people might need both AVLs. I see a conflict (or for the very least unclean code) and I point it out. That's all.

Ok, so we have device which plugs directly to obd and device which uses lv_can.. shame for teltonka that for same functions it will have different avl's don't think ever it will be used in both cases.. so you use one or another..

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
geduxas: For alarm's yes traccar only have event's only for actual case has, been detected.. but don't have status which means ending... But most use cases i think it's needed to detect event.. and it's not so worth to know how long it took.. so if it's jamming it will trigger alarm each time it's detects.. also with towing.. and crash.. all values list in teltonika means same, crash.. but with different scenarious..

victor-butler: Well, actually LV-CAN is a separate addon device. It's like an enhancement and it can be used both with FMBXXX and non FMBXXX devices. 
So it makes sense that the standalone FMBXXX series send the throttle and so do the LV-CAN addon (with a different ID so that we know which is which).

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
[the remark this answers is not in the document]
victor-butler: Well, actually LV-CAN is a separate addon device. It's like an enhancement and it can be used both with FMBXXX and non FMBXXX devices.
So it makes sense that the standalone FMBXXX series send the throttle and so do the LV-CAN addon (with a different ID so that we know which is which).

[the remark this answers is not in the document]
geduxas: > Well, actually LV-CAN is a separate addon device. It's like an enhancement and it can be used both with FMBXXX and non FMBXXX devices. So it makes sense that the standalone FMBXXX series send the throttle and so do the LV-CAN addon (with a different ID so that we know which is which).

Yes,also you can plug Bluetooth obd and connect to teltonika.. its third source..

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
victor-butler: I am not sure I understand. 
All I am saying is that while the code in question captures some of the available alarms on Traccar, it so misses other important data like when the jamming/towing ended, all the device connected events and the crash data trace type. Between the two, the latter is worst because it cannot be fixed with computed attributes.
It took me two months to realize something is wrong. I am sure there are other users who are still unaware.
So the code needs correction one way or another.
Waiting for @tananaev to provide his code suggestion.

geduxas: Ok, so removing everything also is not solution.. also while we have those values decoded don't do any harm.

victor-butler:
> Yes i am agree, there is problem with Teltonika because it doesn't use same for all it's device range.. its hard to maintain old devices and new which split i think in 3-4 ranges.. so in any device range avl's could mean different values.. so we need to split them in code and use by configuration. I am glad that @tananaev implemented seperate fmbxxx model's and i was almost shure that for fmbxxx all devices will share same avl's id's.. so my code changes was to support lv-can device's. So i think we need different group to use for devices with lv-can and other's..

The FMBXXX change was indeed a good idea. It helps to tackle the issue of conflicting parameters between the different device series. You are right that FMBXXX devices all share the same AVLs.
The tricky part about the LV-CAN devices is that they send their data via another "primary" Teltonika device, which can be both FMBXXX or non FMBXXX device. 
So AVL 123 captured by LV-CAN200 will be sent by FMB001 and traccar will decode it as FMBXXX parameter, not LV-CAN200.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
victor-butler: No, there is a problem because currently Traccar is not decoding all data and it's virtually missing from the database. So it  needs to be fixed.

geduxas: Ok, could you give use case scenario where it could lead to problem?

victor-butler: I've already explained in great details above.

geduxas: > I've already explained in great details above.

Yes but what's a point to get end state? Where is a problem without it? Usually it's important to get start, but not length of problem.. so i am asking that, in what cases it's must have to know event length.. 

And for crash.. does the type of crash is so important?  Truly I don't remember from which version it have been detailed.. before that it was only one case.. crash.  Does it mean if crash detected with sensor calibrated or not? How different is full crash from real and limited? :D also what does it means fake crash? 

You only telling facts about states, but don't bring any use case where it's so important.. 

What else i see that you are talking about database.. so it seams that you are working with some extra software which connects directly to database to get it's readings, so maybe you have problems from there? Because I don't see any problem's while using traccar stand alone..

So if you have real problem, could you share to me? Maybe i don't understand exactly what behavior should be acceptable

CarVDB:
Yeap, those events are missing for me as well. Standard traccar install. Thanks for spotting that!

CarVDB:
> > Yeap, those events are missing for me as well. Standard traccar install. Thanks for spotting that!
> 
> What are you missing?

It seems to me you don't really understand what the code changes you've made are doing. 
You are assigning a single alarm key to 4 different IOs, irrespective of the IO value and therefore missing important data. 

Instead, you should be using a switch-case logic to check what the IO value is, as per Teltonika's documentation.
Alternatively, you can leave those IOs unregistered, something which was already suggested, and let the users apply custom logic if needed.

geduxas:
> > > Yeap, those events are missing for me as well. Standard traccar install. Thanks for spotting that!
> > 
> > 
> > What are you missing?
> 
> It seems to me you don't really understand what the code changes you've made are doing. You are assigning a single alarm key to 4 different IOs, irrespective of the IO value and therefore missing important data.
> 
> Instead, you should be using a switch-case logic to check what the IO value is, as per Teltonika's documentation. Alternatively, you can leave those IOs unregistered, something which was already suggested, and let the users apply custom logic if needed.

I am pretty sure i am know what i do. Feel free make changes.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
victor-butler: Hi Anton, any update on fixing the data values that are currently being missed?

tananaev: I think there's a fundamental disagreement, so no updates really.

victor-butler: What do you mean by fundamental disagreements?

Is it true that the server currently deletes the AVL 252 and doesn't record anything in the database if the value of the AVL is `0`, code below:

`register(252, fmbXXX, (p, b) -> {
            p.set(Position.KEY_ALARM, b.readUnsignedByte() > 0 ? Position.ALARM_POWER_CUT : null);
});`

tananaev: The disagreement is that you just want the raw data, but I think most people are interested in decoded attributes, even if it means they technically losing some information.

Issue #1215, opened 2015-05-20
Connect to multiple databases


Starsiss:
Hello,
I am trying to show markers for another kind of devices that i have stored in another database, but i get a 500 traccar/dataService error when trying to recover the data from that database. is there a way to create multiple entries in the conf file to connect to different databases? or I can connect to those DB whenever the user and password supplied in the conf file are correct for that other DB?( althoug the url db name does not match)

thanks,

tananaev:
No, current version of Traccar does not support working with multiple databases.

Starsiss:
I know that SQL server can make selects from another database if it is explicity defined in the FROM clause DB_NAME.dbo.TABLE_NAME and if the user and pwd provided have rights over that DB. I suppose there is nothing more to do than copy the table to traccar db.

thanks,

tananaev:
If it's the same instance of the database engine, then yes, you can probably specify database directly in the SQL statement.

Another option is to set up database replication.

Starsiss:
In the end, I copied the table to the Traccar Database, and worked from there, but what i tought it was a problem od FB connectivity is turning out a little more complex. I set my Class as entity, and I have declared it in the DataService class, interface and Impl. just as any other class, like devices or positions, but every time i get the 500 server error when trying to access to the data through Application.getDataService().METHOD_IMPLEMENTED()

This is the method in DataServiceImpl:
@Transactional    
@Override
    public List<Ascensor> getTodosAscensores()
    {
        EntityManager entitymanager = getSessionEntityManager();        
        List<Ascensor> ascensores = new LinkedList<Ascensor>();     
             String queryString = "SELECT a FROM Ascensor a";    
        TypedQuery<Ascensor> query = entitymanager.createQuery(queryString,Ascensor.class); ascensores = query.setMaxResults(100).getResultList();
             return ascensores;
    }

And my class Ascensor like this
@Entity
@Table(name = "Ascensor", indexes = { @Index(name = "PK_Ascensor", columnList = "Empresa,Codigo") })       

public class Ascensor implements Serializable {    
    private static final long serialVersionUID = 1;
// <editor-fold desc="ATTRIBUTES">
    @Id
    @Column(name = "Codigo", nullable = false, updatable = false, unique = true)
    private String Codigo;

I am totally at loss as I have revised all the code, but unable to find any discordances with other entities.

tananaev:
What is the actual error/exception on the server side?

Starsiss:
[...]

PR #5257, opened 2024-01-24
Update with build of modern and added distribution

hunterino:
This is more inline with the gradle build processs. Rather than the hybred build by gradle and shell currently used.

* It uses gradle to build traccar-web
* It uses gradle application plugin to add 
`gradle run`
* It uses gradle distribution plugin to add
`gradle distZip` which build a zip like the setup shell did, with the exception that you now have a versioned artifact name
* It's read to add maven repository publish of the jar artifact, should you choose.

NOTE: It has a pull request to traccar-web for the other parts.
See [traccar-web pull reqest #1215](https://github.com/traccar/traccar-web/pull/1215)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,17 +1,22 @@
 plugins {
     id "java"
     id "checkstyle"
+    id 'distribution'
+ //   id 'maven-publish'
+    id 'application'
     id "com.google.protobuf" version "0.9.4"
     id "org.kordamp.gradle.project-enforcer" version "0.13.0"
+    id "org.liquibase.gradle" version "2.2.0"
 }
 
 repositories {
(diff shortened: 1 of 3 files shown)


tananaev:
I'm not sure I understand the value. What is the purpose of making a zip file?

hunterino:
> I'm not sure I understand the value. What is the purpose of making a zip file?

Previously, the `setup/package.sh` when run as `package.sh 5.11 Other` would create a zip file. Which is used in docker containters ... that zip file can now be generated by `gradle distZip`

tananaev:
I would understand if it could completely replace packaging script, but just replacing one part doesn't make sense, I think.

PR #5268, opened 2024-02-18
Refactor: Remove duplicated code in get function (ServerResource.java)

MeysamRT76:
Eliminate duplicated code in get function to enhance maintainability and readability.

### DIFF
--- a/src/main/java/org/traccar/api/resource/ServerResource.java
+++ b/src/main/java/org/traccar/api/resource/ServerResource.java
@@ -93,15 +93,10 @@ public class ServerResource extends BaseResource {
         server.setOpenIdEnabled(openIdProvider != null);
         server.setOpenIdForce(openIdProvider != null && openIdProvider.getForce());
         User user = permissionsService.getUser(getUserId());
-        if (user != null) {
-            if (user.getAdministrator()) {
-                server.setStorageSpace(Log.getStorageSpace());
-            }
-        } else {
-            server.setNewServer(UserUtil.isEmpty(storage));
-        }
         if (user != null && user.getAdministrator()) {
             server.setStorageSpace(Log.getStorageSpace());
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/api/resource/ServerResource.java
```diff
@@ -93,15 +93,10 @@ public Server get() throws StorageException {
         server.setOpenIdEnabled(openIdProvider != null);
         server.setOpenIdForce(openIdProvider != null && openIdProvider.getForce());
         User user = permissionsService.getUser(getUserId());
-        if (user != null) {
-            if (user.getAdministrator()) {
-                server.setStorageSpace(Log.getStorageSpace());
-            }
-        } else {
-            server.setNewServer(UserUtil.isEmpty(storage));
-        }
```
tananaev, line 102: This was actually correct. What you did is incorrect because we want to check `user != null` only first.

MeysamRT76: In this case, when the user is not null, you're checking user != null twice along with user.getAdministrator().

The final if statement repeats a check that's already performed within the first if block. This might be unnecessary.

Actually, when the user is not null and is an administrator, server.setStorageSpace is set at line 98, and the same operation is performed at line 104.

tananaev: I understand what you're trying to say, but it seems like you didn't understand what I said. The statement you removed is actually the correct one. You should have removed the other one. There is a very good reason for the if statement to be nested like this.

MeysamRT76: I understand now. I will commit the changes as you said and keep the if statement nested.

tananaev:
Merged, thank you.

PR #5271, opened 2024-02-21
(Tlt2h) Fix for engine hours not being tracked when a packet has a null location

evanpenner:
This is a workaround for the issue where a device's ignition status may change while the device has no GPS signal.

An example of when this would be a problem is when a vehicle drives into a covered building and then turns the vehicle off. Traccar will stop logging positions because the device isn't sending any valid gps data. The issue occurs when looking at the summary report for engine hours, as Traccar will still be counting up even if the device is constantly sending AUTOLOW statuses. We can solve this problem by setting the fix time as the time when the packet is received.


tananaev:
This is not the right fix.

PR #5280, opened 2024-03-08
Huabao encoder Improvement.

Itheras:
Improvement to the HuabaoFrameEncoder. no modification of start and end frames . reordered handlers so computed ignition can be used calculate hours.


src/main/java/org/traccar/BasePipelineFactory.java
```diff
@@ -159,8 +159,8 @@ protected void initChannel(Channel channel) {
                 SpeedLimitHandler.class,
                 MotionHandler.class,
                 CopyAttributesHandler.class,
-                EngineHoursHandler.class,
                 ComputedAttributesHandler.class,
+                EngineHoursHandler.class,
```
tananaev, line 163: This is unrelated to Huabao

src/main/java/org/traccar/protocol/HuabaoFrameEncoder.java
```diff
@@ -23,22 +23,30 @@ public class HuabaoFrameEncoder extends MessageToByteEncoder<ByteBuf> {
...
-                    break;
-                case 0x7e:
-                    out.writeByte(0x7d);
-                    out.writeByte(0x02);
-                    break;
-                default:
-                    out.writeByte(b);
-                    break;
+            index++;
+            if (index > 1 && index < readableBytes) {
```
tananaev, line 32: There must be a cleared way to rewrite this

PR #5281, opened 2024-03-08
Update HuabaoFrameEncoder.java

Itheras:
Improvements to the HuabaoFrameEncoder


src/main/java/org/traccar/BasePipelineFactory.java
```diff
@@ -158,8 +158,8 @@ protected void initChannel(Channel channel) {
                 GeocoderHandler.class,
                 SpeedLimitHandler.class,
                 MotionHandler.class,
-                CopyAttributesHandler.class,
                 EngineHoursHandler.class,
+                CopyAttributesHandler.class,
```
tananaev, line 162: What is going on here again?

Itheras: Sorry let me correct this. Reordered incorrectly . Need to move the enginehour handler under computed attributes handler so calculated ignitions also work with hours

tananaev: This does not belong in this pull request.

PR #5285, opened 2024-03-15
Update SPO2 Key for Minifinder2

sahalghaf:


sahalghaf:
1. SPO2 is the Blood oxygen saturation level calculated by some devices belonging to Eview
2. Here's a snapshot of the document, I've attached the whole file below
Screenshot 2024-03-15 at 1 26 15 PM
3. I've updated this

PR #5289, opened 2024-03-29
Fix INF message on GV310LAU

jcardus:
include iccid

[GV310LAU @Track Air Interface Protocol V2.02.pdf](https://github.com/traccar/traccar/files/14797521/GV310LAU.%40Track.Air.Interface.Protocol.V2.02.pdf)

### DIFF
--- a/src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
@@ -207,7 +207,7 @@ public class Gl200TextProtocolDecoder extends BaseProtocolDecoder {
 
         position.set(Position.KEY_ICCID, parser.next());
         position.set(Position.KEY_RSSI, parser.nextInt());
-        
+
         if (protocolVersion.startsWith("6E")) { // GV310LAU
             position.set(Position.KEY_POWER, parser.nextDouble() / 1000); // odometer or external power
         } else {
--- a/src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
@@ -108,13 +108,13 @@ public class Gl200TextProtocolDecoder extends BaseProtocolDecoder {
     }
 
     private Position initPosition(Parser parser, Channel channel, SocketAddress remoteAddress) {
-if (parser.matches()) {
-        DeviceSession deviceSession = getDeviceSession(channel, remoteAddress, parser.next());
-        if (deviceSession != null) {
-            Position position = new Position(getProtocolName());
-            position.setDeviceId(deviceSession.getDeviceId());
-            return position;
-}
+        if (parser.matches()) {
+            DeviceSession deviceSession = getDeviceSession(channel, remoteAddress, parser.next());
+            if (deviceSession != null) {
+                Position position = new Position(getProtocolName());
+                position.setDeviceId(deviceSession.getDeviceId());
+                return position;
+            }
         }
         return null;
     }
--- a/src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
@@ -52,35 +52,35 @@ public class Gl200TextProtocolDecoder extends BaseProtocolDecoder {
 
     private final DateFormat dateFormat;
 
-    private static final HashMap<String, String> devices = new HashMap<String, String>() {{
-        put("52", "GL50");
-        put("55", "GL50B");
+    private static final HashMap<String, String> devices = new HashMap<String, String>() {{        
         put("02", "GL200");
         put("04", "GV200");
         put("06", "GV300");
         put("08", "GMT100");
         put("09", "GV50P"); // GV50 Plus
         put("0F", "GV55");
-        put("50", "GV55W");
         put("10", "GV55 LITE");
         put("11", "GL500");
         put("1A", "GL300");
         put("1F", "GV500");
-        put("5E", "GV500MAP");
         put("25", "GV300"); // New Version
-        put("35", "GV200"); // New Version
         put("27", "GV300W");
+        put("2C", "GL300W"); // New version
         put("2F", "GV55"); // New Version
         put("30", "GL300"); // New Version
+        put("35", "GV200"); // New Version
         put("36", "GV500"); // New Version
-        put("2C", "GL300W"); // New version
         put("3F", "GMT100"); // New version
-        put("F8", "GV800W");
         put("41", "GV75W");
-        put("FC", "GV600W");
+        put("50", "GV55W");
+        put("52", "GL50");
+        put("55", "GL50B");
+        put("5E", "GV500MAP");
         put("6E", "GV310LAU");
         put("C2", "GV600M");
         put("F1", "GV350M");
+        put("F8", "GV800W");
+        put("FC", "GV600W");
         put("802004", "GV58LAU");
         put("802005", "GV355CEU");
     }};
@@ -248,7 +248,8 @@ public class Gl200TextProtocolDecoder extends BaseProtocolDecoder {
         position.set(Position.KEY_ICCID, parser.next());
         position.set(Position.KEY_RSSI, parser.nextInt());
 
-        if (protocolVersion.startsWith("6E")) { // GV310LAU
+        String model = getDeviceModel(deviceSession, "", protocolVersion);
+        if (model.equals("GV310LAU")) {
             position.set(Position.KEY_POWER, parser.nextDouble() / 1000); // odometer or external power
         } else {
             parser.next(); // odometer or external power
@@ -263,7 +264,7 @@ public class Gl200TextProtocolDecoder extends BaseProtocolDecoder {
 
         position.set(Position.PREFIX_ADC + 1, parser.next());
         position.set(Position.PREFIX_ADC + 2, parser.next());
-        if (protocolVersion.startsWith("6E")) {
+        if (model.equals("GV310LAU")) {
             position.set(Position.PREFIX_ADC + 3, parser.next());
         }
 
--- a/src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
@@ -83,7 +83,7 @@ public class Gl200TextProtocolDecoder extends BaseProtocolDecoder {
 
     private boolean ignoreFixTime;
 
-    private final DateFormat dateFormat;    
+    private final DateFormat dateFormat;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -197,9 +205,14 @@ private Object decodeInf(Channel channel, SocketAddress remoteAddress, String se
                 break;
         }
 
+        position.set(Position.KEY_ICCID, parser.next());
         position.set(Position.KEY_RSSI, parser.nextInt());
 
-        parser.next(); // odometer or external power
+        if (protocolVersion.startsWith("6E")) { // GV310LAU
```
tananaev, line 211: Are you sure there are no conflicts with other devices?

jcardus: Based on the existing tests and the documentation there aren't. I think it's a better option than using an attribute.

This also confirms it:
https://github.com/jaayesta/queclink-parser/blob/master/src/utils.js

tananaev: I don't really want to have two different ways of device identification. Do you want to update other code to use protocol version as well?

jcardus: you mean for GV355CEU?

tananaev: There are several places where we use model name.

jcardus: yes I see. Ok, I'll update for all.

tananaev: Also, the plan is to slowly migrate this decoder away from regex patterns. If you need to make big changes, it might make sense to migrate as well.

jcardus: I think we can do it without big changes. 

I'll make a commit soon to see if I'm on the right direction.

jcardus: should I leave the device name and  attribute as a fallback when we can't identify the model?

jcardus:
Can you take a look?

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -51,6 +52,39 @@ public class Gl200TextProtocolDecoder extends BaseProtocolDecoder {
...
+        put("04", "GV200");
+        put("06", "GV300");
+        put("08", "GMT100");
+        put("09", "GV50P"); // GV50 Plus
+        put("0F", "GV55");
+        put("10", "GV55 LITE");
+        put("11", "GL500");
+        put("1A", "GL300");
+        put("1F", "GV500");
+        put("25", "GV300"); // New Version
```
tananaev, line 66: I don't think we need these new version comments.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -51,6 +52,39 @@ public class Gl200TextProtocolDecoder extends BaseProtocolDecoder {
 
     private final DateFormat dateFormat;
 
+    private static final HashMap<String, String> DEVICE_MODELS = new HashMap<String, String>() {{
```
tananaev, line 55: Statics should be above instance variables.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -122,12 +162,12 @@ private Position decodeAck(Channel channel, SocketAddress remoteAddress, String[
 
     private static final Pattern PATTERN_INF = new PatternBuilder()
             .text("+").expression("(?:RESP|BUFF):GTINF,")
-            .expression("(?:.{6}|.{10})?,")      // protocol version
+            .expression("(.{6}|.{10})?,")      // protocol version
```
tananaev, line 165: I believe we should do this for all types, not just this.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -148,6 +188,7 @@ private Position decodeAck(Channel channel, SocketAddress remoteAddress, String[
             .expression("(?:[01])?,").optional() // pin15 mode
             .number("(d+)?,")                    // adc1
             .number("(d+)?,").optional()         // adc2
+            .number("(d+)?,").optional()         // adc3
```
tananaev, line 191: I'm pretty sure this would break some use cases. Do we have tests for it?

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -163,10 +204,17 @@ private Position decodeAck(Channel channel, SocketAddress remoteAddress, String[
 
     private Object decodeInf(Channel channel, SocketAddress remoteAddress, String sentence) {
         Parser parser = new Parser(PATTERN_INF, sentence);
-        Position position = initPosition(parser, channel, remoteAddress);
```
tananaev, line 166: We need to update this common method instead of doing a one off thing for INF type.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
jcardus: ok, but the patterns are not.

tananaev: They used to be, but it made sense to make an except. Note again that the plan is to get rid of patterns eventually.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
jcardus: Yes, the last one.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
jcardus: ok, should we put the model in the position?

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
tananaev: I think you misunderstood what I said. I'm asking for test cases for old stuff to make sure it doesn't break any case where adc3 is missing.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
tananaev: No. There must be a better solution.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
jcardus: for that I was relying on the existing tests.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
jcardus: ok, I think I need to remove parser.matches() from it right?

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
jcardus: The other types are not using it.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
jcardus: can we put it in the deviceSession?

tananaev: I would avoid touching anything outside of this decoder, unless you want to go and update all other decoders.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
tananaev: We need to make sure it's covered.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
tananaev: The logic should be common.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -42,11 +42,45 @@
...
+        put("55", "GL50B");
+        put("5E", "GV500MAP");
+        put("6E", "GV310LAU");
+        put("C2", "GV600M");
+        put("F1", "GV350M");
+        put("F8", "GV800W");
+        put("FC", "GV600W");
+        put("802004", "GV58LAU");
+        put("802005", "GV355CEU");
+    }};
```
tananaev, line 82: Why are there two brackets?

tananaev:
I'm going to merge it for now because I need to do some other work on this protocol.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
jcardus: it's a little tricky, the method initialises the parser, gets the imei (which comes after the model) and returns a position. I don't see how we can refactor this.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
jcardus: Double brace initialisation so we can use the initialiser block

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -210,6 +264,9 @@ private Object decodeInf(Channel channel, SocketAddress remoteAddress, String se
 
         position.set(Position.PREFIX_ADC + 1, parser.next());
         position.set(Position.PREFIX_ADC + 2, parser.next());
+        if (model.equals("GV310LAU")) {
+            position.set(Position.PREFIX_ADC + 3, parser.next());
```
tananaev, line 268: Looks like this actually breaks the implementation. The group is always present in your regex, but you're only reading it when the model is "GV310LAU". It means if the model is not "GV310LAU", the index gets messed up and everything after it is not read correctly.

jcardus: I thought .optional() in the regex would work in those cases.

tananaev: How would optional know what to skip?

PR #5290, opened 2024-03-30
refactor initPosition

jcardus:


jcardus:
@tananaev can you take a look?

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -1335,8 +1338,9 @@ private Object decodeDtt(Channel channel, SocketAddress remoteAddress, String se
 
     private Object decodeBaa(Channel channel, SocketAddress remoteAddress, String sentence) {
         Parser parser = new Parser(PATTERN_BAA, sentence);
-        Position position = initPosition(parser, channel, remoteAddress);
-        if (position == null) {
+        Position position = new Position();
+        String model = initPosition(parser, channel, remoteAddress, position);
+        if (model == null) {
```
tananaev, line 1343: Position should always be created with the protocol argument, except for test cases.

And `model == null` check doesn't really make sense to me.

tananaev (commented):
Your previous PR is already merged. I think we don't need this for now.

PR #5309, opened 2024-04-25
# feat: Add a new protocol, zr, listening on port 5257.

keDaYao:
[Trak-iot.protocol.TC04-9.20220817.docx](https://github.com/traccar/traccar/files/15105822/Trak-iot.protocol.TC04-9.20220817.docx)
Support device - "TC-01、TC-02、TC-03、TC-04" | Protocol - zr Protocol | Port - 5257


tananaev:
It doesn't look like all of the comments from last PR were addressed.

tananaev:
Also maybe it would be easier to split it into several PRs. For example encoder can be separate from the main decoding.

keDaYao:
Hi, I think this is difficult to split because when the server receives a message from the device, it must reply with different response messages, which requires the encoder to encode the corresponding response messages. Therefore, the encoder code cannot be split.

tananaev:
And where exactly are you using encoder for responses? I don't see it.

keDaYao:
In ZrProtocolDecoder.java, at lines 89, 92, 96,and 99.

keDaYao:
Could you please re-label the suggestions for code modifications? I'm not sure what needs to be changed, and I would greatly appreciate it.

keDaYao:
Yes, I wrote the code.When you mentioned encoder, are you referring to ZrProtocolEncoder.java?I just deleted that part of the file.

tananaev:
Thanks. Let me know once all the other issues are resolved, so we can review it again.

PR #5313, opened 2024-04-30
Update TrvProtocolDecoder.java

Nikutrax:
Correction from CP01 to AP01

### DIFF
--- a/src/main/java/org/traccar/protocol/TrvProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/TrvProtocolDecoder.java
@@ -78,9 +78,9 @@ public class TrvProtocolDecoder extends BaseProtocolDecoder {
             .any()
             .compile();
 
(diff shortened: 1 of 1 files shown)


tananaev:
What is this?

tananaev:
You have to provide more context if you want a send a PR. And if anything it should probably be an addition, not a change.

Nikutrax:
position.set("bloodSugar", Double.parseDouble(values[5]));

insteat of

position.set("bloodSugar", Integer.parseInt(values[5]));

Nikutrax:
because i think the documention was fail and should so

Nikutrax:
> position.set("bloodSugar", Double.parseDouble(values[5]));
> 
> insteat of
> 
> position.set("bloodSugar", Integer.parseInt(values[5]));

on line 348

tananaev:
This is not what we talked about.

PR #5316, opened 2024-05-05
add websocket external access

mr-wolf-gb:
Websocket can be accessed from external Hosts (App):
1. Using Session ID
The session ID can be retrieved as cookies from the response on these API endpoints :

- [api/session](https://www.traccar.org/api-reference/#tag/Session/paths/~1session/post) | Create a new Session
- [api/session?token=](https://www.traccar.org/api-reference/#tag/Session/paths/~1session/get) | Fetch Session information
- [api/devices](https://www.traccar.org/api-reference/#tag/Devices/paths/~1devices/get) | Fetch a list of Devices

```js
const socket = new WebSocket("./api/socket?session=node01d8bcd8o4su6u1ug70qdrena0i1");
socket.onerror = (error) => {
  console.log("socket error: ", error);
};
socket.onmessage = function (event) {
  console.log("socket message : ", data);
};
```
2- Using user Access Token

```js
const socket = new WebSocket("./api/socket?token=SDBGAiEA4SC67Qk5lrCsB2I53EDp5gAR1uips64FRn6W0Dt0jrMCIQDnZ.....");
socket.onerror = (error) => {
  console.log("socket error: ", error);
};
socket.onmessage = function (event) {
  console.log("socket message : ", data);
};
```
3- Add `api/session/check-sid?sid=[SESSION_ID]` to check if the session is still active or not


tananaev:
There are 20+ commits and most of the changes are unrelated to WebSocket.

PR #5320, opened 2024-05-16
Allow Date class in the JEXL Sandbox.

memesaregood1:
Following #5036, this PR adds the Date class to the JEXL sandbox whitelist.


tananaev:
Merged, thanks.

PR #5347, opened 2024-06-15
Photon Geocoder

jinzo:
This is a basic implementation of the open source Photon Geocoder for reverse geocoding. There isn't much documentation, but [the source](https://github.com/komoot/photon/blob/master/src/main/java/de/komoot/photon/query/ReverseRequestFactory.java)  reveals that only the `lang` optional parameter is useful for Traccar.
After some tests, it seems that the fields are mostly present always and there is little to no deviation (for example, Settlement is always in key `city` ect.)

 * https://photon.komoot.io/
 * https://github.com/komoot/photon

### DIFF
--- a/src/main/java/org/traccar/geocoder/GeocodeJSONGeocoder.java
+++ b/src/main/java/org/traccar/geocoder/GeocodeJSONGeocoder.java
@@ -20,7 +20,7 @@ import jakarta.json.JsonArray;
 import jakarta.json.JsonObject;
 import jakarta.ws.rs.client.Client;
 
-public class PhotonGeocoder extends JsonGeocoder {
+public class GeocodeJSONGeocoder extends JsonGeocoder {
 
     private static String formatUrl(String url, String key, String language) {
         if (url == null) {
@@ -36,7 +36,7 @@ public class PhotonGeocoder extends JsonGeocoder {
         return url;
     }
 
-    public PhotonGeocoder(
+    public GeocodeJSONGeocoder(
             Client client, String url, String key, String language, int cacheSize, AddressFormat addressFormat) {
         super(client, formatUrl(url, key, language), cacheSize, addressFormat);
     }
@@ -48,6 +48,9 @@ public class PhotonGeocoder extends JsonGeocoder {
             Address address = new Address();
             JsonObject properties = features.getJsonObject(0).getJsonObject("properties");
 
+            if (properties.containsKey("label")) {
+                address.setFormattedAddress(properties.getString("label"));
+            }
             if (properties.containsKey("housenumber")) {
                 address.setHouse(properties.getString("housenumber"));
             }
@@ -58,7 +61,7 @@ public class PhotonGeocoder extends JsonGeocoder {
                 address.setSettlement(properties.getString("city"));
             }
             if (properties.containsKey("district")) {
-                address.setState(properties.getString("district"));
+                address.setDistrict(properties.getString("district"));
             }
             if (properties.containsKey("state")) {
                 address.setState(properties.getString("state"));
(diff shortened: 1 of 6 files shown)


src/main/java/org/traccar/geocoder/PhotonGeocoder.java
```diff
@@ -0,0 +1,78 @@
...
+        return url;
+    }
+
+    public PhotonGeocoder(
+            Client client, String url, String key, String language, int cacheSize, AddressFormat addressFormat) {
+        super(client, formatUrl(url, key, language), cacheSize, addressFormat);
+    }
+
+    @Override
+    public Address parseAddress(JsonObject json) {
```
tananaev, line 45: The format seems to be similar to `BanGeocoder`. Can we combine the two?

jinzo: Oh damn, I actually flicked thorough some of the Geocoders, looking if anything similar is supported but did not see the Ban one unfortunately. This seems to be GeoJSON 'standard' format (as Nominatim also has this format option).
The only clash I see is the `name` key, which contains different type of data. But we could differentiate this by `osm_id` key (as it looks like the only such difference). I will take a deeper look in a few days and get back to you if I find any other 'ugly' differences.

jinzo: Ok, I did some more research and the `name` difference is avoidable (as we get separate data, and can store that). The only remaining problem is that Ban does not return the country. So I was thinking of creating a [GeocodeJSON](https://github.com/geocoders/geocodejson-spec) geocoder that can be used with the spec adhering geocoders and then a rework of Ban geocoder that extends GeocodeJSONGeocoder and just 'adds' the hardcoded url & countrycode. What do you think?

jinzo:
So I refactored the code as in the previous comment. Renamed Photon Geocoder to "standard" [GeocodeJSON](https://github.com/geocoders/geocodejson-spec) and refactored the Ban Geocoder based on it because Ban does not provide the country with it's output, and every other solution I thought about was worse. 
I changed the test (and Ban geocoder behaviour) in two ways:
 * `name` is no longer used as a `Street` name as this was plain wrong. `name` is `housenumber` + `street` in Ban, therefore it is correct to store/use them separately.
 * `context` is no longer used as `District`, as this looks plain wrong too and we now store `district` key that is provided with Ban in `District`. The context is now discarded as I don't think this can be sanely fit into the current system.

This changes make Ban geocoder more in line with what other Geocoders store in `Address` but still does not break any actual running systems.

src/main/java/org/traccar/geocoder/GeocodeJSONGeocoder.java
```diff
@@ -0,0 +1,81 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.geocoder;
+
+import jakarta.json.JsonArray;
+import jakarta.json.JsonObject;
+import jakarta.ws.rs.client.Client;
+
+public class GeocodeJSONGeocoder extends JsonGeocoder {
```
tananaev, line 23: Names should be camelcase, so `GeocodeJsonGeocoder`

tananaev:
Merged, thank you.

tananaev:
@jinzo looks like something got broken:

https://www.traccar.org/forums/topic/geocoder/

Can you please fix it.

jinzo:
Currently on vacations, but will take a look as soon as I have time.

jinzo:
So this was one of the 'expected' problems, where the previous way of doing things (where 'street' contained also the 'house' number) changed and caused problems (because we are now storing only street name in 'street' and store house number in 'house' in this ban geocoder). So changing the geocoder.format solved their problem. And it seems the encoding problem was just some copy paste error or something, as they confirmed everything works ok after format change.

tananaev:
Sounds good.

Issue #5255, opened 2024-01-24
Add MaxLifeTime configuration for HikariCP to traccar.xml


ghhv:
**Is your feature request related to a problem? Please describe.**
When using Traccar with Docker Swarm, there is a known issue with TCP timeouts between containers of 15mins. This affects other Java based apps like Moby and some Kubernetes installations and any database connection like Postgresql or Mysql. 
In Traccar, this 15min limit is less than the default HikariCP MaxLifeTime value of 30mins and therefore installations using Docker Swarm constantly get this warning:
```WARN: HikariPool-1 - Failed to validate connection com.mysql.cj.jdbc.ConnectionImpl@NNN (No operations allowed after connection closed.). Possibly consider using a shorter maxLifetime value.```
More background below.

**Describe the solution you'd like**
Add the option to configure HikariCP MaxLifetime to traccar.xml configuration like so:
```
<entry key='database.maxLifeTime'>[new_value]</entry>
```
with this added to `Keys.java`
```
 public static final ConfigKey<Integer> DATABASE_MAX_LIFE_TIME = new IntegerConfigKey(
            "database.maxLifeTime",
            List.of(KeyType.CONFIG));
```
And this added to `DatabaseModule.java`
```
 int maxLifeTime= config.getInteger(Keys.DATABASE_MAX_LIFE_TIME );
        if (maxLifeTime!= 0) {
            hikariConfig.setMaxLifetime(maxLifeTime);
        }
```
And anything else I'm not aware of to make this work. See IdleTimeOut below.

**Describe alternatives you've considered**
I've tested changing every other timeout without any resolution. 
A suggested workaround from [here ](https://github.com/moby/moby/issues/37822#issuecomment-422618376) does not work for Traccar as you need to expose specific ports which is incompatible
` EndpointSpec: port published with ingress mode can't be used with dnsrr mode`

**Additional context**
Other (mostly non-Traccar) references to this issue are listed below:
[docker swarm configurations there is an internal TCP idle timeout at 15 minutes](https://github.com/MariaDB/mariadb-docker/issues/113#issuecomment-1243067725)
[[Docker Swarm: TCP connection between containers disconnected after being idle for about 900 seconds](https://github.com/moby/moby/issues/37822#top)](https://github.com/moby/moby/issues/37822)
[...]

PR #5348, opened 2024-06-16
Add maxLifetime for hikariConfig

sfreydin:
Fix for [issue](https://github.com/traccar/traccar/issues/5255)

### DIFF
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -543,6 +543,13 @@ public final class Keys {
             List.of(KeyType.CONFIG));
 
     /**
+     * This property controls the maximum lifetime in miliseconds of a connection in the pool.
+     */
+    public static final ConfigKey<Integer> DATABASE_MAX_LIFETIME = new IntegerConfigKey(
+            "database.maxLifetime",
+            List.of(KeyType.CONFIG));
+ 
+    /**
      * Device limit for self registered users. Default value is -1, which indicates no limit.
      */
     public static final ConfigKey<Integer> USERS_DEFAULT_DEVICE_LIMIT = new IntegerConfigKey(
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/config/Keys.java
```diff
@@ -542,6 +542,13 @@ private Keys() {
             "database.saveEmpty",
             List.of(KeyType.CONFIG));
 
+    /**
+     * This property controls the maximum lifetime in miliseconds of a connection in the pool.
+     */
+    public static final ConfigKey<Integer> DATABASE_MAX_LIFETIME = new IntegerConfigKey(
+            "database.maxLifetime",
+            List.of(KeyType.CONFIG),60000);
```
tananaev, line 550: What is the default value? Can you please link to the documentation.

sfreydin: @tananaev  yep, your right, it's a little bit different value according to [docs](https://github.com/brettwooldridge/HikariCP#gear-configuration-knobs-baby). Fixed.

src/main/java/org/traccar/config/Keys.java
```diff
@@ -542,6 +542,13 @@ private Keys() {
             "database.saveEmpty",
             List.of(KeyType.CONFIG));
 
+    /**
+     * This property controls the maximum lifetime in miliseconds of a connection in the pool.
+     */
+    public static final ConfigKey<Integer> DATABASE_MAX_LIFETIME = new IntegerConfigKey(
```
tananaev, line 548: Can we just do what we do with the `DATABASE_MAX_POOL_SIZE`? Basically only set if the value is not zero.

sfreydin: `DATABASE_MAX_LIFETIME` defines max lifetime for the connection. And I need to control this parameter on the app side.

tananaev: I don't think you understood my comment. I'm saying that we can avoid hardcoding the default value.

sfreydin: done

jinzo:
@sfreydin - without involving @tananaev, you will have to fix the build/checkstyle issues before it can be merged. P.S: if you run the build it will do the same checks locally.

rahuljawale:
Waiting eagerly for this fix/PR to go through.

tananaev:
Once the issues are resolved, we can consider it.

sfreydin:
@tananaev @jinzo fixed

tananaev:
Doesn't look like you fixed it.

PR #5351, opened 2024-06-19
Add functionality to disable forwarding per device

mivtdole:
This is the rewrite of and old pull request

### DIFF
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -865,6 +865,18 @@ public final class Keys {
             List.of(KeyType.CONFIG));
 
     /**
+     * Forwarding attribute to specify if per device position forwarding needs to be done.
+     * When true per device the allowForwarding attribute needs to be specified if positions
+     *  for the device need to be forwarded.
+     * When false the old behaviour is used and for every device positions are forwarded
+     *  independent of the device attribute
+     */
+    public static final ConfigKey<Boolean> FORWARD_ENABLE_PER_DEVICE = new BooleanConfigKey(
+            "forward.enablePerDevice",
+            List.of(KeyType.CONFIG),
+            false);
+
+    /**
      * Position forwarding format. Available options are "url", "json" and "kafka". Default is "url".
      */
     public static final ConfigKey<String> FORWARD_TYPE = new StringConfigKey(
@@ -940,6 +952,18 @@ public final class Keys {
             "forward.retry.limit",
             List.of(KeyType.CONFIG),
             100);
+
+    /**
+     * Forwarding attribute to specify if per device event forwarding needs to be done.
+     * When true per device the allowForwarding attribute needs to be specified if events
+     *  for the device need to be forwarded.
+     * When false the old behaviour is used and for every device events are forwarded
+     *  independent of the device attribute
+     */
+    public static final ConfigKey<Boolean> EVENT_FORWARD_ENABLE_PER_DEVICE = new BooleanConfigKey(
+            "event.forward.enablePerDevice",
+            List.of(KeyType.CONFIG),
+            false);
 
     /**
      * Events forwarding format. Available options are "json" and "kafka". Default is "json".
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -867,9 +867,9 @@ public final class Keys {
     /**
      * Forwarding attribute to specify if per device position forwarding needs to be done.
      * When true per device the allowForwarding attribute needs to be specified if positions
-     *  for the device need to be forwarded.
+     * for the device need to be forwarded.
      * When false the old behaviour is used and for every device positions are forwarded
-     *  independent of the device attribute
+     * independent of the device attribute
      */
     public static final ConfigKey<Boolean> FORWARD_ENABLE_PER_DEVICE = new BooleanConfigKey(
             "forward.enablePerDevice",
@@ -956,9 +956,9 @@ public final class Keys {
     /**
      * Forwarding attribute to specify if per device event forwarding needs to be done.
      * When true per device the allowForwarding attribute needs to be specified if events
-     *  for the device need to be forwarded.
+     * for the device need to be forwarded.
      * When false the old behaviour is used and for every device events are forwarded
-     *  independent of the device attribute
+     * independent of the device attribute
      */
     public static final ConfigKey<Boolean> EVENT_FORWARD_ENABLE_PER_DEVICE = new BooleanConfigKey(
             "event.forward.enablePerDevice",
--- a/src/main/java/org/traccar/database/NotificationManager.java
+++ b/src/main/java/org/traccar/database/NotificationManager.java
@@ -53,6 +53,8 @@ public class NotificationManager {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NotificationManager.class);
 
+    private static final String ATTRIBUTE_DEVICE_ALLOW_FORWARDING = "deviceAllowEventForwarding";
+
     private final Storage storage;
     private final CacheManager cacheManager;
     private final EventForwarder eventForwarder;
@@ -62,6 +64,8 @@ public class NotificationManager {
     private final boolean geocodeOnRequest;
     private final long timeThreshold;
     private final Set<Long> blockedUsers = new HashSet<>();
+
+    private final boolean perDevice;
 
     @Inject
     public NotificationManager(
@@ -80,6 +84,8 @@ public class NotificationManager {
                 blockedUsers.add(Long.parseLong(userIdString));
             }
         }
+
+        this.perDevice = config.getBoolean(Keys.EVENT_FORWARD_ENABLE_PER_DEVICE);
     }
 
     private void updateEvent(Event event, Position position) {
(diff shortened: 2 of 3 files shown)


tananaev:
Where's the old pull request? Provide a link.

mivtdole:
See closed pull request #4778

tananaev:
What is the use case for this? It looks like you originally has "allow forwarding". That actually makes more sense, but probably need another configuration parameters for it. We should also probably do the same thing for event forwarding.

mivtdole:
I created this version because of your comment "That's why I'm thinking that the attribute logic should be inversed. You add it to disable forwarding." in the previous pull request.

tananaev:
I know what I said, but I said it because you only had one parameter, which would break existing logic. But if we add another configuration option, it would be fine.

mivtdole:
> I know what I said, but I said it because you only had one parameter, which would break existing logic. But if we add another configuration option, it would be fine.

What do you mean by other configuration option?

tananaev:
I mean a config parameter that will enable forwarding only for marked devices.

mivtdole:
I will make an updated version with a config parameter

mivtdole:
Do you want a new pull request or an update to this one?

tananaev:
Please use the same pull request. That's always the preference, so we have full conversation here.

mivtdole:
Just to be clear:

- We want a new configuration option with which the system manager can specify if per device forwarding needs to be enabled? Shall we call this configuration option "forward.enablePerDevice" and make it boolean. When this option is missing or false the "old" normal behavior of data for every device is forwarded. Not looking at any forwarding attribute for the device.
- When the configuration option is true. If the user wants to forward data for a device they need to add the attribute "allowForwarding" for the device.

tananaev:
Yes, correct.

src/main/java/org/traccar/config/Keys.java
```diff
@@ -864,6 +864,18 @@ private Keys() {
             "server.forward",
             List.of(KeyType.CONFIG));
 
+    /**
+     * Forwarding attribute to specify if per device position forwarding needs to be done.
+     * When true per device the allowForwarding attribute needs to be specified if positions
+     *  for the device need to be forwarded.
```
tananaev, line 870: Seems to be wrong indentation here. Please fix all the comments.

src/main/java/org/traccar/config/Keys.java
```diff
@@ -864,6 +864,18 @@ private Keys() {
...
+     * When false the old behaviour is used and for every device positions are forwarded
+     *  independent of the device attribute
+     */
+    public static final ConfigKey<Boolean> FORWARD_ENABLE_PER_DEVICE = new BooleanConfigKey(
+            "forward.enablePerDevice",
+            List.of(KeyType.CONFIG),
+            false);
```
tananaev, line 877: I believe false is the default. You don't need to specify it.

src/main/java/org/traccar/config/Keys.java
```diff
@@ -941,6 +953,18 @@ private Keys() {
...
+     * Forwarding attribute to specify if per device event forwarding needs to be done.
+     * When true per device the allowForwarding attribute needs to be specified if events
+     *  for the device need to be forwarded.
+     * When false the old behaviour is used and for every device events are forwarded
+     *  independent of the device attribute
+     */
+    public static final ConfigKey<Boolean> EVENT_FORWARD_ENABLE_PER_DEVICE = new BooleanConfigKey(
+            "event.forward.enablePerDevice",
+            List.of(KeyType.CONFIG),
+            false);
```
tananaev, line 966: Same here.

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -63,6 +65,8 @@ public class NotificationManager {
     private final long timeThreshold;
     private final Set<Long> blockedUsers = new HashSet<>();
 
+    private final boolean perDevice;
```
tananaev, line 68: Let's call it `enablePerDevice`

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -80,6 +84,8 @@ public NotificationManager(
                 blockedUsers.add(Long.parseLong(userIdString));
             }
         }
+
+        this.perDevice = config.getBoolean(Keys.EVENT_FORWARD_ENABLE_PER_DEVICE);
```
tananaev, line 88: Unnecessary `this`.

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -149,21 +155,30 @@ private void updateEvent(Event event, Position position) {
...
-            }
-            if (event.getMaintenanceId() != 0) {
-                eventData.setMaintenance(cacheManager.getObject(Maintenance.class, event.getMaintenanceId()));
-            }
-            eventForwarder.forward(eventData, (success, throwable) -> {
-                if (!success) {
-                    LOGGER.warn("Event forwarding failed", throwable);
+            Device device = cacheManager.getObject(Device.class, event.getDeviceId());
+
+            Object allowForwarding = device.getAttributes().get(ATTRIBUTE_DEVICE_ALLOW_FORWARDING);
```
tananaev, line 160: Why is this an object? Use `getBoolean`.

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -149,21 +155,30 @@ private void updateEvent(Event event, Position position) {
...
+            if ((!this.perDevice)
+                || ((this.perDevice) && (allowForwarding != null) && ((Boolean) allowForwarding))) {
```
tananaev, line 162: Too many unnecessary brackets. Please cleanup.

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -53,6 +53,8 @@ public class NotificationManager {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NotificationManager.class);
 
+    private static final String ATTRIBUTE_DEVICE_ALLOW_FORWARDING = "deviceAllowEventForwarding";
```
tananaev, line 56: Let's rename to "enableEventForwarding".

src/main/java/org/traccar/handler/PositionForwardingHandler.java
```diff
@@ -38,6 +38,8 @@ public class PositionForwardingHandler extends BasePositionHandler {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(PositionForwardingHandler.class);
 
+    private static final String ATTRIBUTE_DEVICE_ALLOW_FORWARDING = "deviceAllowPositionForwarding";
```
tananaev, line 41: allowPositionForwarding

src/main/java/org/traccar/handler/PositionForwardingHandler.java
```diff
@@ -62,6 +65,7 @@ public PositionForwardingHandler(
         this.retryDelay = config.getInteger(Keys.FORWARD_RETRY_DELAY);
         this.retryCount = config.getInteger(Keys.FORWARD_RETRY_COUNT);
         this.retryLimit = config.getInteger(Keys.FORWARD_RETRY_LIMIT);
+        this.perDevice = config.getBoolean(Keys.FORWARD_ENABLE_PER_DEVICE);
```
tananaev, line 68: Let's remove unnecessary `this` for all of the config parameters here

src/main/java/org/traccar/handler/PositionForwardingHandler.java
```diff
@@ -126,10 +130,19 @@ public void run(Timeout timeout) {
...
+            if ((!this.perDevice)
+                || ((this.perDevice) && (allowForwarding != null) && ((Boolean) allowForwarding))) {
+                PositionData positionData = new PositionData();
+                positionData.setPosition(position);
+                positionData.setDevice(device);
+                new AsyncRequestAndCallback(positionData).send();
+            } else {
+                LOGGER.debug("Not forwarding position for device '{}' as forwarding is disabled for this device.",
+                    device.getName());
+            }
```
tananaev, line 145: Same comments as the other handler

src/main/java/org/traccar/database/NotificationManager.java
mivtdole: Changed it to 

```
boolean allowForwarding = (boolean) device.getAttributes().get(ATTRIBUTE_DEVICE_ALLOW_FORWARDING);
```

src/main/java/org/traccar/database/NotificationManager.java
mivtdole: Are you sure as all other attributes start with deviceXxxx?

tananaev: Yes, I am sure.

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -149,21 +155,28 @@ private void updateEvent(Event event, Position position) {
...
-            }
-            if (event.getMaintenanceId() != 0) {
-                eventData.setMaintenance(cacheManager.getObject(Maintenance.class, event.getMaintenanceId()));
-            }
-            eventForwarder.forward(eventData, (success, throwable) -> {
-                if (!success) {
-                    LOGGER.warn("Event forwarding failed", throwable);
+            Device device = cacheManager.getObject(Device.class, event.getDeviceId());
+
+            boolean allowForwarding = (boolean) device.getAttributes().get(ATTRIBUTE_DEVICE_ALLOW_FORWARDING);
```
tananaev, line 160: There should be a `getBoolean` you can use.

src/main/java/org/traccar/handler/PositionForwardingHandler.java
```diff
@@ -126,10 +130,17 @@ public void run(Timeout timeout) {
     @Override
     public void handlePosition(Position position, Callback callback) {
         if (positionForwarder != null) {
-            PositionData positionData = new PositionData();
-            positionData.setPosition(position);
-            positionData.setDevice(cacheManager.getObject(Device.class, position.getDeviceId()));
-            new AsyncRequestAndCallback(positionData).send();
+            Device device = cacheManager.getObject(Device.class, position.getDeviceId());
+
+            boolean allowForwarding = (boolean) device.getAttributes().get(ATTRIBUTE_DEVICE_ALLOW_FORWARDING);
```
tananaev, line 135: Same here.

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -149,21 +155,28 @@ private void updateEvent(Event event, Position position) {
...
+            if (!enablePerDevice || (enablePerDevice && allowForwarding)) {
```
tananaev, line 161: This check is redundant. You can do `!enablePerDevice || allowForwarding`.

src/main/java/org/traccar/handler/PositionForwardingHandler.java
```diff
@@ -126,10 +130,17 @@ public void run(Timeout timeout) {
...
+            if (!enablePerDevice || (enablePerDevice && allowForwarding)) {
```
tananaev, line 136: Same here

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -149,21 +155,28 @@ private void updateEvent(Event event, Position position) {
...
+                if (event.getMaintenanceId() != 0) {
+                    eventData.setMaintenance(cacheManager.getObject(Maintenance.class, event.getMaintenanceId()));
+                }
+                eventForwarder.forward(eventData, (success, throwable) -> {
+                    if (!success) {
+                        LOGGER.warn("Event forwarding failed", throwable);
+                    }
+                });
+            } else {
+                LOGGER.debug("Forwarding events not enabled for {}", event.getDeviceId());
```
tananaev, line 178: It should be unique id, not device id.

src/main/java/org/traccar/handler/PositionForwardingHandler.java
```diff
@@ -126,10 +130,17 @@ public void run(Timeout timeout) {
...
+                PositionData positionData = new PositionData();
+                positionData.setPosition(position);
+                positionData.setDevice(device);
+                new AsyncRequestAndCallback(positionData).send();
+            } else {
+                LOGGER.debug("Forwarding position not enabled for {}.", position.getDeviceId());
```
tananaev, line 142: Same here.

tananaev:
Closing as abandoned.

PR #5354, opened 2024-06-20
Add Amap geocoder and coordinate util

mmzhi:
In China, geolocation encoding is commonly dominated by the use of Gaode Maps(Amap) and Baidu Maps. Other internationally used platforms are not as accurate. Moreover, China has a special system of geographic coordinates(GCJ-02). This merge request implements the geolocation encoding implementation of the Gaode map(Amap).

For this merge request I added two class, CoordinateUtil and AmapGeocoder.
CoordinateUtil is mainly used to convert WGS84 to GCJ-02 coordinates, this code reference: https://github.com/dromara/hutool/blob/v5-master/hutool-core/src/main/java/cn/hutool/core/util/CoordinateUtil.java
AmapGeocoder implements the interpretation of geographic coordinates of Gaode map(Amap).


mmzhi:
Re-worked the code to the code specification and the CI on github has been passed.

src/main/java/org/traccar/geocoder/AmapGeocoder.java
```diff
@@ -0,0 +1,98 @@
...
+                    }
+                }
+                address.setHouse(houseName);
+
+                JsonObject street = addressComponent.getJsonObject("streetNumber");
+                address.setStreet(getJsonNode(street, "street"));
+
+                address.setSuburb(getJsonNode(addressComponent, "township"));
+                address.setSettlement(getJsonNode(addressComponent, "district"));
+                address.setDistrict(getJsonNode(addressComponent, "city"));
```
tananaev, line 65: This doesn't look right. Can you please provide documentation for the API.

src/main/java/org/traccar/geocoder/AmapGeocoder.java
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright 2012 - 2017 Anton Tananaev (anton@traccar.org)
```
tananaev, line 2: Please fix the header.

src/main/java/org/traccar/helper/CoordinateUtil.java
```diff
@@ -0,0 +1,247 @@
...
+        ret += (20.0 * Math.sin(latitude * PI) + 40.0 * Math.sin(latitude / 3.0 * PI)) * 2.0 / 3.0;
+        ret += (160.0 * Math.sin(latitude / 12.0 * PI) + 320 * Math.sin(latitude * PI / 30.0)) * 2.0 / 3.0;
+        return ret;
+    }
+
+
+    /**
+     * Coordinate latitude and longitude
+     */
+    public static class Coordinate implements Serializable {
```
tananaev, line 147: I think we can just use `Pair` class instead of this.

src/main/java/org/traccar/helper/CoordinateUtil.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /**
+     * Calculate latitude coordinates
+     *
+     * @param longitude Longitude
+     * @param latitude Latitude
+     * @return ret Calculated
+     */
+    private static double transLat(double longitude, double latitude) {
```
tananaev, line 134: Did you copy all this code from somewhere?

src/main/java/org/traccar/geocoder/AmapGeocoder.java
mmzhi: English is not my mother tongue. Therefore, my understanding of some English words may be deficient. I really do not quite understand the scope of the area expressed by the word "settlement". Usually in China, the order of area scope is: country, province, city, district, town, street and building. But I read the comment on the "AddressFormat" class, and "district" comes before "settlement". I think "district" expresses a larger scope than "settlement". So I think "settlement" for "district" and "district" for "city" may be more compatible.

src/main/java/org/traccar/geocoder/AmapGeocoder.java
mmzhi: OK, the code here I copied directly from another class in the project. I'll rework it.

src/main/java/org/traccar/helper/CoordinateUtil.java
mmzhi: At the beginning, I also thought of using "Pair". But then I referred to the code of "hutool" toolkit about coordinate system conversion, so I followed the code of this toolkit directly.

src/main/java/org/traccar/helper/CoordinateUtil.java
mmzhi: Yes, most of the code for this class is taken from "https://github.com/dromara/hutool/blob/v5-master/hutool-core/src/main/java/cn/hutool/core/util/ CoordinateUtil.java", the tool library from which it was copied. There are a few major modifications made:
1. I translated the Chinese comments in the original class;
2. I removed the code about BD09 coordinate system. This coordinate system is only used in Baidu map.
3. The code uses a lot of "lat" and "lng" abbreviations to express latitude and longitude, which is not consistent with the original style of the project, so I modified these abbreviations.

mmzhi: Yes, most of the code for this class is taken from "https://github.com/dromara/hutool/blob/v5-master/hutool-core/src/main/java/cn/hutool/core/util/CoordinateUtil.java", the tool library from which it was copied. There are a few major modifications made:
1. I translated the Chinese comments in the original class;
2. I removed the code about BD09 coordinate system. This coordinate system is only used in Baidu map.
3. The code uses a lot of "lat" and "lng" abbreviations to express latitude and longitude, which is not consistent with the original style of the project, so I modified these abbreviations.

src/main/java/org/traccar/geocoder/AmapGeocoder.java
tananaev: Do you have the API documentation?

src/main/java/org/traccar/helper/CoordinateUtil.java
tananaev: Basically we need to update this code to match our style. We only use full words, no shortened versions.

src/main/java/org/traccar/geocoder/AmapGeocoder.java
mmzhi: Yes, the API documentation can be accessed at https://lbs.amap.com/api/webservice/guide/api/georegeo/#t5 . However, this documentation is only available in Chinese.

tananaev: If I understand it correctly, city in the API is only large cities that are basically their own "counties". I think the mapping should be:

country -> country
province -> state
district -> district
township -> settlement
neighborhood -> suburb

mmzhi: This understanding is not quite correct. In China, "city" is a very important administrative unit. It is even more important than "province" in the address.
For example, Guangdong is a province, Guangzhou is a city under Guangdong, and Tianhe is a district under Guangzhou.
In address writing, we would use "Tianhe, Guangzhou, Guangdong, China" for Tianhe District, or "Tianhe, Guangzhou, China "But we will not use "Tianhe, Guangdong, China" for Tianhe District.
However, in some special cases, "city" can also stand for "province". There are four direct-administered municipality in China, which are provincial administrative regions together with provinces. They are: Beijing, Tianjin, Shanghai and Chongqing.

A "neighbourhood" is a very small administrative unit under "township". In most cases, we don't use "neighbourhood" in addresses. Most Chinese people don't even know which "neighbourhood" they belong to.

tananaev: OK, but city only applies to the places where there is a big city an administrative area. So it's not always there.

The mapping should be based on the logical equivalents, not on the order.

mmzhi: Maybe you don't live in China, so you don't understand the importance of "city". In most cases, addresses in China use "city". For historical reasons, there are very few direct-administered counties in China, less than one-twentieth of the number of cities. Therefore, the unit "city" is very important and indispensable.

tananaev: Then we can potentially do something like `district = if city == null ? district : city`

mmzhi: In a few cases, it is possible to express it this way. But in most cases, it is necessary to keep both "city" and "district" in the address, because in more cases, they are upper and lower levels rather than the same level.
And "neighbourhood" is an unnecessary unit. Because addresses in China almost never use "neighbourhood".
You can refer to the address of the U.S. Embassy in China: https://china.usembassy-china.org.cn/contact/ . The address basically only contains building (number), street/road, township, district, city and province.
Or you can also refer to SAP, the software company's documentation for the Chinese address format: https://help.sap.com/docs/SAP_DATA_QUALITY_MANAGEMENT_SDK/7ac66320ca514acc89396a367db6dba8/ e540dc567ee51014a997d681b0e91070.html?version=4.2.9&locale=en-US

Translated with www.DeepL.com/Translator (free version)

mmzhi: In a few cases, it is possible to express it this way. But in most cases, it is necessary to keep both "city" and "district" in the address, because in more cases, they are upper and lower levels rather than the same level.
And "neighbourhood" is an unnecessary unit. Because addresses in China almost never use "neighbourhood".
You can refer to the address of the U.S. Embassy in China: https://china.usembassy-china.org.cn/contact/ . The address basically only contains building (number), street/road, township, district, city and province.
Or you can also refer to SAP, the software company's documentation for the Chinese address format: https://help.sap.com/docs/SAP_DATA_QUALITY_MANAGEMENT_SDK/7ac66320ca514acc89396a367db6dba8/e540dc567ee51014a997d681b0e91070.html?version=4.2.9&locale=en-US

src/main/java/org/traccar/helper/CoordinateUtil.java
mmzhi: OK, I'll modify this part of the code. Also I will adjust the order of the parameters, because the "hutool" library is coded by Chinese. In China, when we use latitude and longitude to express geographic coordinates, we usually write longitude first, then latitude, but in English, we usually write latitude first, then longitude. So I will also change this to unify the style of the code.

src/main/java/org/traccar/geocoder/AmapGeocoder.java
mmzhi: image
China's addressing system is very complex. This is the largest private courier company in China, the function of filling in the mail delivery address, with high reference.

mmzhi:
I resubmitted the code, mainly modifying the copyright description, replacing abbreviations with complete words, and adjusting the order of latitude and longitude.

src/main/java/org/traccar/geocoder/AmapGeocoder.java
mmzhi: After checking some information, I found that the address expression form in the United States is really very different from the address expression form in China. For example, in the United States, "Settlement" can mean a village or a city, and "District" can mean a part of a city or several cities. There is no clear hierarchy between "settlement" and "district". In China, however, "District" is a smaller unit under "City".
So in the end, I think this is a more appropriate mapping.
country -> country
province -> state
district -> district
city -> settlement
township -> suburb

mmzhi:
In the latest commit version, I added the georeferencing interpreters of the three major map providers in China. This way, users can have more choices.

src/main/java/org/traccar/helper/CoordinateUtil.java
tananaev: I still prefer to use pair here.

src/main/java/org/traccar/helper/CoordinateUtil.java
```diff
@@ -0,0 +1,257 @@
...
+public final class CoordinateUtil {
+
+    private CoordinateUtil() {
+
+    }
+
+    /**
+     * Coordinate conversion parameters：π（Pi)
+     */
+    public static final double PI = 3.1415926535897932384626433832795D;
```
tananaev, line 38: Can we use `Math.PI` instead of this?

mmzhi: Yes, I have resubmitted it. Although the PI accuracy in CoordinateUtil.java is higher than Math.PI, it has been tested, but Math.PI is accurate enough to be used in the current business.

mmzhi:
In the project "traccar-web", the company name of the map manufacturer, "autoNavi", is used as the name. "amap" is a product of "autoNavi". In order to standardise the naming convention, "amap" has been changed to "autoNavi".

mmzhi:
> 在“traccar-web”目录中，地图制造商“autoNavi”将用作名称。“amap”是“autoNavi”的产品。为了统一命名规则，“amap”已更改为“autoNavi”。



> In the project "traccar-web", the company name of the map manufacturer, "autoNavi", is used as the name. "amap" is a product of "autoNavi". In order to standardise the naming convention, "amap" has been changed to "autoNavi".

@tananaev In "traccar-web" project, autoNavi (amap) uses GCJ-02 coordinates, but openlayers does not support GCJ-02. As a result, when using autoNavi maps, the position and track of the device will be offset. Has it been considered to support GCJ-02 coordinates on "traccar-web"? I can also contribute code for this feature.

image

tananaev:
> In "traccar-web" project

Let's not discuss it here.

src/main/java/org/traccar/helper/CoordinateUtil.java
tananaev: Just a reminder that this is still not addressed.

602293544:
> 在“traccar-web”项目中，地图制造商“autoNavi”的公司名称被用作名称。“amap”是“autoNavi”的产品。为了统一命名规则，“amap”已更改为“autoNavi”。

大佬，现在代码合并了吗，docker镜像好像没更新？
然后请教下这个配置怎么填
    <entry key='geocoder.key'>1d355462f845e2efb6978b215cb</entry>        // Web服务
    <entry key='geocoder.type'>autonavi</entry>
    <entry key='geocoder.url'>https://restapi.amap.com/v3/geocode/regeo</entry>

meteoral:
> > 在“traccar-web”项目中，地图制造商“autoNavi”的公司名称被用作名称。“amap”是“autoNavi”的产品。为了统一命名规则，“amap”已更改为“autoNavi”。
> 
> 大佬，现在代码合并了吗，docker镜像好像没更新？ 然后请教下这个配置怎么填 `“ <entry key='geocoder.enable'>true</entry>”`
> 
> `“ <entry key='geocoder.key'>1d355462f845e2efb6978b215cb</entry> ” // Web服务`
> 
> `“ <entry key='geocoder.type'>autonavi</entry>”`
> 
> `“ <entry key='geocoder.url'>https://restapi.amap.com/v3/geocode/regeo</entry> ”`
> 
> 现在就是点显示地址没反应

This PR isn't megred yet

602293544:
> > > 在“traccar-web”项目中，地图制造商“autoNavi”的公司名称被用作名称。“amap”是“autoNavi”的产品。为了统一命名规则，“amap”已更改为“autoNavi”。
> > 
> > 
> > 大佬，现在代码合并了吗，docker镜像好像没更新？ 然后请教下这个配置怎么填 `“ <entry key='geocoder.enable'>true</entry>”`
> > `“ <entry key='geocoder.key'>**********</entry> ” // Web服务`
> > `“ <entry key='geocoder.type'>autonavi</entry>”`
> > `“ <entry key='geocoder.url'>https://restapi.amap.com/v3/geocode/regeo</entry> ”`
> > 现在就是点显示地址没反应
> 
> This PR isn't megred yet, and pls hide your geocoder.key or you will leak it

thanks,   the  geocoder.key is not correct  😄

mmzhi:
> > 在“traccar-web”项目中，地图制造商“autoNavi”的公司名称被默认名称。“amap”是“autoNavi”的产品。为了统一命名规则，“amap”已更改为“autoNavi”。
> 
> 大佬，现在代码合并了吗，docker镜像好像没更新？ 然后请教下这个配置怎么填 `“ <entry key='geocoder.enable'>true</entry>”`
> 
> `“ <entry key='geocoder.key'>1d355462f845e2efb6978b215cb</entry> ” // Web服务`
> 
> `“ <entry key='geocoder.type'>autonavi</entry>”`
> 
> `“ <entry key='geocoder.url'>https://restapi.amap.com/v3/geocode/regeo</entry> ”`
> 
> 现在就是点显示地址没反应

Currently the server does not support the address translation of Chinese map manufacturers, although the front-end support for the Gaode map, but there will be a map offset problem, the back-end I have submitted a merger request, the front-end due to the fact that I have more things to do lately, do not have too much time to look at the project, the follow-up I think about how to help the project to solve the problem of Chinese map offset!

carolinechenying:
waiting for  traccar that support chinease map like baidu/amap

cheng2004237:
有支持gcj02的计划吗？

tananaev:
Merged, thanks. I will do a small follow-up cleanup.

PR #5359, opened 2024-06-29
Easier to run locally

jcardus:
This way you can start the server with `./gradlew run`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,6 +3,7 @@ plugins {
     id "checkstyle"
     id "com.google.protobuf" version "0.9.4"
     id "org.kordamp.gradle.project-enforcer" version "0.13.0"
+    id "application"
 }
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -3,6 +3,7 @@ plugins {
     id "checkstyle"
     id "com.google.protobuf" version "0.9.4"
     id "org.kordamp.gradle.project-enforcer" version "0.13.0"
+    id "application"
```
tananaev, line 6: This plugin is pretty heavy. It adds distribution plugin and has some packaging stuff. I think it should already pretty easy to run the app by doing something like:
```
./gradlew build && java -jar ...
```

jcardus: ok, this works:

`./gradlew build && java -jar target/tracker-server.jar`

PR #5364, opened 2024-07-16
Fix out1 on GV310LAU

jcardus:
decodeFri with string array on GV310LAU
add GV310LAU test case

### DIFF
--- a/src/main/java/org/traccar/helper/StringParser.java
+++ b/src/main/java/org/traccar/helper/StringParser.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2015 - 2024 Joaquim Cardeira (joaquim.cardeira@gmail.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.helper;
+
+public class StringParser {
+    int index;
+    String[] values;
+    public StringParser(String[] values) {
+        this.values = values;
+    }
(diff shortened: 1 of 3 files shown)


jcardus:
[Uploading GV310LAU @Track Air Interface Protocol V2.02.pdf…]()

tananaev:
The file link is broken. Looks like you clicked comment before it was fully uploaded.

jcardus:
Ups

jcardus:
[GV310LAU @Track Air Interface Protocol V2.02.pdf](https://github.com/user-attachments/files/16243557/GV310LAU.%40Track.Air.Interface.Protocol.V2.02.pdf)

jcardus:
Maybe you can put it [here](https://www.traccar.org/protocols/), I've done other PR based on this.

src/main/java/org/traccar/helper/StringParser.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.helper;
+
+public class StringParser {
```
tananaev, line 18: If you want to do something like this, we need to migrate existing usages. I would also call it something like `StringArrayParser`.

jcardus: Ok, I will.

I did that because at first I was trying to replace the decodeFri method with the string array version. But it's difficult to replace it, so I made an overload. Is this ok?

tananaev: I think it's ok to have this class. Things to consider:
1. We should see if we can reuse functionality from pattern parser. And maybe have the same base class or something.
2. Migrate existing usages as a separate PR.
3. Do any new work you wanted to do in the first place, as a separate PR as well.

jcardus: I'm talking about the decodeFri method overload, is that ok?

jcardus: Ok, I will close this now. I will first create a new PR with the changes for GV310LAU using the current way of parsing.  I'll do the refactoring after.

PR #5365, opened 2024-07-17
Fix out1 on GV310LAU

jcardus:


src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -933,6 +940,52 @@ private Object decodeFri(Channel channel, SocketAddress remoteAddress, String se
         return positions;
     }
 
+    private Object decodeFri(Channel channel, SocketAddress remoteAddress, String[] v) throws ParseException {
```
tananaev, line 943: It doesn't make sense to have two versions of `decodeFri`

jcardus: that was my question on the previous PR

tananaev: I understood something different. I thought you meant parser overload.

jcardus: It's complex to keep the existing logic of this method with the new parser. I also tried to change the pattern to include the GV310LAU but couldn't make it work.

tananaev: We need to find a way without duplication.

jcardus: can you help me change the pattern?

tananaev: I think we should migrate away from using the pattern.

jcardus: But looking at the pattern it should allow a lot of combinations with different sizes, I'm afraid we loose some old models which I'm not sure where they fit on the pattern.

tananaev: This might be helpful:

https://docs.google.com/spreadsheets/d/1sU7d1aqWc_hVVjprPopS1PewHlfZLlix1Y7Hp5pihnY/edit?gid=2037704330

tananaev: Not sure. Maybe check protocol documents that we have?

jcardus: Nope, can't find those fields in any of the documents.

tananaev: You can use a placeholder model name for now then.

PR #5371, opened 2024-07-25
geocoder: Implement OlaMaps Reverse geocoder

mujeerhashmi:
We have integrated Ola Maps geocoder. Kindly review.

### DIFF
--- a/src/main/java/org/traccar/MainModule.java
+++ b/src/main/java/org/traccar/MainModule.java
@@ -226,7 +226,7 @@ public class MainModule extends AbstractModule {
                 case "maptiler" -> new MapTilerGeocoder(client, key, cacheSize, addressFormat);
                 case "geoapify" -> new GeoapifyGeocoder(client, key, language, cacheSize, addressFormat);
                 case "geocodejson" -> new GeocodeJsonGeocoder(client, url, key, language, cacheSize, addressFormat);
-                default -> new GoogleGeocoder(client, key, language, cacheSize, addressFormat);
+                default -> new GoogleGeocoder(client, url, key, language, cacheSize, addressFormat);
             };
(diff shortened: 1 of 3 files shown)


src/main/java/org/traccar/geocoder/OlaMapsGeocoder.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+        return url;
+    }
+
+    public OlaMapsGeocoder(Client client, String key, int cacheSize, AddressFormat addressFormat) {
+        super(client, formatUrl(key), cacheSize, addressFormat);
+    }
+
+    @Override
+    public Address parseAddress(JsonObject json) {
```
tananaev, line 38: It seems like format matched Google geocoder, so it doesn't make sense to duplicate it.

mujeerhashmi: Yes it's almost the same. How can we use this geocoder which has different url.

tananaev: You can probably modify Google geocoder to allow custom URL. But duplicating code is never a good solution.

mujeerhashmi: Ok will modify the PR.

mujeerhashmi: > You can probably modify Google geocoder to allow custom URL. But duplicating code is never a good solution.

I have done these changes. Kindly review.

mujeerhashmi:
@tananaev done

tananaev:
Merged, thanks.

PR #5392, opened 2024-08-18
Fix ERI for GV350M

AymanElarian:
Hello ,

GV350 fail as we are taking 
 index += 1; // reserved / uart device type


tananaev:
This change doesn't make sense. The documentation says that the field should be there.

AymanElarian:
please try  test case , without this pull request it will fail , with , it will pass

verifyAttributes(decoder, buffer(
                 "+RESP:GTERI,F18501,862599050498599,GV350M_HHP,00000002,,51,1,1,18.9,140,319.5,46.586093,28.985816,20240817120402,0420,0001,03DA,AC81,,0.0,,,0,220100,,0,20240817120401,7860$"));

tananaev:
That doesn't mean that your fix is correct.

AymanElarian:
the position fully decoded. I agree that the reserved fields and UART device type are necessary, but I'm unsure why it functions as expected. It might be due to unnecessary calls to index+ or similar methods being invoked prematurely.

tananaev:
Well, we need to figure out the reason and have a proper fix.

PR #5396, opened 2024-08-25
aovx implement wifi

jcardus:
https://github.com/AOVX-Corporation/AOVX_Share/blob/main/Document/en/01_G%20series/AOVX_G%26A_Series%20Cloud%20Platform%20Protocol.pdf


src/main/java/org/traccar/protocol/HuabaoProtocolDecoder.java
```diff
@@ -699,6 +700,18 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
                         }
                     }
                     break;
+                case 0xF4:
+                    Network network = new Network();
```
tananaev, line 704: I think we need to move `Network` to the top level somewhere, so that we don't accidentally override it.

src/main/java/org/traccar/protocol/HuabaoProtocolDecoder.java
```diff
@@ -699,6 +699,17 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
...
+                case 0xF4:
+                    while (buf.readerIndex() < endIndex) {
+                        WifiAccessPoint wifiAccessPoint = new WifiAccessPoint();
+                        wifiAccessPoint.setMacAddress(String.format("%02x:%02x:%02x:%02x:%02x:%02x",
+                                buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte(),
+                                buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte()));
+                        wifiAccessPoint.setSignalStrength((int) buf.readByte());
+                        network.addWifiAccessPoint(wifiAccessPoint);
+                    }
+                    position.setNetwork(network);
```
tananaev, line 711: I think instead of calling `setNetwork` in multiple places, it should also be moved to the top level. Something similar to what we do in other places, link [this](https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/protocol/EelinkProtocolDecoder.java#L241-L243).

src/main/java/org/traccar/protocol/HuabaoProtocolDecoder.java
```diff
@@ -699,6 +697,16 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
...
                     }
                     break;
+                case 0xF4:
+                    while (buf.readerIndex() < endIndex) {
+                        WifiAccessPoint wifiAccessPoint = new WifiAccessPoint();
+                        wifiAccessPoint.setMacAddress(String.format("%02x:%02x:%02x:%02x:%02x:%02x",
+                                buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte(),
+                                buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte()));
+                        wifiAccessPoint.setSignalStrength((int) buf.readByte());
+                        network.addWifiAccessPoint(wifiAccessPoint);
```
tananaev, line 707: Hopefully the last thing, but there's a more concise way of doing this. Something like this:
```java
                            String mac = ByteBufUtil.hexDump(buf.readSlice(6)).replaceAll("(..)", "$1:");
                            network.addWifiAccessPoint(WifiAccessPoint.from(
                                    mac.substring(0, mac.length() - 1), buf.readByte()));
```
We do it in many other places.

PR #5405, opened 2024-09-14
Arknav Protocol Decoder fix

4lazar:
353425142654345,05*850,000,**PT33**,A,2459.3640,N,12125.2958,E,000.0,224.8,00.8,07:47:26 09-09-05,**1.00,80**

IMEI number,
ID Code (6 Bytes in ASCII), 
Status code (3 Bytes in ASCII), 
Unit number (4 Bytes in ASCII), 
Status (A or V), 
Latitude, N/S indicator,
Longitude, 
E/W Indicator, 
speed, 
Course, 
HDOP, 
GPS time lat/lon value - GPS date lat/lon value, 
Unit Version number, 
battery power level

PR #5408, opened 2024-09-17
Update ArknavProtocol, added battery level

4lazar:


tananaev:
You're missing documentation and description.

4lazar:
all requested info can be found here https://github.com/traccar/traccar/pull/5405

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -47,6 +47,8 @@ public ArknavProtocolDecoder(Protocol protocol) {
             .number("(d+.?d*),")                 // hdop
             .number("(dd):(dd):(dd) ")           // time (hh:mm:ss)
             .number("(dd)-(dd)-(dd),")           // date (dd-mm-yy)
+            .expression(".{4},")                 // Unit Version number
+            .number("(dd)?")                     // Battery level
```
tananaev, line 51: It seems to conflict with the existing protocol.

4lazar: Which protocol is it in conflict with?
Mapping is according to Arknav protocol documentation

tananaev: I see other unit tests have `9.00,D3` in this spot. Which means this is incorrect pattern.

tananaev:
Here's the original documentation.

[R-9-setup2-22-7-2010.pdf](https://github.com/user-attachments/files/17031723/R-9-setup2-22-7-2010.pdf)

4lazar:
Commit is ready with new added device pattern and tests

tananaev:
I don't think it makes sense to duplicate the whole thing.

4lazar:
It make sense because parameters are mostly different between application R9 and R33, except location info.
In future it will help to process correctly the logic separately for each application.

tananaev:
I see that something like 80% of the message is the same, so I don't think your point is valid.

4lazar:
Check the decoding of value.
For application R9, most of the values are HEX encode vs. R33(simplified) where they are ASCII.
Even for Status, decoded values are different

4lazar:
I plan to add new decoding logic for R33 application without affecting R9 because I don't know who is using it and for what purpose.

tananaev:
Well, if you're planning to keep the duplication, we won't be able to merge it. In case you weren't aware, you can still branch within a single regex. I don't know why you're refusing to do it.

4lazar:
I have no problem with this. I can try to do this

4lazar:
New commit is ready with new features for R33 application and one pattern

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -59,23 +62,40 @@ protected Object decode(
...
         position.setLatitude(parser.nextCoordinate());
         position.setLongitude(parser.nextCoordinate());
-        position.setSpeed(parser.nextDouble(0));
+        position.setSpeed(UnitsConverter.kphFromKnots(parser.nextDouble(0)));
         position.setCourse(parser.nextDouble(0));
-
         position.set(Position.KEY_HDOP, parser.nextDouble(0));
-
         position.setTime(parser.nextDateTime(Parser.DateTimeFormat.HMS_DMY));
+        if (model.contains("PT33")) {
```
tananaev, line 81: Why is this not using `equals` instead of `PT33`

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -59,23 +62,40 @@ protected Object decode(
...
+            }
+            if ((status & 0x400L) != 0) {
+                position.addAlarm(Position.ALARM_GEOFENCE_EXIT);
+            }
+            if ((status & 0x010L) != 0) {
+                position.addAlarm(Position.ALARM_SOS);
+            }
+            if ((status & 0x001L) != 0) {
+                position.addAlarm(Position.ALARM_OVERSPEED);
+            }
```
tananaev, line 97: Where can I find this in the doc?

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -47,9 +48,11 @@ public ArknavProtocolDecoder(Protocol protocol) {
             .number("(d+.?d*),")                 // hdop
             .number("(dd):(dd):(dd) ")           // time (hh:mm:ss)
             .number("(dd)-(dd)-(dd),")           // date (dd-mm-yy)
+            .expression(".{4},")                 // Unit Version number
+            .expression("(..)")                  // Battery level
             .any()
             .compile();
-
+
```
tananaev, line 55: Undo this

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: https://github.com/user-attachments/files/17003217/personal.tracker.manual.V2.pdf
pag 39

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: changed to equals

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -59,23 +62,45 @@ protected Object decode(
...
+        String model = parser.next();
+        Boolean valid_geo = parser.next().equals("A");
+        position.setValid(valid_geo);            
+        double latitude = parser.nextCoordinate();
+        double longitude = parser.nextCoordinate();
+        if (valid_geo || (latitude != 0 && longitude != 0)) {
+            position.setLatitude(latitude);
+            position.setLongitude(longitude);
+        }
+        position.setSpeed(UnitsConverter.kphFromKnots(parser.nextDouble(0)));
```
tananaev, line 82: Why are you doing this? This makes no sense to me. And it also breaks naming convention.

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
tananaev: It looks like the value is not HEX, but a specially encoded 3 numbers.

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: I have changed back the code.

4lazar: I found in the documentation that these parameters can be configured in the configuration file.
No more changes to the code

PR #5409, opened 2024-09-18
Allow managers to send announcements

lucasleite01:
Managers should also be allowed to send announcements, since they can have many users bellow them to manage.

Plus, this would enhance the announcements feature to cover scenarios of Traccar servers where managers want to send information to their managed users.

### DIFF
--- a/src/main/java/org/traccar/api/resource/NotificationResource.java
+++ b/src/main/java/org/traccar/api/resource/NotificationResource.java
@@ -28,10 +28,7 @@ import jakarta.ws.rs.core.Response;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.traccar.api.ExtendedObjectResource;
-import org.traccar.model.Event;
-import org.traccar.model.Notification;
-import org.traccar.model.Typed;
-import org.traccar.model.User;
+import org.traccar.model.*;
 import org.traccar.notification.MessageException;
 import org.traccar.notification.NotificationMessage;
 import org.traccar.notification.NotificatorManager;
@@ -42,11 +39,7 @@ import org.traccar.storage.query.Request;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 
 @Path("notifications")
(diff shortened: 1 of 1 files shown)


tananaev:
Do you have a corresponding web app change?

lucasleite01:
Yes, check [here](https://github.com/traccar/traccar-web/pull/1268) the corresponding web app change.

src/main/java/org/traccar/api/resource/NotificationResource.java
```diff
@@ -113,15 +114,27 @@ public Response testMessage(@PathParam("notificator") String notificator)
...
+                if (permissionsService.notAdmin(getUserId())) {
+                    var conditions = new LinkedList<Condition>();
+                    conditions.add(new Condition.Equals("id", userId));
+                    conditions.add(new Condition.Permission(User.class, getUserId(), ManagedUser.class).excludeGroups());
+                    users.add(storage.getObject(
+                            User.class, new Request(new Columns.All(), Condition.merge(conditions))));
+                } else {
+                    users.add(storage.getObject(
+                            User.class, new Request(new Columns.All(), new Condition.Equals("id", userId))));
+                }
```
tananaev, line 137: It probably makes sense to do something like this:

```suggestion
                    var conditions = new LinkedList<Condition>();
                    conditions.add(new Condition.Equals("id", userId));
                if (permissionsService.notAdmin(getUserId())) {
                    conditions.add(new Condition.Permission(User.class, getUserId(), ManagedUser.class).excludeGroups());
                }
                    users.add(storage.getObject(
                            User.class, new Request(new Columns.All(), Condition.merge(conditions))));
```

src/main/java/org/traccar/api/resource/NotificationResource.java
```diff
@@ -113,15 +114,27 @@ public Response testMessage(@PathParam("notificator") String notificator)
...
             NotificationMessage message) throws MessageException, StorageException {
-        permissionsService.checkAdmin(getUserId());
+        permissionsService.checkManager(getUserId());
         List<User> users;
         if (userIds.isEmpty()) {
-            users = storage.getObjects(User.class, new Request(new Columns.All()));
+            users = permissionsService.notAdmin(getUserId()) ?
+                    storage.getObjects(User.class, new Request(new Columns.All(),
+                            new Condition.Permission(User.class, getUserId(), ManagedUser.class).excludeGroups())) :
+                    storage.getObjects(User.class, new Request(new Columns.All()));
```
tananaev, line 123: Let's do if/else here. Tertiary expression we mostly use for short things that make sense to inline.

src/main/java/org/traccar/api/resource/NotificationResource.java
```diff
@@ -113,15 +114,28 @@ public Response testMessage(@PathParam("notificator") String notificator)
...
+                var conditions = new LinkedList<Condition>();
+                conditions.add(new Condition.Equals("id", userId));
+                if (permissionsService.notAdmin(getUserId())) {
+                    conditions.add(new Condition.Permission(User.class, getUserId(), ManagedUser.class).excludeGroups());
+                    users.add(storage.getObject(
+                            User.class, new Request(new Columns.All(), Condition.merge(conditions))));
+                } else {
+                    users.add(storage.getObject(
+                            User.class, new Request(new Columns.All(), Condition.merge(conditions))));
+                }
```
tananaev, line 138: This doesn't look like what I suggested. Why do we need to duplicate `users.add(storage.getObject(...`?

tananaev:
Make sure you compile and test it. Currently build is failing.

lucasleite01:
The error was a style check:
image

The last commit fixes it.

Issue #5413, opened 2024-09-19
SMS Sending error in 6.5


jinzo:
Hi,

continuing from the forum thread: https://www.traccar.org/forums/topic/sms-notification-errorafter-upgrade-to-65-version/page/
The problem is that the constructed JSON (for some reason) inserts another newline (unescaped) at the end of "message" - and this is then invalid JSON and fails with most parsers ( `"Test message\n` ) in below example:

> '{\n'
>  '            "to": "+1234",\n'
>  '            "message": "Test message\n'
>  '"\n'
>  '        }'

(this is a line by line representation of what gets posted quoted with single quotes).

Not entirely sure why this is happening after the upgrade, it for sure did not happen before. Will continue debugging.

jinzo:
Ah, unintended side effect of https://github.com/traccar/traccar/commit/9fc42e046178bbc19e3d7ac1282636a32d565f8e
So I guess, changing this to trim just the right part of the string would 'solve' this problem (as teltonika username and password are always on the left side/start of the string). Or going back up the stack and see why the message has a newline in the first place.

tananaev:
Probably figuring out why there's a new line is what we should do first.

tananaev:
https://github.com/traccar/traccar/issues/5402#issuecomment-2362293674

jcardus:
The problem is just with the test notification?

tananaev:
I don't see any indication that it's limited to just test notifications.

jcardus:
I'll take a look

truppelito:
No, I can confirm it's not. In fact, I found the issue because I stopped receiving notifications during normal operation. I showed the test notification because that's just how I debugged it.

jcardus:
It's related with the templates, when the message comes from the template it comes with new lines, maybe we should trim in the notification part and not on the sms side.

jcardus:
https://github.com/traccar/traccar/pull/5414

this should fix it

truppelito:
Can I suggest that the json payload also be escaped after substituting the strings? It could help prevent similar issues in the future and would make the code more robust.

jcardus:
I don't think that makes sense, you have to escape the string contents in the json, not the json payload. So, maybe the correct way would be to escape the string instead of trimming it.

tananaev:
[...]

PR #5414, opened 2024-09-19
fix sms notifications

jcardus:


tananaev:
I think I have a better fix. See https://github.com/traccar/traccar/issues/5413#issuecomment-2362804196.

PR #5421, opened 2024-09-27
trv protocol - No response for following for some IWAP Commands

Nikutrax:
No response for following IWAP Commands:
AP33
AP34
AP84
AP85


tananaev:
You should at least link to the old PR so we have full context.

Nikutrax:
[Gps Watch protocol_2023.pdf](https://github.com/user-attachments/files/17165428/Gps.Watch.protocol_2023.pdf)

PR #5431, opened 2024-10-07
add data type in huabao (aovx)

jcardus:
With this we can then make some computed attributes to get the rest of the data.

From the comment on the code looks like this is always the data type, I only have the documentation for aovx.

https://github.com/AOVX-Corporation/AOVX_Share/blob/main/Document/en/01_G%20series/AOVX_G%26A_Series%20Cloud%20Platform%20Protocol.pdf

Screenshot 2024-10-07 at 17 00 05


tananaev:
We use type for the overall message type. In this case it seems more like an event. Or maybe just add as a custom attribute.

jcardus:
KEY_EVENT?

PR #5442, opened 2024-10-16
Security: Session fixation

AxelRieben:
### Problem
Session fixation is made possible by the unsafe practice of keeping the same value in session cookies before and after authentication.

In the generic exploitation of session-fixing vulnerabilities, an attacker can obtain a set of session cookies from the target website without first authenticating. He can then force these cookies into the victim's browser using various techniques. If the victim subsequently authenticates on the target website and the cookies are not updated at login, the victim will be identified by the session cookies chosen by the chosen by the attacker. The attacker is then able to impersonate the victim using these known cookies.

See: [OWASP Session Fixation](https://owasp.org/www-community/attacks/Session_fixation )

### Recommendation
This pull request refresh the session cookie after a successful login.

### DIFF
--- a/src/main/java/org/traccar/api/resource/SessionResource.java
+++ b/src/main/java/org/traccar/api/resource/SessionResource.java
@@ -81,6 +81,7 @@ public class SessionResource extends BaseResource {
             LoginResult loginResult = loginService.login(token);
             if (loginResult != null) {
                 User user = loginResult.getUser();
+                request.getSession().invalidate();
                 request.getSession().setAttribute(USER_ID_KEY, user.getId());
                 request.getSession().setAttribute(EXPIRATION_KEY, loginResult.getExpiration());
                 LogAction.login(user.getId(), WebHelper.retrieveRemoteAddress(request));
@@ -128,6 +129,7 @@ public class SessionResource extends BaseResource {
         }
         if (loginResult != null) {
             User user = loginResult.getUser();
+            request.getSession().invalidate();
             request.getSession().setAttribute(USER_ID_KEY, user.getId());
             LogAction.login(user.getId(), WebHelper.retrieveRemoteAddress(request));
             return user;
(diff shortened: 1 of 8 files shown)


src/main/java/org/traccar/api/resource/SessionResource.java
```diff
@@ -81,6 +81,7 @@ public User get(@QueryParam("token") String token) throws StorageException, IOEx
             LoginResult loginResult = loginService.login(token);
             if (loginResult != null) {
                 User user = loginResult.getUser();
+                request.getSession().invalidate();
```
tananaev, line 84: 1. Have you checked that this works correctly?
2. Are you sure you've added it to all the places?
3. If there are more places, it might make sense to add some helper method for this.

tananaev: No, let's do the refactoring here. It's a pretty small change.

AxelRieben: I'd like to know more about your request. In which project folder would you place the extension method and the class? Which lines do you want to include in the method?

AxelRieben: Thanks for the feedback. I've added a helper method for this duplicate code.

src/main/java/org/traccar/api/resource/SessionResource.java
```diff
@@ -105,7 +101,7 @@ public User get(@PathParam("id") long userId) throws StorageException {
         permissionsService.checkUser(getUserId(), userId);
         User user = storage.getObject(User.class, new Request(
                 new Columns.All(), new Condition.Equals("id", userId)));
-        request.getSession().setAttribute(USER_ID_KEY, user.getId());
+        request.getSession().setAttribute(SessionHelper.USER_ID_KEY, user.getId());
```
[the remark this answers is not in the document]
AxelRieben: 👍

PR #5443, opened 2024-10-16
Security: Session cookie is not HttpOnly

AxelRieben:
### Problem
The session cookie does not have the HttpOnly attribute. This parameter tells the browser that the cookie can only be accessed by the server and not by client-side scripts.

See: [OWASP Testing for Cookies Attributes](https://owasp.org/www-project-web-security-testing-guide/v41/4-Web_Application_Security_Testing/06-Session_Management_Testing/02-Testing_for_Cookies_Attributes)

### Recommendation
This pull request adds a parameter to set the HttpOnly attribute to the session cookie. Personally, I would set this attribute to “true” by default, but I don't know the system-wide implications.

### DIFF
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -843,6 +843,14 @@ public final class Keys {
             List.of(KeyType.CONFIG),
             "max-age=3600,public");
 
+     /**
+     * Set HttpOnly attribut to the session cookie.
+     */
+    public static final ConfigKey<Boolean> WEB_COOKIE_HTTP_ONLY = new BooleanConfigKey(
+        "web.cookieHttpOnly",
+        List.of(KeyType.CONFIG),
+        false);
+
     /**
      * Enable TOTP authentication on the server.
      */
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/config/Keys.java
```diff
@@ -843,6 +843,14 @@ private Keys() {
             List.of(KeyType.CONFIG),
             "max-age=3600,public");
 
+     /**
+     * Set HttpOnly attribut to the session cookie.
+     */
+    public static final ConfigKey<Boolean> WEB_COOKIE_HTTP_ONLY = new BooleanConfigKey(
+        "web.cookieHttpOnly",
```
tananaev, line 850: Any reason this needs to be configurable?

AxelRieben: Potentially this could provide backwards compatibility with clients that need to access this cookie client-side. But I'm not aware if such an implementation makes sense.

tananaev: If it's not required currently by the official app, I suggest we remove the configuration option.

AxelRieben: I agree and have removed the parameter. The cookie is now always httpOnly.

PR #5444, opened 2024-10-16
Security: Hide error details

AxelRieben:
### Problem
By default, some pages return technical details on errors, such as the stack trace. While this behavior can be useful for development, it is not user-friendly and may reveal implementation details to an attacker.

See: [OWASP Improper Error Handling](https://owasp.org/www-community/Improper_Error_Handling)

### Recommendation
This pull request add a parameter to hide error details and only show a generic error message.

Linked to pull request : https://github.com/traccar/traccar-web/pull/1283

### DIFF
--- a/src/main/java/org/traccar/api/ResourceErrorHandler.java
+++ b/src/main/java/org/traccar/api/ResourceErrorHandler.java
@@ -15,20 +15,33 @@
  */
 package org.traccar.api;
 
+import org.traccar.config.Config;
+import org.traccar.config.Keys;
 import org.traccar.helper.Log;
 
+import jakarta.inject.Inject;
 import jakarta.ws.rs.WebApplicationException;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.ext.ExceptionMapper;
 
 public class ResourceErrorHandler implements ExceptionMapper<Exception> {
 
+    private static final String HIDDEN_MESSAGE_KEY = "HiddenError";
+    private boolean hideErrorDetail;
+
+    @Inject
+    public ResourceErrorHandler(Config config) {
+        this.hideErrorDetail = config.getBoolean(Keys.WEB_HIDE_ERROR_DETAIL);
+    }
+
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/api/ResourceErrorHandler.java
```diff
@@ -15,20 +15,33 @@
...
 import org.traccar.helper.Log;
 
+import jakarta.inject.Inject;
 import jakarta.ws.rs.WebApplicationException;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.ext.ExceptionMapper;
 
 public class ResourceErrorHandler implements ExceptionMapper<Exception> {
 
+    private static final String HIDDEN_MESSAGE_KEY = "HiddenError";
```
tananaev, line 29: This doesn't make sense to me. I think we should return at least the message.

AxelRieben: How would you do it without revealing technical message ?

tananaev: You don't. The message is needed and intended for the end user in most cases.

harshanasrimal:
I agree with @AxelRieben

Today, one of our users reported receiving a long Java error message when adding new devices. They also provided a screenshot.

While this information is helpful for developers, it may be confusing for non-technical end users. It might be worth improving the error handling to present a more user-friendly message while still logging the technical details for debugging.

tananaev:
@harshanasrimal what version are you using? Please provide the screenshot.

harshanasrimal:
I was using v6.2 for frontend and here is the screenshot I got from the user.

fce7224a-2fe9-4f9d-91b3-2690c9f075e8


 Today I updated it to v6.8.2 and now the error is bit clean but still It confuse non technical users.

Screenshot 2025-08-12 at 19 20 31


when the user click more details he gets the whole error

Screenshot 2025-08-12 at 19 21 10

PR #5455, opened 2024-10-25
Add data compression to PositionForwarderWialon

TheArrex:
If data is bigger then MTU, UDP won't allow to send this. I try to change UDP to TCP, chunk data, but compression do it's best.
Compression is described at the end of this docs: https://extapi.wialon.com/hw/cfg/Wialon%20IPS_en_v_2_0.pdf
Tested on Wialon and build looks fine.

### DIFF
--- a/src/main/java/org/traccar/forward/PositionForwarderWialon.java
+++ b/src/main/java/org/traccar/forward/PositionForwarderWialon.java
@@ -113,10 +113,8 @@ public class PositionForwarderWialon implements PositionForwarder {
         DatagramPacket packet;
 
         if (useCompression) {
-            byte[] compressedBuffer = compressData(buffer);
-            ByteBuf container = createContainer(compressedBuffer);
+            ByteBuf container = compressData(buffer);
             packet = new DatagramPacket(container.array(), container.readableBytes(), address, port);
-            container.release();
         } else {
             packet = new DatagramPacket(buffer, buffer.length, address, port);
         }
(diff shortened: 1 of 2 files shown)


tananaev:
1. Can you rebase on master properly. I see duplicated changes that already exist on the master branch.
2. It doesn't look like Wialon 1.x supports zip.
3. We also don't support zip in our server for Wialon.

src/main/java/org/traccar/forward/PositionForwarderWialon.java
```diff
@@ -111,28 +116,56 @@ public void forward(PositionData positionData, ResultHandler resultHandler) {
...
+        container.write(0xFF);
+
+        int dataLength = data.length;
+        ByteBuffer lengthBuffer = ByteBuffer.allocate(2).order(ByteOrder.LITTLE_ENDIAN);
+        lengthBuffer.putShort((short) dataLength);
+        container.write(lengthBuffer.array(), 0, 2);
+
+        container.write(data, 0, dataLength);
+
+        return container.toByteArray();
```
tananaev, line 168: Using Netty buffers can probably make this much cleaner.

TheArrex:
> 2. It doesn't look like Wialon 1.x supports zip.
> 3. We also don't support zip in our server for Wialon.

Should i just wrap this feature with `version.startsWith("2")` condition?

And maybe we need to change protocol version from hardcode at MainModule to config option?

tananaev:
I think it would be better to just have a constructor parameter to enable/disable ZIP.

TheArrex:
> I think it would be better to just have a constructor parameter to enable/disable ZIP.

Disabled by default?

tananaev:
Yes, I prefer disabled by default because then the format will be compatible with Traccar.

src/main/java/org/traccar/forward/PositionForwarderWialon.java
```diff
@@ -111,6 +129,33 @@ public void forward(PositionData positionData, ResultHandler resultHandler) {
         }
     }
 
+    public static byte[] compressData(byte[] data) {
```
tananaev, line 132: I would probably just combine the two methods to avoid two buffers.

PR #5457, opened 2024-10-29
Add new search condition using Condition.In for optimized database querying

CpdnCristiano:
This pull request introduces a new search condition using Condition.In, which I believe will be more efficient than the current for loop by making a single database request rather than multiple calls.

I've tested this change in MySQL and expect In to work similarly across all SQL databases. While I haven't tested it with in-memory storage, the logic for in-memory storage appears to be correct and should function as intended.

### DIFF
--- a/src/main/java/org/traccar/protocol/DmtProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/DmtProtocolDecoder.java
@@ -16,6 +16,7 @@
 package org.traccar.protocol;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.Channel;
 import org.traccar.BaseProtocolDecoder;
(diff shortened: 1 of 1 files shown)


CpdnCristiano:
I retrieve the latest positions and generate a report based on this information, which requires fetching additional details like the device names.

Below is an example JavaScript code for my use case to check for low battery levels:

```js
const getReportLowPower = async () => {
  const positions = await getPositions();
  const lowPowerPositions = positions.filter(
    (position) =>
      position.attributes?.power &&
      position.attributes.power < LOW_POWER_THRESHOLD
  );
  const devices = await getDevicesByIds(
    lowPowerPositions.map((position) => position.deviceId)
  );

  const data = lowPowerPositions.map((position) => {
    const device = devices.find((device) => device.id === position.deviceId);
    return {
      PLACA: device.name,
      BATERIA: position.attributes.power,
      IMEI: device.uniqueId,
    };
  });

  const reportPath = path.join(
    REPOR_DIR,
    `low_power_${new Date().getTime()}.xlsx`
  );
  const ws = XLSX.utils.json_to_sheet(data);
  const wb = XLSX.utils.book_new();
  XLSX.utils.book_append_sheet(wb, ws, 'Veiculos com bateria baixa');
  XLSX.writeFile(wb, reportPath);

  return reportPath;
}
```

tananaev:
The use case is still not clear. Why can't you just take all devices for the current user?

CpdnCristiano:
Thank you; I’ll go ahead with that approach.

PR #5462, opened 2024-11-08
Add Alarm attribute to Minifinder2ProtocolDecoder.java

markdegrootnl:
Add Alarm attribute to the Minifinder2 protocol to get the raw alarm data.

### DIFF
--- a/src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
@@ -165,14 +165,49 @@ public class Minifinder2ProtocolDecoder extends BaseProtocolDecoder {
                         if (BitUtil.check(alarm, 2)) {
                             position.addAlarm(Position.ALARM_FALL_DOWN);
                         }
+                        if (BitUtil.check(alarm, 4)) {
+                            position.addAlarm(BitUtil.check(alarm, 26) ? Position.ALARM_GEOFENCE_ENTER : Position.ALARM_GEOFENCE_EXIT);
+                            position.set("geo", 1);
+                        }
+                        if (BitUtil.check(alarm, 5)) {
+                            position.addAlarm(BitUtil.check(alarm, 27) ? Position.ALARM_GEOFENCE_ENTER : Position.ALARM_GEOFENCE_EXIT);
+                            position.set("geo", 2);
+                        }
+                        if (BitUtil.check(alarm, 6)) {
+                            position.addAlarm(BitUtil.check(alarm, 28) ? Position.ALARM_GEOFENCE_ENTER : Position.ALARM_GEOFENCE_EXIT);
+                            position.set("geo", 3);
+                        }
+                        if (BitUtil.check(alarm, 7)) {
+                            position.addAlarm(BitUtil.check(alarm, 29) ? Position.ALARM_GEOFENCE_ENTER : Position.ALARM_GEOFENCE_EXIT);
+                            position.set("geo", 4);
+                        }
                         if (BitUtil.check(alarm, 8)) {
                             position.addAlarm(Position.ALARM_POWER_OFF);
                         }
                         if (BitUtil.check(alarm, 9)) {
                             position.addAlarm(Position.ALARM_POWER_ON);
                         }
+                        if (BitUtil.check(alarm, 10)) {
+                            position.addAlarm(Position.ALARM_MOVEMENT);
+                            position.set(Position.KEY_MOTION, true);
+                        }
+                        if (BitUtil.check(alarm, 11)) {
+                            position.set(Position.KEY_MOTION, false);
+                        }
                         if (BitUtil.check(alarm, 12)) {
                             position.addAlarm(Position.ALARM_SOS);
+                        }
+                        if (BitUtil.check(alarm, 13)) {
+                            position.set("side1", true);
+                        }
+                        if (BitUtil.check(alarm, 14)) {
+                            position.set("side2", true);
+                        }
+                        if (BitUtil.check(alarm, 15)) {
+                            position.set(Position.KEY_CHARGE, true);
+                        }
+                        if (BitUtil.check(alarm, 16)) {
+                            position.set(Position.KEY_CHARGE, false);
                         }
                         if (BitUtil.check(alarm, 31)) {
                             position.set("bark", true);
@@ -180,9 +215,7 @@ public class Minifinder2ProtocolDecoder extends BaseProtocolDecoder {
                         if (length == 5) {
                             position.setDeviceTime(new Date(buf.readUnsignedIntLE() * 1000));
                         }
-                        if (alarm > 0) {
-                            position.set("alarmCode", alarm);
-                        }
+                        position.set("alarmCode", alarm);
                         break;
                     case 0x14:
                         position.set(Position.KEY_BATTERY_LEVEL, buf.readUnsignedByte());
@@ -251,9 +284,7 @@ public class Minifinder2ProtocolDecoder extends BaseProtocolDecoder {
                     case 0x24:
                         position.setTime(new Date(buf.readUnsignedIntLE() * 1000));
                         long status = buf.readUnsignedIntLE();
-                        if (BitUtil.check(status, 4)) {
-                            position.set(Position.KEY_CHARGE, true);
-                        }
+                        position.set(Position.KEY_CHARGE, BitUtil.check(status, 4));
                         if (BitUtil.check(status, 7)) {
                             position.set(Position.KEY_ARCHIVE, true);
                         }
--- a/src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
@@ -167,7 +167,11 @@ public class Minifinder2ProtocolDecoder extends BaseProtocolDecoder {
                         }
                         for (int i = 0; i < 4; i++) {
                             if (BitUtil.check(alarm, i + 4)) {
-                                position.addAlarm(BitUtil.check(alarm, i + 26) ? Position.ALARM_GEOFENCE_ENTER : Position.ALARM_GEOFENCE_EXIT);
+                                if (BitUtil.check(alarm, i + 26)) {
+                                    position.addAlarm(Position.ALARM_GEOFENCE_ENTER);
+                                } else {
+                                    position.addAlarm(Position.ALARM_GEOFENCE_EXIT);
+                                }
                                 position.set(Position.KEY_GEOFENCE, i + 1);
                             }
                         }
--- a/src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
@@ -180,6 +180,9 @@ public class Minifinder2ProtocolDecoder extends BaseProtocolDecoder {
                         if (length == 5) {
                             position.setDeviceTime(new Date(buf.readUnsignedIntLE() * 1000));
                         }
+                        if (alarm > 0) {
+                            position.set(Position.KEY_ALARM, alarm);
+                        }
                         break;
                     case 0x14:
                         position.set(Position.KEY_BATTERY_LEVEL, buf.readUnsignedByte());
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
```diff
@@ -180,6 +180,9 @@ protected Object decode(
                         if (length == 5) {
                             position.setDeviceTime(new Date(buf.readUnsignedIntLE() * 1000));
                         }
+                        if (alarm > 0) {
+                            position.set("alarmCode", alarm);
```
tananaev, line 184: Why not decode it properly? Are you missing some specific alarms?

markdegrootnl: There are 18 alarms available for Minifinder2 devices, but most of them are quite specific. 
I could decode them properly but this would add a lot of `ALARM_...` fields in Position.java. 
Or I could hardcode the alarm names in the Minifinder2 decoder. What would you prefer?

tananaev: What alarms are we talking about? Please provide the protocol documentation.

markdegrootnl: Scherm­afbeelding 2024-11-09 om 14 50 14

tananaev: A lot of these can be mapped to the standard alerts. And some of them don't seem like they should be alerts. For example, we can probably save things like "motion" and "no motion" as the standard motion attribute.

markdegrootnl: I agree, but I also think it would be nice to have the raw alarmCode for the edge cases. The same way this is done with the status information in line 260:
https://github.com/traccar/traccar/blob/2c14a8ebd162cfa21d4e366a1d3d988c0423b320/src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java#L248-L261



I'm more than happy to implement the alerts that can be mapped to standard alerts. How about:

| Bit No. | Description                       | Already implemented as | To implement as                                |
|---------|-----------------------------------|------------------------|------------------------------------------------|
| 0       | Low Battery                       | `ALARM_LOW_BATTERY`    |                                                |
| 1       | Overspeed Alert                   | `ALARM_OVERSPEED`      |                                                |
| 2       | Fall Down Alert                   | `ALARM_FALL_DOWN`      |                                                |
| 3       | Tilt Alert                        |                        | _Maybe add new alarm type?_                      |
| 4       | GEO-1 Alert                       |                        | `ALARM_GEOFENCE_ENTER` `ALARM_GEOFENCE_EXIT` * |
| 5       | GEO-2 Alert                       |                        | `ALARM_GEOFENCE_ENTER` `ALARM_GEOFENCE_EXIT` * |
| 6       | GEO-3 Alert                       |                        | `ALARM_GEOFENCE_ENTER` `ALARM_GEOFENCE_EXIT` * |
| 7       | GEO-4 Alert                       |                        | `ALARM_GEOFENCE_ENTER` `ALARM_GEOFENCE_EXIT` * |
| 8       | Power Off Alert                   | `ALARM_POWER_OFF`      |                                                |
| 9       | Power On Alert                    | `ALARM_POWER_ON`       |                                                |
| 10      | Motion Alert                      |                        | `ALARM_MOVEMENT`                               |
| 11      | No-Motion Alert                   |                        | _Maybe add new alarm type?_                      |
| 12      | SOS-Key                           | `ALARM_SOS`            |                                                |
| 13      | Side1 Key                         |                        |                                                |
| 14      | Side2 Key                         |                        |                                                |
| 15      | Battery charging “Start charging” |                        | `Position.KEY_CHARGE` true                     |
| 16      | Battery no charging               |                        | `Position.KEY_CHARGE` false                    |
| 17      | SOS Completed                     |                        |                                                |
| 18      | Reserved                          |                        |                                                |
| 19      | Reserved                          |                        |                                                |
| 20      | Reserved                          |                        |                                                |
| 21      | Fall down Alert Completed         |                        |                                                |
| 22      | Reserved                          |                        |                                                |
| 23      | Reserved                          |                        |                                                |
| 24      | Reserved                          |                        |                                                |
| 25      | Reserved                          |                        |                                                |

*) using bits 26-29 to determine the direction of the geo alerts.

tananaev: - Motion Alert should map to "motion" = true (can also add as an alarm)
- No-Motion Alert should map to "motion" = false
- Side1/2 Key - can probably add a custom attribute for this
- Tilt Alert - we can probably add a new alarm; I think I've seen it in other devices as well

markdegrootnl: Not sure how to add the Tilt alarm, is it sufficient to add a field in Position.java?

tananaev: No, you have to add it to the template file as well and localization in the web app repo.

markdegrootnl:
Aside from the Tilt alarm, would it be possible to merge this PR?

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
```diff
@@ -165,21 +165,57 @@ protected Object decode(
                         if (BitUtil.check(alarm, 2)) {
                             position.addAlarm(Position.ALARM_FALL_DOWN);
                         }
+                        if (BitUtil.check(alarm, 4)) {
+                            position.addAlarm(BitUtil.check(alarm, 26) ? Position.ALARM_GEOFENCE_ENTER : Position.ALARM_GEOFENCE_EXIT);
+                            position.set("geo", 1);
```
tananaev, line 170: Let's call it `geofence` instead of `geo`.

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
```diff
@@ -165,21 +165,57 @@ protected Object decode(
                         if (BitUtil.check(alarm, 2)) {
                             position.addAlarm(Position.ALARM_FALL_DOWN);
                         }
+                        if (BitUtil.check(alarm, 4)) {
```
tananaev, line 168: Can we do a for look for the geofences?

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
```diff
@@ -165,21 +165,57 @@ protected Object decode(
...
+                        }
                         if (BitUtil.check(alarm, 8)) {
                             position.addAlarm(Position.ALARM_POWER_OFF);
                         }
                         if (BitUtil.check(alarm, 9)) {
                             position.addAlarm(Position.ALARM_POWER_ON);
                         }
+                        if (BitUtil.check(alarm, 10)) {
+                            position.addAlarm(Position.ALARM_MOVEMENT);
+                            position.set(Position.KEY_MOTION, true);
```
tananaev, line 192: You shouldn't use both here.

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
```diff
@@ -165,21 +165,57 @@ protected Object decode(
...
+                            position.set(Position.KEY_MOTION, false);
+                        }
                         if (BitUtil.check(alarm, 12)) {
                             position.addAlarm(Position.ALARM_SOS);
                         }
+                        if (BitUtil.check(alarm, 13)) {
+                            position.set("side1", true);
+                        }
+                        if (BitUtil.check(alarm, 14)) {
+                            position.set("side2", true);
```
tananaev, line 204: What are these?

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
```diff
@@ -165,21 +165,57 @@ protected Object decode(
...
+                        if (BitUtil.check(alarm, 16)) {
+                            position.set(Position.KEY_CHARGE, false);
+                        }
                         if (BitUtil.check(alarm, 31)) {
                             position.set("bark", true);
                         }
                         if (length == 5) {
                             position.setDeviceTime(new Date(buf.readUnsignedIntLE() * 1000));
                         }
+                        position.set("alarmCode", alarm);
```
tananaev, line 218: Use key_event.

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
```diff
@@ -248,9 +284,7 @@ protected Object decode(
                     case 0x24:
                         position.setTime(new Date(buf.readUnsignedIntLE() * 1000));
                         long status = buf.readUnsignedIntLE();
-                        if (BitUtil.check(status, 4)) {
-                            position.set(Position.KEY_CHARGE, true);
-                        }
+                        position.set(Position.KEY_CHARGE, BitUtil.check(status, 4));
```
tananaev, line 287: Why are you changing this?

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
markdegrootnl: I think Bit 10 and 11 are actual Motion / No Motion alarms, and do not indicate whether the device is moving. 
You can configure the time after which the alarms trigger.

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
markdegrootnl: The documentation refers to "Side1 Key" and "Side2 Key". Not sure what they do. 
I'm fine with removing them.

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
markdegrootnl: Bit 4 of the status message always indicates if the device is charging or not. In stead of only adding the `true` state when it's charging, my change passes the actual state along.

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
```diff
@@ -165,21 +165,40 @@ protected Object decode(
...
                             position.addAlarm(Position.ALARM_POWER_OFF);
                         }
                         if (BitUtil.check(alarm, 9)) {
                             position.addAlarm(Position.ALARM_POWER_ON);
                         }
+                        if (BitUtil.check(alarm, 10)) {
+                            position.addAlarm(Position.ALARM_MOVEMENT);
+                        }
+                        if (BitUtil.check(alarm, 11)) {
+                            position.addAlarm(Position.ALARM_IDLE);
```
tananaev, line 184: This doesn't look correct.

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
tananaev: Let's remove if we don't know.

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
tananaev: This is one of the attributes that we only include when it's true, so the original code was correct.

src/main/java/org/traccar/protocol/Minifinder2ProtocolDecoder.java
markdegrootnl: Why not? 
Bit 10 is `Motion Alert` and bit 11 is `No-Motion Alert`.

markdegrootnl: Ah got it. I don't think there is a suitable alarm for the `No-Motion Alert`, right?
So what do we do? Delete it for now?

tananaev: Delete it for now

tananaev:
Looks like some checks have failed.

markdegrootnl:
Thank you so much for engaging with me and your quick replies!

PR #5466, opened 2024-11-14
Fix: Set UTF-8 Encoding for Proper Rendering of Chinese Characters in Server Title

### LINKED ISSUE #5465: Server Title in Chinese Appears as Garbled Text Due to Missing charset=utf-8 in Response
When setting the server title to a Chinese string in the Traccar web interface, the title appears as garbled characters in the browser. This issue occurs because the OverrideFilter class does not correctly set the response's Content-Type header with the proper charset=utf-8 encoding.

## Steps to Reproduce:
Set the server title to a Chinese string (e.g., 服务器).
Access the Traccar web interface.
The server title displayed in the browser is garbled or not rendered correctly.
## Expected Behavior:
The server title should be rendered correctly in Chinese, without any garbled characters or encoding issues.

## Actual Behavior:
The server title is displayed as garbled characters, which indicates an encoding issue.

## Cause:
The OverrideFilter does not set the response's Content-Type header with charset=utf-8. This results in the browser not interpreting the content as UTF-8 encoded, leading to incorrect rendering of Chinese characters.

zhanghx0905:
The server title in the Traccar web interface was displaying as garbled characters when set to Chinese strings due to a missing charset=utf-8 in the response Content-Type header. This change ensures that the server title is rendered correctly by explicitly setting the response's character encoding to UTF-8.

Changes:
- Modified the OverrideFilter class to set response.setCharacterEncoding("UTF-8").
- Ensured proper rendering of the title/description in the web interface.

Closes #5465

### DIFF
--- a/src/main/java/org/traccar/web/OverrideFilter.java
+++ b/src/main/java/org/traccar/web/OverrideFilter.java
@@ -49,7 +49,7 @@ public class OverrideFilter implements Filter {
             chain.doFilter(request, response);
             return;
         }
-        response.setCharacterEncoding("UTF-8");
+
         ResponseWrapper wrappedResponse = new ResponseWrapper((HttpServletResponse) response);
 
         chain.doFilter(request, wrappedResponse);
--- a/src/main/java/org/traccar/web/OverrideFilter.java
+++ b/src/main/java/org/traccar/web/OverrideFilter.java
@@ -30,6 +30,7 @@ import jakarta.servlet.ServletResponse;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 
 @Singleton
 public class OverrideFilter implements Filter {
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/web/OverrideFilter.java
```diff
@@ -49,7 +49,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             chain.doFilter(request, response);
             return;
         }
-
+        response.setCharacterEncoding("UTF-8");
```
tananaev, line 52: What if it's not a text? Does it work without it? What's the default value for you?

zhanghx0905: > What if it's not a text? Does it work without it? What's the default value for you?

You are right. This line is unnecessary.

src/main/java/org/traccar/web/OverrideFilter.java
```diff
@@ -70,12 +70,12 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
                 String description = server.getString("description", "Traccar GPS Tracking System");
                 String colorPrimary = server.getString("colorPrimary", "#1a237e");
 
-                String alteredContent = new String(wrappedResponse.getCapture())
+                String alteredContent = new String(wrappedResponse.getCapture(), "UTF-8")
```
tananaev, line 73: Can we use `StandardCharsets.UTF_8` instead of the hardcoded string?

src/main/java/org/traccar/web/OverrideFilter.java
```diff
@@ -70,12 +70,12 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
                         .replace("${title}", title)
                         .replace("${description}", description)
                         .replace("${colorPrimary}", colorPrimary);
 
-                byte[] data = alteredContent.getBytes();
+                byte[] data = alteredContent.getBytes("UTF-8");
```
tananaev, line 78: And same here.

PR #5469, opened 2024-11-21
OpenAPI yaml: User expirationTime nullable: true

adsmith144:
This resolves the issue described in the linked forum thread:
[Link to Traccar Forum topic](https://www.traccar.org/forums/topic/traccar-openapi-client-getusers-expirationtime-can-be-null-but-is-not-marked-as-nullable/
)

Tested and confirmed.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1803,6 +1803,7 @@ components:
           type: string
           description: in IS0 8601 format. eg. `1963-11-22T18:30:00Z`
           format: date-time
+          nullable: true
         deviceLimit:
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -1803,6 +1803,7 @@ components:
           type: string
           description: in IS0 8601 format. eg. `1963-11-22T18:30:00Z`
           format: date-time
+          nullable: true
```
tananaev, line 1806: Can you also update it on the device model?

PR #5471, opened 2024-11-26
Redis cache implementation - Similar to InMemory Cache

bilalcs:
The idea of having a distributed cache (like Redis) is to make traccar more scalable. 
I am not very much familiar with core structure of traccar, so there may be a better way to achieve it. But here is what I changed:

debug.xml: Added entries for redis server
MainModule.java: Add capability to initialize custom cache manager (In future, any type of cache can be initialized there). For this, I removed @Singleton from original Cache class.
CacheManager.java and CacheGraph.java are now interfaces. The implementation type of cache need to implement them in order to bring a new type of cache.

CacheManager and CacheGraph have two implementations; InMemory (existing one) and Redis (new one) with names InMemoryCacheManager, InMemoryCacheGraph and RedisCacheManager, RedisCacheGraph respectively.

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -27,4 +27,13 @@
     <entry key='atrack.custom'>true</entry>
     <entry key='intellitrac.port'>6037</entry>
 
+    <!--<entry key='forward.type'>redis</entry>
+    <entry key='forward.url'>redis://localhost:6379</entry>
+
+    <entry key='cache.type'>redis</entry>
+    <entry key='cache.host'>localhost</entry>
+    <entry key='cache.port'>6379</entry>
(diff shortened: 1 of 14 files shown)


tananaev:
How does Redis cache make it more scalable?

bilalcs:
- Distributed architecture, enabling horizontal scaling across multiple nodes, in-memory Hashtables cant.
- In DR cases, in-memory cache will lost (even in restarts).
- Replication for fault tolerance and load balancing.
- Independent cache, so it is application/language agnostic (for future use cases)

tananaev:
We already support horizontal scaling. I think you should probably study what's available first.

PR #5472, opened 2024-12-03
Don't send geofence notifications for users who don't have access to …

jcardus:
…the geofence

### DIFF
--- a/src/main/java/org/traccar/database/NotificationManager.java
+++ b/src/main/java/org/traccar/database/NotificationManager.java
@@ -140,6 +140,13 @@ public class NotificationManager {
                         LOGGER.info("User {} doesn't have access to geofence {}", user.getId(), event.getGeofenceId());
                         return;
                     }
+                    if (event.getMaintenanceId() != 0
+                            && cacheManager.getUserObjects(user.getId(), Maintenance.class).stream()
+                            .noneMatch(g -> g.getId() == event.getMaintenanceId())) {
+                        LOGGER.info("User {} doesn't have access to maintenance {}",
+                            user.getId(), event.getMaintenanceId());
+                        return;
+                    }
                     for (String notificator : notification.getNotificatorsTypes()) {
                         try {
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -134,6 +134,12 @@ private void updateEvent(Event event, Position position) {
                         LOGGER.info("User {} notification blocked", user.getId());
                         return;
                     }
+                    if (event.getGeofenceId() != 0
+                        && cacheManager.getUserObjects(user.getId(), Geofence.class).stream()
+                            .noneMatch(g -> g.getId() == event.getGeofenceId())) {
```
tananaev, line 139:
```suggestion
                                && cacheManager.getUserObjects(user.getId(), Geofence.class).stream()
                                .noneMatch(g -> g.getId() == event.getGeofenceId())) {
```

src/main/java/org/traccar/database/NotificationManager.java
```diff
@@ -134,6 +134,12 @@ private void updateEvent(Event event, Position position) {
...
+                        LOGGER.info("Geofence {} not in user {}", event.getGeofenceId(), user.getId());
```
tananaev, line 140: Probably better word something like "User X doesn't have access to geofence Y".

tananaev:
If we're doing this, probably makes sense to do a check for maintenances as well.

jcardus:
I copied the getUserObjects from getDeviceObjects, what does proxies and forward parameters do in getObjects?

tananaev:
> I copied the getUserObjects from getDeviceObjects, what does proxies and forward parameters do in getObjects?

Have you tested all the cases? This question makes me worry that you haven't actually tested it properly.

jcardus:
That's why I asked about it. Yes, this needs more testing. I'll close the PR for now.

tananaev:
In the future please only send a pull request for review when it's ready and thoroughly tested. If you want to discuss something related to a change, you can open an issue instead.

PR #5474, opened 2024-12-06
X3Tech NT20 new location message

jcardus:
[NT11 NT20 NT26 NT40 NT46 Protocol rev3.33.pdf](https://github.com/user-attachments/files/18044962/NT11.NT20.NT26.NT40.NT46.Protocol.rev3.33.pdf)

### DIFF
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
@@ -669,7 +669,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
                 || type == MSG_LBS_EXTEND || type == MSG_LBS_WIFI || type == MSG_LBS_2 || type == MSG_LBS_3
                 || (type == MSG_WIFI_3 && variant != Variant.LW4G) || type == MSG_WIFI_5) {
 
-            getLastLocation(position, getDate(buf, deviceSession));
+            getLastLocation(position, decodeDate(buf, deviceSession));
 
             if (variant == Variant.WANWAY_S20 || variant == Variant.SL4X) {
                 buf.readUnsignedByte(); // ta
@@ -796,7 +796,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
             if (type == MSG_GPS_LBS_2 && modelNt20) {
                 buf.readUnsignedByte(); // location source type
                 buf.skipBytes(8); // imei
-                position.setDeviceTime(getDate(buf, deviceSession));
+                position.setDeviceTime(decodeDate(buf, deviceSession));
             }
 
             if (hasGps(type)) {
@@ -1019,7 +1019,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
                         position, buf, false, jc400, jc400, jc400, jc400,
                         deviceSession.get(DeviceSession.KEY_TIMEZONE));
             } else {
-                getLastLocation(position, getDate(buf, deviceSession));
+                getLastLocation(position, decodeDate(buf, deviceSession));
             }
             if (variant == Variant.JC400) {
                 position.set(Position.KEY_POWER, buf.readUnsignedShort() * 0.1);
@@ -1062,7 +1062,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
         return position;
     }
 
-    private static Date getDate(ByteBuf buf, DeviceSession deviceSession) {
+    private static Date decodeDate(ByteBuf buf, DeviceSession deviceSession) {
         DateBuilder dateBuilder = new DateBuilder((TimeZone) deviceSession.get(DeviceSession.KEY_TIMEZONE))
                 .setDate(buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte())
                 .setTime(buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte());
@@ -1149,7 +1149,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
             } else if (subType == 0x05) {
 
                 if (buf.readableBytes() >= 6 + 1 + 6) {
-                    position.setDeviceTime(getDate(buf, deviceSession));
+                    position.setDeviceTime(decodeDate(buf, deviceSession));
                 }
 
                 int flags = buf.readUnsignedByte();
@@ -1252,7 +1252,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
 
         } else if (type == MSG_OBD) {
 
-            getLastLocation(position, getDate(buf, deviceSession));
+            getLastLocation(position, decodeDate(buf, deviceSession));
 
             position.set(Position.KEY_IGNITION, buf.readUnsignedByte() > 0);
 
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
@@ -847,7 +847,10 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
                     if (type == MSG_STATUS && "R11".equals(model)) {
                         position.set(Position.KEY_POWER, buf.readUnsignedShort() * 0.01);
                     } else {
-                        int battery = "NT20".equals(model) ? (int) (buf.readUnsignedByte() * 0.1) : buf.readUnsignedByte();
+                        int battery = buf.readUnsignedByte();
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -786,6 +786,29 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
...
+            position.set(Position.KEY_BATTERY, buf.readUnsignedByte() * 0.1);
+            position.set(Position.KEY_RSSI, buf.readUnsignedByte());
+            position.addAlarm(decodeAlarm(buf.readUnsignedByte(), modelLW));
+
+            buf.readUnsignedByte(); // language
+
+            position.set(Position.KEY_ODOMETER, buf.readMedium());
+            position.set(Position.KEY_HOURS, buf.readMedium() * 60 * 1000);
+
         } else if (isSupported(type)) {
```
tananaev, line 812: You should decode it as a part of this block.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -794,22 +794,33 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
...
+            if ("NT20".equals(model)) {
+                buf.readUnsignedByte(); // location source type
+                buf.skipBytes(8); // imei
+
+                DateBuilder dateBuilder = new DateBuilder((TimeZone) deviceSession.get(DeviceSession.KEY_TIMEZONE))
+                        .setDate(buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte())
+                        .setTime(buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte());
+
+                position.setDeviceTime(dateBuilder.getDate());
+            }
```
tananaev, line 806: Why do we need this instead of just using `decodeGps`?

jcardus: it's not instead, decodeGps is called after that.

jcardus: that's not related to decodeGps, we call that after

tananaev: I see. There are two timestamps. But it looks like it only applies to message type 0x22.

jcardus: That message is used in other models, apparently with a different structure. The tests will fail if we check the message type instead of the model.

tananaev: Yes, I mean that we need to check both. Make sure your decoding logic is valid for the whole protocol.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -794,22 +794,34 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
 
             position.set(Position.KEY_TYPE, type);
 
+            if (type == MSG_GPS_LBS_2 && "NT20".equals(model)) {
```
tananaev, line 797: Let's maybe create a variable for `"NT20".equals(model)`, so we don't need to evaluate it so many times. Like `isNt20` or something.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -794,22 +794,34 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
...
+                buf.readUnsignedByte(); // location source type
+                buf.skipBytes(8); // imei
+
+                DateBuilder dateBuilder = new DateBuilder((TimeZone) deviceSession.get(DeviceSession.KEY_TIMEZONE))
+                        .setDate(buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte())
+                        .setTime(buf.readUnsignedByte(), buf.readUnsignedByte(), buf.readUnsignedByte());
```
tananaev, line 803: To avoid duplication, I suggest we extract this into a method that returns a `Date` object.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -794,22 +794,34 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
...
             if (hasGps(type)) {
                 decodeGps(position, buf, false, deviceSession.get(DeviceSession.KEY_TIMEZONE));
             } else {
                 getLastLocation(position, null);
             }
 
             if (hasLbs(type) && buf.readableBytes() > 6) {
-                boolean hasLength = hasStatus(type)
+                boolean hasLength = (hasStatus(type)
+                        || ("NT20".equals(model) && type == MSG_GPS_LBS_2))
```
tananaev, line 816: Seems like this part should inside `hasStatus` check.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -830,13 +842,16 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
...
-                    if (type == MSG_GPS_LBS_STATUS_5) {
+                    if (type == MSG_GPS_LBS_STATUS_5 || ("NT20".equals(model) && type == MSG_GPS_LBS_2)) {
                         position.set(Position.KEY_POWER, buf.readUnsignedShort() * 0.01);
                     }
                     if (type == MSG_STATUS && "R11".equals(model)) {
                         position.set(Position.KEY_POWER, buf.readUnsignedShort() * 0.01);
                     } else {
                         int battery = buf.readUnsignedByte();
+                        if ("NT20".equals(model) && type == MSG_GPS_LBS_2) {
+                            battery = (int) (battery * 0.1);
```
tananaev, line 853:
```suggestion
                            battery = battery / 10;
```

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -978,7 +993,8 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
                 buf.readUnsignedByte(); // validity
             }
 
-            if (buf.readableBytes() == 3 + 6 || buf.readableBytes() == 3 + 4 + 6) {
+            if ((buf.readableBytes() == 3 + 6 || buf.readableBytes() == 3 + 4 + 6)
+                    && !("NT20".equals(model) && type == MSG_GPS_LBS_2)) {
```
tananaev, line 997: Couldn't we just move the block that's now below to above. And it will solve this problem?

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
jcardus: HasStatus only receives message type

tananaev: Currently yes.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -217,7 +217,7 @@ private static boolean hasLbs(int type) {
         }
     }
 
-    private static boolean hasStatus(int type) {
+    private static boolean hasStatus(int type, boolean modelNt20) {
```
tananaev, line 220: No, I think we should pass the model.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -1051,6 +1062,13 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
         return position;
     }
 
+    private static Date getDate(ByteBuf buf, DeviceSession deviceSession) {
```
tananaev, line 1065: I would rename this to `decodeDate` or `readDate`. Because it's not only gets the date, but also increments indexes in the buffer.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
jcardus: Yes, I thought of that, but then I would hardcode NT20 twice. Maybe I can create a function for the modelNt20.

tananaev: I think hardcoding in two places is ok. If it goes beyond, we can have a constant or something.

PR #5478, opened 2024-12-14
Fix GTCAN in GV355CEU

jcardus:
[GV355CEU@Track_Air_Interface_Protocol_v0704.pdf](https://github.com/user-attachments/files/18134033/GV355CEU%40Track_Air_Interface_Protocol_v0704.pdf)


tananaev:
Needs a description.

jcardus:
An exception was being thrown when parsing adBlueLevel (when it's present) in GTCAN message.

PR #5482, opened 2024-12-19
Add custom groups scope claim name

Klucherev:
Some OIDC providers don't have "groups" claim in id token, but have "roles". With this setting in traccar.xml there is ability to change it to work properly.

Example is
` <entry key='openid.groupsClaimName'>roles</entry>`


tananaev:
Which some providers and do you have some documentation references?

Klucherev:
@tananaev i use logto.io as OIDC provider and there is no "groups" claim in token. Only roles with user roles name array

tananaev:
What about documentation?

Klucherev:
Groups claim is not standard claim. Auth0 also have problems to add it. I've read whole documentation of logto.io and there is no such ability. The only thing that i can to do is add custom claim to access token, not id token

src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -111,13 +114,14 @@ public OpenIdProvider(Config config, LoginService loginService, HttpClient httpC
...
         allowGroup = config.getString(Keys.OPENID_ALLOW_GROUP);
+        groupsClaimName = config.getString(Keys.OPENID_GROUPS_CLAIM_NAME);
     }
 
     public URI createAuthUri() {
         Scope scope = new Scope("openid", "profile", "email");
 
         if (adminGroup != null) {
-            scope.add("groups");
+            scope.add(Objects.requireNonNullElse(groupsClaimName, "groups"));
```
tananaev, line 124: Can we just use `"groups"` as the default value?

Klucherev: Yes. Absolutely. requireNonNullElse sets default value if groupsClaimName is not provided.

tananaev: Yes, but we have a standard mechanism we use everywhere else.

Klucherev: if condition?

tananaev: No, configuration keys can have default values.

PR #5484, opened 2024-12-19
Add quantity of fuel drop/increase in notification

hypoclone:
Added quantity of fuel drop/increase in email notification

### DIFF
--- a/src/main/java/org/traccar/handler/events/FuelEventHandler.java
+++ b/src/main/java/org/traccar/handler/events/FuelEventHandler.java
@@ -23,6 +23,8 @@ import org.traccar.model.Device;
 import org.traccar.model.Event;
 import org.traccar.model.Position;
 import org.traccar.session.cache.CacheManager;
+import java.util.HashMap;
+import java.util.Map;
 
 public class FuelEventHandler extends BaseEventHandler {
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/traccar/handler/events/FuelEventHandler.java
```diff
@@ -50,18 +52,31 @@ public void onPosition(Position position, Callback callback) {
...
 
                 if (change > 0) {
                     double threshold = AttributeUtil.lookup(
                             cacheManager, Keys.EVENT_FUEL_INCREASE_THRESHOLD, position.getDeviceId());
                     if (threshold > 0 && change >= threshold) {
-                        callback.eventDetected(new Event(Event.TYPE_DEVICE_FUEL_INCREASE, position));
+                        Map<String, Object> attributes = new HashMap<>();
+                        attributes.put("before", before);
+                        attributes.put("after", after);
+                        attributes.put("change", change);
```
tananaev, line 64: I don't think it makes sense to have all three parameters. I think either before/after or just the change would be enough.

hypoclone:
Thank you @tananaev, I think before/after should be fine, the files have been updated

templates/full/deviceFuelDrop.vm
```diff
@@ -4,6 +4,8 @@
 <body>
 Device: $device.name<br>
 Time: $dateTool.format("YYYY-MM-dd HH:mm:ss", $event.eventTime, $locale, $timezone)<br>
+Quantity Before: $!event.attributes.before<br>
```
tananaev, line 7: What does the exclamation point do here?

src/main/java/org/traccar/handler/events/FuelEventHandler.java
```diff
@@ -55,13 +58,23 @@ public void onPosition(Position position, Callback callback) {
                     double threshold = AttributeUtil.lookup(
                             cacheManager, Keys.EVENT_FUEL_INCREASE_THRESHOLD, position.getDeviceId());
                     if (threshold > 0 && change >= threshold) {
-                        callback.eventDetected(new Event(Event.TYPE_DEVICE_FUEL_INCREASE, position));
+                        Map<String, Object> attributes = new HashMap<>();
+                        attributes.put("before", before);
+                        attributes.put("after", after);
+                        Event event = new Event(Event.TYPE_DEVICE_FUEL_INCREASE, position);
+                        event.setAttributes(attributes);
+                        callback.eventDetected(event);
```
tananaev, line 66: Instead of providing attributes as a map, you should use `set` function on the even object.

templates/full/deviceFuelDrop.vm
hypoclone: It was supposed to prevent error if null was returned, zero is used instead

tananaev:
Merged, thank you.

PR #5493, opened 2025-01-14
OpenAPI yaml: Set nullable fields as nullable (Device, User schemas)

adsmith144:
### Background

2 months ago I opened a [PR](https://github.com/traccar/traccar/pull/5469) for a specific case of this issue, which was accepted and solved the issue. Now with more experience of Traccar, I address other cases of the same issue. 

This PR marks several fields as nullable in the OpenAPI schema, matched to fields which are regularly null in the traccar server db under normal operation. This is limited to models I am familiar with and anticipate using the API for.

### Issue - current example

Parsing error during `get_devices.sync()` while parsing `last_update (= None)`:

```py 
# Generated python client library snippet (Device)
    @classmethod
    def from_dict(cls: Type[T], src_dict: Dict[str, Any]) -> T:
        from ..models.device_attributes import DeviceAttributes

        d = src_dict.copy()
        ...
        _last_update = d.pop("lastUpdate", UNSET)
        last_update: Union[Unset, datetime.datetime]
        if isinstance(_last_update, Unset):  # evaluates False since "lastUpdate" exists in src_dict and is None
            last_update = UNSET
        else:
            last_update = isoparse(_last_update)  #  error parsing None
```

### Considerations

Several other nullable fields are not marked as nullable in the OpenAPI schemas for Device and User, which could cause similar issues, e.g.:

```yaml
    Device:
      type: object
        ...
        lastUpdate:
          type: string
          description: in IS0 8601 format. eg. `1963-11-22T18:30:00Z`
          format: date-time
        positionId:
          type: integer
        groupId:
          type: integer
        phone:
          type: string
```


tananaev:
Merged, thank you.

Issue #5486, opened 2024-12-31
OwnTracks Bad Request no log entries


pantherale0:
**Describe the bug**
Using OwnTracks as a tracker and Caddy as a reverse proxy (and providing SSL) for the HTTP server running at port 5144, it seems the server returns a HTTP 400 status code, however there are no log entries that show that actually help resolving the issue

**To Reproduce**
1. Setup latest traccar in docker container
2. Setup Caddy server using following config for http app
```
"traccar-owntracks": {
          "listen": [":5144"],
          "routes": [
            {
              "handle": [
                {
                  "handler": "reverse_proxy",
                  "transport": {
                    "protocol": "http"
                  },
                  "upstreams": [
                    {
                      "dial": "127.0.0.1:5144"
                    }
                  ]
                }
              ],
              "match": [
                {
                  "host": [
                    "traccar.example.com"
                  ]
                }
              ],
              "terminal": true
            }
          ]
        }
```
3. Configure OwnTracks in HTTP mode:
`Endpoint = https://traccar.example.com`
`Device ID = 1234`
`Tracker ID = 1`
`Username = example`
4. Create device in Traccar with identifier `owntracks/example/1234`
5. Wait and check logs

**Expected behavior**
An entry in the log file that states why a 400 status code is being raised, but this doesn't happen.

**Screenshots**
N/A, log below
```
2024-12-31 10:24:49  INFO: [Ta2e08722: owntracks < 172.16.55.205] POST / HTTP/1.1\r\nHost: traccar.example.com:5144\r\nUser-Agent: Owntracks-Android/gms/420503003\r\nContent-Length: 127\r\nAccept-Encoding: gzip\r\nContent-Type: application/json; charset=utf-8\r\nX-Forwarded-For: 140.228.92.32\r\nX-Forwarded-Host: traccar.example.com:5144\r\nX-Forwarded-Proto: https\r\nX-Limit-D: 1234\r\nX-Limit-U: jordan\r\n\r\n{"_type":"status","_id":"6491e098","android":{"hib":1,"bo":0,"loc":0,"ps":1,"wifi":1},"topic":"owntracks/example/1234/status"}
2024-12-31 10:24:49  INFO: [Ta2e08722: owntracks > 172.16.55.205] HTTP/1.1 400 Bad Request\r\ncontent-length: 0\r\n\r\n
```

**Smartphone (please complete the following information):**
 - Device: [e.g. iPhone6] S23 Ultra
[...]

PR #5494, opened 2025-01-15
Complete OwnTracks decoder cleanup with updated test cases #5486

ckrey:


src/test/java/org/traccar/protocol/OwnTracksProtocolDecoderTest.java
```diff
@@ -12,16 +12,22 @@ public void testDecode() throws Exception {
         var decoder = inject(new OwnTracksProtocolDecoder(null));
 
         verifyPosition(decoder, request(HttpMethod.POST, "/",
-                buffer("{\"_type\":\"location\",\"acc\":15,\"alt\":440,\"batt\":46,\"conn\":\"w\",\"lat\":46.0681247,\"lon\":11.1512805,\"t\":\"u\",\"tid\":\"5t\",\"tst\":1551874878,\"vac\":2,\"vel\":0}")));
+                buffer("{\"_type\":\"location\",\"topic\":\"owntracks/user/5t\",\"acc\":15,\"alt\":440,\"batt\":46,\"conn\":\"w\",\"lat\":46.0681247,\"lon\":11.1512805,\"t\":\"u\",\"tid\":\"5t\",\"tst\":1551874878,\"vac\":2,\"vel\":0}")));
```
tananaev, line 15: Why are we changing existing test cases?

ckrey:
We are changing existing test cases because you suggested to alway use `topic` and not `tid` as `uniqueId`. The old test cases contained only `tid`:

Bildschirmfoto 2025-01-19 um 11 37 56

if you revert the code back to the old version as suggested in the issue, we can use the old and new test cases:
Bildschirmfoto 2025-01-19 um 11 44 25

tananaev:
You cannot change the test cases to suit your code. You have to change the code to handle test cases. I still don't understand how it should work.

ckrey:
I followed your advice and restored the old test cases, keeping the 2 new test cases. The code is adapted to handle both kinds of messages:
If messages contains `"topic"`, this is used as uniqueid. If not, `"tid"` is used as uniqueid as a fallback.

src/main/java/org/traccar/protocol/OwnTracksProtocolDecoder.java
```diff
@@ -71,6 +80,9 @@ protected Object decode(
         position.setLatitude(root.getJsonNumber("lat").doubleValue());
         position.setLongitude(root.getJsonNumber("lon").doubleValue());
 
+        if (root.containsKey("tid")) {
+                position.set("tid", root.getString("tid"));
```
tananaev: Indentation is wrong here. Also why do we need to save tid?

src/main/java/org/traccar/protocol/OwnTracksProtocolDecoder.java
```diff
@@ -51,7 +51,16 @@ protected Object decode(
...
         }
 
-        String uniqueId = root.getString("tid");
+        String uniqueId;
+        if (root.containsKey("topic")) {
+            uniqueId = root.getString("topic");
+        } else if (root.containsKey("tid")) {
+            uniqueId = root.getString("tid");
+        } else {
+            sendResponse(channel, HttpResponseStatus.BAD_REQUEST);
```
tananaev: Is this a real case?

src/main/java/org/traccar/protocol/OwnTracksProtocolDecoder.java
ckrey: tid is not necessary to be saved per position, removed

src/main/java/org/traccar/protocol/OwnTracksProtocolDecoder.java
ckrey: yes, older devices only report "tid", newer devices include the more granular "topic". Both types are covered in test cases.

tananaev: I think you commented on the wrong thing. I'm asking about line 60.

ckrey: Understand your point now. You are correct, this is not a real case. 
If no id (either topic or tid) is in the message, getDeviceSession will fail anyhow.

PR #5502, opened 2025-01-28
Wialon: interpret LBS data from parameters

justedro:
According to the Wialon protocol specifications, the LBS data is transmitted as parameters. 
https://wialon.com/hw/files/Wialon%20IPS%20v.2.2%20%28RU%29.pdf

Added LBS decoding from the custom parameters
Added device-specific temp and battery level readings

### DIFF
--- a/src/main/java/org/traccar/model/ExtendedModel.java
+++ b/src/main/java/org/traccar/model/ExtendedModel.java
@@ -154,4 +154,7 @@ public class ExtendedModel extends BaseModel {
         }
     }
 
+    public void removeAttribute(String key) {
+        attributes.remove(key);
+    }
 }
--- a/src/main/java/org/traccar/model/ExtendedModel.java
+++ b/src/main/java/org/traccar/model/ExtendedModel.java
@@ -90,71 +90,98 @@ public class ExtendedModel extends BaseModel {
     }
 
     public String getString(String key, String defaultValue) {
-        if (attributes.containsKey(key)) {
-            Object value = attributes.get(key);
-            return value != null ? value.toString() : null;
-        } else {
-            return defaultValue;
-        }
+        return parseAsString(attributes.get(key), defaultValue);
     }
 
     public String getString(String key) {
-        return getString(key, null);
+        return parseAsString(attributes.get(key), null);
     }
 
     public double getDouble(String key) {
-        if (attributes.containsKey(key)) {
-            Object value = attributes.get(key);
-            if (value instanceof Number numberValue) {
-                return numberValue.doubleValue();
-            } else {
-                return Double.parseDouble(value.toString());
-            }
-        } else {
-            return 0.0;
-        }
+        return parseAsDouble(attributes.get(key), 0.0);
     }
 
     public boolean getBoolean(String key) {
-        if (attributes.containsKey(key)) {
-            Object value = attributes.get(key);
-            if (value instanceof Boolean booleanValue) {
-                return booleanValue;
-            } else {
-                return Boolean.parseBoolean(value.toString());
-            }
-        } else {
-            return false;
-        }
+        return parseAsBoolean(attributes.get(key), false);
     }
 
     public int getInteger(String key) {
-        if (attributes.containsKey(key)) {
-            Object value = attributes.get(key);
-            if (value instanceof Number numberValue) {
-                return numberValue.intValue();
-            } else {
-                return Integer.parseInt(value.toString());
-            }
-        } else {
-            return 0;
-        }
+        return parseAsInteger(attributes.get(key), 0);
     }
 
     public long getLong(String key) {
-        if (attributes.containsKey(key)) {
-            Object value = attributes.get(key);
-            if (value instanceof Number numberValue) {
-                return numberValue.longValue();
-            } else {
-                return Long.parseLong(value.toString());
-            }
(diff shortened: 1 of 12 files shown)


src/main/java/org/traccar/model/ExtendedModel.java
```diff
@@ -154,4 +154,7 @@ public long getLong(String key) {
         }
     }
 
+    public void removeAttribute(String key) {
```
tananaev, line 157: If we're adding a new method like this, we should update all the existing places where we could use it currently. We should do a separate PR for that.

src/main/java/org/traccar/protocol/WialonProtocolDecoder.java
```diff
@@ -156,11 +154,54 @@ private Position decodePosition(Channel channel, SocketAddress remoteAddress, St
                     }
                 }
             }
+
+            // interpret accuracy
```
tananaev, line 158: Let's remove all the obvious comments like this

src/main/java/org/traccar/protocol/WialonProtocolDecoder.java
```diff
@@ -156,11 +154,54 @@ private Position decodePosition(Channel channel, SocketAddress remoteAddress, St
...
+
+            if (network.getCellTowers() != null) {
+                position.setNetwork(network);
+            }
         }
 
         return position;
     }
 
+    private static void interpretLBSAndRemoveAttribute(Position position, Network network, String suffix) {
```
tananaev, line 184: I would probably rename this to something like `decodeCellData` or something like that.

src/test/java/org/traccar/ProtocolTest.java
```diff
@@ -11,19 +11,14 @@
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.HttpVersion;
 import org.traccar.helper.DataConverter;
-import org.traccar.model.CellTower;
-import org.traccar.model.Command;
-import org.traccar.model.Position;
-import org.traccar.model.WifiAccessPoint;
+import org.traccar.model.*;
```
tananaev, line 14: We don't do star imports. And also let's separate all the test refactoring in a separate PR as well.

src/test/java/org/traccar/protocol/WialonProtocolDecoderTest.java
```diff
@@ -86,6 +90,18 @@ public void testDecode() throws Exception {
...
+                position(Collections.singletonList(CellTower.from(250, 2, 5901, 45542))));
+
+        verifyNetwork(
+                decoder,
+                text("#D#NA;NA;NA;NA;NA;NA;NA;NA;NA;NA;NA;NA;NA;;NA;mnc1:1:02,mcc1:1:250,lac1:1:5901,cell_id1:1:45542,mnc2:1:22,mcc2:1:222,lac2:1:2222,cell_id2:1:22222"),
+                position(Arrays.asList(
+                        CellTower.from(250, 2, 5901, 45542),
+                        CellTower.from(222, 22, 2222, 22222)
+                ))
+        );
```
tananaev, line 104: Closing brackets should be on the previous line. That's the style we use.

src/test/java/org/traccar/protocol/WialonProtocolDecoderTest.java
```diff
@@ -28,10 +30,10 @@ public void testDecode() throws Exception {
 
         verifyNull(decoder, text(
                 "#L#123456789012345;test"));
-
```
tananaev, line 31: Don't forget to undo all these changes and clean up your PR. Let me know once it's ready for another review.

src/main/java/org/traccar/model/Network.java
```diff
@@ -27,8 +27,10 @@ public class Network {
     public Network() {
     }
 
-    public Network(CellTower cellTower) {
-        addCellTower(cellTower);
+    public Network(CellTower... cellTowers) {
```
tananaev, line 30: If we're changing this, it probably makes sense to change the wifi constructor as well. For consistency

src/main/java/org/traccar/protocol/WialonProtocolDecoder.java
```diff
@@ -156,11 +154,52 @@ private Position decodePosition(Channel channel, SocketAddress remoteAddress, St
...
+            if (position.hasAttribute("bat")) {
+                position.set(Position.KEY_BATTERY_LEVEL, position.removeInteger("bat"));
+            }
+
+            if (position.hasAttribute("temp")) {
+                position.set(Position.KEY_DEVICE_TEMP, position.removeInteger("temp"));
+            }
+
+            Network network = new Network();
+            if (position.hasAttribute("mcc") || position.hasAttribute("mcc1")) {
```
tananaev, line 171: Why do we need this check?

src/main/java/org/traccar/model/Network.java
justedro: oh good point

src/main/java/org/traccar/protocol/WialonProtocolDecoder.java
justedro: no particular reason apart of a few cpu cycles saved sometimes. Removed

PR #5503, opened 2025-01-28
ProtocolTest - network attribute verification

justedro:
Added network to the list of attributes to verify in tests
Unlocks #5502

### DIFF
--- a/src/test/java/org/traccar/ProtocolTest.java
+++ b/src/test/java/org/traccar/ProtocolTest.java
@@ -13,6 +13,7 @@ import io.netty.handler.codec.http.HttpVersion;
 import org.traccar.helper.DataConverter;
 import org.traccar.model.CellTower;
 import org.traccar.model.Command;
+import org.traccar.model.Network;
 import org.traccar.model.Position;
 import org.traccar.model.WifiAccessPoint;
 
@@ -21,9 +22,7 @@ import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Collection;
-import java.util.Date;
 import java.util.List;
-import java.util.Map;
 import java.util.TimeZone;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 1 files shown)


src/test/java/org/traccar/ProtocolTest.java
```diff
@@ -326,7 +327,32 @@ private void verifyDecodedPosition(Object decodedObject, boolean checkLocation,
                 }
             }
         }
+    }
 
+    private void verifyExpectations(Object decodedObject, Position expected) {
```
tananaev, line 332: What is the difference with other verification methods? Why do we need a separate method for this?

justedro: If I understand the logic behind this class correctly, the private method `verifyDecodedPosition` serves as a general set of checks that is configured in all other protected methods. 

I checked its history and noticed it was continuously growing without any clear direction. This resulted in a very complex signature that I found rather challenging to extend. 

In essence it was doing two things: 
1) check if the decoded values in some fields make sense
2) compare the decoded values with exact values set in the parameter

I propose splitting this method into two independent ones: `verifyDecodedPosition` and `verifyExpectations`, respectively. The existing protected methods were adapted to this approach and retained their semantics. 

This also allows me to create a reusable method that would allow me to write isolated tests for the Network data (#5502) and keep the "style" the tests written in

tananaev: What does `verifyExpectations` verify? It is not clear at all from the naming. How is it different from `verifyDecodedPosition`? I don't mind refactoring, but this doesn't seem to make it much better than it was.

justedro: I'm sorry. I just realized that I missed the method `verifyAttribute`, which actually works the way I need it to; I just need to extend it.

tananaev:
This looks reasonable to me. Is it ready to be reviewed and merged?

justedro:
Yes please

PR #5504, opened 2025-01-28
ExtendedModel refactor

justedro:
Introduced the `removeAttribute` and `remove<Type>` methods and replaced all the existing instances of the attribute removal.

Unlocks #5502

### DIFF
--- a/src/main/java/org/traccar/model/ExtendedModel.java
+++ b/src/main/java/org/traccar/model/ExtendedModel.java
@@ -90,83 +90,64 @@ public class ExtendedModel extends BaseModel {
     }
 
     public String getString(String key, String defaultValue) {
-        if (attributes.containsKey(key)) {
-            Object value = attributes.get(key);
-            return value != null ? value.toString() : null;
-        } else {
-            return defaultValue;
-        }
+        return parseAsString(attributes.get(key), defaultValue);
     }
 
     public String getString(String key) {
-        return getString(key, null);
+        return parseAsString(attributes.get(key), null);
     }
 
     public double getDouble(String key) {
-        if (attributes.containsKey(key)) {
-            Object value = attributes.get(key);
-            if (value instanceof Number numberValue) {
-                return numberValue.doubleValue();
-            } else {
-                return Double.parseDouble(value.toString());
-            }
-        } else {
-            return 0.0;
-        }
+        return parseAsDouble(attributes.get(key), 0.0);
     }
 
     public boolean getBoolean(String key) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/traccar/model/ExtendedModel.java
```diff
@@ -154,4 +154,7 @@ public long getLong(String key) {
         }
     }
 
+    public Object removeAttribute(String key) {
```
tananaev, line 157: Should we equivalent for each type? Like `removeString`, `removeLong` etc? For consistency

justedro: I don't think so. 
The purpose and intention of a user of that method is to remove the parameter and have its parameter as an optional "side effect" rather than get the nice typed value (which is covered by the existing methods) and have the removal as a side effect.

tananaev: If you look at the pull request, you'll noticed that we almost always use the result. So I would argue it's not a side effect, but the main purpose.

tananaev: For Java collections it's pretty standard to return value or remove, so I don't see the name as a problem.

justedro:
Ready for review
Left the bare `removeAttribute` as void in place as it still makes sense in some places 
Java 21 pattern matching would make the code cleaner 😞

src/main/java/org/traccar/model/ExtendedModel.java
```diff
@@ -154,4 +154,56 @@ public long getLong(String key) {
...
+
+    public Double removeDouble(String key) {
+        var value = attributes.remove(key);
+        if (value == null) {
+            return null;
+        } else if (value instanceof Number numberValue) {
+            return numberValue.doubleValue();
+        } else {
+            return Double.parseDouble(value.toString());
+        }
```
tananaev, line 174: Also I think we need to reuse the login between `getXxx` and `removeXxx`. There's too much duplication.

justedro: It has the baked-in default values behavior, which does not match the current logic of returning nulls when there's no such key, so nothing comes to my mind apart from the extraction of the few common lines to a function, which feels wrong

tananaev: Duplicating the code is never the right solution. I'm sure there's a way to unify it.

justedro:
Should be good now

tananaev:
Looks good, thank you.

Issue #5511, opened 2025-02-16
report.ignoreOdometer not working at device level


jcardus:
shouldn't we also lookup the device here?

https://github.com/traccar/traccar/blob/5e34d0acba6c43fab49c2d111413d94c90fddfdf/src/main/java/org/traccar/reports/common/ReportUtils.java#L305

tananaev:
Probably should.

PR #5512, opened 2025-02-17
Fix report.ignoreOdometer by device

jcardus:
https://github.com/traccar/traccar/issues/5511


src/main/java/org/traccar/reports/SummaryReportProvider.java
```diff
@@ -101,7 +102,9 @@ private Collection<SummaryReportItem> calculateDeviceResult(
         }
 
         if (first != null && last != null) {
-            boolean ignoreOdometer = config.getBoolean(Keys.REPORT_IGNORE_ODOMETER);
+            boolean ignoreOdometer = AttributeUtil.lookup(
```
tananaev, line 105: Can we get it through trips config here as well?

src/main/java/org/traccar/config/Keys.java
```diff
@@ -1339,7 +1339,7 @@ private Keys() {
      */
     public static final ConfigKey<Boolean> REPORT_IGNORE_ODOMETER = new BooleanConfigKey(
             "report.ignoreOdometer",
-            List.of(KeyType.CONFIG),
+            List.of(KeyType.CONFIG, KeyType.DEVICE),
```
tananaev, line 1342: We also need to add it to the web app.

jcardus: It's already here right?

https://github.com/traccar/traccar-web/blob/e5173d706c0eb463f86c3751d34745341f166c14/src/common/attributes/useCommonDeviceAttributes.js#L17

PR #5517, opened 2025-02-22
Implement mobilogix commands

jcardus:
[EI_2021_0018__MOBILOGIX__MT2000_PROTOCOL_OVER_THE_AIR__19_08_2021__V1_3_7.pdf](https://github.com/user-attachments/files/18919508/EI_2021_0018__MOBILOGIX__MT2000_PROTOCOL_OVER_THE_AIR__19_08_2021__V1_3_7.pdf)

### DIFF
--- a/src/main/java/org/traccar/protocol/MobilogixProtocol.java
+++ b/src/main/java/org/traccar/protocol/MobilogixProtocol.java
@@ -31,9 +31,9 @@ public class MobilogixProtocol extends BaseProtocol {
     @Inject
     public MobilogixProtocol(Config config) {
         setSupportedDataCommands(
-                Command.TYPE_ENGINE_STOP,
+                Command.TYPE_CUSTOM,
                 Command.TYPE_ENGINE_RESUME,
-                Command.TYPE_CUSTOM);
+                Command.TYPE_ENGINE_STOP);
         addServer(new TrackerServer(config, getName(), false) {
             @Override
             protected void addProtocolHandlers(PipelineBuilder pipeline, Config config) {
--- a/src/main/java/org/traccar/protocol/MobilogixProtocol.java
+++ b/src/main/java/org/traccar/protocol/MobilogixProtocol.java
@@ -22,6 +22,7 @@ import org.traccar.CharacterDelimiterFrameDecoder;
 import org.traccar.PipelineBuilder;
 import org.traccar.TrackerServer;
 import org.traccar.config.Config;
+import org.traccar.model.Command;
 
 import jakarta.inject.Inject;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/traccar/protocol/MobilogixProtocolEncoder.java
```diff
@@ -0,0 +1,47 @@
...
+    private Object encodeCommand(Date time, String param) {
+        return String.format("[%s,S6,%s]", DateUtil.formatDate(time, false), param);
+    }
+
+    @Override
+    protected Object encodeCommand(Command command) {
+        return encodeCommand(command, new Date());
+    }
+
+    protected Object encodeCommand(Command command, Date time) {
```
tananaev, line 39: Why do we need a separate method with protected modifier?

src/main/java/org/traccar/protocol/MobilogixProtocol.java
```diff
@@ -22,19 +22,25 @@
...
 import jakarta.inject.Inject;
 
 public class MobilogixProtocol extends BaseProtocol {
 
     @Inject
     public MobilogixProtocol(Config config) {
+        setSupportedDataCommands(
+                Command.TYPE_ENGINE_STOP,
+                Command.TYPE_ENGINE_RESUME,
+                Command.TYPE_CUSTOM);
```
tananaev, line 36: Can we please put custom as the first item here and in the encoder.

src/main/java/org/traccar/protocol/MobilogixProtocolEncoder.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.traccar.Protocol;
+import org.traccar.StringProtocolEncoder;
+import org.traccar.helper.DateUtil;
+import org.traccar.model.Command;
+
+import java.util.Date;
+
+public class MobilogixProtocolEncoder extends StringProtocolEncoder {
+    public MobilogixProtocolEncoder(Protocol protocol) {
```
tananaev, line 26:
```suggestion
public class MobilogixProtocolEncoder extends StringProtocolEncoder {

    public MobilogixProtocolEncoder(Protocol protocol) {
```

src/main/java/org/traccar/protocol/MobilogixProtocolEncoder.java
jcardus: For the test

tananaev: I guess that makes sense.

PR #5518, opened 2025-02-22
Mobilogix

jcardus:
S6 Ack can cause loop.
Single position and reset commands.
Blocked and charge attributes.


src/main/java/org/traccar/protocol/MobilogixProtocolDecoder.java
```diff
@@ -81,12 +81,14 @@ protected Object decode(
         if (channel != null) {
             String time = sentence.substring(1, 20);
             String response;
-            if (type.equals("T1")) {
-                response = String.format("[%s,S1,1]", time);
-            } else {
-                response = String.format("[%s,S%s]", time, type.substring(1));
+            if (!type.equals("T6")) {
```
tananaev, line 84: I think it probably makes sense to move this check next to `if (channel != null)`.

PR #5522, opened 2025-03-09
Teltonika add Auto Geofence Alarm

bkbilly:
Teltonika Devices can be configured to create a Geofence when the ignition is off.
If the device is moved out of the geofence with the ignition off, it creates an alarm.

### DIFF
--- a/src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
@@ -287,12 +287,7 @@ public class TeltonikaProtocolDecoder extends BaseProtocolDecoder {
                 case 3 -> p.addAlarm(Position.ALARM_CORNERING);
             }
         });
-        register(175, fmbXXX, (p, b) -> {
-            switch (b.readUnsignedByte()) {
-                case 0 -> p.addAlarm(Position.ALARM_GEOFENCE_EXIT);
-                case 1 -> p.addAlarm(Position.ALARM_GEOFENCE_ENTER);
-            }
-        });
+        register(175, fmbXXX, (p, b) -> p.addAlarm(b.readUnsignedByte() > 0 ? Position.ALARM_GEOFENCE_ENTER : ALARM_GEOFENCE_EXIT);
         register(636, fmbXXX, (p, b) -> p.set("cid4g", b.readUnsignedInt()));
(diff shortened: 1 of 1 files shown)


bkbilly:
1. https://wiki.teltonika-gps.com/view/FMC920_Teltonika_Data_Sending_Parameters_ID
2. From the list of supported devices, it seems that it supports all devices under the "175 Auto Geofence"

tananaev:
Can you please clarify the second point. I only see that it applies to the FMBXXX series. Nothing about other devices.

bkbilly:
Oh, you are right. I am sorry for the confusion, I've pushed an update to fix this.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -287,6 +287,12 @@ private static void register(int id, Set<String> models, BiConsumer<Position, By
                 case 3 -> p.addAlarm(Position.ALARM_CORNERING);
             }
         });
+        register(175, fmbXXX, (p, b) -> {
+            switch (b.readUnsignedByte()) {
+                case 0 -> p.addAlarm(Position.ALARM_GEOFENCE_EXIT);
+                case 1 -> p.addAlarm(Position.ALARM_GEOFENCE_ENTER);
+            }
+        });
```
tananaev, line 295: I think it's better to do something like this:

```suggestion
        register(175, fmbXXX, (p, b) -> p.addAlarm(b.readUnsignedByte() > 0 ? Position.ALARM_GEOFENCE_ENTER : ALARM_GEOFENCE_EXIT);
```

bkbilly:
This looks much better. Thanks for the suggestion

tananaev:
Checks are failing. You might need to add a line break somewhere.

tananaev:
Still broken.

bkbilly:
I can't think of any reason why it fails...

tananaev:
It tells you exactly what the problem is. Have you compiled the code?

bkbilly:
oh, the Position was missing.

tananaev:
Now we're back to the line length problem. Can you please make sure you build it locally without errors.

bkbilly:
When I compile locally, I don't get the issue with the line length. I've fixed it, hopefully it will pass this time. :)

PR #5523, opened 2025-03-11
Bugfix: Add VGN mode parameter

incognito53:
GV50M tracker was flipping Online - in a minute due to NumberFormatException in decodeIgn's parsing of position data. 
The root cause was a missing index += 1 to skip over the virtual ignition mode parameter.
image


tananaev:
1. We need to add a test case
2. Are you sure it's applicable to all devices?

tananaev:
Looks like existing tests are failing.

incognito53:
Ehh realized that when it broke IGN for some devices. Updated.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -81,13 +81,15 @@ public class Gl200TextProtocolDecoder extends BaseProtocolDecoder {
...
             Map.entry("6E", "GV310LAU"),
+            Map.entry("70", "GV50M"),
             Map.entry("BD", "CV200"),
             Map.entry("C2", "GV600M"),
             Map.entry("DC", "GV600MG"),
             Map.entry("DE", "GL500M"),
             Map.entry("F1", "GV350M"),
             Map.entry("F8", "GV800W"),
             Map.entry("FC", "GV600W"),
+            Map.entry("FE", "GV53MG"),
```
tananaev, line 92: Please attach full documentation for both device models.

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -1074,6 +1077,11 @@ private Object decodeIgn(
         return position;
     }
 
+    private boolean isVirtualIgnitionMode(String type, String model) {
```
tananaev, line 1080: I don't think we need a separate function for this. Also what does `type.contains("VG")` mean?

src/main/java/org/traccar/protocol/Gl200TextProtocolDecoder.java
```diff
@@ -1074,6 +1077,11 @@ private Object decodeIgn(
...
+        return type.contains("VG") &&
+                List.of("GV50M", "GV53MG").contains(model);
```
tananaev, line 1082: A set of direct comparison would make more sense here.

src/test/java/org/traccar/protocol/Gl200TextProtocolDecoderTest.java
```diff
@@ -61,6 +61,9 @@ public void testDecode() throws Exception {
                 "+RESP:GTVGN,C2010D,869653060009939,GV600M,453966,0,0.0,0,163.9,10.239379,45.931389,20231210233723,0222,0010,2F31,006D7621,,,0.0,20240107182623,143F$"),
                 Position.KEY_IGNITION, true);
 
+        verifyPosition(decoder, buffer(
```
tananaev, line 64: Please add new test to the top of the file.

PR #5536, opened 2025-03-30
Flespi add power cut Alarm

bkbilly:


tananaev:
Need some documentation for it. Also, it doesn't seem like an alarm attribute to me.

tananaev:
That seems to map better to the charge attribute. That's not an alarm.

bkbilly:
The charge attribute is the external.powersource.voltage and it is already shown as an attribute.
The external.powersource.status is very important safety alarm because we can get alerted when the device has disconnected in case a thief cuts the cables of the device. 
These alerts get delivered immediately so there is an early warning that something is up with the vehicle.

tananaev:
> The charge attribute is the external.powersource.voltage and it is already shown as an attribute.

No, voltage is not the same as the charge attribute.

I would still need to see the documentation on what the `external.powersource.status` attribute mean.

bkbilly:
I am not sure if there is a documentation for the relation between flespi and the device, but for the teltonika FMC920 device I am using it is called unplug detection and it's under the parameter ID 252.

The documentation on how this is configured is located here: https://wiki.teltonika-gps.com/view/FMC920_Accelerometer_Features_settings#Unplug_Detection

tananaev:
An event is not the same thing as status. Status is something that is reported constantly. Event is something that usually happens once. From what I can tell so far, you're trying to convert a status into an event. That's not something we usually do. Hopefully you understand the difference between the two concepts.

bkbilly:
I am not sure that I understand because it can be configured as an event on teltonika devices so it should be an event.
It is very similar to the other alarms that they are already used like gnss.antenna.cut.status and hood.open.status which both can be configured as an event or as a status.

In my opinion this is a very important alarm that the user wants to be informed, not open the app and see the attribute.

tananaev:
It might be an important event, but so far I don't understand what exactly this attribute represents. You keep referring to Teltonika, but they have a pretty clear separation between statuses and events, which is not what I see here so far.

bkbilly:
I have tested it on teltonika, this is why I keep on mentioning it. I don't have any experience with other devices.

Thank you taking the time to explain, but it seems that we don't share the same opinion for this so I'll close this PR.

tananaev:
It's not about opinions. I just need documentation to understand what this attribute really means. What makes you think that it's the same thing as Teltonika ID 252?

bkbilly:
This conclusion came from my tests on the FMC920 devices which are configured to be charged on need. The only instance that the `external.powersource.status` changes status is when I disconnect the power.

Unfortunately I can't find any documentation on Flespi that can prove the results of my tests, but if you want me to try anything, I would be happy to do so.

tananaev:
Can you see the raw data somewhere in Flespi to confirm that the value matches the IO and present all the time?

bkbilly:
Oh, you are right, I found it when I clicked on the 3 dots:
image

tananaev:
Do you have Teltonika documentation for ID 252?

bkbilly:
This is the parameter list of the FMC920 device:
https://wiki.teltonika-gps.com/view/FMC920_Teltonika_Data_Sending_Parameters_ID

PR #5543, opened 2025-04-10
Implement Getscale RFID reader with AOVX

jcardus:
https://www.getscale.com.br/gs-404
[GS404_protocolo.pdf](https://github.com/user-attachments/files/19676439/GS404_protocolo.pdf)
[V_Series Protocol.pdf](https://github.com/user-attachments/files/19676449/V_Series.Protocol.pdf)

### DIFF
--- a/src/main/java/org/traccar/protocol/HuabaoProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/HuabaoProtocolDecoder.java
@@ -1012,7 +1012,22 @@ public class HuabaoProtocolDecoder extends BaseProtocolDecoder {
 
         int type = buf.readUnsignedByte();
 
-        if (type == 0x41) {
+        if (type == 0x40) {
+            Position position = new Position(getProtocolName());
+            position.setDeviceId(deviceSession.getDeviceId());
+
+            getLastLocation(position, null);
+            String data = buf.readCharSequence(buf.readableBytes(), StandardCharsets.US_ASCII).toString().trim();
+            if (data.startsWith("GTSL")) {
+                String[] values = data.split("\\|");
+                if (values.length > 4) {
+                    position.set(Position.KEY_DRIVER_UNIQUE_ID, values[4]);
+                }
+            }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/HuabaoProtocolDecoder.java
```diff
@@ -1224,9 +1224,24 @@ private Position decodeTransparent(DeviceSession deviceSession, ByteBuf buf) {
 
             return position;
 
+        } else if (type == 0x40) {
```
tananaev, line 1227: This should probably be first, not last.

src/test/java/org/traccar/protocol/HuabaoProtocolDecoderTest.java
```diff
@@ -11,6 +11,10 @@ public void testDecode() throws Exception {
 
         var decoder = inject(new HuabaoProtocolDecoder(null));
 
+        verifyAttribute(decoder, binary(
+                        "7e0900001b41305888975883194000184754534c7c367c317c307c31303231383039347c317c0d0ab47e"),
```
tananaev, line 15: Incorrect indentation

src/main/java/org/traccar/protocol/HuabaoProtocolDecoder.java
```diff
@@ -1012,7 +1012,22 @@ private Position decodeTransparent(DeviceSession deviceSession, ByteBuf buf) {
...
+            getLastLocation(position, null);
+            String data = buf.readCharSequence(buf.readableBytes(), StandardCharsets.US_ASCII).toString().trim();
+            if (data.startsWith("GTSL")) {
+                String[] values = data.split("\\|");
+                if (values.length > 4) {
+                    position.set(Position.KEY_DRIVER_UNIQUE_ID, values[4]);
+                }
+            }
+
+            return position;
```
tananaev, line 1028: I think we should probably only return position if we have some data.

PR #5553, opened 2025-05-10
suntech ST340UR and HBM=2 support

CpdnCristiano:
This PR fixes multiple NullPointerExceptions that occurred when the device sent position data without required timestamps such as fixTime or deviceTime.


tananaev:
Please add a test case.

CpdnCristiano:
I'm not sure what kind of test I should perform, but I noticed that when my GPS tracker sends a command response like this:

```
ST300CMD;Res;xxxxxxxxx;319H;Disable1
```

…it does **not** include a timestamp or location data. Because of that, the `Position` object created by Traccar has a `null` value for the `fixTime`.

As a result, when Traccar tries to process this response using the `FilterHandler`, it throws a **NullPointerException** because it attempts to call `.getTime()` on a `null` `Date` object.

Here is the error I found in my logs:

```
2025-05-10 00:35:03  WARN: Position handler failed - Cannot invoke "java.util.Date.getTime()" because the return value of "org.traccar.model.Position.getFixTime()" is null - NullPointerException (FilterHandler:126 < *:212 < *:279 < BasePositionHandler:34 < ProcessingHandler:158 < DistanceHandler:74 < ...)
```

This pull request fixes the issue by copying the `fixTime` from the last known valid position when handling command responses that do not include a timestamp.

That way, the `Position` object will have a non-null `fixTime`, avoiding the `NullPointerException` in `FilterHandler` and ensuring smooth processing of such messages.

I hope I explained it clearly. My English is not very good.

tananaev:
I'm asking for a unit test to cover this scenario.

tananaev:
It looks like you piled up a lot more changes on top of your original fix. You should probably change title. And also please provide the protocol documentation.

CpdnCristiano:
[protocol manual](https://github.com/user-attachments/files/20383275/ilide.info-opermanual-sb-st340ur-rev109-20220125-pr_dd9931c3fa3d9454c4a8c21d9db5346e.pdf)

src/main/java/org/traccar/protocol/SuntechProtocolDecoder.java
```diff
@@ -360,9 +361,9 @@ private Position decode2356(
                 && !type.equals("ALT") && !type.equals("UEX")) {
             return null;
         }
-
+        Integer model = null;
         if (protocol.startsWith("ST3") || protocol.equals("ST500") || protocol.equals("ST600")) {
-            index += 1; // model
+           model = Integer.parseInt(values[index++]);
```
tananaev, line 366: indentation seems incorrect here

src/main/java/org/traccar/protocol/SuntechProtocolDecoder.java
```diff
@@ -360,9 +361,9 @@ private Position decode2356(
                 && !type.equals("ALT") && !type.equals("UEX")) {
             return null;
         }
-
+        Integer model = null;
```
tananaev, line 364: I would prefer no initial value and have else case instead

src/main/java/org/traccar/protocol/SuntechProtocolDecoder.java
```diff
@@ -439,7 +440,8 @@ private Position decode2356(
                 position.set(Position.KEY_RPM, Integer.parseInt(values[index++]));
             }
 
-            if (values.length - index >= 2) {
+            if ((values.length - index >= 2 && !protocol.startsWith("ST3"))
+                    || (protocol.startsWith("ST3") && model == 45)) {
```
tananaev, line 444: Can you explain why we need this check?

src/main/java/org/traccar/protocol/SuntechProtocolDecoder.java
```diff
@@ -458,6 +460,18 @@ private Position decode2356(
 
             }
 
+            if (getHbm(deviceSession.getDeviceId()) >= 2) {
```
tananaev, line 463: I think it's better to call `getHbm` once somewhere at the beginning.

src/main/java/org/traccar/protocol/SuntechProtocolDecoder.java
```diff
@@ -458,6 +460,18 @@ private Position decode2356(
...
+                if (index + 5 <= values.length) {
+                    int cid = Integer.parseInt(values[index++]);
+                    int mcc = Integer.parseInt(values[index++]);
+                    int mnc = Integer.parseInt(values[index++]);
+                    int rssi = Integer.parseInt(values[index++]);
+                    int lac = Integer.parseInt(values[index++]);
+                    position.setNetwork(new Network(CellTower.from(mcc, mnc, lac, cid, rssi)));
+                    position.set(Position.KEY_RSSI, rssi);
```
tananaev, line 471: If you add a cell, don't save rssi separately.

src/main/java/org/traccar/protocol/SuntechProtocolDecoder.java
CpdnCristiano: "If HBM2 is enabled and the protocol has no driver, the 'Cell ID' value will be incorrectly set to 'driver'. The documentation only mentions that the ST340UR(45) model has a driver. For a moment, I thought this applied only to models with the 'ST3' prefix, but after further testing, I concluded that this assumption was wrong. Based on the analysis, I believe the best alternative would be to create a new method, `isIncludeDriver`, similar to `isIncludeRpm`, to follow the existing pattern. However, I’m concerned about the possibility of breaking systems that are already in production. I think it’s better to close this PR and revert to the initial approach."

CpdnCristiano: ## Alternative Approach Considered

I have thought of another approach:

```java
int need = hbm == 1 ? 2 : 7;

if (values.length - index >= need) {
String driverUniqueId = values[index++];
if (!driverUniqueId.isEmpty()) {
position.set(Position.KEY_DRIVER_UNIQUE_ID, driverUniqueId);
}
index += 1; // registered
}
```

The idea here is to dynamically calculate how many values ​​are needed based on whether HBM2 is enabled (`hbm == 1`).

However, I am still considering the potential side effects and whether this could introduce unexpected issues into existing systems.

I am open to feedback on whether this would be a better solution or if we should stick with the previous method of creating a new method

src/main/java/org/traccar/protocol/SuntechProtocolDecoder.java
```diff
@@ -439,7 +441,9 @@ private Position decode2356(
                 position.set(Position.KEY_RPM, Integer.parseInt(values[index++]));
             }
 
-            if (values.length - index >= 2) {
+            int need = hbm == 1 ? 2 : 7;
```
tananaev, line 444: I like the approach, but I think we need a better name for this variable. Or maybe just inline it to avoid naming.

PR #5558, opened 2025-05-20
Improve address search with line segment distance calculation

GeniLogic:
# Improved Address Search with Line Segment Distance Calculation

## Description
This PR improves the address search functionality by implementing a line segment distance calculation algorithm. This enhancement provides better results when searching for positions along streets, especially for long streets where positions might be outside the initial radius but still on the same street.

## Changes
- Added line segment distance calculation for better street coverage
- Improved position filtering for long streets
- Added proper documentation and code cleanup
- Added author attribution

## Technical Details
The new algorithm calculates the distance between a position and a street segment instead of just using a point-to-point distance. This provides more accurate results when searching for positions along streets.

## Author
LECOMTE Benjamin - GeniLogic

## Testing
The changes have been tested with various street lengths and positions.

### DIFF
--- a/src/main/java/org/traccar/api/resource/AddressSearchResource.java
+++ b/src/main/java/org/traccar/api/resource/AddressSearchResource.java
@@ -1,0 +1,179 @@
+package org.traccar.api.resource;
+
+import jakarta.inject.Inject;
+import jakarta.ws.rs.Consumes;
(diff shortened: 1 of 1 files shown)


tananaev:
It looks like you included a lot of unrelated changes. Some added and removed config files etc. You have to clean all that up first.

PR #5563, opened 2025-05-25
Scheduled commands

CpdnCristiano:
This pull request introduces a new feature to schedule and execute commands based on calendars, along with several supporting changes to the codebase. The most important changes include adding a new database schema for calendar-based commands, implementing a new task for managing scheduled commands, and extending the `Command` model to support scheduling. Additionally, utility methods for calendar period checks and new conditions for database queries have been added.

### Database Schema Changes:
* Added a new Liquibase changelog (`changelog-6.8.0.xml`) to introduce a `calendarid` column in the `tc_commands` table, along with a foreign key constraint referencing the `tc_calendars` table.
* Updated `changelog-master.xml` to include the new changelog file `changelog-6.8.0.xml`.

### Command Scheduling Feature:
* Introduced `TaskCommands` in `src/main/java/org/traccar/schedule/TaskCommands.java`, a scheduled task to execute commands based on calendar periods. This task checks for eligible commands and sends them to associated devices or groups.
* Extended the `Command` model to implement the `Schedulable` interface, adding support for `calendarId` and its getter/setter methods. [[1]](diffhunk://#diff-d8b09d2708eaf5d65ca382881df5a0647b733854861402efb288c230b6bdde75L25-R25) [[2]](diffhunk://#diff-d8b09d2708eaf5d65ca382881df5a0647b733854861402efb288c230b6bdde75R110-R121)

### Calendar Utility Enhancements:
* Enhanced the `Calendar` model with new methods to find periods between two dates (`findPeriodsBetween`) and check moments within a date range (`checkMomentBetween`).
* Added a test case `testCalendarBetween` in `CalendarTest` to verify the behavior of the new calendar utility methods.

These changes collectively enable the scheduling and execution of commands based on calendar configurations, enhancing the application's functionality and flexibility.


src/main/java/org/traccar/storage/query/Condition.java
```diff
@@ -208,4 +208,22 @@ public long getDeviceId() {
         }
     }
 
+    class IsNull implements Condition {
```
tananaev, line 211: Can we reuse `Compare` condition for this somehow?

schema/changelog-master.xml
```diff
@@ -47,5 +46,5 @@
   <include file="changelog-6.3.xml" relativeToChangelogFile="true" />
   <include file="changelog-6.6.xml" relativeToChangelogFile="true" />
   <include file="changelog-6.7.0.xml" relativeToChangelogFile="true" />
-
-</databaseChangeLog>
+  <include file="changelog-6.7.2.xml" relativeToChangelogFile="true" />
```
tananaev, line 49: This should be 6.8.0. We want to avoid schema changes between minor releases.

schema/changelog-master.xml
```diff
@@ -2,8 +2,7 @@
 <databaseChangeLog
   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
-                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
```
tananaev, line 5: Please don't reformat the rest of the file

schema/changelog-6.7.2.xml
```diff
@@ -0,0 +1,29 @@
...
+    <addColumn tableName="tc_commands">
+      <column name="calendarid" type="INT">
+        <constraints nullable="true" />
+      </column>
+    </addColumn>
+    <addForeignKeyConstraint baseTableName="tc_commands"
+                            baseColumnNames="calendarid"
+                            referencedTableName="tc_calendars"
+                            referencedColumnNames="id"
+                            constraintName="fk_tc_commands_calendarid"
```
tananaev, line 25: This is not the formatting style we use. Please match other styles.

schema/changelog-6.7.2.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd"
+  logicalFilePath="changelog-6.7.2">
+
+  <changeSet author="author" id="changelog-6.7.2">
+    <preConditions onFail="MARK_RAN">
```
tananaev, line 10: This check is not needed.

schema/changelog-6.7.2.xml
```diff
@@ -0,0 +1,29 @@
...
+      <not>
+        <columnExists tableName="tc_commands" columnName="calendarid" />
+      </not>
+    </preConditions>
+
+    <addColumn tableName="tc_commands">
```
tananaev, line 16: Would it make more sense to use command queue table for this?

src/main/java/org/traccar/schedule/TaskCommands.java
```diff
@@ -0,0 +1,134 @@
...
+            List<Command> commands = storage.getObjects(Command.class, new Request(
+                    new Columns.All(), new Condition.IsNotNull("calendarid")));
+
+            for (Command command : commands) {
+                var calendar = storage.getObject(org.traccar.model.Calendar.class, new Request(
+                        new Columns.All(), new Condition.Equals("id", command.getCalendarId())));
+                if (calendar == null || calendar.checkMoment(lastCheck.getTime())) {
+                    continue; // this command already executed
+                }
+                if (calendar.checkMomentBetween(lastCheck.getTime(), currentCheck.getTime())) {
```
tananaev, line 83: Why can't we check that the current time is within the calendar?

src/main/java/org/traccar/schedule/TaskCommands.java
```diff
@@ -0,0 +1,134 @@
...
+import org.traccar.storage.query.Request;
+
+import java.util.Calendar;
+import java.util.Date;
+import java.util.List;
+
+public class TaskCommands extends SingleScheduleTask {
+    private static final Logger LOGGER = LoggerFactory.getLogger(TaskCommands.class);
+
+    private static final long CHECK_PERIOD_MINUTES = 5;
```
tananaev, line 45: I think this is too frequent. We usually do 15 minutes.

src/main/java/org/traccar/schedule/TaskCommands.java
```diff
@@ -0,0 +1,134 @@
...
+    public void run() {
+        Calendar currentCheck = Calendar.getInstance();
+        currentCheck.setTime(new Date());
+        currentCheck.set(Calendar.SECOND, 0);
+        Calendar lastCheck = Calendar.getInstance();
+        lastCheck.setTime(new Date());
+        lastCheck.add(java.util.Calendar.MINUTE, (int) -CHECK_PERIOD_MINUTES);
+        try {
+            List<Command> commands = storage.getObjects(Command.class, new Request(
+                    new Columns.All(), new Condition.IsNotNull("calendarid")));
```
tananaev, line 75: I suspect we don't even need `IsNotNull` condition at all. We can just extract all and filter in the code.

src/main/java/org/traccar/schedule/TaskCommands.java
```diff
@@ -0,0 +1,134 @@
...
+
+        var groupIds = storage.getObjects(Group.class, new Request(
+                new Columns.Include("id"),
+                new Condition.Permission(Group.class, Command.class, command.getId())))
+                .stream().map(BaseModel::getId).toList();
+
+        for (long groupId : groupIds) {
+            var devices = storage.getObjects(Device.class, new Request(
+                    new Columns.Include("id"),
+                    new Condition.Equals("groupId", groupId)));
```
tananaev, line 109: This is not going to give you all devices.

src/main/java/org/traccar/schedule/TaskCommands.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+        for (long deviceId : deviceIds) {
+            try {
+                Command newCmd = QueuedCommand.fromCommand(command).toCommand();
+                newCmd.setDeviceId(deviceId);
+                commandsManager.sendCommand(newCmd);
+                actionLogger.command(null, 0, 0, deviceId, command.getType(), true);
+            } catch (Exception e) {
+                LOGGER.warn("Failed to send command to device {}", deviceId, e);
+            }
```
tananaev, line 129: We need to avoid this duplication.

tananaev (commented):
First pass

src/main/java/org/traccar/schedule/TaskCommands.java
CpdnCristiano: Some users tend to enter an exact time in the calendar, as shown in the image (e.g., "From: 23/05/2025 18:30" to "To: 23/05/2025 18:30"). Because of this, their scheduled events may fall outside of the verification interval if the system doesn’t check precisely at that time.

For example:

The system's last check was at 18:25.

The system's current check is at 18:35.

The event is scheduled exactly for 18:30.

If the system only looks for events at the exact time of the current check (18:35), this event would be missed.

Therefore, I believe the best approach is to check whether the event occurred between the last check and the current check. This ensures that events scheduled for exact moments — even without a time range — are still properly detected and executed, even if they don’t coincide with the system's periodic checks.
image

src/main/java/org/traccar/schedule/TaskCommands.java
CpdnCristiano: I understand that the usual interval is 15 minutes, but some users might require higher precision, especially when actions need to be executed shortly after a scheduled event.

Perhaps the best approach is to make this interval configurable, allowing each deployment to adjust it based on their specific needs and system load.

If you agree, I can implement this change.

src/main/java/org/traccar/schedule/TaskCommands.java
tananaev: But in other places it works like this. I think it's better to prevent users from creating a calendar with less than 15 minutes.

src/main/java/org/traccar/schedule/TaskCommands.java
tananaev: Configurable with default 15 minutes makes sense to me.

schema/changelog-master.xml
```diff
@@ -47,5 +47,5 @@
   <include file="changelog-6.3.xml" relativeToChangelogFile="true" />
   <include file="changelog-6.6.xml" relativeToChangelogFile="true" />
   <include file="changelog-6.7.0.xml" relativeToChangelogFile="true" />
-
+  <include file="changelog-6.8.0.xml" relativeToChangelogFile="true" />
```
tananaev, line 50:
```suggestion
  <include file="changelog-6.8.0.xml" relativeToChangelogFile="true" />

```

src/main/java/org/traccar/helper/LogAction.java
```diff
@@ -114,10 +114,18 @@ public void resetAccumulators(HttpServletRequest request, long userId, long devi
...
     public void command(HttpServletRequest request, long userId, long groupId, long deviceId, String type) {
+        command(request, userId, groupId, deviceId, type, false);
+    }
+
+    public void command(HttpServletRequest request, long userId, long groupId, long deviceId, String type,
+            boolean scheduled) {
         Action action = new Action();
         action.setAddress(WebHelper.retrieveRemoteAddress(request));
-        action.setUserId(userId);
+        if (userId > 0) {
```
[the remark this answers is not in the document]
CpdnCristiano: As I don't know which user scheduled that command, it goes as 0, I believe that in the database it should be null and not 0

CpdnCristiano:
I made the adjustments

src/main/java/org/traccar/helper/LogAction.java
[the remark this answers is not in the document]
CpdnCristiano: fix

src/main/java/org/traccar/schedule/TaskCommands.java
tananaev: Looks like this is not resolved yet.

PR #5566, opened 2025-05-27
feat: position convert

sgpublic:
This PR incorporates commits from #5354.

Added coordinate conversion function, integrating the coordinate conversion APIs of the following platforms:

+ Amap: GCJ-02
+ Tencent Map: GCJ-02
+ Baidu Map: BD-09

Considering that these platforms have QPS and periodic usage limits (monthly or daily), this PR creates a background service for batch API requests for asynchronous conversion. After the API Key of the corresponding platform is configured, the background task is allowed to start.

This PR also modifies the API to allow the frontend to support this change.


tananaev:
Why do we need to convert positions? We had a discussion about it in another thread and I don't think it's the right approach.

sgpublic:
I understand what you mean. You want to map the tiles provided by the map service in China to WGS-84 before displaying them.

But the actual situation is that the algorithms used by GCJ-02 and BD-09 are non-linear and irreversible (reference: https://github.com/hujiulong/gcoord/issues/50). These algorithms are not public, so it is extremely difficult to restore them. Even if a tile can be restored to the WGS-84 coordinate system, the size of each tile may not be square or even rectangular, and it is difficult for MapLibre to display it.

GCJ-02 tiles just don’t line up properly with WGS-84 tiles near China’s borders. Since the offset algorithm can’t be reversed, even using fitting methods to approximate back to WGS-84 makes it hard to match original WGS-84 tiles. That’s why I think the better solution is to simply use Chinese map platforms’ tiles directly and convert coordinates before displaying.

tananaev:
Even if it's not possible to do as a projection, the conversion should done only at the last stage for map rendering only.

I also don't think we should use some API for conversion when the algorithm is already well known and there are open source options available.

sgpublic:
In other words, is your idea that we should only use open source algorithms for coordinate transformations?

In fact, the accuracy of existing open source algorithms is not high, and it may not be possible to achieve acceptable errors throughout China. Many reports on excessive errors can be found:

+ https://github.com/wandergis/coordtransform/issues/42
+ https://github.com/wandergis/coordtransform/issues/37
+ https://github.com/chinabugotech/hutool/issues/643#issuecomment-562026945

There are also reports that the error of the open source algorithm is around 10m: https://github.com/wandergis/coordtransform/issues/16

My opinion is that this error might be acceptable when determining rough locations, but is hardly acceptable when used for tracking positions in Traccar.

sgpublic:
I just did an experiment and randomly selected a few points I know. The coordinate error after using the API and using `CoordinateUtil.java` is about 0.1m, which is indeed very small and completely acceptable. At least the above conclusion is true for GCJ-02. Please allow me to apologize for not investigating more carefully.

Just to confirm, can I summarize that we have reached the following consensus: 

1. It may be difficult for us to remap the tiles provided by the Chinese local map service back to the original coordinate system and let MapLibre display it correctly.
2. In order to make the coordinates display correctly on the Chinese local map, we can use open source algorithms for coordinate conversion.

tananaev:
1. Yes.
2. Yes. But it need to be done at the last possible moment. Basically when we render on the map only.

tananaev:
We have implemented GCJ-02 support in Traccar. It's implemented on the web side:

https://github.com/traccar/traccar-web/commit/f71884a29993772f49135d118f29d4f5a9c48564

PR #5587, opened 2025-06-17
remove unused return in method

LuizFJP:


src/main/java/org/traccar/ProcessingHandler.java
```diff
@@ -207,17 +207,16 @@ private void finishedProcessing(ChannelHandlerContext ctx, Position position, bo
...
+    Queue<Position> queue = getQueue(deviceId);
+    Position nextPosition;
+    synchronized (queue) {
+        queue.remove(); // remove current position
+        nextPosition = queue.peek();
     }
-
+    if (nextPosition != null) {
+        processPositionHandlers(ctx, nextPosition);
+    } else {
```
tananaev: What happened to the indentation?

LuizFJP: My bad. IDE problem. I sent a new commit fixing that.

src/main/java/org/traccar/ProcessingHandler.java
```diff
@@ -210,7 +210,7 @@ private void processNextPosition(ChannelHandlerContext ctx, long deviceId) {
         Queue<Position> queue = getQueue(deviceId);
         Position nextPosition;
         synchronized (queue) {
-            queue.poll(); // remove current position
+            queue.remove(); // remove current position
```
tananaev, line 213: What is the point of this change?

PR #5590, opened 2025-06-22
Add release delivery to Docker Hub and GHCR

magna-z:
@tananaev,
I have migrated and refactored the build and publication of Docker traccar/tracker images from the [traccar/traccar-docker repository](https://github.com/traccar/traccar-docker). Along with the creation of the new release, Docker images will be published on the Docker Hub and GitHub Container Registry.

To parallelize the publication, I split the submission into S3 and the Docker build. In the future can add another task to create a release on GitHub.

Before the first launch, you need to add the secrets `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,23 +1,24 @@
-name: Build Installers
+name: Build Release
 
 on:
   workflow_dispatch:
     inputs:
       version:
-        description: 'Version'
+        description: Version
         required: true
-        default: 'preview'
+        default: preview
 
 jobs:
-  release:
-    runs-on: ubuntu-22.04
+  build:
+    name: Build installers
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
           submodules: true
-      - run: git checkout ${{ github.ref_name }}
-        working-directory: ./traccar-web
+      - working-directory: ./traccar-web
+        run: git checkout ${{ github.ref_name }}
       - uses: actions/setup-java@v4
         with:
           distribution: temurin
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -29,24 +31,89 @@ jobs:
...
+    steps:
+      - uses: actions/download-artifact@v4
+      - env:
           S3_ACCESS_KEY: ${{ secrets.S3_ACCESS_KEY }}
           S3_SECRET_KEY: ${{ secrets.S3_SECRET_KEY }}
-        run: s3cmd --acl-public put traccar-*.zip s3://traccar/builds/ --host=nyc3.digitaloceanspaces.com --host-bucket=traccar --access_key="$S3_ACCESS_KEY" --secret_key="$S3_SECRET_KEY"
+        run: |
+          sudo apt-get install s3cmd
+          s3cmd --acl-public put installers/traccar-*.zip s3://traccar/builds/ --host=nyc3.digitaloceanspaces.com --host-bucket=traccar --access_key="$S3_ACCESS_KEY" --secret_key="$S3_SECRET_KEY"
+  docker-build-push:
```
tananaev: we should probably skip this on a preview build?

.github/workflows/release.yml
```diff
@@ -1,23 +1,25 @@
-name: Build Installers
+---
+name: Build Release
```
tananaev:
```suggestion
name: Build Release
```

tananaev:
Are we completely deprecating docker repo?

magna-z:
> Are we completely deprecating docker repo?

Yes, the `traccar-docker` repository will become unnecessary and may be deleted after some time.

.github/workflows/release.yml
magna-z: > we should probably skip this on a preview build?

There were no such exceptions in the current workflow, and I assumed that you would use this workflow only when the release is ready.

.github/workflows/release.yml
magna-z: ОК, but `---` on first line in YAML document is the best practice.

.github/workflows/release.yml
magna-z: @tananaev, describe the functionality of this workflow that you would like to receive.

I can add a job with "github.com/marketplace/actions/gh-release" and change the trigger to "push tags", then the installers will be assembled and uploaded to GitHub Release, GHCR, DockerHub and the site.

tananaev: By default this builds a preview version that basically overwrites files. I think we should completely skip docker task if the version is set to "preview".

I don't want to use push tags to generate builds. Usually we do final testing on release build and there might be last minute fixes. Now thinking about it, we should probably split docker into separate task. Can docker be triggered by publishing release?

.github/workflows/release.yml
tananaev: Do you have a reference to support this? ChatGPT says that it's not a best practice. It's something you need if you have multiple YAML documents in a single file, but completely optional if it's just one document.

.github/workflows/release.yml
```diff
@@ -29,24 +31,116 @@ jobs:
...
+            })
+      - uses: actions/download-artifact@v4
+      - uses: softprops/action-gh-release@v2
+        with:
+          tag_name: v${{ github.event.inputs.version }}
+          files: installers/traccar-*.zip
+  docker:
+    name: Build and push docker images
+    needs: build
+    if: ${{ 1 != 1 && github.event.inputs.version != 'preview' }} # job disabled for testing
```
tananaev: What is `1 != 1` for?

.github/workflows/release.yml
```diff
@@ -29,24 +31,116 @@ jobs:
...
+      - uses: actions/upload-artifact@v4
+        with:
+          name: installers
+          path: setup/traccar-*.zip
+          compression-level: 0
+          retention-days: 1
+  upload:
+    name: Upload installers
+    needs: build
+    if: ${{ 1 != 1 }} # job disabled for testing
```
tananaev: And here

.github/workflows/release.yml
magna-z: In order not to repeatedly shift artifacts, I left one workflow and closed release jobs with the condition `${{ github.event.input.version != 'preview' }}`. I also added job `release`, which creates a tag with the specified version and then creates a release with all installers.

Watch the test run: https://github.com/magna-z/traccar/actions/runs/15851363221 and https://github.com/magna-z/traccar/releases/tag/v1.2.4

.github/workflows/release.yml
magna-z: ChatGPT is right, `---` is used in YAML files with multiple documents. It is with such documents that I work more often, and therefore I just consider the `---` at the beginning of the file to be a general good practice in any case.
In this file, I returned it as it was.

.github/workflows/release.yml
magna-z: It is only for testing in my fork repository (https://github.com/magna-z/traccar) with minimal changes in file. At now removed in PR.

.github/workflows/release.yml
magna-z: And this only for testing.

tananaev: Is it ready to be reviewed now?

tananaev:
@magna-z any update? Is it ready for a review?

magna-z:
> @magna-z any update? Is it ready for a review?

I've finished

.github/workflows/release.yml
```diff
@@ -29,24 +30,115 @@ jobs:
...
+    if: ${{ github.event.inputs.version != 'preview' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            github.rest.git.createRef({
```
tananaev, line 77: I don't think we want to create a tag. You can expect the tag to be already created. That's because we need to tag both server and the web app. I usually do it manually because I also need to update the submodule commit.

magna-z: To create a release via `softprops/action-gh-release`, an tag is required. How do you create tags in these repositories?

tananaev: Usually I create tags manually. So you can assume there are tags.

magna-z: On the website github.com and the `gh` utility does not have the functionality to create tags. They can only be done through `git tag ...` + `git push` or by creating a release.
How do you create tags in this and the web repository? I'm asking to understand which is the best trigger to use for this workflow.

tananaev: I do it locally using `git tag` + `git push --tags`.

tananaev: @magna-z any update? Let's leave manual trigger for the whole task. We can automate it further at a later point.

tananaev:
I'll merge it for now.

PR #5591, opened 2025-06-23
# Add JMAK JR11 & JR12 Protocol Decoder (STRING Mode)

renansoaress:
**Overview**
This PR adds initial support for the JMAK JR11 and JR12 GPS trackers in Traccar’s “STRING” protocol mode by introducing a dedicated decoder. These devices send their location and status messages as ASCII strings; this change enables Traccar to parse those incoming frames correctly.

**Key Changes**

**- Protocol Decoder**

  - Added JmakFrameDecoder.java to split raw message frames.
  
  - Introduced JmakProtocol.java to register the “jmak” protocol.
   
  - Implemented JmakProtocolDecoder.java to parse JR11/JR12 string messages and extract fields (device ID, timestamp, latitude/longitude, speed, status, etc.).

**- Testing**

  - Created JmakProtocolDecoderTest.java with unit tests covering valid location reports;

**- Future Work**
  
  - ZIP Mode — Support for compressed binary messages.
  
  - JSON Mode — Support for JSON-formatted telematics frames.

Please review the decoder implementation, configuration changes, and test coverage. Feedback on naming, edge-case handling, or additional test scenarios is welcome!


tananaev:
You have to provide the protocol documentation.

renansoaress:
> You have to provide the protocol documentation.

I’m sorry; I forgot to send the documentation.
[J_R12_Protocolo_de_comunicação.pdf](https://github.com/user-attachments/files/20869760/J_R12_Protocolo_de_comunicacao.pdf)
[J_R11_Protocolo_de_comunicação.pdf](https://github.com/user-attachments/files/20869761/J_R11_Protocolo_de_comunicacao.pdf)

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
```diff
@@ -0,0 +1,103 @@
...
+package org.traccar.protocol;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.ByteToMessageDecoder;
+import io.netty.handler.codec.TooLongFrameException;
+
+import java.util.List;
+
+public class JmakFrameDecoder extends ByteToMessageDecoder {
```
tananaev: You should use `BaseFrameDecoder`

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private final int maxFrameLength;
```
tananaev: Let's skip frame limit for now.

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
```diff
@@ -0,0 +1,103 @@
...
+            ByteBuf frame = in.readRetainedSlice(frameLength);
+            out.add(frame);
+
+        } else {
+            // Skip unknown data
+            in.readByte();
+        }
+    }
+
+    private int findJsonEnd(ByteBuf buf) {
```
tananaev: We already have `JsonFrameDecoder`. I suggest we reuse that.

src/main/java/org/traccar/protocol/JmakProtocol.java
```diff
@@ -0,0 +1,39 @@
...
+
+import io.netty.handler.codec.string.StringDecoder;
+import org.traccar.BaseProtocol;
+import org.traccar.PipelineBuilder;
+import org.traccar.TrackerServer;
+import org.traccar.config.Config;
+
+import jakarta.inject.Inject;
+
+public class JmakProtocol extends BaseProtocol { // JMAK 0.0.2
```
tananaev: Remove unnecessary comments

src/main/java/org/traccar/protocol/JmakProtocol.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Inject
+    public JmakProtocol(Config config) {
+        addServer(new TrackerServer(config, getName(), false) {
+            @Override
+            protected void addProtocolHandlers(PipelineBuilder pipeline, Config config) {
+                pipeline.addLast(new JmakFrameDecoder(3072));
```
tananaev: Where does this number come from?

src/main/java/org/traccar/protocol/JmakProtocol.java
```diff
@@ -0,0 +1,39 @@
...
+                pipeline.addLast(new StringDecoder());
```
tananaev, line 35: This probably won't work for binary format?

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
```diff
@@ -0,0 +1,318 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+import java.math.BigInteger;
+
+public class JmakProtocolDecoder extends BaseProtocolDecoder {
+
+    private static final Map<Integer, String> STANDARD_DATA_MAP = new HashMap<>();
+    private static final Map<Integer, String> CAN_DATA_MAP = new HashMap<>();
```
tananaev: Why do we need this map instead of mapping directly in the code? Also the keys you're using don't follow our naming convention.

src/test/java/org/traccar/protocol/JmakProtocolDecoderTest.java
```diff
@@ -0,0 +1,24 @@
+package org.traccar.protocol;
+
+import org.junit.jupiter.api.Test;
+import org.traccar.ProtocolTest;
+
+public class JmakProtocolDecoderTest extends ProtocolTest {
```
tananaev, line 6: You also need to add a test for the frame decoder.

src/main/java/org/traccar/protocol/JmakProtocol.java
renansoaress: I removed the magic number and extracted it into a constant:

`private static final int DEFAULT_MAX_FRAME_LENGTH = 2000;`

This establishes the maximum allowed length for the data frame.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: Keeping the STANDARD_DATA_MAP and CAN_DATA_MAP makes the decoder more modular, easier to maintain and evolve, without compromising the clarity of the parsing logic.
This nomenclature is intended for use by our clients.

renansoaress:
The changes have been made and the questions have been answered. Could you please review them again?

tananaev:
Looks like it's broken. Please make sure you run `./gradlew check` before pushing code.

renansoaress:
> Looks like it's broken. Please make sure you run `./gradlew check` before pushing code.

Resolved!

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
tananaev: This is not resolved. Please do not mark things as resolved. I will mark them once I review.

src/main/java/org/traccar/protocol/JmakProtocol.java
tananaev: Not resolved.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
tananaev: Not resolved.

src/test/java/org/traccar/protocol/JmakFrameDecoderTest.java
```diff
@@ -0,0 +1,50 @@
+package org.traccar.protocol;
+
+import org.junit.jupiter.api.Test;
+import org.traccar.ProtocolTest;
+
+public class JmakFrameDecoderTest extends ProtocolTest {
+
+    @Test
+    public void testDecode() throws Exception {
```
tananaev: Please only include real data from a device or documentation. No made up tests.

Also please remove comments.

tananaev:
Looks like most of the comments are not resolved.

renansoaress:
> Looks like most of the comments are not resolved.

Ready!

tananaev:
Most of the comments are still not addressed.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: > Not resolved.

I don't understand. Please give an example of what you mean.

src/main/java/org/traccar/protocol/JmakProtocol.java
renansoaress: I modified the code and it's working correctly.

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
renansoaress: So the idea is not to handle the message limit?

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
tananaev: If you don't understand, why did you say that you resolved all the comments? You have to remove all these maps and just do something like a switch in the decoding method.

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
tananaev: Yes.

src/main/java/org/traccar/protocol/JmakProtocol.java
tananaev: How does it work for the binary format? Please show logs.

renansoaress: Here are the logs from testing the equipment sending data over all protocols—JSON, STRING, and ZIP. Please note that JSON and ZIP currently only return an ACK and do not save the data (they haven’t been implemented yet); only STRING MODE is implemented.
[ZIP_MODE.txt](https://github.com/user-attachments/files/20981902/ZIP_MODE.txt)
[STRING_MODE.txt](https://github.com/user-attachments/files/20981903/STRING_MODE.txt)
[JSON_MODE.txt](https://github.com/user-attachments/files/20981904/JSON_MODE.txt)

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: Thanks for the clarification and sorry for my limited familiarity with Java and the pull-request workflow. I’ve removed all the map-based logic and replaced it with a simple switch in the decoding method, as requested. Would you mind taking another look and letting me know if everything looks good now?

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
renansoaress: It was removed as requested;

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
```diff
@@ -0,0 +1,59 @@
...
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.TooLongFrameException;
+import org.traccar.BaseFrameDecoder;
+
+public class JmakFrameDecoder extends BaseFrameDecoder {
+
+    private final JsonFrameDecoder jsonDecoder;
```
tananaev: You can just extend JSON decoder instead.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
```diff
@@ -0,0 +1,286 @@
...
+                    case 15:
+                        position.set("canOilTemperature", Double.parseDouble(value));
+                        break;
+                    default:
+                        break;
+                }
+            }
+        }
+
+        position.setNetwork(new org.traccar.model.Network());
```
tananaev: What is this for?

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
```diff
@@ -0,0 +1,286 @@
...
+                        break;
+                    case 13:
+                        position.set("canFuelConsumption", Double.parseDouble(value));
+                        break;
+                    case 14:
+                        position.set("canFuelUsed", Double.parseDouble(value));
+                        break;
+                    case 15:
+                        position.set("canOilTemperature", Double.parseDouble(value));
+                        break;
```
tananaev: Please make sure you reuse standard attributes where possible. For example, we definitely have an attribute for RPM and can speed.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
```diff
@@ -0,0 +1,286 @@
...
+        String bin4 = new StringBuilder(
+                String.format("%4s", Integer.toBinaryString(value)).replace(' ', '0')
+        ).reverse().toString();
+        Map<String, Integer> info = new HashMap<>();
+        info.put("input1", bin4.charAt(0) == '1' ? 1 : 0);
+        info.put("input2", bin4.charAt(1) == '1' ? 1 : 0);
+        info.put("output1", bin4.charAt(2) == '1' ? 1 : 0);
+        info.put("output2", bin4.charAt(3) == '1' ? 1 : 0);
+        return info;
+    }
```
tananaev: This needs to be inlined. Also this is not the correct way for checking bits. Use BitUtil.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
```diff
@@ -0,0 +1,286 @@
...
+        int index = 1;
+        for (int i = 0; i < bin.length(); i++) {
+            if (bin.charAt(bin.length() - 1 - i) == '1' && index < parts.length) {
+                map.put(i + 1, parts[index++]);
+            }
+        }
+        return map;
+    }
+
+    private static Map<String, Integer> parseOnOffCan(String decimal) {
```
tananaev: Same here

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
```diff
@@ -0,0 +1,286 @@
...
+        Map<String, Integer> info = new HashMap<>();
+        info.put("canParkingBrake", bin6.charAt(0) == '1' ? 1 : 0);
+        info.put("canBrake",         bin6.charAt(1) == '1' ? 1 : 0);
+        info.put("canClutch",        bin6.charAt(2) == '1' ? 1 : 0);
+        info.put("canAirConditioning", bin6.charAt(3) == '1' ? 1 : 0);
+        info.put("canSeatBelt",        bin6.charAt(4) == '1' ? 1 : 0);
+        return info;
+    }
+
+    private static Map<String, Integer> parseEstadoIo(String decimal) {
```
tananaev: Also what is `Estado`?

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.Map;
+import java.math.BigInteger;
+
+public class JmakProtocolDecoder extends BaseProtocolDecoder {
+
+    public JmakProtocolDecoder(Protocol protocol) {
+        super(protocol);
+    }
+
+    private static String hexToBinString(String hex) {
```
tananaev: What does this do?

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
```diff
@@ -0,0 +1,286 @@
...
+        BigInteger value = new BigInteger(hex, 16);
+        String bin = value.toString(2);
+        int numBits = hex.length() * 4;
+        if (bin.length() < numBits) {
+            bin = String.format("%" + numBits + "s", bin).replace(' ', '0');
+        }
+        return bin;
+    }
+
+    private static Map<Integer, String> parseFlaggedData(String hex, String[] parts) {
```
tananaev: And this also?

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: hexToBinString(String hex)
Converts the hexadecimal flag mask (from both the main packet and the CAN block) into a full binary string, preserving leading zeros. Each hex digit equals 4 bits, so the method:

1. Parses the hex into a BigInteger.
 
2. Calls toString(2) to get the binary, which omits leading zeros.

3. Pads the result with zeros on the left up to hex.length()*4 bits.
This produces a fixed-length bitmask for flag inspection.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: parseFlaggedData(String hex, String[] parts)
Takes that same hex mask and the array of data values parsed from the packet. It walks through the binary mask from LSB to MSB, and each time it finds a ‘1’ bit, it maps that bit position to the next string in parts. Steps:

1. Convert hex to full binary via hexToBinString.

2. Loop over each bit; keep an index into parts.

3. On bit=1, do map.put(bitPosition, parts[index++]).
This yields a Map<Integer,String> of only the active fields in the standard or CAN section.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: It was the state of ios, it was added inlined.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: Ok, it was added inlined;

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: It was added inlined

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: Removed;

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
renansoaress: The idea is that we don't support JSON for now, we just return ack for a first configuration;

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: ok changed!

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
tananaev: I understand that, but you have still extend it.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
tananaev: Why would you want to convert it into a string? Computers can work just fine with bits directly.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
tananaev: This doesn't make sense either.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
```diff
@@ -0,0 +1,269 @@
...
+                        position.set(Position.KEY_OBD_ODOMETER, Double.parseDouble(value) * 1000);
+                        break;
+                    case 2:
+                        position.set("canHourmeter", Double.parseDouble(value));
+                        break;
+                    case 3:
+                        position.set(Position.KEY_OBD_SPEED, Double.parseDouble(value));
+                        break;
+                    case 4:
+                        int rpmInt = (int) Math.round(Double.parseDouble(value));
```
tananaev: Why do you need a variable here? Why rounding?

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
```diff
@@ -0,0 +1,269 @@
...
+                        position.set(Position.KEY_RPM, rpmInt);
+                        break;
+                    case 5:
+                        int canVal = Integer.parseInt(value);
+                        position.set("canParkingBrake", BitUtil.check(canVal, 0) ? 1 : 0);
+                        position.set("canBrake", BitUtil.check(canVal, 1) ? 1 : 0);
+                        position.set("canClutch", BitUtil.check(canVal, 2) ? 1 : 0);
+                        position.set("canAirConditioning", BitUtil.check(canVal, 3) ? 1 : 0);
+                        position.set("canSeatBelt", BitUtil.check(canVal, 4) ? 1 : 0);
```
tananaev: I would prefer this as a single bitmask.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
tananaev: We only use English in the codebase.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
tananaev: It seems like you literally just took examples that I provided and nothing else. You have to review all the parameters.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: The value received is a float, I'm just rounding it to an integer.

tananaev: You should just use it as is.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: fixed and added inline

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: All parameters were reviewed.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+        Position position = new Position(getProtocolName());
+        position.setDeviceId(session.getDeviceId());
+
+        Map<Integer, String> data = parseFlaggedData(standard[0], standard);
+        int eventId = 0;
+        int eventStatus = 0;
+        String eventName = null;
+        for (Map.Entry<Integer, String> entry : data.entrySet()) {
```
tananaev: Actually I realized what you're doing here and it's unnecessary. You should just check bits and check them as you go. Check for example `Gl200TextProtocolDecoder` lines 1060 - 1111. You should do something similar.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
tananaev: I don't see any updates.

renansoaress: https://github.com/traccar/traccar/pull/5591/commits/b6f1a60139f36c2232816f9ab16d8397f56c3adf

tananaev: You should review it again. You're missing a lot of standard attributes.

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
renansoaress: Could you explain?

tananaev: Instead of creating an instance, you just extend it. Then for JSON call super.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: I made this modification following this example, you can check.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: The idea here is to keep them separate at the clients’ request.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: Ok, modified.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: This logic was removed and bitUtil was added.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: This logic was removed and bitUtil was added.

src/main/java/org/traccar/protocol/JmakProtocolDecoder.java
renansoaress: Where can I get the list of these standard attributes? I believe that all possible attributes have already been used.

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
renansoaress: I’ve made the change as requested. Could you verify that it’s correct?

src/main/java/org/traccar/protocol/JmakFrameDecoder.java
```diff
@@ -0,0 +1,51 @@
...
+public class JmakFrameDecoder extends JsonFrameDecoder {
+
+    @Override
+    protected Object decode(ChannelHandlerContext ctx, Channel channel, ByteBuf buf) throws Exception {
+        if (buf.readableBytes() == 0) {
+            return null;
+        }
+        buf.markReaderIndex();
+        byte first = buf.readByte();
+        buf.resetReaderIndex();
```
tananaev, line 31: This is not the way to do it. You should just call `getByte`.

tananaev:
Looks like you haven't done the check again.

renansoaress:
> Looks like you haven't done the check again.

It was really missing, ready!

PR #5593, opened 2025-06-30
bugfix: ensure decodeExtras gets called when needed

commandini:


src/main/java/org/traccar/protocol/AlematicsProtocolDecoder.java
```diff
@@ -147,7 +147,8 @@ protected Object decode(
 
         if (parser.hasNext()) {
             position.set("text", parser.next());
-        } else if (parser.hasNext()) {
+        }
+        if (parser.hasNext()) {
```
commandini, line 151: Hi @tananaev

Original code block:
```
  if (parser.hasNext()) {
      position.set("text", parser.next());
  } else if (parser.hasNext()) {
      decodeExtras(position, parser);
  }
```

If the 1st if-block gets executed, the 2nd if-block cannot get executed.
If the 1st if-block doesn't get executed, the 2nd if-block cannot get executed.
It means 2nd if-block is dead code. I guess the intention was using two consecutive if-blocks and not if-else block.
This change also looks consistent with the bodies of the if-blocks, i.e. `parser.next()` is called in the 1st if-block and if parser still has next items, `decodeExtras` gets called in the 2nd if-block. What do you think?

tananaev: > If the 1st if-block doesn't get executed, the 2nd if-block cannot get executed.

This is incorrect.

commandini: @tananaev Does it mean 2 consecutive calls to `parser.hasNext() `can return different results? E.g. first call returning false and second call returning true?

tananaev: Yes. Please check what `hasNext` does. It skips to the next element.

commandini: ok, closing this PR then. thanks for reviewing.

PR #5594, opened 2025-06-30
some code improvements

commandini:


src/main/java/org/traccar/api/BaseObjectResource.java
```diff
@@ -100,7 +100,7 @@ public Response update(T entity) throws Exception {
         if (entity instanceof User after) {
             User before = storage.getObject(User.class, new Request(
                     new Columns.All(), new Condition.Equals("id", entity.getId())));
-            permissionsService.checkUserUpdate(getUserId(), before, (User) entity);
+            permissionsService.checkUserUpdate(getUserId(), before, after);
```
commandini, line 103: reusing the variable created via pattern matching (line 100).

src/main/java/org/traccar/protocol/CastelProtocolDecoder.java
```diff
@@ -583,7 +583,7 @@ private Object decodeMpip(
 
         if (type == 0x4001) {
 
-            sendResponse(channel, remoteAddress, version, id, (short) type, null);
+            sendResponse(channel, remoteAddress, version, id, type, null);
```
commandini, line 587: removed unnecessary casting.

src/main/java/org/traccar/protocol/GranitProtocolDecoder.java
```diff
@@ -209,7 +209,7 @@ protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg
                     if (buf.getUnsignedByte(buf.readerIndex()) != 0xFE) {
                         Position position = new Position(getProtocolName());
                         position.setDeviceId(deviceSession.getDeviceId());
-                        position.setTime(new Date((unixTime + i * timeIncrement) * 1000));
+                        position.setTime(new Date((unixTime + (long) i * timeIncrement) * 1000));
```
commandini: i * increment is multiplication of two integers. Casting one of them to long prevents integer overflow (long overflow is still possible though).

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -103,16 +103,16 @@ private String decodeEvent(String event, Position position, String model) {
             char inputStatus = event.charAt(0);
             if (inputStatus >= 'A' && inputStatus <= 'D') {
                 int inputStatusInt = inputStatus - 'A';
-                position.set(Position.PREFIX_IN + 1, (boolean) BitUtil.check(inputStatusInt, 0));
-                position.set(Position.PREFIX_IN + 2, (boolean) BitUtil.check(inputStatusInt, 1));
+                position.set(Position.PREFIX_IN + 1, BitUtil.check(inputStatusInt, 0));
+                position.set(Position.PREFIX_IN + 2, BitUtil.check(inputStatusInt, 1));
                 if ("SF-Lite".equals(model)) {
```
commandini, line 108: removed unnecessary casting.

src/main/java/org/traccar/protocol/LaipacProtocolDecoder.java
```diff
@@ -103,16 +103,16 @@ private String decodeEvent(String event, Position position, String model) {
...
                     position.set(Position.PREFIX_IN + 3, false);
                 }
                 return null;
             } else if (inputStatus >= 'O' && inputStatus <= 'R') {
                 int inputStatusInt = inputStatus - 'O';
-                position.set(Position.PREFIX_IN + 1, (boolean) BitUtil.check(inputStatusInt, 0));
-                position.set(Position.PREFIX_IN + 2, (boolean) BitUtil.check(inputStatusInt, 1));
+                position.set(Position.PREFIX_IN + 1, BitUtil.check(inputStatusInt, 0));
+                position.set(Position.PREFIX_IN + 2, BitUtil.check(inputStatusInt, 1));
                 if ("SF-Lite".equals(model)) {
```
commandini, line 116: removed unnecessary casting.

src/main/java/org/traccar/protocol/NavigilFrameDecoder.java
```diff
@@ -35,10 +35,7 @@ protected Object decode(
         }
 
         // Check for preamble
-        boolean hasPreamble = false;
-        if (buf.getUnsignedIntLE(buf.readerIndex()) == PREAMBLE) {
-            hasPreamble = true;
-        }
+        boolean hasPreamble = buf.getUnsignedIntLE(buf.readerIndex()) == PREAMBLE;
```
commandini, line 39: shorter initialization of the boolean variable.

src/main/java/org/traccar/protocol/TelemaxProtocolDecoder.java
```diff
@@ -97,7 +97,7 @@ protected Object decode(
                 time = new SimpleDateFormat("yyMMddHHmmss").parse(readValue(sentence, index, 12));
                 position.set(Position.KEY_STATUS, readValue(sentence, index, 8));
             } else {
-                time = new Date(time.getTime() + interval * 1000);
+                time = new Date(time.getTime() + interval * 1000L);
```
tananaev: This is incorrect. `interval` is an int, so if we don't multiply by a long explicitly, it will overflow.

src/main/java/org/traccar/protocol/NoranProtocolDecoder.java
```diff
@@ -72,12 +72,9 @@ protected Object decode(
...
                 || type == MSG_CONTROL_RESPONSE || type == MSG_ALARM) {
 
-            boolean newFormat = false;
-            if (type == MSG_UPLOAD_POSITION && buf.readableBytes() == 48
+            boolean newFormat = type == MSG_UPLOAD_POSITION && buf.readableBytes() == 48
                     || type == MSG_ALARM && buf.readableBytes() == 48
-                    || type == MSG_CONTROL_RESPONSE && buf.readableBytes() == 57) {
-                newFormat = true;
-            }
+                    || type == MSG_CONTROL_RESPONSE && buf.readableBytes() == 57;
```
commandini, line 77: shorter initialization of the boolean variable.

src/main/java/org/traccar/protocol/TelemaxProtocolDecoder.java
```diff
@@ -97,7 +97,7 @@ protected Object decode(
...
             }
```
commandini: prevent integer overflow.

src/main/java/org/traccar/protocol/GranitProtocolDecoder.java
tananaev: Why do we need this?

commandini: `i `and `timeIncrement` are both integers, their multiplication can overflow.

tananaev: It cannot because the numbers we expect here are small.

src/main/java/org/traccar/protocol/TelemaxProtocolDecoder.java
commandini: ok, so you agree with this change, right? Because it suggest to multiply by a long.

src/main/java/org/traccar/protocol/GranitProtocolDecoder.java
commandini: ok, shall I revert this line and the rest of the PR looks good to you?

src/main/java/org/traccar/protocol/TelemaxProtocolDecoder.java
tananaev: Sorry, I read it backwards. It's the same situation as the other place. It's not required because we don't expect large numbers here.

src/main/java/org/traccar/protocol/GranitProtocolDecoder.java
tananaev: I have some other comments, but yes, this one you can simply revert.

tananaev:
Merged, thank you.

PR #5596, opened 2025-07-02
Add ACK response for Sinotrack ST-901L binary packets [protocol H02]

dontbug:
Testing assumptions to solve the problem with Sinotrack ST-901L 
Details: https://www.traccar.org/forums/topic/sinotrack-st-901-config-issues/page/9/#post-112372

### DIFF
--- a/src/main/java/org/traccar/protocol/H02ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/H02ProtocolDecoder.java
@@ -155,6 +155,10 @@ public class H02ProtocolDecoder extends BaseProtocolDecoder {
 
         processStatus(position, buf.readUnsignedInt());
 
+        if (channel != null) {
+            channel.writeAndFlush(new NetworkMessage("ON", remoteAddress));
+        }
+
         return position;
     }
 


tananaev:

dontbug:
Probabily need response ACK.

From my observations I noticed that Sinotrack ST-901L sends 2 formats, both in ASCII and in bit.
For ASCII, traccar responds, and takes over and remembers the GPS client and does not repeat sendings, and for a bit request, unfortunately, there is no answer in the code.
I don’t know where it came from using ChatAI, but they said that it is enough to answer ON or LOAD.
So as a solution, you can try this moment for the file src\main\java\org\traccar\protocol\H02ProtocolDecoder.java in the function decodeBinary add a response. This is how it will look.

Here, from the logs and from the packets, you can see that there is a response only for ASCII packets that start with *HQ
Log files:
logs

Packet Sender:
2025-07-02 19:12:33.232 - send Package *HQ
2025-07-02 19:21:29.859 - send Package binary
packet_sender

You can notice that the TRACCAR server did not respond to the TCP packet. Presumably, this is why the Sinotrack ST-901L took the sending of this packet for retry.
I am not sure about the answer, but I need to test it, since I did not find documentation with binary processing and what kind of ACK response is needed in this case

tananaev:
Please provide logs before and after. Does it actually fix anything?

dontbug:
> Please provide logs before and after. Does it actually fix anything?

Unfortunately, these are not solutions, I have been struggling to find the right answer for several hours now. But so far without success. If anything, I will reopen a new pull request when I definitely succeed

PR #5597, opened 2025-07-03
Add ACK response for Sinotrack ST-901L binary packets [protocol H02]

dontbug:
Based on the analysis of servers from pro.sinotrack.com and their responses, we added the same response that is required for binary queries. In the end, the solution works for Sinotrack ST-901L and no more repeating old queries accumulate.

I am attaching screenshots of requests and responses to my Traccar server for processing 2 different types of requests from Sinotrack ST-901L:
packet_sniffer

Related old pull request https://github.com/traccar/traccar/pull/5596

### DIFF
--- a/src/main/java/org/traccar/protocol/H02ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/H02ProtocolDecoder.java
@@ -155,10 +155,8 @@ public class H02ProtocolDecoder extends BaseProtocolDecoder {
 
         processStatus(position, buf.readUnsignedInt());
 
-        if (channel != null) {
-            sendResponse(channel, remoteAddress, id, "R12");
-        }
-        
+        sendResponse(channel, remoteAddress, id, "R12");
+
         return position;
     }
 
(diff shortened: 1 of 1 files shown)


tananaev:
And you still don't have any documentation on what `R12` means and why we need to send it? How do we know it won't break some devices?

dontbug:
> And you still don't have any documentation on what `R12` means and why we need to send it? How do we know it won't break some devices?

I took the answers based on what the server from https://sinotrack.com/
For example, the PRO server: 45.112.204.245:8090 that processes ST-901L responds for binary requests, with the answer R12.
packet_sniffer2
Screenshot 2025-07-03 062943


As for other devices, there has not been a single answer for binnary so far.
I will also enter into negotiations with Sinotrack and they said that they will send the protocol by mail, and I will return with confirmation also from Sinotrack representatives.

dontbug:
Analyzing the [forum](https://www.traccar.org/forums/topic/sinotrack-st-901-config-issues/page/9/#post-112372), I noticed many problems with Sinotrack ST-901L, and others also have the same problem in the logs that I encountered when processing binary data sent from the GPS tracker.
After implementing this functionality, my server processes correctly, and repeated cyclic requests every second from the GPS tracker stopped coming because they were processed.
Cyclic requests affected the traffic consumption of the GPS tracker.

Traffic consumption per month when sending data to Traccar Server (two identical configurations related to the frequency of data sending):
ST-901: 88.95 MB 
ST-901L-4G: 212 MB

tananaev:
Make sure you run `gradle check` and fix any issues.

PR #5598, opened 2025-07-04
Optimize Docker images

tananaev:
@magna-z what do you think about doing images like this?

Basically 3 stages:
1. Unzip Traccar package. Separate stage means that we can reuse this code between all images and don't need to worry about installing and cleaning up unzip.
2. Package minimal version of JRE. This is what we do for official installers. Makes sense to do it for docker as well. It can save us quite a bit in size. It's also reusable. The only difference is that alpine needs a different base package.
3. Final docker image. It's basically the same across all with different base.

In my local testing it saved about 100MB on alpine image size.

### DIFF
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -1,14 +1,16 @@
-FROM alpine:3.22
-
+FROM alpine AS package
 ARG VERSION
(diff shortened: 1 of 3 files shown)


tananaev:
@magna-z I'll merge this for now as I'm going to do a release. But if there are any concerns, we can change back later.

Issue #5592, opened 2025-06-28
[feature request] Include id in position forwarding.


jcardus:
I'm forwarding the positions to another storage to use as an archive.

PR #5599, opened 2025-07-04
Add feature to include id in position forwarding

DhruvTheDev1:
Addesses #5592 
- Added JsonProperty("id") in Position to include 'id' in JSON serialization for forwarding

I reviewed the full hierarchy for Position class (baseModel -> ExtendedModel -> Message -> Position) 
I checked Position Forward Implementations to ensure it correctly handles the updated JSON with the new id field

### DIFF
--- a/src/main/java/org/traccar/model/Position.java
+++ b/src/main/java/org/traccar/model/Position.java
@@ -20,6 +20,8 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import org.traccar.storage.QueryIgnore;
 import org.traccar.storage.StorageName;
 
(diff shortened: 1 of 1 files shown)


tananaev:
Have you actually tested it? Please provide JSON before and after your change.

DhruvTheDev1:
@tananaev I did test it, I made a test class in test/forwarding as there was no Model and all tests passed - It included the ID in the JSON.

tananaev:
> Please provide JSON before and after your change.

DhruvTheDev1:
@tananaev 
I tested the JSON before and after adding the code.

Before:
```json
{
  "id": 12345,
  "attributes": {},
  "deviceId": 42,
  "protocol": null,
...
}
```
After:
```json
{
  "attributes": {},
  "deviceId": 42,
  "protocol": null,
...
  "id": 12345
}
```
Even without adding the code it does show id still. The annotation just changes the order of the id.

Is my change actually causing the forwarded to include the id field when sending the position data or was it already being forwarded without this change?

tananaev:
You don't know what your change did? I'm very confused.

DhruvTheDev1:
To be honest I'm still fairly new to the codebase. I reviewed the positioning class hierarchy and based on the tests, the id field appears in the JSON both before and after my change, so I'm not sure if my change actually affects the forwarding behaviour or if it's already included by default.

PR #5605, opened 2025-07-09
Streaming in storage layer

jcardus:


src/test/java/org/traccar/reports/ReportUtilsTest.java
```diff
@@ -422,6 +426,7 @@ public void testDetectTripAndStopByGap() throws Exception {
         assertEquals(7, itemTrip.getMaxSpeed(), 0.01);
         assertEquals(600, itemTrip.getDistance(), 0.01);
 
+        when(storage.getObjects(eq(Position.class), any())).thenReturn(data.stream());
```
tananaev: Why not make data a stream instead of calling `.stream` everywhere?

tananaev:
It's more messy than I expected it to be. Let's just change QueryBuilder as the first step. And add helper method to return old format. That way we're only changing one class. But I think QueryBuilder implementation needs improvement. It doesn't feel right with all the catches and `ResultSet` outside. There must be a cleaner way to do it.

jcardus:
I tried but couldnt find a cleaner way.

Ok.

tananaev:
Let's start by just changing the QueryBuilder.

jcardus:
What about creating a new method in QueryBuilder instead of changing the current one? Like executeQueryStreamed.

tananaev:
Yes, we can create a new method for now.

src/main/java/org/traccar/storage/QueryBuilder.java
```diff
@@ -461,6 +443,90 @@ public <T> List<T> executeQuery(Class<T> clazz) throws SQLException {
...
+
+    private void close(ResultSet resultSet) {
+        try {
+            if (resultSet != null) {
+                resultSet.close();
+            }
+            statement.close();
+            connection.close();
+        } catch (SQLException e) {
+            throw new RuntimeException(e);
```
jcardus, line 526: Maybe here we should just log the exception...

tananaev:
I'll merge for now and do a bit of a cleanup.

PR #5607, opened 2025-07-11
Solve the problem that administrators cannot query other users' devices based on deviceIds and uniqueIds

zhizunyoulan:
When non-administrators query devices, they only check those whose userId is their own.

Administrators can query devices with any userId. When no userId is passed, if all is true, all users' will be queried; otherwise, only their own will be queried.

Between deviceIds and uniqueIds, there is an OR relationship, and with userId, there is an AND relationship


tananaev:
What is the use case for this?

zhizunyoulan:
> What is the use case for this?

In the current code, the userId is not an independent query condition. It only takes effect when uniqueIds are empty, deviceIds are empty, all=false, and userId is not equal to 0. In my new code, userId becomes an independent query condition. As long as all=false and userId is not equal to 0, it will take effect. This allows querying specific devices of a particular user.

PR #5608, opened 2025-07-11
Solve the problem that administrators cannot query other users' devices based on deviceIds and uniqueIds

zhizunyoulan:
When non-administrators query devices, they only check those whose userId is their own.

Administrators can query devices with any userId. When no userId is passed, if all is true, all users' will be queried; otherwise, only their own will be queried.

Between deviceIds and uniqueIds, there is an OR relationship, and with userId, there is an AND relationship


nilei:
@zhizunyoulan  Are you going to resolve the conflicts? I am also having this issue that I want to retrieve device information with an admin account based on the unique device ID instead of internal device ID.

PR #5609, opened 2025-07-13
getObjectsStreamed in Storage

jcardus:


src/main/java/org/traccar/storage/DatabaseStorage.java
```diff
@@ -73,12 +74,19 @@ public <T> List<T> getObjects(Class<T> clazz, Request request) throws StorageExc
...
             }
-            return builder.executeQuery(clazz);
+            return builder.executeQueryStreamed(clazz);
         } catch (SQLException e) {
             throw new StorageException(e);
         }
     }
 
+    @Override
+    public <T> List<T> getObjects(Class<T> clazz, Request request) throws StorageException {
```
tananaev: Let's put this method above `getObjectsStreamed` for consistency.

src/main/java/org/traccar/storage/Storage.java
```diff
@@ -20,11 +20,14 @@
 import org.traccar.storage.query.Request;
 
 import java.util.List;
+import java.util.stream.Stream;
 
 public abstract class Storage {
 
     public abstract <T> List<T> getObjects(Class<T> clazz, Request request) throws StorageException;
 
+    public abstract <T> Stream<T> getObjectsStreamed(Class<T> clazz, Request request) throws StorageException;
```
tananaev: Let's call it `getObjectsStream`.

jcardus: it's executeQueryStreamed in QueryBuilder

tananaev: I know, but I think in this case it sounds better if you just call it stream. Because it's a getter and not called "execute".

jcardus: alright

PR #5610, opened 2025-07-14
Streaming in PositionResource

jcardus:
This does not optimize for the case where only positionIds are used or when we request just latest positions, as those datasets are typically small.
I'm slightly concerned about flushing after every record, but I fear that if we don't flush explicitly, the stream will only flush once it's closed.

### DIFF
--- a/src/test/java/org/traccar/reports/ReportUtilsTest.java
+++ b/src/test/java/org/traccar/reports/ReportUtilsTest.java
@@ -25,6 +25,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.TimeZone;
+import java.util.stream.Stream;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
@@ -104,7 +105,7 @@ public class ReportUtilsTest extends BaseTest {
     @Test
     public void testDetectTripsSimple() throws Exception {
 
-        List<Position> data = Arrays.asList(
+        Stream<Position> data = Stream.of(
                 position("2016-01-01 00:00:00.000", 0, 0),
                 position("2016-01-01 00:01:00.000", 0, 0),
                 position("2016-01-01 00:02:00.000", 10, 0),
@@ -113,7 +114,7 @@ public class ReportUtilsTest extends BaseTest {
                 position("2016-01-01 00:05:00.000", 0, 3000),
                 position("2016-01-01 00:15:00.000", 0, 3000),
                 position("2016-01-01 00:25:00.000", 0, 3000));
-        when(storage.getObjectsStream(eq(Position.class), any())).thenReturn(data.stream());
+        when(storage.getObjectsStream(eq(Position.class), any())).thenReturn(data);
 
         Device device = mockDevice(500, 300, 180, 900, false);
         ReportUtils reportUtils = new ReportUtils(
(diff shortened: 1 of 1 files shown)


tananaev:
I think instead of implementing `StreamingOutput` for every method, we should do something like this:
```java
@Provider
@Produces(MediaType.APPLICATION_JSON)
public class StreamWriter implements MessageBodyWriter<Stream<?>> {

    private final ObjectMapper mapper = new ObjectMapper();

    @Override
    public boolean isWriteable(Class<?> type, Type gType, Annotation[] a, MediaType m) {
        return Stream.class.isAssignableFrom(type);
    }

    @Override
    public void writeTo(Stream<?> stream, Class<?> t, Type gt, Annotation[] a,
                        MediaType m, MultivaluedMap<String,Object> h, OutputStream os)
                        throws IOException {
        try (JsonGenerator g = mapper.createGenerator(os)) {
            g.writeStartArray();
            stream.forEach(o -> {
                try { mapper.writeValue(g, o); }
                catch (IOException io) { throw new UncheckedIOException(io); }
            });
            g.writeEndArray();
        }
    }
}
```

jcardus:
I had to put stream inside the try otherwise we would have a connection leak

src/test/java/org/traccar/reports/ReportUtilsTest.java
```diff
@@ -290,7 +294,7 @@ public void testDetectStopsOnly() throws Exception {
                 position("2016-01-01 00:03:00.000", 0, 0),
                 position("2016-01-01 00:04:00.000", 1, 0),
                 position("2016-01-01 00:05:00.000", 0, 0));
-        when(storage.getObjects(eq(Position.class), any())).thenReturn(data);
+        when(storage.getObjectsStream(eq(Position.class), any())).thenReturn(data.stream());
```
tananaev: We can probably create a stream directly.

src/test/java/org/traccar/reports/ReportUtilsTest.java
```diff
@@ -422,6 +426,7 @@ public void testDetectTripAndStopByGap() throws Exception {
         assertEquals(7, itemTrip.getMaxSpeed(), 0.01);
         assertEquals(600, itemTrip.getDistance(), 0.01);
 
+        when(storage.getObjectsStream(eq(Position.class), any())).thenReturn(data.stream());
```
tananaev: It might make sense to split this into two tests

src/main/java/org/traccar/api/StreamWriter.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Override
+    public void writeTo(Stream<?> stream, Class<?> type, Type genericType, Annotation[] annotations,
+                        MediaType mediaType, MultivaluedMap<String, Object> httpHeaders, OutputStream entityStream)
+            throws IOException {
+        try (stream;
+             JsonGenerator jsonGenerator = objectMapper.createGenerator(entityStream)) {
+            jsonGenerator.writeStartArray();
+            stream.forEach(object -> {
```
tananaev: Can we do `for (T element : stream.iterator()) {` here and avoid try/catch?

src/main/java/org/traccar/api/StreamWriter.java
```diff
@@ -0,0 +1,46 @@
...
+    public boolean isWriteable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return Stream.class.isAssignableFrom(type);
+    }
+
+    @Override
+    public void writeTo(Stream<?> stream, Class<?> type, Type genericType, Annotation[] annotations,
+                        MediaType mediaType, MultivaluedMap<String, Object> httpHeaders, OutputStream entityStream)
+            throws IOException {
+        try (stream;
+             JsonGenerator jsonGenerator = objectMapper.createGenerator(entityStream)) {
```
tananaev:
```suggestion
        try (stream; var generator = objectMapper.createGenerator(entityStream)) {
```

PR #5615, opened 2025-07-17
Add curl to docker images to fix healthcheck failures

Shadoweee77:
Healthchecks were failing due to missing curl inside the containers. This commit updates the final stages of Dockerfiles (based on Alpine, Debian, and Ubuntu) to include curl, ensuring healthchecks can run properly.

### DIFF
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -9,6 +9,7 @@ RUN jlink --module-path $JAVA_HOME/jmods \
     --strip-debug --no-header-files --no-man-pages --compress=2 --output /jre
 
 FROM alpine:3.22
+RUN apk add --no-cache curl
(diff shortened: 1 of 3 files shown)


tananaev:
Is this for the docker compose? I'm pretty sure it's working for me. What's the error?

Shadoweee77:
@tananaev Yes this is for compose healthchecks. As for the error I just get the unhealthy status when checking the container via `docker inspect --format='{{.State.Health.Status}}' traccar_server`. Noticed this because traefik wouldn't proxy when the container was reporting as starting/unhealthy.

tananaev:
This should solve the problem:

https://github.com/traccar/traccar/commit/e5c8a223d5c142de4c79ced86dbc0b6a7be7ec1b

PR #5616, opened 2025-07-18
Include Parsing Control State Flags and Security State Flags from CAN Modules

troianvasile:
Parsing Control State Flags and Security State Flags on devices that have ALL-CAN / LV-CAN  module connected.


tananaev:
We don't usually decode individual bits of information like that. It's because it takes too much spaces in the database and in most cases people don't need all of it.

PR #5617, opened 2025-07-20
Fix for failing docker compose healthchecks failing

Shadoweee77:
In reference to #5615 and commit e5c8a22, the `wget` command in the Docker healthcheck used an invalid `-s` option (not supported in BusyBox wget). This caused the healthcheck to fail silently.

This update:

- Replaces `-s` with the correct `-S` to show server response
- Adds `--spider` to prevent wget from attempting to download and save the /api/health response as a file named `health,` which can lead to file conflicts and failing healthchecks

### DIFF
--- a/docker/compose/traccar-mysql.yaml
+++ b/docker/compose/traccar-mysql.yaml
@@ -33,7 +33,7 @@ services:
       DATABASE_USER: traccar
       DATABASE_PASSWORD: traccar
     healthcheck:
-      test: [ "CMD", "wget", "-q", "-s", "http://localhost:8082/api/health" ]
+      test: [ "CMD", "wget", "-q", "-S", "http://localhost:8082/api/health" ]
       interval: 2m
       timeout: 5s
       start_period: 1h
--- a/docker/compose/traccar-mysql.yaml
+++ b/docker/compose/traccar-mysql.yaml
@@ -33,7 +33,7 @@ services:
       DATABASE_USER: traccar
       DATABASE_PASSWORD: traccar
     healthcheck:
-      test: [ "CMD", "wget", "-q", "-S", "http://localhost:8082/api/health" ]
+      test: [ "CMD", "wget", "--spider", "-q", "-S", "http://localhost:8082/api/health" ]
       interval: 2m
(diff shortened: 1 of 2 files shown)


tananaev:
Can you please reference to the manual with available commands.

Shadoweee77:
My bad must've found some shitty man before - clearly states the `-s` should act as a `--spider`. Either way the PR fixed healthcheck issues for me.

tananaev:
> Either way the PR fixed healthcheck issues for me.

Do you mean your PR fixes it? But what is broken exactly?

tananaev:
OK, this is what's expected:
```
# docker compose exec traccar wget --help
BusyBox v1.37.0 (2025-05-26 20:04:45 UTC) multi-call binary.

Usage: wget [-cqS] [--spider] [-O FILE] [-o LOGFILE] [--header STR]
	[--post-data STR | --post-file FILE] [-Y on/off]
	[-P DIR] [-U AGENT] [-T SEC] URL...

Retrieve files via HTTP or FTP

	--spider	Only check URL existence: $? is 0 if exists
	--header STR	Add STR (of form 'header: value') to headers
	--post-data STR	Send STR using POST method
	--post-file FILE	Send FILE using POST method
	-c		Continue retrieval of aborted transfer
	-q		Quiet
	-P DIR		Save to DIR (default .)
	-S    		Show server response
	-T SEC		Network read timeout is SEC seconds
	-O FILE		Save to FILE ('-' for stdout)
	-o LOGFILE	Log messages to FILE
	-U STR		Use STR for User-Agent header
	-Y on/off	Use proxy
```

tananaev:
So I think we should just replace `-s` with `--spider`. Can you please update the PR.

Shadoweee77:
Initially the container reports `starting` - which is fine. After that we get a `healthy` status, but when the next check comes it will report `unhealthy` as the wget fails because it already has downloaded the api/`health` file.

tananaev:
Updated here: https://github.com/traccar/traccar/commit/adcc42280733ff9aa34e9f7dacc8429a5f4aa1ac.

PR #5621, opened 2025-07-28
Update Minifinder2ProtocolEncoder.java

markdegrootnl:
Actually send Minifinder config read command.

Also added an option to send arbitrary config data to the device.

### DIFF
--- a/src/main/java/org/traccar/protocol/Minifinder2ProtocolEncoder.java
+++ b/src/main/java/org/traccar/protocol/Minifinder2ProtocolEncoder.java
@@ -20,6 +20,7 @@ import io.netty.buffer.Unpooled;
 import org.traccar.BaseProtocolEncoder;
 import org.traccar.Protocol;
 import org.traccar.helper.Checksum;
+import org.traccar.helper.DataConverter;
 import org.traccar.model.Command;
 
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


tananaev:
Need a bit more context here. Also please attach the documentation.

markdegrootnl:
I'm afraid I cannot share the documentation with you since I got it under NDA. 

The idea is you send a packet with type `MSG_CONFIGURATION` (`0x02`). 

If you send the command with key `0xF0` the device will return the current configuration. This was already partially implemented in 64b2d09ecc58ba68eea39fdb538f22accf386e72 but the command was never sent because the line `return encodeContent(content);` was missing. 

I've expanded the implementation so you can send the desired configuration in the `KEY_DATA` parameter as hex string and it will be sent to the device. If you don't supply any `KEY_DATA`, the old implementation of reading the current configuration will be used.

There are a ton of configuration settings you can change. The structure of the data is the same as the decode implementation you implemented in 1e0cc806b7386680492cb3b6b266bb4f958b2f1e

markdegrootnl:
Found some documentation on your site: https://www.traccar.org/protocol/5187-minifinder2/Eview%20Pet%20Tracker%20Communication%20Protocol%20V20230506.pdf

Page 21 describes the CONFIGURATION COMMAND(0X02) KEY LIST and the subsequent pages have the details for each configuration paramater

PR #5623, opened 2025-07-30
Huabao and RS232

felipeparente:
Added RS232 data to the hexData field, in case it is not mapped by the Huabao protocol.


src/main/java/org/traccar/protocol/HuabaoProtocolDecoder.java
```diff
@@ -1162,12 +1162,25 @@ private Position decodeTransparent(DeviceSession deviceSession, ByteBuf buf) {
             position.setDeviceId(deviceSession.getDeviceId());
 
             getLastLocation(position, null);
-            String data = buf.readCharSequence(buf.readableBytes(), StandardCharsets.US_ASCII).toString().trim();
+
+            // Capture the remaining bytes from the buffer BEFORE attempting to convert them to String
+            int readableBytes = buf.readableBytes();
+            byte[] rawBytes = new byte[readableBytes];
+            buf.readBytes(rawBytes);
```
tananaev: Instead of doing this, I think you should check the header first and then either read the string or dump the raw data hex.

src/main/java/org/traccar/model/Position.java
```diff
@@ -103,6 +103,7 @@ public class Position extends Message {
 
     public static final String KEY_DRIVER_UNIQUE_ID = "driverUniqueId";
     public static final String KEY_CARD = "card";
+    public static final String KEY_HEX_DATA = "hexData";
```
tananaev: We don't new attributes unless they are used by at least 2 different protocol. Also in this case it's not clear what this HEX even means.

felipeparente: This field would send in attributes the hexadecimal data read from RS232.
I came to look for another field in the Protocol class, but I didn't find any suitable for it.

tananaev: You can probably just store it as "data", but you should define it here. It should just be a custom attribute.

src/main/java/org/traccar/protocol/HuabaoProtocolDecoder.java
felipeparente: I am reading the whole content of the buffer initially, in case the mapped header is not found, all its content is sent in the Position. KEY_HEX_DATA field.

In any case, the whole content of the buffer will be read.

src/main/java/org/traccar/model/Position.java
felipeparente: Now I didn't understand.
And isn't that what I did?
In this case, are you suggesting that I only change the name "Hexdata" to "date"?

tananaev: You cannot add it to `Position.java`. Instead you should just do it inline. Something like this:
```java
positions.set("data", ...);
```

felipeparente: OK, I'll change that.

felipeparente (commented):
Fixed field as instructed.

src/main/java/org/traccar/protocol/HuabaoProtocolDecoder.java
tananaev: This is still not resolved.

felipeparente: I did it this way because I wouldn't need to return byte read pointers if I needed to retrieve all the data, as in the "else" case.
Also, in the future, another "startsWith" check might appear, and the capture would already have been performed.

felipeparente: The original has this line:
String data = buf.readCharSequence(buf.readableBytes(), StandardCharsets.US_ASCII).toString().trim();

Right away, it already converts to ASCII; I'd have to convert it back to send the raw HEX data, which is what I need.

tananaev: That needs to be changed, but you can easily just peek the header with incrementing the index.

felipeparente: Change submitted.
I had to return the buffer index and read the buffer again to send the raw data in hexadecimal.

tananaev: I don't think this makes sense. Why do you need to reset index? What are resetting it to? Why do you need a byte array?

felipeparente: I need the raw data in hexadecimal.
For example, in this buf object, I receive the following data: 40000A415303210420000000DC
With this data arriving in the "data" field of the "attributes" field, I can further process it.

If I retrieve it, converted to ASCII, as the line (String data = buf.readCharSequence(buf.readableBytes(), StandardCharsets.US_ASCII).toString().trim();) does, I would get the following data: @ AS ! Ü
Which makes no sense at all.

And since this line already reads data from the buffer, I need to return the read pointer to get the original raw data in hexadecimal.

tananaev: I think you need to read my earlier comments a bit more carefully. I also recommend checking how it's done in other places.

felipeparente: I read and reread your initial comment, I also checked how they are done in other cases, and see that they use different forms, but in any case, the process of capturing raw data has now been simplified and committed.

tananaev: You should never need to use `resetReaderIndex`.

felipeparente: I need it because I don't want to change the initial code, and if you notice, the existing code:
String data = buf.readCharSequence(buf.readableBytes(), StandardCharsets.US_ASCII).toString().trim();

This already consumes the entire buffer; to get the raw data, the index must be reset.

tananaev: That's why I said earlier that you should peek the header instead of reading the whole string upfront.

felipeparente: I understood that, but the code that reads the entire string isn't mine; it was already in the code, and I didn't want to change it to avoid problems.
And the reading is still converted to ASCII, which isn't useful in my case.
By resolving this, I can use multiple sensors with this device model.

tananaev: You should update the existing code if it makes sense. In this case I think it makes perfect sense.

PR #5628, opened 2025-08-09
Add support to decode car acc status for h02 protocol

samm-git:
This PR should add support for the ST906 battery reporting provided as reply to CXZT command:

```
2025-08-09 21:19:15  INFO: [T13cb679d: h02 > 1.2.3.4] CXZT
2025-08-09 21:19:15  INFO: [T13cb679d: h02 < 1.2.3.4] *HQ,11223344,SMS,ST906(70SACD)_TQ_V_2.0 2024/06/07\nID:11223344\nIP:1.2.3.4 5013\nUT:30,30,300\nVOLT:13.1V\nAPN:internet.example.com\nGPS:A-22-20\nGSM:26#
```
My plan is to run CXZT command using API every hour and use computed attributes to have it in the logs and screen.

### DIFF
--- a/src/test/java/org/traccar/protocol/H02ProtocolDecoderTest.java
+++ b/src/test/java/org/traccar/protocol/H02ProtocolDecoderTest.java
@@ -267,25 +267,6 @@ public class H02ProtocolDecoderTest extends ProtocolTest {
         
         verifyPosition(decoder, binary(
                 "24971305007205201916101533335008000073206976000000effffbffff000252776566060000000000000000000049"));
-
-        verifyAttribute(decoder, buffer(
-                "*HQ,135790246811220,HTBT,100#"),
-                Position.KEY_BATTERY_LEVEL, 100);                
-    }
-
-    @Test
-    public void testDecodeStatus() throws Exception {
-
-        var decoder = inject(new H02ProtocolDecoder(null));
-
-        verifyAttribute(decoder, buffer(
-                "*HQ,2705171109,V1,213324,A,5002.5849,N,01433.7822,E,0.00,000,140613,FFFFFFFF#"),
-                Position.KEY_STATUS, 0xFFFFFFFFL);
-
-        verifyAttribute(decoder, binary(
-                "2441091144271222470112142233983006114026520E000000FFFFFBFFFF0014060000000001CC00262B0F170A"),
-                Position.KEY_STATUS, 0xFFFFFBFFL);
-
         verifyAttribute(decoder, buffer(
                 "*HQ,5226073533,SMS,ST906(70SACD)_TQ_V_2.0 2024/06/07\\n" + //
                                                 "ID:5226073533\\n" + //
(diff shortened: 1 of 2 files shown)


tananaev:

samm-git:
@tananaev, although there is no exact documentation, the result is consistent and can be found on the forums for all St906 4 G trackers. I added a comment about it.

An alternative approach could be to log any SMS type as a device command and later use compute attributes to catch messages with voltage, making the code less firmware-dependent. Please let me know your thoughts.

tananaev:
Can you please reference all the forum threads you're talking about.

samm-git:
@tananaev  sure

https://www.traccar.org/forums/topic/secrets-revealed-sinotrack-st906l-4g-new-latest-firmware-release-st90670elascdtqv10-20231213-via-cxzt-command/

https://www.traccar.org/forums/topic/sinotrack-st-906l4g-unknown-command-code-for-utxxxx300-upload-timer-300-value-in-cxzt/

https://www.traccar.org/forums/topic/st-901-sinotrack-to-traccar/

tananaev:
It seems like all the samples are from the same person.

samm-git:
Looks like it is. I am not affiliated with him; I just received this tracker from AliExpress and found that it has a voltage reading that looks pretty accurate (tested on a lab power supply first). I decided it would be nice to have it in the Traccar UI and alerts. If you think it would be better to be more generic, I can dump all SMS types reply to the KEY_Result instead, so the user will have to use computed attributes to extract parts they need. I understand your concerns that there is no good documentation, but i think h02 doc is overall crap, some very random resources on the net, half of them chinese, etc.

tananaev:
It would be nice, but I think the pattern is too specific and I'm afraid it will break very often. So I want to know that the format is generic enough.

samm-git:
Ok, now it's done. I modified the patch to send all SMS data received to the "result" string. I was able to get power as a computed attribute with such an expression:

```java
result.contains('VOLT:') && result ? new("java.lang.Double",result.substring(result.indexOf('VOLT:') + 5,result.indexOf('V', result.indexOf('VOLT:') + 5))) : null
```
I also have this in my config:
```xml
    <entry key='processing.computedAttributes.deviceAttributes'>true</entry>
    <entry key='processing.computedAttributes.newInstanceCreation'>true</entry>

    <entry key='processing.copyAttributes.enable'>true</entry>
    <entry key='processing.copyAttributes'>power</entry>
```

And also in the cron, every 20 minutes:

```bash
/usr/local/bin/curl --silent 'https://traccar.example.com/api/commands/send'  \
  -X POST  -H 'Content-Type: application/json' -H "Authorization: Bearer XXXXXX" \
  --data-raw '{"id":1,"attributes":{"data":"CXZT"},"deviceId":1,"type":"custom","textChannel":false,"description":"Get status"}'
```


In this way, it should be safe to include this patch in H02.

1. If the device does not use the "SMS" command at all, the code will never be called.
2. If it is in use, it's likely to reply to one of the status commands (`RCONF`, `CXZT`, maybe some others)
3. It is now easy to parse it and extract any data the user needs, and there is no need to recompile the code if the format changes.
 
Please let me know your thoughts.

src/test/java/org/traccar/protocol/H02ProtocolDecoderTest.java
```diff
@@ -284,8 +287,22 @@ public void testDecodeStatus() throws Exception {
                 Position.KEY_STATUS, 0xFFFFFBFFL);
 
         verifyAttribute(decoder, buffer(
-                "*HQ,4210051415,V1,164549,A,0956.3869,N,08406.7068,W,000.00,000,221215,FFFFFBFF,712,01,0,0,6#"),
```
tananaev: why are we removing tests?

src/test/java/org/traccar/protocol/H02ProtocolDecoderTest.java
```diff
@@ -268,6 +268,9 @@ public void testDecode() throws Exception {
         verifyPosition(decoder, binary(
                 "24971305007205201916101533335008000073206976000000effffbffff000252776566060000000000000000000049"));
 
+        verifyAttribute(decoder, buffer(
+                "*HQ,135790246811220,HTBT,100#"),
```
tananaev: What is this test for?

src/test/java/org/traccar/protocol/H02ProtocolDecoderTest.java
samm-git: My mistake. Fixed now.

src/test/java/org/traccar/protocol/H02ProtocolDecoderTest.java
samm-git: ...Also fixed, mistakenly editied this ID instead of original one

PR #5630, opened 2025-08-10
Add support for S5L GPS protocol

harshanasrimal:
### Summary
This PR adds support for the S5L GPS protocol used by SEEWORLD trackers.
It includes a full implementation of the decoder based on the official protocol documentation.

### Details
- Implemented `S5lProtocol` with `S5lProtocolDecoder` and `S5lProtocolEncoder`.
- Added login, heartbeat, location, and alarm packet parsing.
- Added server command encoder based on protocol spec.
- Tested with live devices; all packet types work correctly.
- Added Fuel level monitoring based on Expanded info and custom attribute called `FuelMax` which must be number(maximum oil level we get when full tank). 

### Reference
[Protocol Documentation PDF](https://drive.google.com/file/d/1p5UYAfuO8FvY6-MjER0CIJmpAX58xbBZ/view?usp=sharing)


tananaev:
How is it different from GT06?

harshanasrimal:
> How is it different from GT06?

S5L appends alarm extension + expansion info (oil, temperature, mileage). The GT06 decoder doesn’t expect these, so its parser gets misaligned and interprets, say, an oil byte as the terminal-status/alarm bits. Result: random alarms.

Also S5L doesn't out battery info but current gt06 always detect battery level as 66% and show warning.
On the other hand, S5L return operating external voltage. that details also not extract by the current protocol.

In this attachment, you will notice the alarms and the battery level. but the alarm is completely wrong. I can guarantee that I use the device model as `SEEWORLD` under Extra fields.

Screenshot 2025-08-10 at 23 11 04

tananaev:
Yeah, but duplicating the rest of the protocol is not a solution for that. You should instead differentiate based on model type.

harshanasrimal:
> Yeah, but duplicating the rest of the protocol is not a solution for that. You should instead differentiate based on model type.

you mean within the same protocol?

tananaev:
Yes, if you haven't noticed, we already support a very large number of protocol variations there.

harshanasrimal:
> Yes, if you haven't noticed, we already support a very large number of protocol variations there.

I referred the current code before implementing a new. I was afraid to touch the current code because you have already handled  `SEEWORD` model with `S5` variation. My code may break your implementation if there is another model which named same. Is it the same device I'm trying? if then I can modify it to work with current.

tananaev:
You should check that.

harshanasrimal:
As I figured out, the issue is in `HEARTBEAT` packet. Anyhow My issue has solved with my implementation and I'm currently using it. Do you need me to modify your gt06 to make it compatible with SEEWORD S5 model?

tananaev:
If you want to merge it into mainline - yes.

harshanasrimal:
I would like to contribute mainline because I use traccar since 2023 and I really appreciate your effort. for sure I will look into this and update this PR once I get a free time.

harshanasrimal:
@tananaev Here I modified current `GT06` protocol. Please refer the changes and let me know if you have any suggestions. 

Also I noticed an issue in original `GT06`. When the current protocol decode Terminal information, it take 4 bits to figure out alarm but official document says only 3 bits. I applied a fix to that one also.

On the other hand, as `SEEWORD S5` supports fuel monitoring, they ask end user to set `FuelMax` from the portal which is differ from vehicle to vehicle. Is that okay to implement that one also within the current `GT06` protocol? (please refer my `S5lProtocolDecoder.java`)

src/main/java/org/traccar/protocol/S5lProtocolDecoder.java
```diff
@@ -0,0 +1,262 @@
...
+import org.traccar.model.CellTower;
+import org.traccar.model.Device;
+import org.traccar.model.Network;
+import org.traccar.model.Position;
+
+import java.net.SocketAddress;
+import java.nio.charset.StandardCharsets;
+import java.util.TimeZone;
+
+public class S5lProtocolDecoder extends BaseProtocolDecoder {
```
tananaev: Why is this still here?

harshanasrimal: I was waited for your confirmation. just deleted them

tananaev: Confirmation of what? Let me know when the PR is fully ready for review.

harshanasrimal: `SEEWORD S5` supports fuel monitoring, they ask end user to set `FuelMax` from the portal which is differ from vehicle to vehicle. I want your approval to implement it within the current GT06 protocol similar way to I implemented it in that new file. is that okay to add FuelMax attribute?

tananaev: You shouldn't be adding it to the protocol decoder.

harshanasrimal: okay. I reverted that commit and now you can review the branch

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -399,9 +401,9 @@ private void decodeStatus(Position position, ByteBuf buf) {
         position.set(Position.KEY_STATUS, status);
         position.set(Position.KEY_IGNITION, BitUtil.check(status, 1));
         position.set(Position.KEY_CHARGE, BitUtil.check(status, 2));
-        position.set(Position.KEY_BLOCKED, BitUtil.check(status, 7));
+        position.set(Position.KEY_BLOCKED, !BitUtil.check(status, 6)); // GPS Signal OK
 
-        switch (BitUtil.between(status, 3, 6)) {
+        switch (BitUtil.between(status, 3, 5)) {
```
tananaev: Why are we changing all this?

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -804,6 +806,25 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
             double temperature = BitUtil.to(value, 15) * 0.1;
             position.set(Position.PREFIX_TEMP + 1, BitUtil.check(value, 15) ? temperature : -temperature);
 
+        } else if (type == MSG_STATUS && variant == Variant.S5) {
```
tananaev: Why do we need another else case? This looks very similar to the standard status message.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -936,18 +969,6 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
                     position.set(Position.KEY_POWER, buf.readUnsignedShort() * 0.01);
                     position.set(Position.KEY_RSSI, buf.readUnsignedByte());
                     buf.readUnsignedByte(); // alarm extension
-                } else if (variant == Variant.S5) {
```
tananaev: Why are you moving this?

harshanasrimal: in S5, both location and Alarm packets are same. just the protocol byte is different.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
harshanasrimal: previous code was wrong. if you notice, it check 4 bits. but in GT06, there is only 3 bits for alarm.

on the other hand, 6th index is the GPS OK boolean(7th bit). But in current code, It get 7th index which is 8th bit.

above two fixes are common for gt06

tananaev: It checks only 3 in the original code. I'm becoming very concerned now with all your changes.

harshanasrimal: You are correct. I request my apologies for the mistake. I was thinking about index. I reverted them to original and added `KEY_GPS` to confirm GPS signal OK.

Screenshot 2025-08-11 at 14 21 45

please review now

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
harshanasrimal: I used that to keep HEARTBEAT packet seperately. I removed that codeblock and added a condition on standard decode

harshanasrimal:
@tananaev Please review now. I fixed all the issues and also implemented condition to handle below cases.

`1. The alarm extension is only valid in the alarm packet, not in the positioning packet and the heartbeat packet.`

`2. Bit5-Bit3 The alarm bit is only valid in the alarm package, not in the positioning package and heartbeat package.`

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -392,41 +394,45 @@ private boolean decodeLbs(Position position, ByteBuf buf, int type, boolean hasL
...
-            case 7:
-                if (variant == Variant.VXT01) {
-                    position.addAlarm(Position.ALARM_OVERSPEED);
-                } else {
-                    position.addAlarm(Position.ALARM_REMOVING);
-                }
-                break;
-            default:
-                break;
+        if( (variant != Variant.S5) || (variant == Variant.S5 && type == MSG_GPS_LBS_STATUS_1)) {
```
tananaev: Why only `variant == Variant.S5 && type == MSG_GPS_LBS_STATUS_1`?

harshanasrimal: `  if( (variant != Variant.S5) || (variant == Variant.S5 && type == MSG_GPS_LBS_STATUS_1)) {`

if variant is not `S5` then the switch execute as it is. But, if the variant is `S5`, then the switch will only check if the packet is Alarm packet.

tananaev: What does it send for non alarm messages?

harshanasrimal: we must ignore this alarm bits if the packet is not an alarm packet. It mentioned in the official documentation. So we only decode other bits except alarm bits for non alarm packets

tananaev: But what does it actually send? Typically it just sends zero, which means no alarms.

harshanasrimal: the document has mentioned regarding 00 separately. And at the end of the chart, they have mentioned that we must ignore alarms if the packet is not an alarm packet. I feel it like the device could send a cache or something. 

What did you get regarding this alarm packets when you refer the protocol documentation?

tananaev: For all other variations we just read alarm as alarm. I've never seen a case where device would send an old value or something like this. Do you have a sample from a real device?

harshanasrimal: I have a sample device. I will check and let you know. Im outside right now. 
Anyhow, When I contact SEEWORD technical team, They told me that those bits are only valid with alarm packages as mentioned in the protocol documentation.

tananaev: Some other documents say the same, but it hasn't been a problem so far.

harshanasrimal: mm. sometimes even the technical team doesn't know about the product they sell. I will try with the real device and let you know soon. like within an hour.

harshanasrimal: Finally I was able to understand that what the document trying to say.

Alarm: `78782b1619080b0d13159f00c72d2f08a2a21c011400019d01a840c7aa0d51000002000004ff0000001e015b5c550d0a`

Location: `78782b1219080b0d131f9f00c72d0708a2a1ef041400019d01a840c7aa0d41000002000004ff0000001e015ded110d0a`

Heartbeat:  `78780a1341000002000164d9200d0a`

I got above packets under a power cut scenario. Alarm packet is perfectly fine. but when we consider Location packet and the heartbeat packet, both says `Alarm Normal` but its not. That's why the document says only the alarm in `Alarm Packet` is valid.

So you are correct bro. we don't need to waste time for another if condition. I reverted the commit. Please review now.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -239,6 +239,8 @@ private boolean hasStatus(int type, String model) {
                 return "NT20".equalsIgnoreCase(model);
             case 0xA3: // MSG_FENCE_SINGLE / MSG_STATUS_3
                 return variant == Variant.SEEWORLD;
+            case MSG_GPS_LBS_1:
```
tananaev: We should probably add another type instead of reusing this.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -399,6 +401,7 @@ private void decodeStatus(Position position, ByteBuf buf) {
         position.set(Position.KEY_STATUS, status);
         position.set(Position.KEY_IGNITION, BitUtil.check(status, 1));
         position.set(Position.KEY_CHARGE, BitUtil.check(status, 2));
+        position.set(Position.KEY_GPS, BitUtil.check(status, 6)); // GPS Signal OK
```
tananaev: What does this mean?

harshanasrimal: It means the terminal has GPS coverage

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
harshanasrimal: as we consider about the variant using below code, do we still need another type?

`return variant == Variant.S5;`

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
tananaev: GPS coverage? What is that?

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
tananaev: Yes. Because it has status.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
harshanasrimal: Screenshot 2025-08-11 at 19 28 09


If the device failed to get position from GPS, it return 0

tananaev: Where do you see that? It says GPS tracking on/off. To me it sounds very different.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
harshanasrimal: it has status only when the variant is `S5` right? Honestly I didn't get you bro.

tananaev: It should be `MSG_GPS_LBS_STATUS_...`.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
harshanasrimal: Check this out

`78782b1219080b0e0821010000000000000000000400019d01a840070c000003e801000004ff00000000000536620d0a`

IMG_7127


the SMS confirms that the device is still searching for GPS coordinates, and the location packet confirms it using that GPS bit

tananaev: How is it different from the GPS validity flag?

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
harshanasrimal: Screenshot 2025-08-11 at 19 46 17

When we consider standard `GT06` Location packet, it doesn't contain `STATUS` but both standard and the `S5` model use `0x12` protocol number for location packets.

So how we update the type below?
`MSG_GPS_LBS_1 = 0x12;`

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
harshanasrimal: seems like both are same.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
tananaev: You can have two types with the same id.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
tananaev: Then we don't need this.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
harshanasrimal: Oh got your point and implemented. I thought its a bad practice to duplicate. Thanks for the suggestion.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
harshanasrimal: Okay. I removed it.

harshanasrimal:
@tananaev I see one failed test case.

`Gt06ProtocolDecoderTest > testDecode() FAILED`
`org.opentest4j.AssertionFailedError at Gt06ProtocolDecoderTest.java:111`

when I check the test file I saw it's regarding Voltage. 

`verifyAttribute(decoder, binary("78780a130604ea04000006bc8a0d0a"),Position.KEY_POWER, null);`

it expect null value but actually the packet has a voltage. (12.58V)

Am I authorized to edit the test file?

tananaev:
You're not. You can't break tests, unless you can provide an undeniable proof that the test was incorrect.

harshanasrimal:
> You're not. You can't break tests, unless you can provide an undeniable proof that the test was incorrect.

@tananaev could you please check `Gt06ProtocolDecoderTest.java:565`

it says below
`decoder.setModelOverride("SEEWORLD");`

`verifyAttribute(decoder, binary("78780a134506560b0102242b850d0a"),Position.KEY_POWER, 11.0);`

but according to SEEWORLD documentation, `KEY_POWER` must be `16.22`

harshanasrimal:
@tananaev also please check `Gt06ProtocolDecoderTest.java:189`

`verifyAttribute(decoder,binary("78782b1215050d03041bcf031ff30a0be795bc001c17014e14a065dd95314504b6040000001c00000cd90ab8fb6f0d0a"),Position.PREFIX_TEMP + 1, 0x1c);`

In current GT06, if protocol number is `0x12` and if the length is `0x2b`, then it considered as `Variant.S5`. But in the test case, that packet pass as standard `GT06` (Model is not defined)

My question is, Do we have another S5 model or Is that the same SEEWORLD model. If it's the same one, then we must set Model before run the test case.

tananaev:
Pretty much all tests were implemented based on real samples and/or documentation.

harshanasrimal:
> Pretty much all tests were implemented based on real samples and/or documentation.

I feel that above two test cases are wrong. could you dig deeper please.

tananaev:
Unfortunately I don't have time to dig deeper at this time.

harshanasrimal:
I can guarantee that these test cases are wrong according to SEEWORD documentation. and also current code confirm that the location pack is an S5 variant

harshanasrimal:
@tananaev Any update regarding this?

tananaev:
No update. Waiting for you to either fix tests or provide some evidence that those were incorrect.

harshanasrimal:
> No update. Waiting for you to either fix tests or provide some evidence that those were incorrect.

@tananaev I recently found another SEEWORLD protocol document and It describes another use of `GT06` protocol by `SEEWORLD`. Accouding to that document, protocol number and the packet length is exactly similar to S5L but the structure is different.

[GT06 Protocol by SEEWORLD](https://drive.google.com/file/d/12H5wfJCCeB-mBsy3Osq9xHhxmzXYVJyt/view?usp=sharing)

it’s clear that we cannot reliably determine the decode type using only the `SEEWORLD` model name or packet length. Different SEEWORLD device models share the same protocol numbers and sometimes even similar packet sizes, but the field layout and meaning can differ.

To avoid incorrect parsing and misinterpretation of data, I recommend we either:

1. Use the exact device model name in `Model` field (`NS5L`, `S5L`, `S5E`)
2. Introduce a combined model as `SEEWORLD_S5`
3. Implement completely different protocol at another port.

could you please suggest the best approach to continue. I will modify the code once you choose the best approach.

tananaev:
How would second point solve the problem?

harshanasrimal:
The current issue is we are unable to detect the variant considering SEEWORLD model and the packet length. But if user define that the device is SEWORLD_S5 then we don't need to determine it using packet behavior. instead, we can directly set variant.

May I implement the second point?

tananaev:
I'm a bit confused here. What is the difference between SEWORLD and S5?

harshanasrimal:
SEEWORLD is the Brand name and S5 is the model. some SEEWORLD devices use standard GT06 and S5L/S5E use different.

tananaev:
Then why are we trying to combine them?

harshanasrimal:
I saw `S5` as the model from some other brands as well.

tananaev:
In this case it obviously refers to a very specific model.

harshanasrimal:
Yes. If we use `SEEWORLD_S5` as the model, then we obviously know the device is from SEEWORLD and the model could be S5L or S5E. Then we can decode without any issues. 

Do you have any better suggestion or do you agree with me?

tananaev:
Why can't we just use `SEEWORLD`?

harshanasrimal:
Refer below protocol documents as example scenario

[GT06 by SEEWORLD](https://drive.google.com/file/d/12H5wfJCCeB-mBsy3Osq9xHhxmzXYVJyt/view)

[S5 by SEEWORLD](https://drive.google.com/file/d/1p5UYAfuO8FvY6-MjER0CIJmpAX58xbBZ/view)

In above protocols, both are for `SEEWORLD` devices but the decoding method is different.

Even we can't identify based on packet length because it also same.

Please refer heartbeat packet in both document as a proof.

tananaev:
If it's a standard GT06, it shouldn't need differentiation. It's only needed for a custom vendor variant that is in conflict with another format, and there's no other way to determine how it needs to be decoded.

harshanasrimal:
I just shared only the document that I found and when I refer your test cases, 

it says below

```
decoder.setModelOverride("SEEWORLD");

verifyAttribute(decoder, binary("78780a134506560b0102242b850d0a"),Position.KEY_POWER, 11.0);
```

This test also contain same packet length as `S5` variant but the thing is it expect different output.

So If this test case is correct, then how we determine `S5`variant?

tananaev:
Then it sounds to me that there are more than 2 variations of SEEWORLD? One standard GT06, your variant and the one we have the test for. What am I missing?

harshanasrimal:
Exactly. That's why we stuck if we use just the brand name `SEEWORLD`.

tananaev:
But have you found all the relevant documents?

harshanasrimal:
No. only have `S5L/S5E` and the standard `GT06` one. There may have many more but without documents we are unable to decide.

tananaev:
I found the documentation for the sample `78780a134506560b0102242b850d0a`:

[06_D11LSPV3.3_20231208.pdf](https://github.com/user-attachments/files/21791866/06_D11LSPV3.3_20231208.pdf)

harshanasrimal:
> I found the documentation for the sample `78780a134506560b0102242b850d0a`:
> 
> [06_D11LSPV3.3_20231208.pdf](https://github.com/user-attachments/files/21791866/06_D11LSPV3.3_20231208.pdf)

Seems like its completely different from the `S5L/S5E` protocol document. not only the heartbeat packet. what do you suggest? If the best thing is to keep modifying the current `GT06`, then we must go with a combined model like `SEEWORLD_S5` or else we have to set and determine by Exact model like `S5L`, `S5E` etc.

Personally I feel that the current `GT06ProtocolDecoder` getting too complicated. It will be harder to maintain the file in future if we don't separate the protocols. On the other hand, the system will also have to go through all the cases to identify the models each time the device send packets. But if we use a seperate protocol, the system already know the way to handle exact model and it can decode without any extra weight. Just an idea. 

Please let me know your thought. as we can see we may expect more variants in future as well.

tananaev:
I'll repeat one last time - a separate decoder is not an option.

harshanasrimal:
> I'll repeat one last time - a separate decoder is not an option.

okay okay bro. just shared my idea. then we have 2 options

1. use `SEEWORLD_S5` as model
2. use exact model name like `S5L`, `S5E`, `NS5L` etc

What do you think?

tananaev:
Ideally we use model series. They usually the same format. For example we can call it `S5`.

harshanasrimal:
Some other brands also have devices in S5 series. 

So is that okay to use `S5` only? 

Or else What if we use `SEEWORLD_S5`?

harshanasrimal:
@tananaev you haven't replied to my previous comment. could you please

tananaev:
I already told you exactly what to call it.

harshanasrimal:
As I understood, you asking me to use `S5` as the model. I will modify the code once I got a free time.

PR #5635, opened 2025-08-18
Add time offset attribute support to fix Issues with MTK 2503D chip (with a built-in MT3333 positioning chip)

riccefarias:
Summary

  This PR adds support for configurable time offset on devices and groups, allowing administrators to correct timestamp discrepancies from GPS devices.

  Problem

  Some GPS devices report incorrect timestamps due to:
  - Incorrect time zone settings
  - Clock drift
  - Firmware issues
  - Different time synchronization methods

  Currently, Traccar only supports global time.override which affects all devices uniformly.

  Solution

  Introduces a new time.offset attribute that can be configured:
  - Globally in the configuration file
  - Per device as a device attribute
  - Per group with automatic inheritance to child devices

  The offset is specified in seconds and can be positive or negative.

  Implementation Details

  - Added TIME_OFFSET configuration key in Keys.java with support for device/group hierarchy
  - Extended TimeHandler to apply the offset after processing time.override
  - Offset is applied to both deviceTime and fixTime
  - Uses existing AttributeUtil.lookup() for hierarchical attribute resolution

  Usage Examples

  Global configuration (traccar.xml):

  <entry key='time.offset'>3600</entry> 

  Device attribute (via API or UI):

  {
    "attributes": {
      "time.offset": -7200  // Subtract 2 hours for this device
    }
  }

  Group attribute:

  Configure on a group and all devices in that group will inherit the offset unless overridden.

  Testing

  - Added unit tests in TimeHandlerTest.java
  - Tests cover positive and negative offsets
  - Verified compilation and all tests pass

  Backward Compatibility

  - Fully backward compatible
  - Default offset is 0 (no change)
  - Works alongside existing time.override configuration

  Files Changed

  - src/main/java/org/traccar/config/Keys.java - Added TIME_OFFSET configuration
  - src/main/java/org/traccar/handler/TimeHandler.java - Implemented offset logic
  - src/main/java/org/traccar/MainModule.java - Updated dependency injection
  - src/test/java/org/traccar/handler/TimeHandlerTest.java - Added tests

### DIFF
--- a/src/main/java/org/traccar/MainModule.java
+++ b/src/main/java/org/traccar/MainModule.java
@@ -326,9 +326,9 @@ public class MainModule extends AbstractModule {
 
     @Singleton
     @Provides
-    public static TimeHandler provideTimeHandler(Config config) {
-        if (config.hasKey(Keys.TIME_OVERRIDE)) {
-            return new TimeHandler(config);
+    public static TimeHandler provideTimeHandler(Config config, CacheManager cacheManager) {
+        if (config.hasKey(Keys.TIME_OVERRIDE) || config.hasKey(Keys.TIME_OFFSET)) {
+            return new TimeHandler(config, cacheManager);
         }
         return null;
     }
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -1570,6 +1570,15 @@ public final class Keys {
             List.of(KeyType.CONFIG));
 
     /**
+     * Time offset in seconds to apply to device time. Can be positive or negative.
+     * Can be configured globally, per device, or per group.
+     */
+    public static final ConfigKey<Integer> TIME_OFFSET = new IntegerConfigKey(
+            "time.offset",
+            List.of(KeyType.CONFIG, KeyType.DEVICE),
+            0);
+
+    /**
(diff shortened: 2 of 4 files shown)


tananaev:
I don't think this is a good solution for a rollover problem. Ideally we should detect and fix it automatically.

riccefarias:
> I don't think this is a good solution for a rollover problem. Ideally we should detect and fix it automatically.

I’m struggling to find a reliable detection method, considering the sheer number of devices, many of which use the same ports/protocols.

riccefarias:
Are you suggesting something like: if (YEAR === 2006) { timestamp += 619315200; }?

tananaev:
Something like that, but more generic.

riccefarias:
> Something like that, but more generic.

I agree we could explore something along those lines. Unfortunately, I’m sticking with my solution—it fits my needs perfectly and gives me more freedom and control over any other potential offset-related issues.

tananaev:
OK, let's close it for now then.

josuelmm:
Look this @tananaev https://github.com/josuelmm/traccar/commit/b8cd85a08757f47d9316ccc43dda723750392037

mohamedemam223:
The problem has been solved after updating the GPS MT3333 module, but you need to connect to the tracking device through uart and then the update is done, now find out how the update works online without the need to connect the device physically.

PR #5636, opened 2025-08-18
fix bug MTK 2503D chip (with a built-in MT3333 positioning chip)

michaelloliveira:
## Description
This Pull Request applies an automatic correction for invalid GPS timestamps (prior to 01/01/2016) in the `DatabaseHandler` handler. Now, whenever a position arrives with a `fixtime` less than 1451606400 (UNIX timestamp), the adjustment +619315200 seconds will be applied to correct the time.

## Changes made
- Added the `correctFixTime(Position position)` method to correct invalid timestamps.

- Apply the correction before saving the position to the database (`storage.addObject(...)`).

## Benefits
- Prevents inconsistencies in reports and graphs due to incorrect timestamps.

- Maintains compatibility with the rest of the system.

## Tests performed
- Test with GPS positions prior to 2016 → timestamp corrected and `fixTimeCorrected` flag added.
- Test with GPS positions after 2016 → timestamp maintained unchanged and without flag.

Statement on Positioning Issues of GPS Devices equipped with MTK 2503D Chips

The specific manifestations are as follows: the
positioning time resets to start from January 1, 2006, and the positioning status
cannot remain stable continuously.
It should be specially noted that this issue only affects 2G locator products equipped with the aforementioned chips, and 4G locators are not affected by this problem.

Method for correcting GPS time anomalies on the server side:
Trigger condition: When the positioning timestamp (UNIX format) reported by the
device is less than 1451606400 (corresponding to 00:00:00 UTC on January 1, 2016),
the correction mechanism will be activated; if the reported positioning timestamp
(UNIX format) is 1451606400 or greater, no processing will be performed.
Correction algorithm: Corrected timestamp = Device-reported timestamp (UNIX
format) + 619315200
Correction example: The data reported by the device is 78 78 26 22 06 01 01 10 33 05
CC 02 88 E6 04 09 B5 29 66 00 15 60 00 00 00 00 00 00 00 00 00 00 01 00 00 00 00 00
04 C6 0F 0D 0A
Parsed GPS timestamp: 2006/01/01 16:51:5 (1136105465 in UNIX format)
Corrected GPS timestamp (UNIX format): 1136105465 + 619315200 = 175542066


tananaev:
Database handler is definitely not the right place for this.

michaelloliveira:
Very good, please suggest the best place to apply this solution

tananaev:
I don't know yet.

michaelloliveira:
OK. Let's think of something. thank you for your attention

michaelloliveira:
I put it in FilterHandler, with a key to activate the filter.

What do you think? What's more appropriate?

tananaev:
Slightly better, but I think it's still not ideal. I'll probably work on a solution myself shortly.

PR #5638, opened 2025-08-19
Add filter.zeroAccuracy filter option.

jakiki6:
One of my friends has the issue that some coordinates are completely wrong but all wrong ones have an accuracy that is 0 while all valid ones have an accuracy that is bigger than 0.

I know that there is the "filter.accuracy" key but that only filters out points with an accuracy higher than a threshold.

This PR adds a new filter with the "filter.zeroAccuracy" key that filters out all points with an accuracy of 0.

This is probably a niche use case but I know at least one user who needs this.

### DIFF
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -1457,6 +1457,13 @@ public final class Keys {
             List.of(KeyType.CONFIG));
 
     /**
+     * Filter positions with a zero accuracy.
+     */
+    public static final ConfigKey<Boolean> FILTER_ZERO_ACCURACY = new BooleanConfigKey(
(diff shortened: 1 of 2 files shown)


tananaev:
This filter doesn't really make sense to me. I would instead just recommend using computed attributes to adjust the accuracy value.

laura240406:
He tried that but he wants to remove all points with a 0 accuracy value by marking them as invalid and enabling the filter that removes invalid points but that doesn't work as the filter is applied before the computed attribute.

tananaev:
We have early computed attribute option if you're not aware. If you set priority to negative, it will be calculated before everything else.

laura240406:
that does seem to work, thanks :)

PR #5645, opened 2025-08-22
Polish translation added.

lukashaxx:
I've added report files translated into Polish.

IMPORTANT: I'm not sure if I need to add translation information anywhere else besides the files.


tananaev:
It looks fine to me. Have you tested that it works?

lukashaxx:
Unfortunately, I haven't checked if it works - I don't have such an opportunity, but looking at the code, it should work.

tananaev:
Please verify before we merge.

mundeczek:
@lukashaxx very good work!
I would change the date format to dd-MM-yyyy.

lukashaxx:
I tested the following events (after changing the date format):
- device online
- device moving
- device stopped
- status unknown
- device entered geofence
- device left geofence
All tested notifications work fine.

I tested using the Traccar client - unfortunately, I can't generate other events (ignition, refueling, etc.), but I did them identically to the ones tested, and I think they will work correctly.

PR #5655, opened 2025-08-31
El language

RafailStratiotis:
El language For Traccar


tananaev:
Have you tested and it works?

RafailStratiotis:
yes Anton

PR #5679, opened 2025-09-30
Add Italian language Files

netluca1:
IT language For Traccar

### DIFF
--- a/templates/notifications/it/alarm.vm
+++ b/templates/notifications/it/alarm.vm
@@ -1,0 +1,15 @@
+#set($subject = "$device.name: Allarme!")
+#set($alarmKey = $event.getString('alarm'))
+#set($alarmName = $translations.get("alarm${alarmKey.substring(0, 1).toUpperCase()}${alarmKey.substring(1)}"))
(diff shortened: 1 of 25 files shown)


tananaev:
Have you tested to make sure it works?

netluca1:
Hi Anton, I was able to test it via email, and it works correctly:
move, stop, turn on, turn off.
As for "alarm," everything works, but it doesn't translate the alarm type. It remains in English, but it is translated in web events.
What can I check?

eosfomogame:
@netluca1 Great job! it’s nice to finally have Italian support. I just have a few suggestions for the next version:
 - Use “è” instead of “e'”.
 - Be more consistent with capitalization. For example, now there is “e' disconnesso” and “si è Connesso”.
 - There seems to be a translation issue with the overspeed limit: instead of “velocità”, it shows “indirizzo.”
 - LLMs can be very helpful for translations; the important thing is to keep the structure unchanged.

Grazie!

tananaev:
Feel free to send a pull request.

Issue #5647, opened 2025-08-25
[Feature Request] User email in audit report


lukashaxx:
It would be useful to have an additional field in the audit report, namely: user email, as having only the user ID would require manual searching in the database to find out who it is.

Thank you.
Lukasha

Vansh1228:
I would like to work on this issue. Can you assign it to me? @lukashaxx

tananaev:
Feel free to send a pull request.

Vansh1228:
[PR-5682](https://github.com/traccar/traccar/pull/5682) is raised for the above feature request.

PR #5682, opened 2025-10-07
[Feature Request] Adding user email to the audit report. 

Vansh1228:
[Fix for #5647 ](https://github.com/traccar/traccar/issues/5647)

### DIFF
--- a/src/main/java/org/traccar/api/BaseResource.java
+++ b/src/main/java/org/traccar/api/BaseResource.java
@@ -42,4 +42,12 @@ public class BaseResource {
         return 0;
     }
 
+    protected String getUserEmail() {
+        UserPrincipal principal = (UserPrincipal) securityContext.getUserPrincipal();
+        if (principal != null) {
+            return principal.getUserEmail();
+        }
+        return "NA";
+    }
+
 }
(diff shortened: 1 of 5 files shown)


src/main/java/org/traccar/api/BaseResource.java
```diff
@@ -42,4 +42,12 @@ protected long getUserId() {
         return 0;
     }
 
+    protected String getUserEmail() {
```
tananaev, line 45: What is this for?

src/main/java/org/traccar/api/resource/AuditResource.java
```diff
@@ -41,10 +42,23 @@ public class AuditResource extends BaseResource {
...
+
+        Stream<Action> actions = storage.getObjectsStream(Action.class, new Request(
                 new Columns.All(),
                 new Condition.Between("actionTime", from, to),
                 new Order("actionTime")));
-    }
 
+        return actions.map(action -> {
+            try {
+                User user = storage.getObject(User.class, new Request(
```
tananaev, line 53: I don't think it makes sense to request it one by one.

src/main/java/org/traccar/model/Action.java
```diff
@@ -48,6 +48,16 @@ public long getUserId() {
         return userId;
     }
 
+    private String userEmail;
+
+    public String getUserEmail() {
```
tananaev, line 53: Functions here need to be marked to be ignored for database.

Vansh1228:
@tananaev I will resolve the review comments and gain more clarity on the codebase

PR #5683, opened 2025-10-08
fix priority to notifications

2ecomp:


tananaev:
What is this?

2ecomp:
I think I messed up. I wasn't supposed to commit there, it was supposed to be in my location. Sorry, but the push wasn't arriving with priority, so it wasn't waking up on the smartphone screen. I had to pass priority in the return to get the push with attention.
But it was an old version, I believe it's been refactored already.
"return new NotificationMessage((String) velocityContext.get("subject"), writer.toString(), priority);"
I apologize.

PR #5685, opened 2025-10-10
Fix stack overflow, flatten various recursive paths

### LINKED ISSUE #5684: Bulk import via OsmAnd causes stack overflow
I created a script to programmatically add points via the HTTP OsmAnd API. The script averages somewhere between 50-100 points/sec.

I run the script for a little while and kill it, while watching the server logs. traccar starts to process the points, gets through about 100, and then crashes:

```
traccar-1   | 2025-10-09 22:06:06  INFO: [T721aacd9] id: 69, time: 2024-10-17 23:51:49, lat: <scrub>, lon <scrub>, course: 0.0
traccar-1   | 2025-10-09 22:06:07  INFO: [T721aacd9] id: 69, time: 2024-10-17 23:51:50, lat: <scrub>, lon <scrub>, course: 0.0
traccar-1   | 2025-10-09 22:06:07  INFO: [T721aacd9] id: 69, time: 2024-10-17 23:51:51,  lat: <scrub>, lon <scrub>, course: 0.0
traccar-1   | 2025-10-09 22:06:07  INFO: [T721aacd9] id: 69, time: 2024-10-17 23:51:52,  lat: <scrub>, lon <scrub>, course: 0.0
traccar-1   | 2025-10-09 22:06:07  WARN: A task raised an exception. Task: org.traccar.database.BufferingManager$$Lambda/0x000000080081a3c8@670fdb58 - StackOverflowError (... < Log:160 < ... < PositionLogger:94 < ProcessingHandler:195 < PostProcessHandler:64 < ...)
```

I set up Traccar today using the recommended Docker compose setup, so the software version is whatever today's docker `:latest` tag is.

qlyoung:
This fixes two recursive paths. One of them fixes #5684 and the other fixes potential recursion issues when there are many handlers.

I have tested both setups under my own deployment and the first patch fixes the issue I described in #5684.

---

The call path for processing new points is recursive, with recursion depth
controlled by the depth of the point queue. This results in a stack overflow
under heavy load. Fix it by eliminating the problematic recursive path.

The previous path was:
```
  processNextPosition()
    -> processPositionHandlers()
       -> finishedProcessing()
          -> processNextPosition()
```
The base case was "no more positions to process"; so with a large queue this
could easily exceed the stack limit.

It is now:
```
  processNextPosition()
    -> schedule(processPositionHandlers())

  processPositionHandlers()
  -> finishedProcessing()
     -> processNextPosition()
```

Points are still processed one at a time in order, but each processing job
happens in a new stack.

### DIFF
--- a/src/main/java/org/traccar/ProcessingHandler.java
+++ b/src/main/java/org/traccar/ProcessingHandler.java
@@ -211,7 +211,7 @@ public class ProcessingHandler extends ChannelInboundHandlerAdapter implements B
             nextPosition = queue.peek();
         }
         if (nextPosition != null) {
-            processPositionHandlers(ctx, nextPosition);
+            ctx.executor().execute(() -> processPositionHandlers(ctx, nextPosition));
         }
     }
 


src/main/java/org/traccar/ProcessingHandler.java
```diff
@@ -160,27 +160,23 @@ public void onReleased(ChannelHandlerContext context, Position position) {
...
-                    continuation.run();
-                } else {
-                    ctx.executor().execute(continuation);
-                }
+                isFiltered = filtered;
             }
-        });
+        };
+
+        while (iterator.hasNext() && !cb.isFiltered) {
```
tananaev: This doesn't make sense to me. It seems like you're just running everything in parallel, which is definitely incorrect.

qlyoung: Can you elaborate at all on why you think this? What is being run in parallel?

If there is a problem here, I'd say it's the opposite; this code blocks on calling each event handler rather than setting up a recursion chain and yielding immediately.

qlyoung:
Dropped the second commit; I wasn't super sure that is a problem path anyway since I don't have a complete grasp on the event loop / async setup in this codebase. Thoughts on the patch as it is now?

tananaev:
It looks ok now. Have you tested and everything works?

qlyoung:
Seems stable in my spot testing, no longer crashes and I see the points ingested. If you have specific test recommendations happy to follow them.

tananaev:
Merged, thanks.

Issue #5687, opened 2025-10-15
Include `com.kohlschutter.junixsocket` to allow UNIX domain socket based connection to Postgres servers


wucke13:
As described in https://jdbc.postgresql.org/documentation/use/#unix-sockets , there is not much else left to do, this should work pretty much out of the box then.

tananaev:
You have to provide some context.

wucke13:
Sure thing!

I'd like to operate Traccar with postgresql. I'm on a monolithic host (like most hobbyists, I guess?) so both Traccar and Postgres will be running on one and the same machine.

For this scenario, I think a UNIX domain socket connection from Traccar to Postgres are preferable:

- They are faster [^1][^2][^3]
- Access control to the DB is implemented without explicit authentication (as in username and password), instead just the ambient authentication already present in the form of the Linux user which runs Traccar, which Postgres uses to enforce access control [^4].
  - No need for passwords in configuration files
  - No rotating of password
  - ... all the secret management stuff re/ database auth goes away
- For local development running Postgres + Traccar as a user (on a shared machine), or for security hardening, it can be beneficial to not have each Postgres instance globally listen on a network port. I guess this is pretty niche, but if I can I disable Postgres listening on TCP completely, which lowers the attack surface.
- It's a small change! As in, it most likely requires no change in the Traccar code whatsoever. If the `com.kohlschutter.junixsocket` package becomes part of the Traccar release `.jar`, then using it is just a matter of adjusting the jdbc url. https://jdbc.postgresql.org/documentation/use/#unix-sockets has the details on the matter.

I hope this is more elaborately making the point why this might be a small yet desirable addition to Traccar.

[^1]: https://www.cybertec-postgresql.com/en/postgresql-performance-advice-unix-sockets-vs-localhost/
[^2]: https://momjian.us/main/blogs/pgblog/2012.html#June_6_2012
[^3]: https://dr-knz.net/local-overheads-in-postgresql-and-cockroachdb.html
[^4]: https://www.postgresql.org/docs/current/auth-peer.html

tananaev:
Seems reasonable. Will we be able to use MySQL via unix socket as well?

wucke13:
Seemigly, yes! https://code.google.com/archive/p/junixsocket/wikis/ConnectingToMySQL.wiki

tananaev:
Cool, then feel free to send a pull request.

wucke13:
@tananaev Done :), lets continue the discussion in #5692.

PR #5692, opened 2025-10-26
feat: add junixsocket

wucke13:
Add support for JUnixSocket, allowing database connections via UNIX domain sockets without all the username and password fuzz. For a detailed description on how to configure this, see the following page:

https://kohlschutter.github.io/junixsocket/dependency.html#jdbc

Fixes: https://github.com/traccar/traccar/issues/5687

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,6 +51,7 @@ ext {
     nettyVersion = "4.2.6.Final"
     protobufVersion = "4.32.1"
     jxlsVersion = "2.14.0" // version 3 has breaking changes
+    junixsocketVersion = "2.10.1"
 }
 
 def resolveTransitiveVersion = { String root, String group, String name ->
@@ -85,6 +86,8 @@ dependencies {
     implementation "com.mysql:mysql-connector-j:9.4.0"
     implementation "org.mariadb.jdbc:mariadb-java-client:3.5.6"
     implementation "org.postgresql:postgresql:42.7.8"
+    implementation "com.kohlschutter.junixsocket:junixsocket-core:$junixsocketVersion"
+    implementation "com.kohlschutter.junixsocket:junixsocket-mysql:$junixsocketVersion"
     implementation "com.microsoft.sqlserver:mssql-jdbc:13.2.0.jre11"
     implementation "com.zaxxer:HikariCP:7.0.2"
     implementation "io.netty:netty-buffer:$nettyVersion"


wucke13:
Worth noting: https://kohlschutter.github.io/junixsocket/dependency.html#gradle describes a bunch of plugins that could be used. I opted to only add the mysql plugin. If upstream prefers, we can go all-in on this, but I think the current selection is the most sensible one.

build.gradle
```diff
@@ -138,6 +139,8 @@ dependencies {
     implementation "com.warrenstrange:googleauth:1.5.0"
     implementation "com.google.openlocationcode:openlocationcode:1.0.4"
     implementation "io.modelcontextprotocol.sdk:mcp:0.13.1"
+    implementation "com.kohlschutter.junixsocket:junixsocket-core:$junixsocketVersion"
```
tananaev: I would add these next to database dependencies.

tananaev:
Is this really all that needs to be done? Have you tested it? What's an example config for testing?

build.gradle
wucke13: Sure thing, I didn't at first understand the sorting principle. Fixed now.

wucke13:
> Is this really all that needs to be done?

I think yes.

> Have you tested it?

Admittedly not yet.

> What's an example config for testing?

I'm working on one.

wucke13:
That felt too complicated, for what it is in the end (took me almost an hour). Well, yes, it works. Here is a minimal repro for my smoke-test:

```bash
gradle build
rm -rf -- postgres-data
initdb -D postgres-data
postgres -D postgres-data/ -k "$PWD" &
sleep 3
CONFIG_USE_ENVIRONMENT_VARIABLES=true DATABASE_USER="$USER" DATABASE_DRIVER=org.postgresql.Driver DATABASE_URL="jdbc:postgresql://localhost/template1?socketFactory=org.newsclub.net.unix.AFUNIXSocketFactory\$SystemProperty" java -Dorg.newsclub.net.unix.socket.default="$PWD/.s.PGSQL.5432" -jar target/tracker-server.jar
```

I run this in a nix-shell to get the relevant dependencies: `nix-shell -p gradle openjdk postgresql`.

tananaev:
Why do we need this part?
```
-Dorg.newsclub.net.unix.socket.default="$PWD/.s.PGSQL.5432"
```

wucke13:
To point the socket factory to the path where the UNIX domain socket file lays. This can also be specified as get parameter in the JDBC URL.

tananaev:
What would the URL look like?

wucke13:
`jdbc:postgresql://localhost/<DB_NAME>?socketFactory=org.newsclub.net.unix.AFUNIXSocketFactory$FactoryArg&socketFactoryArg=<PATH_TO_SOCKETFILE>`

So a real example could be:
`jdbc:postgresql://localhost/traccar?socketFactory=org.newsclub.net.unix.AFUNIXSocketFactory$FactoryArg&socketFactoryArg=/var/run/postgresql/.s.PGSQL.5432`

PR #5695, opened 2025-11-01
Restore position validation checks removed in commit 16b8958

Natan-M64:
This fixes a critical bug where GPS coordinates get stuck after device reconnection. The issue was introduced in commit 16b895849922efeb81433c1b6e374095307c8f69 which removed position.getValid() checks from multiple handlers.

The bug manifested as:
- Device loses GPS signal and goes offline
- Device moves to a new location while offline
- Device reconnects and sends new position with valid GPS fix
- Coordinates remain stuck at last position before disconnect
- Only speed and course update correctly

Root cause in DistanceHandler:
The distance filter was rejecting valid positions when the calculated distance exceeded maxError, even when the position had a valid GPS fix. The previous logic allowed valid positions to bypass distance filtering, which makes sense - we should trust a valid GPS fix over distance-based filtering.

Changes:
- DistanceHandler: Restore || position.getValid() to distance filter
- GeocoderHandler: Only geocode valid positions to avoid wasting API calls
- MotionEventHandler: Only process motion events from valid positions
- OverspeedEventHandler: Only detect overspeed from valid positions

Tested:
- Coordinates now update correctly after reconnection
- Devices no longer get stuck coordinates
- All unit tests pass

### DIFF
--- a/src/main/java/org/traccar/handler/DistanceHandler.java
+++ b/src/main/java/org/traccar/handler/DistanceHandler.java
@@ -57,7 +57,7 @@ public class DistanceHandler extends BasePositionHandler {
             }
             if (filter && last.getLatitude() != 0 && last.getLongitude() != 0) {
                 boolean satisfiesMin = minError == 0 || distance > minError;
-                boolean satisfiesMax = maxError == 0 || distance < maxError;
+                boolean satisfiesMax = maxError == 0 || distance < maxError || position.getValid();
                 if (!satisfiesMin || !satisfiesMax) {
                     position.setValid(last.getValid());
                     position.setLatitude(last.getLatitude());
--- a/src/main/java/org/traccar/handler/GeocoderHandler.java
+++ b/src/main/java/org/traccar/handler/GeocoderHandler.java
@@ -41,7 +41,7 @@ public class GeocoderHandler extends BasePositionHandler {
 
     @Override
     public void onPosition(Position position, Callback callback) {
-        if (!ignorePositions) {
+        if (!ignorePositions && position.getValid()) {
             if (reuseDistance != 0) {
                 Position lastPosition = cacheManager.getPosition(position.getDeviceId());
                 if (lastPosition != null && lastPosition.getAddress() != null
--- a/src/main/java/org/traccar/handler/events/MotionEventHandler.java
+++ b/src/main/java/org/traccar/handler/events/MotionEventHandler.java
@@ -54,6 +54,9 @@ public class MotionEventHandler extends BaseEventHandler {
         if (device == null || !PositionUtil.isLatest(cacheManager, position)) {
             return;
         }
+        if (!position.getValid()) {
+            return;
+        }
 
         TripsConfig tripsConfig = new TripsConfig(new AttributeUtil.CacheProvider(cacheManager, deviceId));
         MotionState state = MotionState.fromDevice(device);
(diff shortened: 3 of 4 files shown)


tananaev:
1. How is maxError related to validity check?
2. The validity check was removed absolutely internally. If you want to exclude those positions, you should use our standard filtering.

Natan-M64:
Hi @tananaev,

Thank you for the feedback! Let me clarify the two points:

1. Relationship between maxError and validity:

The relationship is explicitly documented in Keys.java:1588-1589:
```
/**
 * Distance in meters. Distances above this value gets handled like explained in 
 * 'coordinates.filter', but only if Position is also marked as 'invalid'.
 */
public static final ConfigKey<Integer> COORDINATES_MAX_ERROR
```

This means maxError should only filter invalid positions. Valid GPS fixes should always pass through, regardless of distance. The logic || position.getValid() implements exactly this documented behavior.

2. Why this differs from other handlers:

In commit 16b8958, position.getValid() was removed from multiple handlers, but it had different meanings:

- MotionEventHandler/GeocoderHandler: Checked if we should process invalid positions (configurable via EVENT_MOTION_PROCESS_INVALID_POSITIONS and GEOCODER_PROCESS_INVALID_POSITIONS)
- DistanceHandler: Implements the documented logic that maxError only filters invalid positions

The removal was appropriate for handlers that had configuration flags, but in DistanceHandler, || position.getValid() is the core filtering logic, not an optional feature.

The bug scenario:
1. Device loses GPS → last valid position cached
2. Device moves physically while offline
3. GPS reconnects → new position with valid fix arrives
4. DistanceHandler calculates large distance (device moved)
5. Without || position.getValid(): coordinates get replaced with cached position (frozen)
6. With || position.getValid(): valid GPS fix is trusted (correct behavior)

The "standard filtering" (coordinates.filter) is already active, but it needs the || position.getValid() condition to implement the documented maxError behavior.

tananaev:
1. What you're referring is outdated comment. This was removed.
2. It seems like you misunderstand what `maxError` configuration is intended for. It is a very dangerous option that should almost never be needed. The only valid use case for it is when there's a device malfunction and device jumps due to that.

Once again, the change was absolutely intentional and there's no plan to roll it back.

Natan-M64:
Hi @tananaev,

I understand the change in commit 16b8958 was intentional, but I found an inconsistency between the code and the official documentation at https://www.traccar.org/configuration-file/

The documentation explicitly states:

> **coordinates.maxError**: "Distance in meters. Distances above this value gets handled like explained in 'coordinates.filter', **but only if Position is also marked as 'invalid'**."

The current code (after 16b8958) applies `maxError` filtering regardless of position validity:

```java
// Current code
boolean satisfiesMax = maxError == 0 || distance < maxError;
// This ALWAYS filters when distance >= maxError, even if position.valid == true
```

This contradicts the documented behavior. According to the documentation, the maxError filter should only apply when the position is INVALID.

My configuration:
<entry key='coordinates.maxError'>500</entry>

Observed behavior (bug):
- Device loses GPS → reconnects with valid GPS signal
- Device moved >500m while offline
- New position has valid=true but coordinates are replaced with last position
- Speed/course update correctly, creating data inconsistency

Expected behavior (per documentation):
- Position has valid=true → trust the GPS coordinates regardless of distance
- Only filter if valid=false

Should we:
1. Update the documentation to match the code, or
2. Update the code to match the documentation?

I believe option 2 is correct because trusting valid GPS signals is more reliable than distance-based filtering.

Config use case: maxError protects against GPS glitches (jumps to 0,0 or random coordinates during signal loss), but should not override a valid GPS fix after reconnection.

tananaev:
I already updated the documentation.

PR #5698, opened 2025-11-04
Filter overpass speeding by highway

ChrisCope:
I have seen the overpass api return speed limits for railway tracks that run parallel to a road which means that if the GPS position is not very accurate the first response from overpass can sometimes be the railway and not the road.

This change updates the overpass query so that only features tagged with 'highway' and 'maxspeed' are returned instead of just 'maxspeed'

### DIFF
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -1585,8 +1585,7 @@ public final class Keys {
             List.of(KeyType.CONFIG));
 
     /**
-     * Distance in meters. Distances above this value gets handled like explained in 'coordinates.filter', but only if
-     * Position is also marked as 'invalid'.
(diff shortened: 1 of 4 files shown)


tananaev:
What about regular roads?

ChrisCope:
From our testing it appears all roads have the highway tag independent of the type of road.

Here is a random position in the UK with a railway next to it: [link to position](https://overpass-api.de/api/interpreter?data=[out:json][timeout:25];way[maxspeed](around:100,52.841773,-2.035132);out%20tags;)

You can see that highway is returned for all roads.

A random position in Prague: [link to position](https://overpass-api.de/api/interpreter?data=[out:json][timeout:25];way[maxspeed](around:100,50.074266,14.450911);out%20tags;)

Again you can see that all roads are tagged with highway, in this case the value for the highway tag are 'residential'

tananaev:
Merged, thanks.

PR #5708, opened 2025-11-21
Add The Things Network Webhook Decoder

Steve-Tech:
Hi,
The Things Network (TTN) provides a webhook to publish messages from LoRaWAN devices on their network, this PR will decode those messages for Traccar.

The `decoded_payload` sent by TTN in the webhook must follow the format specified by TTN Mapper:
> Make sure you have a Payload decoder function enabled on the TTN Console that decodes the raw payload into a json object. The JSON object should contain the keys “latitude”, “longitude” and “altitude” and one of “hdop”, “accuracy” or “sats”. When using the Cayenne LPP data format, the GPS coordinates will be decoded into a different JSON format, but this format is also supported.

`speed` and `heading`, while not apart of the TTN Mapper format, are also supported. All other keys will be added as a custom attribute in Traccar; if you are writing your own payload formatter, it is a good idea to use the keys specified in [`Position.java`](https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/model/Position.java), but that is not required.

If using `speed`, a `ttnhttp.speed` entry should be added to your traccar.xml to configure your units, supported values are [listed here](https://github.com/traccar/traccar/blob/95151f242b3957e7e902c42e5a2b750e95897b28/src/main/java/org/traccar/BaseProtocolDecoder.java#L110-L117).

TTN Console Webhook Config:
- Webhook format: `JSON`
- Base URL: `<Your publicly available Traccar url>:5261`
- Enabled event types
  - Uplink message (path can be left empty)

Thanks,
Steve

### DIFF
--- a/src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
@@ -21,7 +21,9 @@ import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpHeaderValues;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import jakarta.json.Json;
+import jakarta.json.JsonNumber;
 import jakarta.json.JsonObject;
+import jakarta.json.JsonString;
 import jakarta.json.JsonValue;
 import org.traccar.BaseHttpProtocolDecoder;
 import org.traccar.Protocol;
@@ -32,7 +34,6 @@ import org.traccar.session.DeviceSession;
 import java.io.StringReader;
 import java.net.SocketAddress;
 import java.nio.charset.StandardCharsets;
-import java.util.Date;
 
 public class TtnHttpProtocolDecoder extends BaseHttpProtocolDecoder {
 
@@ -94,81 +95,33 @@ public class TtnHttpProtocolDecoder extends BaseHttpProtocolDecoder {
                 position.setLongitude(coordinates.getJsonNumber("longitude").doubleValue());
                 position.setAltitude(coordinates.getJsonNumber("altitude").doubleValue());
                 foundGps = true;
-            } else if (jsonKey.startsWith("digital_in_")) {
-                int suffix = Integer.parseInt(jsonKey.substring(("digital_in_").length()));
-                suffix += 1;
-                position.set(Position.PREFIX_IN + suffix, payload.getJsonNumber(jsonKey).doubleValue());
-            } else if (jsonKey.startsWith("digital_out_")) {
-                int suffix = Integer.parseInt(jsonKey.substring(("digital_out_").length()));
-                suffix += 1;
-                position.set(Position.PREFIX_OUT + suffix, payload.getJsonNumber(jsonKey).doubleValue());
-            } else if (jsonKey.startsWith("analog_in_")) {
-                int suffix = Integer.parseInt(jsonKey.substring(("analog_in_").length()));
-                suffix += 1;
-                position.set(Position.PREFIX_ADC + suffix, payload.getJsonNumber(jsonKey).doubleValue());
-            } else if (jsonKey.startsWith("temperature_")) {
-                int suffix = Integer.parseInt(jsonKey.substring(("temperature_").length()));
-                suffix += 1;
-                position.set(Position.PREFIX_TEMP + suffix, payload.getJsonNumber(jsonKey).doubleValue());
-            } else if (jsonKey.startsWith("humidity_")) {
-                position.set(Position.KEY_HUMIDITY, payload.getJsonNumber(jsonKey).doubleValue());
             } else {
                 switch (jsonKey) {
                     case "latitude":
-                    case "lat":
                         position.setLatitude(payload.getJsonNumber(jsonKey).doubleValue());
                         foundGps = true;
                         break;
                     case "longitude":
-                    case "lng":
                         position.setLongitude(payload.getJsonNumber(jsonKey).doubleValue());
                         foundGps = true;
                         break;
                     case "altitude":
-                    case "alt":
                         position.setAltitude(payload.getJsonNumber(jsonKey).doubleValue());
                         break;
-                    case "sat":
+                    case "accuracy":
+                        position.setAccuracy(payload.getJsonNumber(jsonKey).doubleValue());
+                        break;
                     case "sats":
-                    case "satellites":
                         position.set(Position.KEY_SATELLITES, payload.getJsonNumber(jsonKey).intValue());
                         break;
                     case "speed":
                         position.setSpeed(convertSpeed(payload.getJsonNumber(jsonKey).doubleValue(), "kn"));
                         break;
-                    case "bearing":
                     case "heading":
                         position.setCourse(payload.getJsonNumber(jsonKey).doubleValue());
                         break;
-                    case "accuracy":
-                        position.setAccuracy(payload.getJsonNumber(jsonKey).doubleValue());
-                        break;
-                    case "hdop":
-                        position.set(Position.KEY_HDOP, payload.getJsonNumber(jsonKey).doubleValue());
-                        break;
-                    case "time":
-                        switch (payload.get("time").getValueType()) {
-                            case STRING:
-                                position.setFixTime(DateUtil.parseDate(payload.getString("time")));
-                                break;
-                            case NUMBER:
-                                long timestamp = payload.getJsonNumber("time").longValue();
-                                if (timestamp > 10000000000L) {
-                                    position.setFixTime(new Date(timestamp));
-                                } else {
-                                    position.setFixTime(new Date(timestamp * 1000));
-                                }
-                                break;
-                        }
-                        break;
-                    case "humidity":
-                        position.set(Position.KEY_HUMIDITY, payload.getJsonNumber(jsonKey).doubleValue());
-                        break;
-                    case "battery":
-                        position.set(Position.KEY_BATTERY, payload.getJsonNumber(jsonKey).doubleValue());
-                        break;
-                    case "alarm":
-                        position.set(Position.KEY_ALARM, payload.getString(jsonKey));
+                    default:
+                        decodeJsonValue(position, jsonKey, payload.get(jsonKey));
                         break;
                 }
             }
@@ -182,5 +135,24 @@ public class TtnHttpProtocolDecoder extends BaseHttpProtocolDecoder {
 
         sendResponse(channel, HttpResponseStatus.OK);
         return position;
+    }
+
+    private void decodeJsonValue(Position position, String key, JsonValue value) {
+        switch (value.getValueType()) {
+            case STRING:
+                position.set(key, ((JsonString) value).getString());
+                break;
+            case NUMBER:
+                position.set(key, ((JsonNumber) value).doubleValue());
+                break;
+            case TRUE:
+                position.set(key, true);
+                break;
+            case FALSE:
+                position.set(key, false);
+                break;
+            default:
+                break;
+        }
     }
 }


tananaev:
Do you have some documentation for the format?

Steve-Tech:
Yeah, for the whole TTN message this is [their documentation](https://www.thethingsindustries.com/docs/integrations/data-formats/#uplink-messages), and there's few examples that include latitude and longitude in my [tests](https://github.com/traccar/traccar/pull/5708/files#diff-154cf65daf33195b5374f2690384a3a8c994f4f6850b1c2c875e00eb141fa6f2).

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+            // Prevent 0,0 coordinates if lat/lng were not provided
+            getLastLocation(position, null);
+        }
+
+        sendResponse(channel, HttpResponseStatus.OK);
+        return position;
+    }
+
+    @Override
+    protected void sendQueuedCommands(Channel channel, SocketAddress remoteAddress, long deviceId) {
```
tananaev: Why do we need this?

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+    protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg) throws Exception {
+
+        FullHttpRequest request = (FullHttpRequest) msg;
+        String contentType = request.headers().get(HttpHeaderNames.CONTENT_TYPE);
+        if (contentType != null && contentType.startsWith(HttpHeaderValues.APPLICATION_JSON.toString())) {
+            return decodeJson(channel, remoteAddress, request);
+        } else {
+            sendResponse(channel, HttpResponseStatus.BAD_REQUEST);
+            return null;
+        }
```
tananaev, line 52: I would do something like this instead:

```suggestion
        String contentType = request.headers().get(HttpHeaderNames.CONTENT_TYPE);
        if (contentType == null || !contentType.startsWith(HttpHeaderValues.APPLICATION_JSON.toString())) {
            sendResponse(channel, HttpResponseStatus.BAD_REQUEST);
            return null;
        }
        
        // inline decoding here
```

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    private void positionSetFromJson(Position position, String key, JsonValue value) {
```
tananaev:
```suggestion
    private void decodeJsonValue(Position position, String key, JsonValue value) {
```

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: Sorry, I must've left it there when I copied OsmAnd.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+        position.setDeviceId(deviceSession.getDeviceId());
+
+        JsonObject message = root.getJsonObject("uplink_message");
+        if (message == null) {
+            sendResponse(channel, HttpResponseStatus.BAD_REQUEST);
+            return null;
+        }
+
+        JsonObject payload = message.getJsonObject("decoded_payload");
+        // Join messages don't have a payload, but we still need to respond with success to prevent being disabled
```
tananaev: Please remove all unnecessary comments.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+        if (payload == null) {
+            sendResponse(channel, HttpResponseStatus.NO_CONTENT);
```
tananaev: Why not just OK response?

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+            return null;
+        }
+
+        Network network = new Network();
+        network.setRadioType("lora");
```
tananaev: This is not a valid radio type we should be using:

https://developers.google.com/maps/documentation/geolocation/requests-geolocation

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+        network.setCarrier(message.getJsonObject("network_ids").getString("tenant_id"));
```
tananaev: If we don't have cell or wifi info, we should not set network object at all.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+            position.setTime(DateUtil.parseDate(message.getJsonArray("rx_metadata")
+                    .getJsonObject(0).getString("time")));
+        } else {
+            // Fallback to message "received_at" for TTN's Network Server
+            position.setTime(DateUtil.parseDate(message.getString("received_at")));
+        }
+
+        // Loop over all position keys in the Position class
+        for (Field field : Position.class.getDeclaredFields()) {
+            int mods = field.getModifiers();
```
tananaev: use full words for variables

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+            // rx_metadata doesn't seem consistent across gateways, I don't know if time is always present
+            position.setTime(DateUtil.parseDate(message.getJsonArray("rx_metadata")
+                    .getJsonObject(0).getString("time")));
+        } else {
+            // Fallback to message "received_at" for TTN's Network Server
+            position.setTime(DateUtil.parseDate(message.getString("received_at")));
+        }
+
+        // Loop over all position keys in the Position class
+        for (Field field : Position.class.getDeclaredFields()) {
```
tananaev: Actually we should not do this at all

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: It was mostly for debugging as it shows as 204 No Content in the logs but still gets treated as success by TTN, but I can change it.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+            for (String jsonKey : payload.keySet()) {
+                // Check if the jsonKey matches the posKey or has an underscore after (for CayenneLPP's index)
+                if (posKey.equals(jsonKey) || jsonKey.startsWith(posKey + '_')) {
+                    positionSetFromJson(position, posKey, payload.get(jsonKey));
+                }
+            }
+        }
+
+        // Map some CayenneLPP values and other common variations
+        Map<String, String> prefixMappings = new HashMap<>() {{
```
tananaev: Why do we need a map instead of checking directly?

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: Idk, that one seems somewhat necessary, as even though there's no payload and nothing for us to do, if we respond with an error TTN will disable the webhook.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+                if (jsonKey.startsWith(prefix + "_")) {
+                    int suffix = Integer.parseInt(jsonKey.substring((prefix + "_").length()));
+                    // CayenneLPP uses 0-based indexing, "Start with 1 not 0"
+                    suffix += 1;
+                    String posKey = prefixMappings.get(prefix) + suffix;
+                    positionSetFromJson(position, posKey, payload.get(jsonKey));
+                }
+            }
+            // TTN's CayenneLPP decoder uses a 'gps_*' key for coordinates
+            if (jsonKey.startsWith("gps_")) {
```
tananaev, line 89: I don't see this in the documentation that you provided.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+                        position.setLongitude(payload.getJsonNumber(jsonKey).doubleValue());
+                        foundLng = true;
+                        break;
+                    case "altitude":
+                    case "alt":
+                        position.setAltitude(payload.getJsonNumber(jsonKey).doubleValue());
+                        break;
+                    case "sat":
+                    case "sats":
+                    case "satellites":
```
tananaev: Don't see this in the documentation either.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+                                break;
+                        }
+                        break;
+                }
+            }
+        }
+
+        if (!foundSats) {
+            // If no satellites were provided, assume valid if we have lat/lng
+            position.setValid(foundLat && foundLng);
```
tananaev: We should not do it like this. If device doesn't report the validity flag, we should just set it to true when location is available.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        if (!(foundLat && foundLng)) {
```
tananaev: Why do we need two variables for coordinates?

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -0,0 +1,258 @@
...
+            put("temp", Position.PREFIX_TEMP);
+            put("temperature", Position.PREFIX_TEMP);
+        }};
+
+        boolean foundLat = false;
+        boolean foundLng = false;
+        boolean foundSats = false;
+
+        // Common variations and coordinates
+        for (String jsonKey : payload.keySet()) {
```
tananaev, line 88: Documentation shows that JSON had pretty deep nesting, but here you're just looping through top level. I'm not sure how this is supposed to work.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: Yeah that doesn't seem well documented, but that's what TTN's own payload tester generates:

image

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: It's looping though the `decoded_payload`, and the user can select or write their own decoder/formatter for that element, so there's only nesting there if the device's formatter puts nesting there.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
tananaev: The code should be self-documenting. We only add comments in extremely rare cases. I expect that we need close to zero comments for this implementation. Everything seems pretty obvious from the code.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
tananaev: We would need some documentation. Also it seems to me that you could have multiple GPS positions in a single message. How is that handled?

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: In case one of them is missing for some reason.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
tananaev: Where did you get all the possible parameters for the `decoded_payload`? We would need some documentation for that.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
tananaev: What would be the use case for that? Have you seen in reality?

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: There isn't a standard for `decoded_payload` since data is sent to TTN in the device's specific binary format, and then formatted on TTN's end by a user/vendor supplied decoder/formatter written in JS. The output of that formatter is the content of `decoded_payload`. CayenneLPP is the closest there is to a standard (it's also only a binary standard, whatever TTN's decoder spits out is whatever they want), but it also isn't too common, which is why I went with this shotgun sort of approach. For example, here is someone else's [CayennaLPP decoder](https://github.com/ElectronicCats/CayenneLPP/blob/master/decoders/decoder.js), or [Seeed's T1000 decoder](https://github.com/Seeed-Solution/SenseCAP-Decoder/blob/main/T1000/TTN/SenseCAP_T1000_TTN_Decoder.js) which has an array of `messages` and isn't currently compatible.

I could simplify it to only the keys that [TTN Mapper](https://docs.ttnmapper.org/FAQ.html#using-a-gps-tracker-device-transmitting-gps-coordinates-in-the-lora-payload) supports, but I like the idea of being able to tag any sort of sensor with the location.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: Currently this PR will just use the last one present in the JSON, and TTN's default CayenneLPP decoder will sort the highest channel/index to be last.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
tananaev: I say we explicitly decode only what TTN Mapper supports and then everything else will be as custom attributes.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
tananaev: Why not first? Can't we just statically get "gps_0"?

Steve-Tech: Its not guaranteed to be 0, some devices might increase the channel number globally on each sensor (e.g. `temperature_0`, `gps_1`), and some might start at 1.

The channel number is basically arbitrary:
image

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: Wdym by custom attributes? Do I just pass them straight into `position.set` (I do like this!), or do you mean to ignore them?

tananaev: Yes, just pass as is.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
tananaev: OK, then we probably have to keep this case.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: Cool thanks! Also can I keep `"speed"` and `"heading"`, or do you want strict TTN Mapper “latitude”, “longitude”, “altitude”, “hdop”, “accuracy”, “sats”, and everything else custom?

tananaev: Sounds good

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: Not in this PR, but would I be able to add a `LorawanGateway` class, similar to `CellTower` and `WifiAccessPoint`? As TTN does provide a list of gateways that received the message including the rssi and (if published) their lat/lng in the metadata.

tananaev: Let's discuss it separately.

Steve-Tech:
Thanks for your work! That was a very thorough review!

PR #5712, opened 2025-11-25
Add support for GET method in HTTP SMS client

juneeighteen:
- Add SMS_HTTP_METHOD configuration key with default value 'POST'
- Update HttpSmsClient to respect the HTTP method configuration
- When method is GET, append template as query parameters to URL
- Maintains backward compatibility with existing POST-based configs

This change allows Traccar to work with SMS APIs that only support GET requests with query parameters (such as voip.ms REST API).

### DIFF
--- a/src/main/java/org/traccar/sms/HttpSmsClient.java
+++ b/src/main/java/org/traccar/sms/HttpSmsClient.java
@@ -39,10 +39,12 @@ public class HttpSmsClient implements SmsManager {
     private final String authorization;
     private final String template;
     private final MediaType mediaType;
+    private final String method;
 
     public HttpSmsClient(Config config, Client client) {
         this.client = client;
         url = config.getString(Keys.SMS_HTTP_URL);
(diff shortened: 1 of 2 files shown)


juneeighteen:
The voip.ms API only supports sending messages via GET

src/main/java/org/traccar/sms/HttpSmsClient.java
```diff
@@ -86,8 +88,14 @@ private String preparePayload(String phone, String message) {
         }
     }
 
-    private Invocation.Builder getRequestBuilder() {
-        Invocation.Builder builder = client.target(url).request();
+    private Invocation.Builder getRequestBuilder(String phone, String message) {
+        String targetUrl = url;
+        if ("GET".equalsIgnoreCase(method)) {
+            // For GET requests, append template as query parameters
```
tananaev, line 94: I don't this this makes sense. Instead we should support templated URL.

kmk142789 (approved):

PR #5719, opened 2025-12-07
fix: Implemented user email in audit logs with batch fetching (#5647)

### LINKED ISSUE #5647: [Feature Request] User email in audit report

mayank1008-tech:
Fixes #5647

### Changes
* Added `userEmail` field to the `Action` model.
* Annotated the new field with `@QueryIgnore` to prevent database persistence errors.
* Updated `AuditResource` to populate user emails.
* Implemented **batch fetching** for users to avoid the N+1 query problem (fetching all users in a single query instead of inside the loop).

### DIFF
--- a/src/main/java/org/traccar/api/resource/AuditResource.java
+++ b/src/main/java/org/traccar/api/resource/AuditResource.java
@@ -56,7 +56,10 @@ public class AuditResource extends BaseResource {
                         new Columns.All(),
                         new Condition.Between("actionTime", from, to),
                         new Order("actionTime")))
-                .peek(action -> action.setUserEmail(userEmailMap.get(action.getUserId())));
+                .map(action -> {
+                    action.setUserEmail(userEmailMap.get(action.getUserId()));
+                    return action;
+                });
     }
 
 }


src/main/java/org/traccar/api/resource/AuditResource.java
```diff
@@ -41,10 +46,35 @@ public class AuditResource extends BaseResource {
...
                 new Columns.All(),
                 new Condition.Between("actionTime", from, to),
                 new Order("actionTime")));
+
+        Set<Long> userIds = actions.stream()
+                .map(Action::getUserId)
+                .filter(id -> id > 0)
+                .collect(Collectors.toSet());
+
+        var userEmailMap = new HashMap<Long, String>();
```
tananaev: I think we can just get all users instead of getting them one by one.

mayank1008-tech: @tananaev updated to fetch all users in a single query

tananaev: Can we use `Collectors.toMap` to create a map directly using a stream

src/main/java/org/traccar/api/resource/AuditResource.java
```diff
@@ -41,10 +46,28 @@ public class AuditResource extends BaseResource {
...
+        for (User user : users) {
+            if (user.getEmail() != null) {
+                userEmailMap.put(user.getId(), user.getEmail());
+            }
+        }
+
+        for (Action action : actions) {
+            if (userEmailMap.containsKey(action.getUserId())) {
```
tananaev: And same here

tananaev: This way we can return a stream directly

mayank1008-tech: @tananaev Sir I have updated both the loops to streams. It definitely looks cleaner this way

tananaev: Unfortunately it's still not exactly what I wanted. You should use stream based API from storage durectly.

mayank1008-tech: @tananaev Sir, I updated it to use Storage's ObjectStream.

tananaev: Sorry for this back and forth, but you have to do the same thing for users. Also don't forget to close that stream at the end.

mayank1008-tech: @tananaev Sir i switched both Users and actions to use storageObjectStream() with try resource management. And did a style change from import java.util.* to java.util.Map and Date

src/main/java/org/traccar/api/resource/AuditResource.java
```diff
@@ -41,10 +44,19 @@ public class AuditResource extends BaseResource {
...
+        }
+
         return storage.getObjectsStream(Action.class, new Request(
-                new Columns.All(),
-                new Condition.Between("actionTime", from, to),
-                new Order("actionTime")));
+                        new Columns.All(),
+                        new Condition.Between("actionTime", from, to),
+                        new Order("actionTime")))
+                .peek(action -> action.setUserEmail(userEmailMap.get(action.getUserId())));
```
tananaev: Why are you using `peek` here?

mayank1008-tech: @tananaev Thank you for being patient with my silly mistakes, I upgraded the code to use map() instead of peek().

PR #5737, opened 2026-01-04
Update ArknavProtocolDecoder.java

4lazar:
1. Model Compatibility Fix Added PT33 device support
2. Added new attributes
- HW version
- FW version
- Battery level
- Status
3. Added alerts for PT33
- Low battery
- Movement
- Geofence
- Button SOS
- Max speed

### DIFF
--- a/src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
@@ -35,20 +35,20 @@ public class ArknavProtocolDecoder extends BaseProtocolDecoder {
     private static final Pattern PATTERN = new PatternBuilder()
             .number("(d+),")                     // imei
             .expression(".{6},")                 // id code
-            .expression("(.{3}),")               // status
+            .number("(ddd),")                    // status
             .expression("(.{4}),")               // unit number
             .expression("([AV]),")               // validity
             .number("(dd)(dd.d+),")              // latitude
-            .expression("([NS]),")               // latitude hemisphere
+            .expression("([NS]),")
             .number("(ddd)(dd.d+),")             // longitude
-            .expression("([EW]),")               // longitude hemisphere
+            .expression("([EW]),")
             .number("(d+.?d*),")                 // speed
             .number("(d+.?d*),")                 // course
             .number("(d+.?d*),")                 // hdop
             .number("(dd):(dd):(dd) ")           // time (hh:mm:ss)
             .number("(dd)-(dd)-(dd),")           // date (dd-mm-yy)
             .expression("(.{4}),")               // unit version number
-            .expression("(.{2})")                // battery level
+            .number("(xx)")                      // battery level
             .any()
             .compile();
 
--- a/src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
@@ -35,18 +35,20 @@ public class ArknavProtocolDecoder extends BaseProtocolDecoder {
     private static final Pattern PATTERN = new PatternBuilder()
             .number("(d+),")                     // imei
             .expression(".{6},")                 // id code
-            .number("ddd,")                      // status
-            .number("Lddd,")                     // version
+            .expression("(.{3}),")               // status
+            .expression("(.{4}),")               // unit number
             .expression("([AV]),")               // validity
             .number("(dd)(dd.d+),")              // latitude
-            .expression("([NS]),")
+            .expression("([NS]),")               // latitude hemisphere
             .number("(ddd)(dd.d+),")             // longitude
-            .expression("([EW]),")
+            .expression("([EW]),")               // longitude hemisphere
             .number("(d+.?d*),")                 // speed
             .number("(d+.?d*),")                 // course
             .number("(d+.?d*),")                 // hdop
             .number("(dd):(dd):(dd) ")           // time (hh:mm:ss)
             .number("(dd)-(dd)-(dd),")           // date (dd-mm-yy)
+            .expression("(.{4}),")               // unit version number
+            .expression("(.{2})")                // battery level
             .any()
             .compile();
 
@@ -67,16 +69,33 @@ public class ArknavProtocolDecoder extends BaseProtocolDecoder {
         }
         position.setDeviceId(deviceSession.getDeviceId());
 
+        position.set(Position.KEY_STATUS, parser.next());
+        position.set(Position.KEY_VERSION_HW, parser.next());
         position.setValid(parser.next().equals("A"));
         position.setLatitude(parser.nextCoordinate());
         position.setLongitude(parser.nextCoordinate());
         position.setSpeed(parser.nextDouble(0));
         position.setCourse(parser.nextDouble(0));
-
         position.set(Position.KEY_HDOP, parser.nextDouble(0));
-
         position.setTime(parser.nextDateTime(Parser.DateTimeFormat.HMS_DMY));
-
+        position.set(Position.KEY_VERSION_FW, parser.next());
+        if (position.getString(Position.KEY_VERSION_HW).equals("PT33")) {
+            String status = position.getString(Position.KEY_STATUS);
+            switch (status.charAt(0)) {
+                case '1' -> position.addAlarm(Position.ALARM_LOW_BATTERY);
+                case '2' -> position.addAlarm(Position.ALARM_MOVEMENT);
+                case '4' -> position.addAlarm(Position.ALARM_GEOFENCE_EXIT);
+            }
+            switch (status.charAt(1)) {
+                case '1' -> position.addAlarm(Position.ALARM_SOS);
+            }
+            switch (status.charAt(2)) {
+                case '1' -> position.addAlarm(Position.ALARM_OVERSPEED);
+            }
+            position.set(Position.KEY_BATTERY_LEVEL, parser.nextInt());
+        } else {
+            parser.next();
+        }
         return position;
     }
 
--- a/src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
@@ -36,7 +36,7 @@ public class ArknavProtocolDecoder extends BaseProtocolDecoder {
             .number("(d+),")                     // imei
             .expression(".{6},")                 // id code
             .number("(ddd),")                    // status
-            .expression("(.{4}),")               // unit number
+            .expression("(.{4}),")               // device model
             .expression("([AV]),")               // validity
             .number("(dd)(dd.d+),")              // latitude
             .expression("([NS]),")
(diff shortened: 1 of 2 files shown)


tananaev:

4lazar:
> Please provide the protocol documentation.

[personal.tracker.manual.V2.pdf](https://github.com/user-attachments/files/24424599/personal.tracker.manual.V2.pdf)

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -35,18 +35,20 @@ public ArknavProtocolDecoder(Protocol protocol) {
     private static final Pattern PATTERN = new PatternBuilder()
             .number("(d+),")                     // imei
             .expression(".{6},")                 // id code
-            .number("ddd,")                      // status
-            .number("Lddd,")                     // version
+            .expression("(.{3}),")               // status
```
tananaev: Why is this not a number?

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -35,18 +35,20 @@ public ArknavProtocolDecoder(Protocol protocol) {
...
+            .expression("(.{4}),")               // unit number
```
tananaev: Why do you call this unit number, but then map it hardware version?

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -35,18 +35,20 @@ public ArknavProtocolDecoder(Protocol protocol) {
...
             .expression("([AV]),")               // validity
             .number("(dd)(dd.d+),")              // latitude
-            .expression("([NS]),")
+            .expression("([NS]),")               // latitude hemisphere
```
tananaev: I think these comments are redundant.

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -35,18 +35,20 @@ public ArknavProtocolDecoder(Protocol protocol) {
...
             .number("(ddd)(dd.d+),")             // longitude
-            .expression("([EW]),")
+            .expression("([EW]),")               // longitude hemisphere
             .number("(d+.?d*),")                 // speed
             .number("(d+.?d*),")                 // course
             .number("(d+.?d*),")                 // hdop
             .number("(dd):(dd):(dd) ")           // time (hh:mm:ss)
             .number("(dd)-(dd)-(dd),")           // date (dd-mm-yy)
+            .expression("(.{4}),")               // unit version number
```
tananaev: What does unit version number mean?

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -35,18 +35,20 @@ public ArknavProtocolDecoder(Protocol protocol) {
...
+            .expression("(.{2})")                // battery level
```
tananaev: Are you sure these last two don't conflict with other version of the protocol?

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -67,16 +69,33 @@ protected Object decode(
...
         position.setSpeed(parser.nextDouble(0));
         position.setCourse(parser.nextDouble(0));
-
         position.set(Position.KEY_HDOP, parser.nextDouble(0));
-
         position.setTime(parser.nextDateTime(Parser.DateTimeFormat.HMS_DMY));
-
+        position.set(Position.KEY_VERSION_FW, parser.next());
+        if (position.getString(Position.KEY_VERSION_HW).equals("PT33")) {
+            String status = position.getString(Position.KEY_STATUS);
```
tananaev, line 84: It seems like the codes don't match between different devices.

[R-9-setup2-22-7-2010.pdf](https://github.com/user-attachments/files/24425804/R-9-setup2-22-7-2010.pdf)

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: According to documentation, "status" is ASCII string pag.39

tananaev: Yes, but listed possible values are ONLY digits.

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: This value is named as Unit number in documentation, but in the same time it represents the name of hardware ( the same unit can be named R-33 or PT33)

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: No problem, can be deleted

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: It represents the software version of the device

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
tananaev: Hardware model name is not the same thing as hardware version.

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: In documentation provided by you, battery level is represented by HEX value, but for personal tracker it is represented by number level.
Below in code this value is processed only for personal tracker.

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: It will be processed only for PT33 and leave  as is for other units

tananaev: It sounds like you want to use model instead of hardware version. And you probably don't need to save it at all.

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
tananaev: sounds like it should be `number("(xx)")`

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: I would belive so much to these values, despite in fact that this is a chinese device and values can differ. This variant will ensure compatibility and correct decoding in case of string presences.

tananaev: Typically we make regex patterns as specific as possible. In this case I don't see any good reason not to.

4lazar: I will change it to 
.number("(ddd),")               // status

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: What attribute is suitable to display it the menu "More details" of device?
image

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: ok, i will change it accordingly

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
4lazar: It will be displayed in device details (to understand which version it is)

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
tananaev: We don't save device model in the position attributes.

tananaev: But if you really want, we can use `model` to save it.

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
tananaev: let's update the comment to firmware version

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
tananaev: btw another thing we don't usually read variables from position. instead just add java variables.

src/test/java/org/traccar/protocol/ArknavProtocolDecoderTest.java
```diff
@@ -19,7 +20,30 @@ public void testDecode() throws Exception {
...
+                "358266016278447,05*827,100,PT33,V,4821.6584,N,01053.8650,E,000.0,000.0,00.0,20:26:24 17-09-24,2.35,16"),
+                Position.KEY_ALARM, Position.ALARM_LOW_BATTERY);
+
+        verifyAttribute(decoder, text(
+                "358266016278447,05*827,010,PT33,V,4821.6584,N,01053.8650,E,000.0,000.0,00.0,20:26:24 17-09-24,2.35,16"),
+                Position.KEY_ALARM, Position.ALARM_SOS);
 
+        verifyAttribute(decoder, text(
+                "358266016278447,05*827,001,PT33,V,4821.6584,N,01053.8650,E,000.0,000.0,00.0,20:26:24 17-09-24,2.35,16"),
+                Position.KEY_ALARM, Position.ALARM_OVERSPEED);
```
tananaev: I don't think we need a test case for every alarm. Just one alarm test is enough.

4lazar: Test cases to ensure that each char position is processed correctly.

4lazar: Test cases to ensure that each char position is processed correctly.

tananaev: Understand that, but I don't think it's necessary.

4lazar: Ok, low battery alert will remain.

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -67,16 +69,34 @@ protected Object decode(
         }
         position.setDeviceId(deviceSession.getDeviceId());
 
+        position.set(Position.KEY_STATUS, parser.next());
+        String model = parser.next();
+        position.set("model", model);
```
tananaev, line 74: We typically don't store model.

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
tananaev: Looks like the last comment is not resolved.

src/main/java/org/traccar/protocol/ArknavProtocolDecoder.java
```diff
@@ -67,16 +69,34 @@ protected Object decode(
...
+                case '2' -> position.addAlarm(Position.ALARM_MOVEMENT);
+                case '4' -> position.addAlarm(Position.ALARM_GEOFENCE_EXIT);
+            }
+            switch (status.charAt(1)) {
+                case '1' -> position.addAlarm(Position.ALARM_SOS);
+            }
+            switch (status.charAt(2)) {
+                case '1' -> position.addAlarm(Position.ALARM_OVERSPEED);
+            }
+            position.set(Position.KEY_BATTERY_LEVEL, parser.nextInt());
```
tananaev, line 96: Why is this `nextInt` and not `nextHexInt`?

PR #5742, opened 2026-01-13
Fix case-sensitive email login issue

### LINKED ISSUE #5723: PostgreSQL is case sensitive - login input for email address's are problematic
If a user email address is **Example**@ domain, then he can not login with **example**@ domain. since the username field is case sensitive.

[PostgreSQL is a case-sensitive database by default ](https://www.npgsql.org/efcore/misc/collations-and-case-sensitivity.html)

I'm looking at applying [CITEXT data type](https://aws.amazon.com/blogs/database/manage-case-insensitive-data-in-postgresql/), but I'm not sure if that would introduce any issues with traccar?

michaeldflowers:
## Summary
  - Normalize email addresses to lowercase when storing (`User.setEmail()`)
  - Normalize email addresses to lowercase when querying (`LoginService.login()`)
  - Ensures case-insensitive login regardless of database type

  Fixes #5723

  ## Test plan
  - [x] Register a user with mixed-case email (e.g., `Test@Example.COM`)
  - [x] Verify login works with different case variations (e.g., `test@example.com`)

### DIFF
--- a/src/main/java/org/traccar/api/security/LoginService.java
+++ b/src/main/java/org/traccar/api/security/LoginService.java
@@ -92,7 +92,7 @@ public class LoginService {
             return null;
         }
 
-        email = email.trim();
+        email = email.trim().toLowerCase();
         User user = storage.getObject(User.class, new Request(
                 new Columns.All(),
                 new Condition.Or(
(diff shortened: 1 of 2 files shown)


tananaev:
This is an incorrect fix. Instead of mutating the inputs, we should make database search case insensitive, as discussed in the ticket you linked.

PR #5743, opened 2026-01-14
Fix case-sensitive email login issue

### LINKED ISSUE #5723: PostgreSQL is case sensitive - login input for email address's are problematic

michaeldflowers:
## Summary

- Add `Condition.EqualsIgnoreCase` for case-insensitive database queries
- Generate `LOWER(column) = LOWER(?)` SQL for email lookups
- No mutation of input data - comparison handled at database level

Fixes #5723

## Test plan

- Register user with mixed-case email (e.g., `Test@Example.COM`)
- Login with different case variation (e.g., `test@example.com`)
- Existing users unaffected - no data migration needed

### DIFF
--- a/debug.xml
+++ b/debug.xml
@@ -2,7 +2,7 @@
 <!DOCTYPE properties SYSTEM 'http://java.sun.com/dtd/properties.dtd'>
 <properties>
 
-    <entry key='web.path'>./traccar-web/simple</entry>
+    <entry key='web.path'>./traccar-web/build</entry>
     <entry key='web.localizationPath'>./traccar-web/src/resources/l10n</entry>
     <entry key='web.debug'>true</entry>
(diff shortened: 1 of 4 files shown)


tananaev:
This is not what we discussed either. We want to use `citext`.

michaeldflowers:
No problem. I saw that was a suggestion. I wasn't aware that it was a requirement. Let me take another look.

PR #5772, opened 2026-01-29
Fix H02 and Watch protocol bounds checking vulnerabilities

renaudallard:
An extensive description is in each commit. The health parameters of the watch protocol are parsed, so they could also be exploited even though the devices page doesn't mention the support of devices of that type.


tananaev:
Let's split it in separate PRs and add test cases.

renaudallard:
I am running it without issue, but I will add proper tests and split it.

renaudallard:
I will close this one to split it.

PR #5773, opened 2026-01-29
Fix H02 protocol bounds checking vulnerabilities

renaudallard:
Add bounds validation to prevent denial of service attacks through malformed protocol messages:

- decodeV3: Add bounds check for cell tower parsing loop to handle cases where count field exceeds actual array data
- decodeVp1: Add length validation before accessing cell segment indices to handle empty or malformed segments

These vulnerabilities allow denial of service but not remote code execution due to Java's memory safety.

Added test cases for malformed inputs.


src/test/java/org/traccar/protocol/H02ProtocolDecoderTest.java
```diff
@@ -276,6 +276,18 @@ public void testDecode() throws Exception {
...
+        verifyAttributes(decoder, buffer(
+                "*HQ,353111080001055,V3,044855,28403,10,001450,011473,158,-62,0292,0,X,030817,FFFFFBFF#"));
+
+        // Test VP1 with malformed cell data containing empty segments (bounds checking)
+        verifyNotNull(decoder, buffer(
+                "*hq,356327081001239,VP1,V,470,002,Y92,3565,0Y#"));
+
+        // Test VP1 with cell segment having fewer than 3 values (bounds checking)
+        verifyNotNull(decoder, buffer(
+                "*hq,356327081001239,VP1,V,470,002,92,3565Y92Y#"));
```
tananaev, line 289: Is this real data from devices?

renaudallard: These are crafted test cases, not captured from real devices. I verified the vulnerabilities by running the tests against the codebase without the fixes and they all cause ArrayIndexOutOfBoundsException. As with Watch protocol, this is mainly against DOS attacks.

tananaev: Let's discuss this in the other PR first.

renaudallard:
As discussed in watch protocol

PR #5774, opened 2026-01-29
Fix Watch protocol bounds checking vulnerabilities

renaudallard:
Add bounds validation to prevent denial of service attacks through malformed protocol messages:

- Cell tower parsing: Add check to ensure values array has enough elements before accessing indices in parsing loop
- WiFi parsing: Add similar bounds check for WiFi data loop
- Health data handlers (TEMP, btemp2, oxygen, BPHRT, BLOOD): Add length validation before accessing array elements
- JXTK audio handler: Add validation for indexOf result and array length before accessing header values

These vulnerabilities allow denial of service but not remote code execution due to Java's memory safety.

Added test cases for malformed inputs.

### DIFF
--- a/src/test/java/org/traccar/protocol/WatchProtocolDecoderTest.java
+++ b/src/test/java/org/traccar/protocol/WatchProtocolDecoderTest.java
@@ -171,6 +171,38 @@ public class WatchProtocolDecoderTest extends ProtocolTest {
         verifyPosition(decoder, buffer(
                 "[3G*8800000015*00DD*UD,010120,025946,V,0.0,N,0.0,E,22.0,0,-1,0,100,98,0,0,00000000,0,5,eduroam,f4:db:e6:d2:a8:00,-53,eduroam,f4:db:e6:da:d0:80,-79,eduroam,78:0c:f0:24:f9:80,-82,Lions,b0:be:76:0a:05:9a,-82,tubs-guest,f4:db:e6:d2:a8:01,-53,0.0]"));
 
+        // Test position with cellCount > actual cell data (bounds checking)
+        verifyPosition(decoder, buffer(
+                "[3G*9031853319*0040*UD2,220322,055105,A,22.761162,N,114.360192,E,0,0,47,14,100,64,0,0,00000008,10,1,460,0,1234]"));
+
+        // Test position with wifiCount > actual WiFi data (bounds checking)
+        verifyPosition(decoder, buffer(
+                "[3G*9031853319*0050*UD2,220322,055105,A,22.761162,N,114.360192,E,0,0,47,14,100,64,0,0,00000008,0,0,10,wifi1,aa:bb:cc:dd:ee:ff,-50]"));
+
+        // Test TEMP with empty data (bounds checking)
+        verifyNull(decoder, buffer(
+                "[ZJ*5678901234*0001*0004*TEMP]"));
+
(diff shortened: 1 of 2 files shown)


src/test/java/org/traccar/protocol/WatchProtocolDecoderTest.java
```diff
@@ -171,6 +171,38 @@ public void testDecode() throws Exception {
...
+                "[ZJ*357653059860416*0007*0009*BLOOD,109]"));
+
+        // Test JXTK with truncated header - missing commas (bounds checking)
+        verifyNull(decoder, buffer(
+                "[3G*1234567890*0010*JXTK,0,test]"));
+
+        // Test JXTK with insufficient values in header (bounds checking)
+        verifyNull(decoder, buffer(
+                "[3G*1234567890*0015*JXTK,0,name,1]"));
+
```
tananaev, line 205: Are these all failing currently and are they all real messages from devices?

renaudallard: Without the fix, these test cases would cause ArrayIndexOutOfBoundsException crashes.
Those are crafted test cases specifically designed to demonstrate the bounds checking vulnerabilities                                                                   
They represent malformed protocol messages that could be sent by a malicious actor attempting a denial-of-service attack or a buggy device or firmware sending corrupted data.

tananaev: Wait wait wait. What is the denial of service problem exactly?

renaudallard: When the decoder throws an uncaught ArrayIndexOutOfBoundsException, it propagates to MainEventHandler.exceptionCaught(), which calls closeChannel() and terminates the TCP connection.                                                                                                                                                           
                                                                                                                                                                                
An attacker could exploit this by sending malformed packets to disconnect devices. While the server itself doesn't crash (Netty handles exceptions per-channel), it causes:
  - Device disconnection requiring reconnection 
  - Gaps in tracking data 
  - Log spam from repeated exceptions

It's a low DoS, not a full server crash, but enough to disrupt tracking for targeted devices. The fix makes the decoder resilient to malformed input rather than crashing on bad data.
This is defensive hardening. Even if no attacker exploits it, a buggy device sending corrupted data shouldn't be able to crash its own connection. And don't forget that many/most trackers are cheap chinese gear with unverifiable firmware.

tananaev: This is not a vulnerability that needs fixing. When there's a malformed message, we throw exception and close connection. That is expected behavior.

If a real device sends something like this, then we could potentially look into it, but that's not the same thing as DoS.

renaudallard:
I understand your point of view, but isn't graceful handling more robust than crashing? Besides bounds checks are cheap and don't change behaviour for valid messages. So if a device sends something like this we are already prepared.
But if you prefer to close those PRs, I can understand.

tananaev:
Nothing is crashing. Exceptions are the standard way in Java to handle runtime errors.

Again, I think if there are some real devices sending data like this, we should fix it. If it's just contrived examples, then it's not something we should handle explicitly.

renaudallard:
OK, I understand your point of view, I am not that much a java guy.

PR #5777, opened 2026-02-01
Add dynamic routing key resolution for AMQP forwarders

AlejandroRM-DEV:
### Summary
This PR introduces a robust and centralized template resolution system for forwarding positions and events. 

### Key Changes
- **`Interpolator` Utility**: A new, thread-safe centralized class that encapsulates the Velocity engine. It ensures consistent template handling across all forwarders and provides a clear API for position and event data interpolation.
- **Velocity Engine Integration**: Configured with strict reference handling (`RUNTIME_REFERENCES_STRICT`), ensuring that invalid templates or missing variables trigger clear errors rather than silent failures.
- **Universal Forwarder Support**:
    - **AMQP**: Dynamic `routingKey` for RabbitMQ `topic` exchanges.
    - **MQTT & Kafka**: Dynamic topic names, allowing for protocol-based or device-based partitioning.
    - **JSON & URL**: Dynamic endpoint resolution.
    - **Redis**: Customizable keys via `forward.topic`, moving away from the hardcoded `positions.{uniqueId}` format.

### Supported Variables
The `Interpolator` exposes the following variables to the Velocity context:
- `$uniqueId`: The `uniqueId` from the associated `Device`.
- `$protocol`: The protocol string from the `Position` (also available in `Event` if a position is attached).
- `$eventType`: The type of the `Event` (e.g., `deviceOnline`, `alarm`). *Available in event forwarders only.*

### Template Syntax
We use standard **Apache Velocity** syntax:
- **Simple Reference**: `$uniqueId` — used for most basic cases.
- **Formal Reference**: `${uniqueId}` — **Highly recommended** when the variable is adjacent to other text to prevent parsing ambiguity (e.g., `${uniqueId}_v1`).

#### 1. Multi-Protocol Kafka Partitioning
Instead of one "positions" topic, you can split data by protocol:
```xml
<entry key='forward.topic'>traccar.positions.$protocol</entry>
```
*   **Result**: Positions from `osmand` go to `traccar.positions.osmand`.

#### 2. RESTful JSON Routing
Dynamically route positions to device-specific API endpoints:
```xml
<entry key='forward.url'>https://api.myapp.com/v1/devices/${uniqueId}/data</entry>
```
*   **Output**: For device `ABC-123`, the target becomes `https://api.myapp.com/v1/devices/ABC-123/data`.

#### 3. Advanced Redis Key Structure
Customize Redis namespace:
```xml
<entry key='forward.topic'>app:sensor:${protocol}:${uniqueId}:latest</entry>
```
*   **Output**: `app:sensor:gl200:99000123:latest`

#### 4. Event Filtering in AMQP
Route events by type for specialized consumers:
```xml
<entry key='event.forward.topic'>traccar.events.${eventType}</entry>
```
*   **Output**: An SOS alarm becomes `traccar.events.alarm`. A consumer can bind a queue to `traccar.events.alarm` to only receive critical alerts.

### Backward Compatibility & Safety
1.  **Zero-Impact for Existing Users**: If a configuration string (like `positions`) does not contain Velocity variables, the engine returns the string as-is.
2.  **Legacy URL Support**: In `PositionForwarderUrl`, the traditional `{uniqueId}` and `{latitude}` style placeholders are still processed *after* the Velocity interpolation, ensuring existing configurations continue to work perfectly.
3.  **Strict Error Handling**: If a template refers to a variable that doesn't exist (e.g., a typo like `$unique_id`), the system throws an `IllegalArgumentException` with a descriptive message, making configuration debugging easier.
4.  **Null Safety**: In `EventData`, if no position is associated with the event, `$protocol` resolves to an empty string to prevent template crashes.

### DIFF
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -921,14 +921,6 @@ public final class Keys {
             "positions");
 
     /**
-     * Position forwarding Redis key.
-     */
-    public static final ConfigKey<String> FORWARD_REDIS_KEY = new StringConfigKey(
-            "forward.topic",
-            List.of(KeyType.CONFIG),
-            "positions.{uniqueId}");
-
-    /**
      * URL to forward positions. Data is passed through URL parameters. For example, {uniqueId} for device identifier,
      * {latitude} and {longitude} for coordinates.
      */
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -926,7 +926,7 @@ public final class Keys {
     public static final ConfigKey<String> FORWARD_REDIS_KEY = new StringConfigKey(
             "forward.topic",
             List.of(KeyType.CONFIG),
-            "position.{uniqueId}");
+            "positions.{uniqueId}");
 
     /**
      * URL to forward positions. Data is passed through URL parameters. For example, {uniqueId} for device identifier,
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -921,6 +921,14 @@ public final class Keys {
             "positions");
 
     /**
+     * Position forwarding Redis key.
+     */
+    public static final ConfigKey<String> FORWARD_REDIS_KEY = new StringConfigKey(
+            "forward.topic",
+            List.of(KeyType.CONFIG),
+            "position.{uniqueId}");
+
+    /**
      * URL to forward positions. Data is passed through URL parameters. For example, {uniqueId} for device identifier,
      * {latitude} and {longitude} for coordinates.
      */
--- a/src/main/java/org/traccar/forward/AmqpClient.java
+++ b/src/main/java/org/traccar/forward/AmqpClient.java
@@ -15,6 +15,7 @@
  */
 package org.traccar.forward;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.rabbitmq.client.BuiltinExchangeType;
 import com.rabbitmq.client.Channel;
 import com.rabbitmq.client.Connection;
@@ -26,8 +27,12 @@ import java.net.URISyntaxException;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.concurrent.TimeoutException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class AmqpClient {
+    private final Pattern PLACEHOLDER_PATTERN = Pattern.compile("\\{\\{([^}]+)}}");
+
     private final Channel channel;
     private final String exchange;
     private final String topic;
@@ -52,7 +57,41 @@ public class AmqpClient {
         }
     }
 
-    public void publishMessage(String message) throws IOException {
-        channel.basicPublish(exchange, topic, MessageProperties.PERSISTENT_TEXT_PLAIN, message.getBytes());
+    public void publishMessage(String message, String routingKey) throws IOException {
+        channel.basicPublish(exchange, routingKey, MessageProperties.PERSISTENT_TEXT_PLAIN, message.getBytes());
+    }
+
+    public String resolveRoutingKey(JsonNode root) {
+        Matcher matcher = PLACEHOLDER_PATTERN.matcher(topic);
+        StringBuilder result = new StringBuilder();
+
+        while (matcher.find()) {
+            String path = matcher.group(1).trim();
+            String value = resolvePath(root, path);
+
+            matcher.appendReplacement(
+                    result,
+                    Matcher.quoteReplacement(value)
+            );
+        }
+
+        matcher.appendTail(result);
+        return result.toString();
+    }
+
+    private String resolvePath(JsonNode root, String path) {
+        String[] parts = path.split("\\.");
+        JsonNode current = root;
+
+        for (String part : parts) {
+            if (current == null) {
+                return "null";
+            }
+            current = current.get(part);
+        }
+
+        return (current != null && !current.isNull())
+                ? current.asText()
+                : "null";
     }
 }
--- a/src/main/java/org/traccar/forward/AmqpClient.java
+++ b/src/main/java/org/traccar/forward/AmqpClient.java
@@ -15,7 +15,6 @@
  */
 package org.traccar.forward;
 
-import com.fasterxml.jackson.databind.JsonNode;
 import com.rabbitmq.client.BuiltinExchangeType;
 import com.rabbitmq.client.Channel;
 import com.rabbitmq.client.Connection;
@@ -27,19 +26,13 @@ import java.net.URISyntaxException;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.concurrent.TimeoutException;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 public class AmqpClient {
-    private final Pattern PLACEHOLDER_PATTERN = Pattern.compile("\\{\\{([^}]+)}}");
-
     private final Channel channel;
     private final String exchange;
-    private final String topic;
 
-    AmqpClient(String connectionUrl, String exchange, String topic) {
+    AmqpClient(String connectionUrl, String exchange) {
         this.exchange = exchange;
-        this.topic = topic;
 
         ConnectionFactory factory = new ConnectionFactory();
         try {
@@ -59,39 +52,5 @@ public class AmqpClient {
 
     public void publishMessage(String message, String routingKey) throws IOException {
         channel.basicPublish(exchange, routingKey, MessageProperties.PERSISTENT_TEXT_PLAIN, message.getBytes());
-    }
-
-    public String resolveRoutingKey(JsonNode root) {
-        Matcher matcher = PLACEHOLDER_PATTERN.matcher(topic);
-        StringBuilder result = new StringBuilder();
-
-        while (matcher.find()) {
-            String path = matcher.group(1).trim();
-            String value = resolvePath(root, path);
-
-            matcher.appendReplacement(
-                    result,
-                    Matcher.quoteReplacement(value)
-            );
-        }
-
-        matcher.appendTail(result);
-        return result.toString();
-    }
-
-    private String resolvePath(JsonNode root, String path) {
-        String[] parts = path.split("\\.");
-        JsonNode current = root;
-
-        for (String part : parts) {
-            if (current == null) {
-                return "null";
-            }
-            current = current.get(part);
-        }
-
-        return (current != null && !current.isNull())
-                ? current.asText()
-                : "null";
     }
 }
--- a/src/main/java/org/traccar/forward/AmqpClient.java
+++ b/src/main/java/org/traccar/forward/AmqpClient.java
@@ -50,7 +50,7 @@ public class AmqpClient {
         }
     }
 
-    public void publishMessage(String message, String routingKey) throws IOException {
-        channel.basicPublish(exchange, routingKey, MessageProperties.PERSISTENT_TEXT_PLAIN, message.getBytes());
+    public void publishMessage(String message, String topic) throws IOException {
+        channel.basicPublish(exchange, topic, MessageProperties.PERSISTENT_TEXT_PLAIN, message.getBytes());
     }
 }
--- a/src/main/java/org/traccar/forward/EventForwarderAmqp.java
+++ b/src/main/java/org/traccar/forward/EventForwarderAmqp.java
@@ -16,6 +16,7 @@
 package org.traccar.forward;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 
 import org.traccar.config.Config;
 import org.traccar.config.Keys;
@@ -32,17 +33,28 @@ public class EventForwarderAmqp implements EventForwarder {
         String exchange = config.getString(Keys.EVENT_FORWARD_EXCHANGE);
         String topic = config.getString(Keys.EVENT_FORWARD_TOPIC);
         this.objectMapper = objectMapper;
-        amqpClient = new AmqpClient(connectionUrl, exchange, topic);
+        this.amqpClient = new AmqpClient(connectionUrl, exchange, topic);
     }
 
     @Override
     public void forward(EventData eventData, ResultHandler resultHandler) {
         try {
             String value = objectMapper.writeValueAsString(eventData);
-            amqpClient.publishMessage(value);
+            String routingKey = buildRoutingKey(eventData);
+            amqpClient.publishMessage(value, routingKey);
             resultHandler.onResult(true, null);
         } catch (IOException e) {
             resultHandler.onResult(false, e);
         }
     }
+
+    private String buildRoutingKey(EventData eventData) {
+        ObjectNode root = objectMapper.createObjectNode();
+
+        root.set("event", objectMapper.valueToTree(eventData.getEvent()));
+        root.set("device", objectMapper.valueToTree(eventData.getDevice()));
+
+        return amqpClient.resolveRoutingKey(root);
+    }
+
 }
--- a/src/main/java/org/traccar/forward/EventForwarderAmqp.java
+++ b/src/main/java/org/traccar/forward/EventForwarderAmqp.java
@@ -16,7 +16,6 @@
 package org.traccar.forward;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ObjectNode;
 
 import org.traccar.config.Config;
 import org.traccar.config.Keys;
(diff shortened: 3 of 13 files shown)


stylder (approved):
Well done on adding dynamic routing key resolution — this adds useful flexibility to the AMQP forwarding logic

tananaev:
1. I think if we do support placeholders, we need to support it for all forwarders.
2. Supporting all attributes seems like an overkill. I think we can select some reasonable set of things to support.
3. It would be nice to use some existing placeholder handling instead of building our own.

AlejandroRM-DEV:
Okay, I'll provide an update within the next few days.

tananaev:
Great. I would also recommend discussing before implementing.

AlejandroRM-DEV:
Hey, me again 👋 Just wanted to get your thoughts on this:

1. Implement this only for forwarders that use `forward.topic` and `event.forward.topic` (AMQP, MQTT, Kafka).
2. For placeholder interpolation, I’d use `StringSubstitutor` from Apache Commons Text:
   [https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringSubstitutor.html](https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringSubstitutor.html)
3. Limit supported properties to values that are static or don’t change often, for example:

   * Device: `id`, `groupId`, `uniqueId`, `status`, `model`?
   * Position: `protocol`, `valid`?
   * Event: `type`

tananaev:
1. I think we should apply it to both topic and URL.
2. Do we already have it as an explicit dependency?
3. Are you really planning to use all of those options? For example I think device status and model don't make much sense. One is a temporary status unrelated to the data you're sending and model is optional and not sent in most cases.

AlejandroRM-DEV:
1. So, something like this?

   * AMQP / MQTT / Kafka → topic
   * HTTP / JSON / Redis → URL
     Wialon doesn’t seem support dynamic changes.

2. No, we don’t have it as an explicit dependency. I see Apache Velocity already; it could work.

3. I’m good with keeping it to the following options:

   * Device: `uniqueId`
   * Position: `protocol`
   * Event: `type`

tananaev:
1. This makes sense to me.
2. I don't think we should add a new dependency just for this. If there's nothing existing, we should just implement some common helper. We should probably reuse it for URL forwarder as well.
3. This makes sense to start with.

AlejandroRM-DEV:
Excellent, I'll work on it.

AlejandroRM-DEV:
Hi, I updated the PR description and added a new commit.

src/main/java/org/traccar/forward/Interpolator.java
```diff
@@ -0,0 +1,61 @@
...
+import org.traccar.model.Device;
+import org.traccar.model.Position;
+
+import java.io.StringWriter;
+import java.util.Objects;
+import java.util.Optional;
+
+public final class Interpolator {
+    private static final String LOG_TAG = Interpolator.class.getSimpleName();
+    private static final VelocityEngine ENGINE = createEngine();
```
tananaev: Seems like a huge overkill using velocity here.

AlejandroRM-DEV: You're right. Since there are few options, I made a new version dependency-free using the replace method. Working only for notation ${uniqueId}

src/main/java/org/traccar/forward/Interpolator.java
```diff
@@ -0,0 +1,38 @@
...
+
+        return resolve(template, uniqueId, protocol, eventType);
+    }
+
+    public static String resolve(String template, EventData data) {
+        Objects.requireNonNull(data, "Event data cannot be null");
+
+        String uniqueId = data.getDevice() != null ? data.getDevice().getUniqueId() : "";
+        String protocol = data.getPosition() != null ? data.getPosition().getProtocol() : "";
+        String eventType = data.getEvent() != null ? data.getEvent().getType() : "";
```
tananaev: Why do we need all these fallbacks?

src/main/java/org/traccar/forward/Interpolator.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Objects;
+
+public final class Interpolator {
+
+    private Interpolator() {
+    }
+
+    private static String resolve(String template, String uniqueId, String protocol, String eventType) {
+        Objects.requireNonNull(template, "Template string cannot be null");
```
tananaev: Remove all unnecessary checks.

src/main/java/org/traccar/forward/PositionForwarderRedis.java
```diff
@@ -24,19 +24,30 @@
...
     private final ObjectMapper objectMapper;
 
     public PositionForwarderRedis(Config config, ObjectMapper objectMapper) {
         this.objectMapper = objectMapper;
         this.url = config.getString(Keys.FORWARD_URL);
+        if (config.hasKey(Keys.FORWARD_TOPIC)) {
+            this.topic = config.getString(Keys.FORWARD_TOPIC);
+        } else {
+            this.topic = null;
+        }
```
tananaev:
```suggestion
        topic = config.getString(Keys.FORWARD_TOPIC);
```

src/main/java/org/traccar/forward/PositionForwarderRedis.java
```diff
@@ -24,19 +24,30 @@
...
     @Override
     public void forward(PositionData positionData, ResultHandler resultHandler) {
 
         try {
-            String key = "positions." + positionData.getDevice().getUniqueId();
+            String key;
+            if (topic != null) {
+                key = Interpolator.resolve(topic, positionData);
+            } else {
+                key = "positions." + positionData.getDevice().getUniqueId();
```
tananaev: Can't we have this as default topic?

src/main/java/org/traccar/forward/Interpolator.java
```diff
@@ -0,0 +1,38 @@
...
+
+        return template
+                .replace("${uniqueId}", uniqueId)
```
tananaev: Seems like the format is inconsistent with url forwarder.

src/main/java/org/traccar/forward/EventForwarderAmqp.java
```diff
@@ -26,20 +26,22 @@ public class EventForwarderAmqp implements EventForwarder {
 
     private final AmqpClient amqpClient;
     private final ObjectMapper objectMapper;
+    private final String topic;
 
     public EventForwarderAmqp(Config config, ObjectMapper objectMapper) {
         String connectionUrl = config.getString(Keys.EVENT_FORWARD_URL);
         String exchange = config.getString(Keys.EVENT_FORWARD_EXCHANGE);
-        String topic = config.getString(Keys.EVENT_FORWARD_TOPIC);
+        this.topic = config.getString(Keys.EVENT_FORWARD_TOPIC);
```
tananaev: Remove unnecessary `this` here and other places.

src/main/java/org/traccar/forward/Interpolator.java
AlejandroRM-DEV: Done

src/main/java/org/traccar/forward/Interpolator.java
AlejandroRM-DEV: Done with format {uniqueId}

src/main/java/org/traccar/forward/Interpolator.java
AlejandroRM-DEV: Could they not be null sometimes?

src/main/java/org/traccar/forward/PositionForwarderRedis.java
AlejandroRM-DEV: You're right, we could use `position.{uniqueId}`, but there is another problem I didn't see previously. The topic isn't null; I found that FORWARD_TOPIC has `positions` as the default value. 

Maybe declaring a local ConfigKey with default `position.{uniqueId}`

```java
ConfigKey<String> FORWARD_TOPIC = new StringConfigKey(
            "forward.topic",
            List.of(KeyType.CONFIG),
            "position.{uniqueId}");
```

src/main/java/org/traccar/forward/Interpolator.java
tananaev: You should check, but I don't think all of them can be null.

src/main/java/org/traccar/forward/PositionForwarderRedis.java
tananaev: Sounds reasonable.

src/main/java/org/traccar/forward/Interpolator.java
AlejandroRM-DEV: `EventData`
*   Mandatory: `event`, `device`.
*   Optional: `position`. It is `null` for non-GPS events:
    *   Status changes (`ConnectionManager.java`).
    *   Inactivity alerts (`TaskDeviceInactivityCheck.java`).
    *   Command confirmations (`CommandsManager.java`).

`PositionData`
*   Mandatory: `position`, `device`.

src/main/java/org/traccar/forward/PositionForwarderRedis.java
AlejandroRM-DEV: In the Redis context, `key` sounds more appropriate than `topic` so I change it, but it still reads from `forward.topic` defaulting to `positions.{uniqueId}` 

2 tests:

Image

Edit: The image says "position," but I correct it to "positions."

src/main/java/org/traccar/config/Keys.java
```diff
@@ -920,6 +920,14 @@ private Keys() {
             List.of(KeyType.CONFIG),
             "positions");
 
+    /**
+     * Position forwarding Redis key.
+     */
+    public static final ConfigKey<String> FORWARD_REDIS_KEY = new StringConfigKey(
```
tananaev: This doesn't make sense to me.

AlejandroRM-DEV: Why not? 

We can't reuse the FORWARD_TOPIC variable.

tananaev: Why is that?

AlejandroRM-DEV: It will be a breaking change because FORWARD_TOPIC defaults to `positions`, and this forwarder has been appending uniqueId.

tananaev: Now that it's configurable, you can set it to whatever you want. Am I missing something?

AlejandroRM-DEV: Users already forwarding with Redis expect positions.uniqueId. However, using FORWARD_TOPIC overrides this, setting the topic default to positions. They must define explicitly FORWARD_TOPIC.

tananaev: I understand that.

AlejandroRM-DEV: Great, just confirming we're on the same page and that this approach is OK.

tananaev: To be clear, I still want to remove the `FORWARD_REDIS_KEY`.

AlejandroRM-DEV: Done

src/main/java/org/traccar/forward/AmqpClient.java
```diff
@@ -52,7 +50,7 @@ public class AmqpClient {
         }
     }
 
-    public void publishMessage(String message) throws IOException {
-        channel.basicPublish(exchange, topic, MessageProperties.PERSISTENT_TEXT_PLAIN, message.getBytes());
+    public void publishMessage(String message, String routingKey) throws IOException {
```
tananaev: Why is this called routingKey instead of topic?

src/main/java/org/traccar/forward/EventForwarderKafka.java
```diff
@@ -48,7 +48,8 @@ public void forward(EventData eventData, ResultHandler resultHandler) {
         try {
             String key = Long.toString(eventData.getDevice().getId());
             String value = objectMapper.writeValueAsString(eventData);
-            producer.send(new ProducerRecord<>(topic, key, value));
+            String resolvedTopic = Interpolator.resolve(topic, eventData);
```
tananaev: inline this

src/main/java/org/traccar/forward/EventForwarderMqtt.java
```diff
@@ -38,7 +38,8 @@ public EventForwarderMqtt(Config config, ObjectMapper objectMapper) {
     public void forward(EventData eventData, ResultHandler resultHandler) {
         try {
             String payload = objectMapper.writeValueAsString(eventData);
-            mqttClient.publish(topic, payload, (message, e) -> resultHandler.onResult(e == null, e));
+            String resolvedTopic = Interpolator.resolve(topic, eventData);
```
tananaev: same here

src/main/java/org/traccar/forward/Interpolator.java
```diff
@@ -0,0 +1,30 @@
+package org.traccar.forward;
+
+public final class Interpolator {
```
tananaev, line 7: Maybe TopicInterpolator?

src/main/java/org/traccar/forward/PositionForwarderAmqp.java
```diff
@@ -26,20 +26,22 @@ public class PositionForwarderAmqp implements PositionForwarder {
...
+        amqpClient = new AmqpClient(connectionUrl, exchange);
         this.objectMapper = objectMapper;
     }
 
     @Override
     public void forward(PositionData positionData, ResultHandler resultHandler) {
         try {
             String value = objectMapper.writeValueAsString(positionData);
-            amqpClient.publishMessage(value);
+            String routingKey = Interpolator.resolve(topic, positionData);
```
tananaev: inline

src/main/java/org/traccar/forward/PositionForwarderJson.java
```diff
@@ -52,7 +52,8 @@ public void forward(PositionData positionData, ResultHandler resultHandler) {
             return;
         }
 
-        var requestBuilder = client.target(url).request();
+        String resolvedUrl = Interpolator.resolve(url, positionData);
```
tananaev: inline

src/main/java/org/traccar/forward/PositionForwarderKafka.java
```diff
@@ -49,7 +49,8 @@ public void forward(PositionData positionData, ResultHandler resultHandler) {
         try {
             String key = Long.toString(positionData.getDevice().getId());
             String value = objectMapper.writeValueAsString(positionData);
-            producer.send(new ProducerRecord<>(topic, key, value));
+            String resolvedTopic = Interpolator.resolve(topic, positionData);
```
tananaev: here as well

src/main/java/org/traccar/forward/PositionForwarderRedis.java
```diff
@@ -24,22 +24,24 @@
...
 
     private final String url;
+    private final String key;
 
     private final ObjectMapper objectMapper;
 
     public PositionForwarderRedis(Config config, ObjectMapper objectMapper) {
         this.objectMapper = objectMapper;
         this.url = config.getString(Keys.FORWARD_URL);
+        key = config.getString(Keys.FORWARD_TOPIC);
```
tananaev: topic

src/main/java/org/traccar/forward/AmqpClient.java
AlejandroRM-DEV: Depending on the system, "topic" can mean multiple things:

AMQP, or RabbitMQ:

* Topic = an exchange type (routing mechanism).
 * The topic exchange uses the routing key, which is the message label, to determine where to send it.

Kafka:

  * Topic: the real stream or destination of the message.
  * (No key notion for routing.)

distinct messaging systems have distinct meanings for the same word, but it's okay, we can make it `topic` for all forwarders.

tananaev: If you're not ok with topic, we can change it to key or something. I just want consistency. Currently in your code you have topic first and then it suddenly transforms into routing key.

src/main/java/org/traccar/forward/Interpolator.java
AlejandroRM-DEV: It is used to interpolate URLs too (URL & JSON).

tananaev: OK, let's keep it.

tananaev:
Can you please test everything locally before asking for another review.

AlejandroRM-DEV:
Captura de pantalla 2026-02-05 a la(s) 11 18 34 p m

src/main/java/org/traccar/forward/Interpolator.java
```diff
@@ -0,0 +1,38 @@
...
+    private static String resolve(String template, String uniqueId, String protocol, String eventType) {
+        return template
+                .replace("{uniqueId}", uniqueId)
+                .replace("{protocol}", protocol)
+                .replace("{eventType}", eventType);
+    }
+
+    public static String resolve(String template, PositionData data) {
+        String uniqueId = data.getDevice().getUniqueId();
+        String protocol = Optional.ofNullable(data.getPosition())
```
tananaev, line 21: Why have you changed these? I think it was perfectly fine as it was before.

AlejandroRM-DEV: URL test failed because Position has null protocol. 

Validation of this would require another ternary or an if. I think this approach is cleaner 🙂

PR #5788, opened 2026-02-13
Add external voltage extraction for h02 protocol (replaces #5703)

benjamingray123:
This PR replaces #5703 which was accidentally closed due to a force-push error.

It addresses all maintainer feedback from the previous PR:
- Uses relative `readableBytes()` checks instead of absolute indexes
- Removes duplicate unit tests
- Keeps frame decoder logic configurable via config (no code changes to `H02FrameDecoder`)

Original description below:

---

Following my post in the forum: https://www.traccar.org/forums/topic/automotive-relay-style-cj7320-cj730-lk720-not-displaying-vehicle-battery-voltage/

My code adds support for extracting external/vehicle battery voltage from H02 protocol binary messages with extended format (47 bytes).

The frame decoder now correctly detects and reads extended format messages when configured with `protocol.h02.messageLength=47`.

**Changes**
- **H02ProtocolDecoder.java**: Extract voltage from bytes -3 and -2 (before record number byte) using relative buffer operations
- **H02ProtocolDecoderTest.java**: Add unit test for voltage extraction

**Configuration**
Users with extended format devices should set:
```xml
<entry key='protocol.h02.messageLength'>47</entry>
```

### DIFF
--- a/src/main/java/org/traccar/protocol/H02ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/H02ProtocolDecoder.java
@@ -155,6 +155,11 @@ public class H02ProtocolDecoder extends BaseProtocolDecoder {
 
         processStatus(position, buf.readUnsignedInt());
 
+        if (buf.readableBytes() >= 3) {
+            buf.skipBytes(buf.readableBytes() - 3);
+            position.set(Position.KEY_POWER, buf.readUnsignedShort() / 10.0);
+        }
+
         if (getConfig().getBoolean(Keys.PROTOCOL_ACK.withPrefix(getProtocolName()))) {
             sendResponse(channel, remoteAddress, id, "R12");
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/H02ProtocolDecoder.java
```diff
@@ -155,6 +155,11 @@ private Position decodeBinary(ByteBuf buf, Channel channel, SocketAddress remote
 
         processStatus(position, buf.readUnsignedInt());
 
+        if (buf.readableBytes() >= 3) {
+            buf.skipBytes(buf.readableBytes() - 3);
+            position.set(Position.KEY_POWER, buf.readUnsignedShort() / 10.0);
+        }
```
tananaev, line 161: This definitely conflicts with other protocols.

PR #5789, opened 2026-02-14
Enhance summary report with additional fields and improvements

juniormelo26:
This PR enhances the summary report functionality by adding new fields and utility functions to improve data analysis and reporting capabilities.

Changes Made:
📊 Enhanced Report Models:

Added new fields to SummaryReportItem for extended data capture
Enhanced TripReportItem with additional trip metrics
Improved data structure for more comprehensive reporting
🔧 Utility Functions:

Added new utility functions in ReportUtils for common report operations
Simplified complex calculations and data processing
Improved code reusability across report providers
⚙️ Configuration Updates:

Extended Keys.java with new configuration options
Added support for customizable report parameters
Enhanced flexibility for report generation
📈 Report Provider Improvements:

Enhanced SummaryReportProvider with new calculation methods
Improved performance and accuracy of summary reports
Added support for additional metrics and aggregations
Benefits:
More detailed and informative summary reports
Better data analysis capabilities
Improved performance through optimized calculations
Enhanced configurability for different use cases
Testing:
All existing tests continue to pass
New functionality tested with sample data
No breaking changes introduced


src/main/java/org/traccar/reports/SummaryReportProvider.java
```diff
@@ -64,6 +64,84 @@ public SummaryReportProvider(
...
 
+
+    private long calculateIdleTime(long deviceId, Date from, Date to) throws StorageException {
+
+        final long minIdleDuration = config.getLong(Keys.REPORT_IDLE_MIN_DURATION);
+        final long maxGapDuration = config.getLong(Keys.REPORT_IDLE_MAX_GAP);
+        final double idleRpmThreshold = config.getDouble(Keys.REPORT_IDLE_RPM_THRESHOLD);
+
+        long idleTime = 0;
+        var positions = PositionUtil.getPositions(storage, deviceId, from, to);
```
tananaev, line 90: It definitely doesn't make sense to extract all positions just to calculate idle time. It we add idle state, it should be something similar to motion state. Calculated live and reconstructed based on events.

juniormelo26:
Got it, so this way we need to have an IdleState, IdleProcessor and an Idle event. Or you suggest another approach. I agree that the search for events would be more performative.

tananaev:
I think ideally it's all generate as a part of motion processor.

juniormelo26:
ok, I can implement it and bring it here for us to discuss.

tananaev:
I think it's better to discuss the exact plan first before implementing it.

juniormelo26:
Hello Anton, I don't know if I understand exactly how you would like it, I made the changes that way. Sorry, I hadn't seen your previous message and I already implemented it.

tananaev:
We should discuss it first before making any further changes. So far it looks like very low quality AI slop.

PR #5793, opened 2026-02-17
Fix missing openapi.yaml endpoints and params

njoguamos:
This PR improves `openapi.yaml` API documentation for several endpoints.

### Positions endpoint improvements:

- Current `openapi.yaml` state that `deviceId is optional, but requires the from and to parameters when use`, this is not correct as per the latest source code.
https://github.com/traccar/traccar/blob/91a33caf4b28391045607c7b1c1249c5d109fc77/src/main/java/org/traccar/api/resource/PositionResource.java#L81-L92 Clarified that providing `deviceId` without `from` and `to` parameters returns the latest position.
- Added the missing  `geofenceId` query parameters to positions endpoint
https://github.com/traccar/traccar/blob/91a33caf4b28391045607c7b1c1249c5d109fc77/src/main/java/org/traccar/api/resource/PositionResource.java#L66-L68

###  Users endpoint enhancements

- Changed the `/users` `userId` query parameter type from `string` to `integer`, 
- Added  missing `deviceId` and `excludeAttributes` to filter params.
https://github.com/traccar/traccar/blob/91a33caf4b28391045607c7b1c1249c5d109fc77/src/main/java/org/traccar/api/resource/UserResource.java#L72-L74
- Added an empty `security` array to the "Create a User" endpoint  (I saw the @PermitAll).

### Events endpoint update

* Added a missing `404 Not Found` response to the event retrieval endpoint.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -434,7 +434,6 @@ paths:
           content: {}
     post:
       summary: Create a Device
-      description: Create a new device and link it to the current user
       tags:
         - Devices
       requestBody:
@@ -454,7 +453,6 @@ paths:
   /devices/{id}:
     put:
       summary: Update a Device
-      description: Update the specified device
       tags:
         - Devices
       parameters:
@@ -479,7 +477,6 @@ paths:
       x-codegen-request-body-name: body
     delete:
       summary: Delete a Device
-      description: Delete the specified device
       tags:
         - Devices
       parameters:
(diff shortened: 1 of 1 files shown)


njoguamos:
With this PR: 

The following API endpoints documentation will be update to date with the source code. 

-  [x] Users
-  [x] Positions
-  [x] Events
-  [x] Users
-  [x] Server

I will also look at the rest of the endpoints.

tananaev:
Don't know about this. Most of the descriptions seem to be either duplicative or just unnecessary.

njoguamos:
I see the point. I have removed the recently pushed descriptions.

PR #5795, opened 2026-02-19
Refactor ArnaviBinaryProtocolDecoder constants and methods

NL1ghtman:
As previously recommended, I made it inline.
Fixed sendResponse, tags 5 and 9.
Fixed message handling — Arnavi can send multiple packets at once.
Added CAN and LLS tags.
Tested on real data.
Protocol description attached.
https://docs.google.com/spreadsheets/d/15s-2ZbqOQ1bZvAtFFm9sIEuKy3jbJzxdeynp72sjoYU/edit?pref=2&pli=1&gid=12#gid=12


src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,88 +34,407 @@
...
         position.setDeviceId(deviceSession.getDeviceId());
-
         position.setTime(time);
 
         int readBytes = 0;
         while (readBytes < length) {
-            short tag = buf.readUnsignedByte();
+
+            if (buf.readableBytes() < 5) {
+                break; // incomplete tag — stop safely
```
tananaev: why do we need this?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,88 +34,407 @@
...
+                // Bytes 2-3: value LE uint16
+                case 0x06: {
+                    int mode = buf.readUnsignedByte();
+
+                    switch (mode) {
+                        case 0x01: {
+                            int virtualSensors = buf.readUnsignedByte();  // byte[1]
+                            int physicalInputs = buf.readUnsignedShortLE(); // bytes[2-3]
+                            position.set("virtualignition", BitUtil.check(virtualSensors, 0));
+                            //position.set(Position.KEY_IGNITION, virtualIgnition); //optional | опционально, лучше использовать вычисляемый атрибут
```
tananaev: English comments?

tananaev:
Can we please also cut almost all comments in the code, unless absolutely necessary.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: This is a check for a corrupted packet. When the connection was interrupted, an incomplete packet was received and the parser crashed with an error.

tananaev: Please provide logs.

NL1ghtman:
I encountered an issue with coordinate freezing; the decoder breaks when enabling the parameter `<entry key='coordinates.filter'>true</entry>` in the Traccar config file. The error is intermittent. I assume it is due to coordinate filtering in the decoder and the logic built into the parameter itself. I will make corrections after testing.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: During testing, I found this check redundant.



[tracker-server.log](https://github.com/user-attachments/files/25522743/tracker-server.log)

tananaev: So, it's not needed?

NL1ghtman: Yes, it is not needed. This function is performed by the code in FrameDecoder
```java
if (buf.readableBytes() >= length) {
                return buf.readRetainedSlice(length);
```
The loop jumps from PACKET to PACKET until it finds 0x5D, while reading the length field of each PACKET (getUnsignedShortLE(index + 1)) — this is how the framer knows where each nested packet ends. Only after 0x5D is found, the buffer is sliced entirely and passed to ArnaviBinaryProtocolDecoder.

NL1ghtman:
Removed unnecessary comments, updated tag numbering, and fixed the handling of LLS and discrete inputs. The coordinate processing error occurred after changing the tracker settings. The distance between some coordinates did not satisfy the conditions. When using the filter.zero parameter, coordinates are processed correctly.

tananaev:
I still see a lot of unnecessary comments. Our policy is that the code should be self-documenting. Comments are reserved for extremely rare cases where something is not obvious from the code.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,88 +34,326 @@
...
+                    }
                     break;
+                }
 
-                case 4:
-                    position.setLongitude(buf.readFloatLE());
-                    position.setValid(true);
+                case 4: {
+                    float longitude = buf.readFloatLE();
+                    if (longitude > -180.0f && longitude < 180.0f) {
```
tananaev: why do we need this check?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,88 +34,326 @@
...
+                    }
                     break;
+                }
 
-                case 3:
-                    position.setLatitude(buf.readFloatLE());
-                    position.setValid(true);
+                case 3: {
+                    float latitude = buf.readFloatLE();
+                    if (latitude > -90.0f && latitude < 90.0f) {
```
tananaev: And this?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,88 +34,326 @@
...
+                    int course = buf.readUnsignedByte() * 2;
+                    int altitude = buf.readUnsignedByte() * 10;
+                    int satByte = buf.readUnsignedByte();
+                    int speedKnots = buf.readUnsignedByte();
+
+                    if (satByte == 0xFE || satByte == 0xFF) {
+                        position.setValid(false);
+                    } else {
+                        int gpsSat = satByte & 0x0F;
+                        int gloSat = (satByte >> 4) & 0x0F;
```
tananaev: Use BitUtil instead

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,88 +34,326 @@
...
+
+                    switch (mode) {
+                        case 0x01: {
+                            int virtualSensors = buf.readUnsignedByte();
+                            int physicalInputs = buf.readUnsignedShortLE();
+                            position.set("virtualignition", BitUtil.check(virtualSensors, 0));
+                            position.set("callButton", BitUtil.check(virtualSensors, 1));
+                            for (int i = 0; i < 8; i++) {
+                                position.set(Position.PREFIX_IN + (i + 1), (physicalInputs & (1 << i)) != 0);
+                            }
```
tananaev: Typically we don't decode all bits like this. It will include too much unnecessary data in JSON.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,88 +34,326 @@
...
+                            break;
+                        }
+                        case 0x06: {
+                            int inputNum = buf.readUnsignedByte();
+                            int value = buf.readUnsignedShortLE();
```
tananaev: Remove unnecessary variables like this

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,88 +34,326 @@
...
+                    break;
+                }
+
+                case 8: {
+                    int mnc = buf.readUnsignedByte();
+                    int mcc = buf.readUnsignedShortLE();
+                    int signal = buf.readUnsignedByte();
+                    position.set(Position.KEY_RSSI, signal);
+                    position.set("mcc", mcc);
+                    position.set("mnc", mnc);
```
tananaev: This is not how we store cell data.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,88 +34,326 @@
...
+                    break;
+                }
+
+                case 9: {
                     long status = buf.readUnsignedIntLE();
-                    position.set(Position.KEY_POWER, BitUtil.from(status, 24) * 150 / 1000.0);
                     position.set(Position.KEY_STATUS, status);
+
+                    int inputs = (int) (status & 0xFF);
```
tananaev: Why `(int)` casting needed?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,88 +34,326 @@
...
+                    long flags = buf.readUnsignedIntLE();
+                    position.set("doorDriver", BitUtil.check(flags, 8));
+                    position.set("doorPassenger", BitUtil.check(flags, 9));
+                    position.set("trunkOpen", BitUtil.check(flags, 10));
+                    position.set("hoodOpen", BitUtil.check(flags, 11));
+                    position.set("handbrake", BitUtil.check(flags, 12));
+                    position.set("brakePedal", BitUtil.check(flags, 13));
+                    position.set("engineRunning", BitUtil.check(flags, 14));
+                    position.set("canIgnition", BitUtil.check(flags, 16));
+                    position.set("keyInIgnition", BitUtil.check(flags, 19));
```
tananaev: Same as earlier. We don't decode all bits like this usually.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -126,19 +364,31 @@ protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg
...
+
+            String imei = String.valueOf(imeiLong);
+            DeviceSession deviceSession = getDeviceSession(channel, remoteAddress, imei);
             if (deviceSession != null) {
                 sendResponse(channel, version, 0);
             }
+            return null;
+        }
 
+        if (startSign != PACKAGE_START_SIGN) {
```
tananaev: Why is this needed?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -148,35 +398,41 @@ protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg
...
 
         List<Position> positions = new LinkedList<>();
+        int packageNumber = buf.readUnsignedByte();
+
+        while (buf.isReadable()) {
+
+            int recordType = buf.readUnsignedByte();
 
-        int index = buf.readUnsignedByte();
+            if (recordType == PACKAGE_END_SIGN) {
```
tananaev: And this

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -148,35 +398,41 @@ protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg
...
                 }
+                default:
+                    return positions.isEmpty() ? null : positions;
             }
-
-            recordType = buf.readByte();
         }
 
-        sendResponse(channel, HEADER_VERSION_1, index);
+        sendResponse(channel, HEADER_VERSION_2, packageNumber);
```
tananaev: Why are we changing response?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: Coordinates falling outside the specified bounds are considered invalid. This serves as an additional verification step

tananaev: Does device actually report invalid coordinates? Do you have logs?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: We could add a condition to display only inputs in the 'true' state. However, in some cases, it's important to record the 'false' state as well. Is this really critical?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: Could you elaborate on the specific requirements? I don't quite understand

tananaev: You should use network property. See other decoders for examples.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: Originally, tag 9 holds more data than fits in an int, hence the long to int conversion in the code. Currently, only a subset is used, making long unnecessary. I can refactor to use int, but if we need to add new statuses later, we'll have to reintroduce the conversion logic

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: I agree with this point. We can log this data only when the state is 'true'

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: For HEADER responses where (packageNumber == 0). Otherwise, there is no difference. Also, for future reference, HEADER version 1 will no longer be supported by Arnavi developers

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: PACKAGE_END_SIGN is used to terminate the loop. When the while (buf.isReadable()) loop reads up to 0x5D, we call sendResponse

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: It was originally designed as a safeguard against garbage data and to ensure correct message parsing when multiple packets are contained in a single message. Let's keep this check. I've seen similar implementations in the Teltonica protocol.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: I haven't

tananaev: Then please remove. Remove all unnecessary defensive checks.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
tananaev: This is not how we usually do this. Typically you would just do loop until readable bytes > 1.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
tananaev: Logs please to show this actually happens.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
tananaev: I guess it's ok, but make sure you use BitUtil for all the bit manipulation.

NL1ghtman: ok, I`ll use it

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: removed

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: removed

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: removed

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: removed

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: removed

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: изображение
Keeping this implementation because the device allows configurable data sets.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -148,35 +398,41 @@ protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg
...
 
-        byte recordType = buf.readByte();
-        while (buf.readableBytes() > 0) {
             switch (recordType) {
-                case RECORD_PING, RECORD_DATA, RECORD_TEXT, RECORD_FILE, RECORD_BINARY -> {
+                case RECORD_PING:
+                case RECORD_DATA:
+                case RECORD_TEXT:
+                case RECORD_FILE:
+                case RECORD_BINARY: {
```
tananaev: Why are we changing this?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -148,35 +398,41 @@ protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg
...
                     int length = buf.readUnsignedShortLE();
-                    Date time = new Date(buf.readUnsignedIntLE() * 1000);
+                    Date time = new Date(buf.readUnsignedIntLE() * 1000L);
```
tananaev: And this?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -148,35 +398,41 @@ protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg
...
 
-                    if (recordType == RECORD_DATA) {
+                    if (recordType == RECORD_DATA && length > 0) {
```
tananaev: And this

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
tananaev: I don't understand the reason.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: "L" indicates the long type, ensuring the multiplication is performed as long * long without risk of overflow.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: length > 0 prevents decodePosition call on empty body. Safeguards against firmware issues.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
tananaev: It's completely redundant. It doesn't actually do anything in this case.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
tananaev: Have you actually seen messages like this?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: Switch expressions with -> need Java 14+, but Traccar's minimum is Java 11

tananaev: > but Traccar's minimum is Java 11

Where did you get this information from? It's completely false.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: With this data, just like input states, we might need to track 'false' values for notifications and external integrations. Example: analyzing how often and how long the driver presses the brake. Those who don't need this data can disable sending it in the device settings

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: изображение

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: No, haven't seen it. We're on the stable firmware build

NL1ghtman: removed

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: removed

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman:
``` java
case RECORD_PING, RECORD_DATA, RECORD_TEXT, RECORD_FILE, RECORD_BINARY : {
```
ok?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
tananaev: Doesn't it mean we should remove?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
tananaev: https://github.com/traccar/traccar/blob/master/build.gradle#L14-L15

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: No, I actually find this data useful for my use case

NL1ghtman: Actually, I don't understand why this needs to be removed if this functionality might be useful to other users besides me

NL1ghtman:
Found an example of an incomplete packet: 145 bytes declared, 144 received — missing 2 tags + checksum + 0x5D. I didn't see any buffer bounds protection in ArnaviFrameDecoder, only 
`if (buf.getByte(index) != PACKAGE_END_SIGN) {
    return null;
}` 


Attaching the error logs. 
[log.txt](https://github.com/user-attachments/files/25712801/log.txt)

An additional check index >= buf.readableBytes() should fix TCP fragmentation.

`if (index >= buf.readableBytes() || buf.getByte(buf.readerIndex() + index) != PACKAGE_END_SIGN) {
    return null;
}`

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
tananaev: I already explained earlier. This all too much bloat to the attributes. Typically we save bitmasks as is and then you can extract values using computed attributes. In this case especially because all of these attributes are custom.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: I switched to using the Network model. But RSSI isn't saving, and I can't understand why.

изображение

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: Got it now. I've pushed a commit.

tananaev:
One big missing thing is unit tests. We need to add some tests for the new data you've added.

NL1ghtman:
> One big missing thing is unit tests. We need to add some tests for the new data you've added.

done

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
```diff
@@ -1,34 +1,137 @@
...
+
+                verifyNull(decoder, binary("ff24f30c45f5c90f03000102030405060708"));
+
+                verifyPositions(decoder, binary(
+                                "5b01012800a3175f5903513934420447221c42055402781E0900f0c5215b4e0084005c00007c005d0000a300fa37010000295d"),
+                                position("2017-07-07 05:09:55.000", true, 45.05597, 39.03347));
+        }
+
+        @Test
+        public void testFullTagsAttributes() throws Exception {
```
tananaev: Please keep the standard format for decoder unit tests.

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
```diff
@@ -1,34 +1,137 @@
...
 import org.traccar.ProtocolTest;
+import org.traccar.model.CellTower;
+import org.traccar.model.Position;
 
 public class ArnaviBinaryProtocolDecoderTest extends ProtocolTest {
 
-    @Test
-    public void testHeader1Decode() throws Exception {
+        @Test
+        public void testHeader1Decode() throws Exception {
```
tananaev: There's something wrong with formatting. Why are we changing existing stuff?

NL1ghtman: I used automatic code formatting in VS Code.
Fixed now

NL1ghtman:
Please let me know if anything else needs to be adjusted

tananaev:
The test file is still completely inconsistent with other tests.

NL1ghtman:
> The test file is still completely inconsistent with other tests.

Got it. I will follow the pattern used in the other tests.

tananaev:
I don't see any changes. It seems like you forgot to push any updates.

NL1ghtman:
Done

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
```diff
@@ -2,33 +2,122 @@
...
 
-        verifyPositions(decoder, binary(
-                "5b01012800a3175f5903513934420447221c42055402781E0900f0c5215b4e0084005c00007c005d0000a300fa3701000029012800a3175f5903513934420447221c42055402781E0900f0c5215b4e0084005c00007c005d0000a300fa37010000295d"),
-                position("2017-07-07 05:09:55.000", true, 45.05597, 39.03347));
-    }
+                verifyNull(decoder, binary("ff23ae52a969f30c45f5"));
 
-}
+                verifyPositions(decoder, binary(
+                                "5bae0196008a52a9690373f35c42049746ac420574147529973e00000007ca325cb061ca3200000863fa001e96e553fe006353801a0201b50f5d6c0601010000474e0000005b4e0000005c000017005d000017003203000000330040110034d83b05003579b7200136700e090037c301000039ce0300003a4c0000003b4e0000003c8a0101003db8ab0000404c010000be1e4b00c045002d0000fa3b010000d75d"),
```
tananaev: We don't need so many test cases. It's fine to check 1-2 attributes.

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
```diff
@@ -2,33 +2,122 @@
...
+                                position("2017-07-07 05:09:55.000", true, 45.05597, 39.03347));
+        }
 
-        verifyNull(decoder, binary(
-                "ff23f30c45f5c90f0300"));
+        @Test
+        public void testFullTagsAttributes() throws Exception {
+                var decoder = inject(new ArnaviBinaryProtocolDecoder(null));
 
-        verifyPositions(decoder, binary(
```
tananaev, line 29: Why are we removing this test?

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
```diff
@@ -2,33 +2,122 @@
 
 import org.junit.jupiter.api.Test;
 import org.traccar.ProtocolTest;
+import org.traccar.model.Position;
 
 public class ArnaviBinaryProtocolDecoderTest extends ProtocolTest {
 
-    @Test
-    public void testHeader1Decode() throws Exception {
+        @Test
```
tananaev: What happened with indentation?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -126,19 +282,25 @@ protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg
 
         ByteBuf buf = (ByteBuf) msg;
 
-        byte startSign = buf.readByte();
+        if (!buf.isReadable()) {
```
tananaev: Why do we need this check?

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,90 +36,244 @@
...
     }
 
-    private void sendResponse(Channel channel, byte version, int index) {
-        if (channel != null) {
-            ByteBuf response = Unpooled.buffer();
-            response.writeByte(0x7b);
+    private void sendResponse(Channel channel, int version, int packageNumber) {
+        if (channel == null) {
+            return;
+        }
```
tananaev: Please don't do unnecessary reformatting like this

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
```diff
@@ -34,90 +36,244 @@
...
 
-    private static final byte RECORD_PING = 0x00;
-    private static final byte RECORD_DATA = 0x01;
-    private static final byte RECORD_TEXT = 0x03;
-    private static final byte RECORD_FILE = 0x04;
-    private static final byte RECORD_BINARY = 0x06;
+    private static final int HEADER_VERSION_1 = 0x22;
+    private static final int HEADER_VERSION_2 = 0x23;
+    private static final int HEADER_VERSION_3 = 0x24;
+    // private static final int HEADER_VERSION_4 = 0x25;
```
tananaev: Do we need this one?

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
NL1ghtman: VS Code auto-formatting. This is not critical, but I can revert the changes if preferred.

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
NL1ghtman: I didn't remove it — I swapped it with my own data packet to test the new tags.

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: removed

src/main/java/org/traccar/protocol/ArnaviBinaryProtocolDecoder.java
NL1ghtman: removed

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
NL1ghtman: fixed

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
tananaev: It is critical for preserving change history. Please revert any unnecessary formatting changes.

NL1ghtman:
Done

tananaev:
It doesn't look like the formatting issue is fixed. Have you checked the GitHub diff first before asking to review?

NL1ghtman:
Could you please specify where the error is? What exactly looks off? I'm using the built-in VS Code formatter. Where did I make a mistake?

tananaev:
Just open the diff here on GitHub and you should see the problem.

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
```diff
@@ -20,15 +21,43 @@ public void testHeader1Decode() throws Exception {
...
+                "5b01012800a3175f5903513934420447221c42055402781E0900f0c5215b4e0084005c00007c005d0000a300fa37010000295d"),
+                position("2017-07-07 05:09:55.000", true, 45.05597, 39.03347));
+    }
 
+    @Test
+    public void testHeader3Decode() throws Exception {
+        var decoder = inject(new ArnaviBinaryProtocolDecoder(null));
+        verifyNull(decoder, binary("ff24f30c45f5c90f03000102030405060708"));
         verifyPositions(decoder, binary(
-                "5b01012800a3175f5903513934420447221c42055402781E0900f0c5215b4e0084005c00007c005d0000a300fa3701000029012800a3175f5903513934420447221c42055402781E0900f0c5215b4e0084005c00007c005d0000a300fa37010000295d"),
```
tananaev, line 30: What happened to this test? Why was it removed?

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
```diff
@@ -20,15 +21,43 @@ public void testHeader1Decode() throws Exception {
 
     @Test
     public void testHeader2Decode() throws Exception {
-
         var decoder = inject(new ArnaviBinaryProtocolDecoder(null));
-
```
tananaev, line 25: Revert unnecessary changes.

src/test/java/org/traccar/protocol/ArnaviBinaryProtocolDecoderTest.java
```diff
@@ -20,15 +21,43 @@ public void testHeader1Decode() throws Exception {
...
         verifyNull(decoder, binary(
                 "ff23f30c45f5c90f0300"));
+        verifyPositions(decoder, binary(
+                "5b01012800a3175f5903513934420447221c42055402781E0900f0c5215b4e0084005c00007c005d0000a300fa37010000295d"),
+                position("2017-07-07 05:09:55.000", true, 45.05597, 39.03347));
+    }
```
tananaev, line 33: Missing spacing.

NL1ghtman:
Fixed testHeader2Decode and testHeader3Decode, added missing spaces.

NL1ghtman:
Thank you for your help and the valuable lesson

PR #5801, opened 2026-02-24
Fix OpenID administrator flag not syncing for existing users

vaishnavbhosale:
###Issue:

When OpenID group membership changes, the administrator flag
is only set during initial user creation and is not updated
for existing users on subsequent logins.

###Steps to reproduce:

1. Login with a user in adminGroups (administrator = true)
2. Remove the user from adminGroups in the OpenID provider
3. Login again
4. Administrator flag remains unchanged

###Fix:

Synchronize the administrator flag on every OpenID login
by comparing and updating it if necessary.

### DIFF
--- a/src/main/java/org/traccar/api/security/LoginService.java
+++ b/src/main/java/org/traccar/api/security/LoginService.java
@@ -35,6 +35,7 @@ import jakarta.inject.Singleton;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.security.GeneralSecurityException;
+import java.util.Objects;
 
 @Singleton
 public class LoginService {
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/api/security/LoginService.java
```diff
@@ -129,7 +130,25 @@ public LoginResult login(String email, String name, boolean administrator) throw
...
+            boolean updated = false;
+
+            if (!name.equals(user.getName())) {
+                user.setName(name);
+                updated = true;
+            }
+            if (user.getAdministrator() != administrator) {
+                user.setAdministrator(administrator);
+                updated = true;
+            }
```
tananaev: I think we can do this under a single condition. If either one changed, update both.

vaishnavbhosale: Thanks for the suggestion. I've updated the logic to use a single conditional check and synchronize both fields together as recommended.

src/main/java/org/traccar/api/security/LoginService.java
```diff
@@ -129,11 +130,23 @@ public LoginResult login(String email, String name, boolean administrator) throw
             user.setFixedEmail(true);
             user.setAdministrator(administrator);
             user.setId(storage.addObject(user, new Request(new Columns.Exclude("id"))));
+        } else {
```
tananaev: Collapse this now to use `else if {`

src/main/java/org/traccar/api/security/LoginService.java
```diff
@@ -129,11 +130,23 @@ public LoginResult login(String email, String name, boolean administrator) throw
...
+
+            if (!java.util.Objects.equals(name, user.getName())
```
tananaev: Use import

vaishnavbhosale: done...

PR #5806, opened 2026-03-01
Create geofence realtime

jubarcik:
Adds a real-time geofence if the vehicle is within the perimeter; can be used to create an anchor function in the webview.


tananaev:
This needs a better description.

jubarcik:
When creating a geofence, even if the device was inside it, it would only update if there was a status change, such as turning the ignition on or off, or a periodic device update. Now, if the device is inside the geofence, it updates in real time without needing any device changes, and it also updates the geofence IDs in positions. Since I use the geofence as an anchor function, it only became active if the device changed state.

tananaev:
I don't think that's how most people expect things to work. You basically retroactively updating the geofence. What about the events? What if the position is really old?

jubarcik:
Since this is a feature I worked hard to develop, I would be grateful if you included it in the project. Thank you.

jubarcik:
In the anchor function, I create a perimeter with the device stationary, with a radius between 50m and 150m only. If the vehicle leaves this established perimeter, it notifies and blocks the vehicle by exiting the geofence. When the client removes the geofence, it removes the geofence and unlocks the vehicle.

tananaev:
I don't think you answered my questions.

PR #5807, opened 2026-03-01
ttnhttp: Add downlink support

Steve-Tech:
Hi, The Things Network provides an API for [scheduling 'downlinks'](https://www.thethingsindustries.com/docs/integrations/webhooks/scheduling-downlinks/), which are messages sent back to the device (e.g. to sound a buzzer, etc.).

To make the configuration easy, the HTTP headers of the normal 'uplink' webhooks can optionally provide Traccar an API key (`X-Downlink-Apikey`) and 2 URLs to POST to (`X-Downlink-Push` - pushes to the queue, `X-Downlink-Replace` - replaces the queue [not used]). I've implemented this auto configuration, but they can also be manually configured from the device attributes.

I am unsure if this is the best way of passing those headers from the decoder to the command sender (though it does work), but I am happy to receive feedback.

As there are no standard commands, only custom commands are handled. If the command data entered by the user is valid JSON, then it is passed straight to TTN's downlink encoder, otherwise the data is wrapped into `{"command": "<data>"}` to become valid JSON.

Thanks,
Steve

### DIFF
--- a/src/main/java/org/traccar/command/CommandSenderManager.java
+++ b/src/main/java/org/traccar/command/CommandSenderManager.java
@@ -30,7 +30,8 @@ public class CommandSenderManager {
     private static final Map<String, Class<? extends CommandSender>> SENDERS_ALL = Map.of(
             "firebase", FirebaseCommandSender.class,
             "traccar", TraccarCommandSender.class,
-            "findHub", FindHubCommandSender.class);
+            "findHub", FindHubCommandSender.class,
+            "ttnhttp", TtnHttpCommandSender.class);
 
     private final Config config;
     private final Injector injector;
@@ -51,6 +52,8 @@ public class CommandSenderManager {
             } else if (config.hasKey(Keys.NOTIFICATOR_TRACCAR_KEY)) {
                 return injector.getInstance(TraccarCommandSender.class);
             }
+        } else if (device.hasAttribute("X-Downlink-Apikey")) {
+            return injector.getInstance(TtnHttpCommandSender.class);
         }
         return null;
     }
--- a/src/main/java/org/traccar/command/CommandSenderManager.java
+++ b/src/main/java/org/traccar/command/CommandSenderManager.java
@@ -52,7 +52,7 @@ public class CommandSenderManager {
             } else if (config.hasKey(Keys.NOTIFICATOR_TRACCAR_KEY)) {
                 return injector.getInstance(TraccarCommandSender.class);
             }
-        } else if (device.hasAttribute("X-Downlink-Apikey")) {
+        } else if (device.hasAttribute(Keys.COMMAND_TTNHTTP_APIKEY.getKey())) {
             return injector.getInstance(TtnHttpCommandSender.class);
         }
         return null;
--- a/src/main/java/org/traccar/command/TtnHttpCommandSender.java
+++ b/src/main/java/org/traccar/command/TtnHttpCommandSender.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2026 Stephen Horvath (me@stevetech.au)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


src/main/java/org/traccar/command/CommandSenderManager.java
```diff
@@ -51,6 +52,8 @@ public CommandSender getSender(Device device) {
             } else if (config.hasKey(Keys.NOTIFICATOR_TRACCAR_KEY)) {
                 return injector.getInstance(TraccarCommandSender.class);
             }
+        } else if (device.hasAttribute("X-Downlink-Apikey")) {
```
tananaev: This is not how we name attributes.

src/main/java/org/traccar/command/TtnHttpCommandSender.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public void sendCommand(Device device, Command command) throws Exception {
+        if (!device.hasAttribute("X-Downlink-Apikey")) {
+            throw new RuntimeException("Missing X-Downlink-Apikey header");
+        }
+        String key = device.getString("X-Downlink-Apikey");
+
+        String url = device.getString("X-Downlink-Push");
+        if (command.getBoolean(Command.KEY_NO_QUEUE) && device.hasAttribute("X-Downlink-Replace")) {
```
tananaev: What is the difference between these two keys?

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
```diff
@@ -63,6 +64,16 @@ protected Object decode(Channel channel, SocketAddress remoteAddress, Object msg
             return null;
         }
 
+        Device device = getCacheManager().getObject(Device.class, deviceSession.getDeviceId());
```
tananaev, line 68: Are headers always there or they're sent only once?

src/main/java/org/traccar/command/TtnHttpCommandSender.java
Steve-Tech: TTN has it's own queue, push adds to the end of the queue, replace effectively clears the queue before adding.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: They're always there as long as 'Downlink API Key' is enabled in the TTN console.

tananaev: Then I think we need to do something better than writing on every position.

src/main/java/org/traccar/command/TtnHttpCommandSender.java
tananaev: Then why do we need to support it?

Steve-Tech: I find it useful to change a message before it has actually been sent to the tracker using that 'No queue' checkbox. But I can remove replace support if I'm just abusing the 'No queue' checkbox.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
Steve-Tech: Fair enough, would you prefer writing if it doesn't exist, or writing if it's different?

src/main/java/org/traccar/command/TtnHttpCommandSender.java
tananaev: Yeah, it's definitely not designed for that. Let's remove for now.

src/main/java/org/traccar/protocol/TtnHttpProtocolDecoder.java
tananaev: I think it makes sense if it's different.

PR #5808, opened 2026-03-03
Gt06ProtocolDecoder: populate power from adc1 if power is not present

chrystianmartins:
Set power from adc1 for MSG_INFO (subType 0x00) when power is not already present.

### DIFF
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
@@ -1159,7 +1159,14 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
 
             if (subType == 0x00) {
 
-                position.set(Position.PREFIX_ADC + 1, buf.readUnsignedShort() * 0.01);
+                double value = buf.readUnsignedShort() * 0.01;
+
+                position.set(Position.PREFIX_ADC + 1, value);
+
+                if (value > 0 && position.getAttributes().get(Position.KEY_POWER) == null) {
+                    position.set(Position.KEY_POWER, value);
+                }
+
                 return position;
 
             } else if (subType == 0x04) {


tananaev:
Need a bit more context here.

chrystianmartins:
Sure, here is the context.

Some GT06-based devices report the external power voltage through **ADC1** in the `MSG_INFO` message (`subType == 0x00`). In these cases the device does not explicitly populate the `power` attribute in the protocol payload.

Before this change, the decoder only stored the ADC value as:

`Position.PREFIX_ADC + 1`

This means the voltage was available only as an ADC attribute and `Position.KEY_POWER` could remain `null`, even though the device had already provided the information indirectly.

What this change does:

1. It reads the ADC1 value exactly the same way as before (`buf.readUnsignedShort() * 0.01`).
2. It still stores this value in `Position.PREFIX_ADC + 1`, preserving the original behavior.
3. Additionally, if the decoded position does not already contain a `Position.KEY_POWER` value, the decoder uses the ADC1 value to populate `Position.KEY_POWER`.

The important part is that we **do not overwrite** an existing `power` value. The assignment only happens when:

* `value > 0`
* `Position.KEY_POWER` is still `null`

So the goal is simply to ensure that `power` is available when the device reports it indirectly via ADC1, while keeping the original ADC attribute and avoiding overriding explicit power values that might appear in other messages.

tananaev:
Please provide logs and the protocol documentation.

chrystianmartins:
This appears to be relevant for some Concox devices.

Protocol manual:
http://traccar.org/protocol/5023-gt06/JM-LL301%20GPS%20Tracker%20Communication%20Protocol_v1.0_20210823.pdf

Related forum report:
https://www.traccar.org/forums/topic/concox-crx-pro-4g-external-battery-voltage-gt06/

In these cases, the external power voltage is reported through ADC1 in the GT06 `MSG_INFO` message, but the decoder only keeps it as `adc1`. As a result, `Position.KEY_POWER` remains unset even though the device is already providing the voltage indirectly.

I am not sure why the computed/calculated value does not already resolve this, but based on the device behavior above, mapping ADC1 to power when power is absent seems necessary for these Concox models.

PR #5812, opened 2026-03-11
Implement daily summary scheduled report

AussieTraccar:
**Purpose:** Scheduled summary report only supports type `Summary` and not type `Daily Summary`.

**Solution:** This PR implements scheduled Summary report type `Daily Summary` in addition to type `Summary`.

### DIFF
--- a/src/main/java/org/traccar/schedule/TaskReports.java
+++ b/src/main/java/org/traccar/schedule/TaskReports.java
@@ -45,7 +45,6 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.Set;
-import java.util.Objects;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/schedule/TaskReports.java
```diff
@@ -128,6 +129,9 @@ private void executeReport(Report report, Date from, Date to) throws StorageExce
         if (!groupIdsPart.isEmpty()) {
             url.append(groupIdsPart).append('&');
         }
+        if (Objects.equals(report.getType(), "summary") && report.hasAttribute("daily")) {
```
tananaev: Why not just check `hasAttribute`?

src/main/java/org/traccar/schedule/TaskReports.java
```diff
@@ -128,6 +129,9 @@ private void executeReport(Report report, Date from, Date to) throws StorageExce
...
+            url.append("daily=").append(report.getAttributes().get("daily")).append('&');
```
tananaev:
```suggestion
            url.append("daily=").append(report.getBoolean("daily")).append('&');
```

src/main/java/org/traccar/schedule/TaskReports.java
AussieTraccar: This attribute is really only relative to `summary` report. Just catches future attributes in other reports that maybe set as `daily`.

tananaev: Yeah, but I don't think we need this extra safety here.

src/main/java/org/traccar/schedule/TaskReports.java
AussieTraccar: Lacking code insight and missed `getBoolean()`. After digging deeper in `ExtendedModel.java`, agree that this probably the preferred method.

PR #5817, opened 2026-03-14
Update XexunProtocol.java

ctucker1984:
Corrects a math.floor to int calculation which causes the erroneous plotting of the position of the tracker approx. 1 degree east of the real position.


tananaev:
I think you updated the wrong file. Also you would need to add a test case.

ctucker1984:
Lord, you're right. I'm going to close this and re-file. My appologies.

PR #5826, opened 2026-03-16
Add support for H08E external battery voltage readings

FoxExe:
Device info: https://www.traccar.org/forums/topic/unknown-gt06s-h08e-device/#post-83291
No internal battery (Always 100%).
External battery voltage stored in extended statistics message (module type 0x27, uint16_t, maybe 1 or 4 byte possible too)

Alternative (short) variant:
`case 0x0027 -> position.set(Position.KEY_POWER, buf.readUnsignedShort() / 100.0);`

### DIFF
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
@@ -913,6 +913,14 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
                     int moduleLength = buf.readUnsignedByte();
                     switch (moduleType) {
                         case 0x0018 -> position.set(Position.KEY_BATTERY, buf.readUnsignedShort() / 100.0);
+                        case 0x0027 -> {
+                            // H08E - possible 1 or 2 bytes (maybe 4)
+                            long value = 0;
+                            for (int i = 0; i < moduleLength && buf.isReadable(); i++) {
+                                value = (value << 8) | buf.readUnsignedByte();
+                            }
+                            position.set(Position.KEY_POWER, value / 100.0);
+                        }
                         case 0x0032 -> position.set("startupStatus", buf.readUnsignedByte());
                         case 0x006A -> position.set(Position.KEY_BATTERY_LEVEL, buf.readUnsignedByte());
                         default -> buf.skipBytes(moduleLength);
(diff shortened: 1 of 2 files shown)


tananaev:
Please attach documentation. And add a real test case.

FoxExe:
Ok, i just added test, using real raw data from server logs.

For example:
`78780f36040602000200270212ce004c80370d0a` -> `0x12ce` -> 4814(mV) / 100 = 48.14v (eBike 13s battery, ~56% charge)

I found this information through observation, but I also found a reference to it in this file: https://www.traccar.org/protocol/5023-gt06/JM-LL301%20GPS%20Tracker%20Communication%20Protocol_v1.0_20210823.pdf ("example" on page 9)

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -913,6 +913,14 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
                     int moduleLength = buf.readUnsignedByte();
                     switch (moduleType) {
                         case 0x0018 -> position.set(Position.KEY_BATTERY, buf.readUnsignedShort() / 100.0);
+                        case 0x0027 -> {
+                            // H08E - possible 1 or 2 bytes (maybe 4)
```
tananaev: Where did you get this information from? I don't see it in the doc.

FoxExe: image
Page 9.

FoxExe: Or you talking about "1 or 4 bytes is possible"?
There is "module length" byte. So, size of data (potentially) can change.

tananaev: I understand there is a length, but where did you get the information that it's variable for this specific value? And what the values are?

FoxExe: As I mentioned, the sample data and documentation are already provided above. My assumption about variable length is based on the fact that the payload itself is dynamic. That’s exactly why I provided two versions of the code: one with a length check and an alternative one without it.
I’m not sure what else is missing. Could you point out specifically what part of my previous answer is unclear?

tananaev: I suspect power value is always 2 bytes and that's what we should implement. Unless you have some clear evidence to the contrary.

PR #5827, opened 2026-03-17
Improve code readability and modularity in report utilities and geofence logic

Sri27-ram:
Refactoring only — no behavior or API changes; all existing tests pass.

This PR makes several code quality improvements across the report generation 
and geofence modules.

Changes in `ReportUtils`:
-Extracted repeated address-resolution logic into a dedicated `resolveAddress()` 
helper to avoid duplication across trip and stop report builders
-Moved address-resolution responsibility into a new `ReportAddressResolver` class 
to keep concerns separated
-Replaced the report-type conditional in `calculateTripOrStop()` with a 
`TripStopReportCalculator` interface and two implementations, making it easier 
to extend in the future
Renamed the `trips` boolean to `isTripReport` for clarity
-Introduced named boolean variables in `calculateFuel()` to make the branching 
logic easier to follow

Changes in `GeofenceGeometry`:
-Decomposed the compound bounds-check condition in `containsPoint()` into 
named booleans to make the date-line handling and spatial logic more readable

All existing tests pass. No behaviour changes.

### DIFF
--- a/REFACTORING.md
+++ b/REFACTORING.md
@@ -1,0 +1,34 @@
+# Refactoring Summary
+
+This document describes the refactoring techniques applied to the Traccar codebase. Each refactoring was applied in a separate commit with clear, easy-to-understand commit messages and 10-minute intervals between commits.
+
+## Commit History
+
+| Time  | Commit message | Technique |
+|-------|----------------|-----------|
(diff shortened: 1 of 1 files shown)


tananaev:
Sorry, but I don't think I see the value in this. You just added more complexity and indirection. You extracted one-liners into separate classes, which doesn't make sense to me.

PR #5831, opened 2026-03-23
Support proxy configuration for Telegram notifications

klim42:
## Summary

Add optional proxy configuration for Telegram notifications.

This allows Telegram requests to work in deployments where outbound access to `api.telegram.org` is restricted.

## Changes

- add Telegram proxy host, port, type, username, and password settings
- support optional HTTP and SOCKS5 proxies for Telegram notifications
- keep existing behavior unchanged when proxy is not configured

## Verification

- `./gradlew compileJava`

### DIFF
--- a/src/main/java/org/traccar/notificators/NotificatorTelegram.java
+++ b/src/main/java/org/traccar/notificators/NotificatorTelegram.java
@@ -32,6 +32,8 @@ import org.traccar.model.User;
 import org.traccar.notification.NotificationFormatter;
 import org.traccar.notification.NotificationMessage;
 
+import java.net.URI;
+
 @Singleton
 public class NotificatorTelegram extends Notificator {
 
@@ -75,20 +77,18 @@ public class NotificatorTelegram extends Notificator {
         chatId = config.getString(Keys.NOTIFICATOR_TELEGRAM_CHAT_ID);
         sendLocation = config.getBoolean(Keys.NOTIFICATOR_TELEGRAM_SEND_LOCATION);
 
-        String proxyHost = config.getString(Keys.NOTIFICATOR_TELEGRAM_PROXY_HOST);
-        if (proxyHost != null) {
-            int proxyPort = config.getInteger(Keys.NOTIFICATOR_TELEGRAM_PROXY_PORT);
+        String proxyUrl = config.getString(Keys.NOTIFICATOR_TELEGRAM_PROXY_URL);
+        if (proxyUrl != null) {
             ClientBuilder clientBuilder = ClientBuilder.newBuilder()
                     .register(objectMapperContextResolver)
-                    .property(
-                            ClientProperties.PROXY_URI,
-                            String.format("http://%s:%d", proxyHost, proxyPort));
-            String proxyUser = config.getString(Keys.NOTIFICATOR_TELEGRAM_PROXY_USER);
-            if (proxyUser != null) {
-                clientBuilder.property(ClientProperties.PROXY_USERNAME, proxyUser);
-                clientBuilder.property(
-                        ClientProperties.PROXY_PASSWORD,
-                        config.getString(Keys.NOTIFICATOR_TELEGRAM_PROXY_PASSWORD, ""));
+                    .property(ClientProperties.PROXY_URI, proxyUrl);
+            String userInfo = URI.create(proxyUrl).getUserInfo();
+            if (userInfo != null) {
+                String[] credentials = userInfo.split(":", 2);
+                clientBuilder.property(ClientProperties.PROXY_USERNAME, credentials[0]);
+                if (credentials.length > 1) {
+                    clientBuilder.property(ClientProperties.PROXY_PASSWORD, credentials[1]);
+                }
             }
             this.client = clientBuilder.build();
         } else {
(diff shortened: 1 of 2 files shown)


tananaev:
Is there a way to configure `jakarta.ws.rs.client.Client` to go through a proxy? I think we also need to simplify this a lot. It's a very niche function, so I don't want to add too much complexity.

klim42:
> Is there a way to configure `jakarta.ws.rs.client.Client` to go through a proxy? I think we also need to simplify this a lot. It's a very niche function, so I don't want to add too much complexity.

Yes, for HTTP proxy we can configure the existing `jakarta.ws.rs.client.Client`. The custom code here was only added for SOCKS5 support, because Jersey proxy configuration appears to cover HTTP proxy, but not SOCKS5.

If you'd prefer to keep this simpler, I can rework the PR to support HTTP proxy only and remove the SOCKS5-specific code.

tananaev:
Works for me. The simpler the better.

klim42:
> Works for me. The simpler the better.

Reworked to keep this simpler.
This version only supports HTTP proxy and uses a dedicated configured `jakarta.ws.rs.client.Client` with Jersey proxy properties, so the custom SOCKS-specific code path is removed.

src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -62,15 +65,35 @@ public static class LocationMessage {
...
         sendLocation = config.getBoolean(Keys.NOTIFICATOR_TELEGRAM_SEND_LOCATION);
+
+        String proxyHost = config.getString(Keys.NOTIFICATOR_TELEGRAM_PROXY_HOST);
+        if (proxyHost != null) {
+            int proxyPort = config.getInteger(Keys.NOTIFICATOR_TELEGRAM_PROXY_PORT);
+            ClientBuilder clientBuilder = ClientBuilder.newBuilder()
+                    .register(objectMapperContextResolver)
+                    .property(
+                            ClientProperties.PROXY_URI,
+                            String.format("http://%s:%d", proxyHost, proxyPort));
```
tananaev: I think it's better to just have a single proxy URL.

src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -62,15 +65,35 @@ public static class LocationMessage {
...
+            String proxyUser = config.getString(Keys.NOTIFICATOR_TELEGRAM_PROXY_USER);
+            if (proxyUser != null) {
+                clientBuilder.property(ClientProperties.PROXY_USERNAME, proxyUser);
+                clientBuilder.property(
+                        ClientProperties.PROXY_PASSWORD,
+                        config.getString(Keys.NOTIFICATOR_TELEGRAM_PROXY_PASSWORD, ""));
```
tananaev: You shouldn't use a version with default empty value here. Default should be defined in the key.

src/main/java/org/traccar/notificators/NotificatorTelegram.java
```diff
@@ -62,15 +67,33 @@ public static class LocationMessage {
...
         urlSendLocation = String.format(
                 "https://api.telegram.org/bot%s/sendLocation", config.getString(Keys.NOTIFICATOR_TELEGRAM_KEY));
         chatId = config.getString(Keys.NOTIFICATOR_TELEGRAM_CHAT_ID);
         sendLocation = config.getBoolean(Keys.NOTIFICATOR_TELEGRAM_SEND_LOCATION);
+
+        String proxyUrl = config.getString(Keys.NOTIFICATOR_TELEGRAM_PROXY_URL);
+        if (proxyUrl != null) {
+            ClientBuilder clientBuilder = ClientBuilder.newBuilder()
+                    .register(objectMapperContextResolver)
+                    .property(ClientProperties.PROXY_URI, proxyUrl);
```
tananaev, line 84: Is setting this not enough for username and password?

klim42: As far as I can tell, no.
In Jersey, `ClientProperties.PROXY_URI` is only used for the proxy endpoint itself. Proxy credentials are read separately from `ClientProperties.PROXY_USERNAME` and `ClientProperties.PROXY_PASSWORD`, and they are not populated automatically from the URI user info.
That is why I parse `user:pass@` from the proxy URL and set those properties explicitly.

tananaev: But the URL still works for it as is? Have you checked that it's working correctly?

klim42: I checked it with a local smoke test against an HTTP proxy.
With a plain `ClientProperties.PROXY_URI`, the request succeeds through a non-authenticated proxy.
With `ClientProperties.PROXY_URI = "http://user:pass@host:port"` against a proxy requiring Basic authentication, Jersey did not send `Proxy-Authorization` and the request failed with `407 Proxy Authentication Required`.
With the current approach in this PR, where credentials are parsed from `proxy.url` and mapped to `ClientProperties.PROXY_USERNAME` and `ClientProperties.PROXY_PASSWORD`, the request succeeded and the proxy received the expected `Proxy-Authorization` header.

So `PROXY_URI` alone is enough for a plain HTTP proxy, but it does not appear to be enough for proxy authentication.

tananaev:
Merged, thanks.

PR #5834, opened 2026-03-24
GT06 NT20 protocol correction, hours should be long and odometer shou…

jcardus:
…ld be read as unsigned

### DIFF
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
@@ -1028,7 +1028,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
 
             if (modelNT20 && type == MSG_GPS_LBS_2) {
                 position.set(Position.KEY_ODOMETER, buf.readUnsignedMedium());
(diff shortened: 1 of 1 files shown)


tananaev:
Need proper explanation.

jcardus:
> Need proper explanation.

If we don't read unsigned we will get negative values on higher values


Int (4 bytes) is not enough for milliseconds, it overlaps at 597 hours.

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -1027,8 +1027,8 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
             }
 
             if (modelNT20 && type == MSG_GPS_LBS_2) {
-                position.set(Position.KEY_ODOMETER, buf.readMedium());
-                position.set(Position.KEY_HOURS, buf.readMedium() * 60 * 1000);
+                position.set(Position.KEY_ODOMETER, buf.readUnsignedMedium());
+                position.set(Position.KEY_HOURS, (long) buf.readUnsignedMedium() * 60 * 1000);
```
tananaev:
```suggestion
                position.set(Position.KEY_HOURS, buf.readUnsignedMedium() * 60 * 1000L);
```

tananaev:
Thanks

PR #5836, opened 2026-03-26
feat: add offline grace period to suppress transient disconnect events

eosfomogame:
## Summary

This PR introduces an **offline grace period** mechanism that delays the `device_offline` event when a device disconnects. If the device reconnects within the configured grace period, no offline event is fired — avoiding spurious offline/online notification spam caused by brief network glitches or device reconnects.

## Motivation

In real-world deployments, GPS trackers frequently drop and re-establish TCP connections due to cellular network instability, protocol timeouts, or brief signal loss. Without a grace period, each such reconnect triggers a pair of `device_offline` / `device_online` events, flooding notification channels (email, SMS, push) with false alarms.

This is a commonly requested feature in the community and a practical necessity for production deployments.

## Configuration

The feature is controlled by the existing config key:

```
event.deviceOfflineMinDuration = 60
```

- Value is in **seconds**
- Default is `0` → grace period disabled (fully backward compatible)
- When enabled, the `device_offline` event is delayed by the configured duration

### Changes in `ConnectionManager.java`

- Added `offlineGracePeriod` field read from `Keys.EVENT_DEVICE_OFFLINE_MIN_DURATION`
- Added `pendingOfflineTasks` (`ConcurrentHashMap<Long, ScheduledFuture<?>>`) to track pending offline timers per device
- Added `offlineGraceScheduler` (`ScheduledExecutorService`) with a named daemon thread for clean lifecycle management
- Modified `deviceDisconnected()`: when grace period > 0, schedules a delayed `STATUS_OFFLINE` update instead of firing immediately; cancels any existing pending task for the same device first
- Modified `updateDevice()`: when `STATUS_ONLINE` is received, cancels any pending offline task atomically via `ConcurrentHashMap.remove()`
- Added `@PreDestroy shutdown()` to gracefully stop the scheduler on server shutdown

### Race condition handling

The lambda uses an atomic `ConcurrentHashMap.remove()` check:

```java
if (pendingOfflineTasks.remove(deviceId) != null) {
    updateDevice(deviceId, Device.STATUS_OFFLINE, null);
}
```

This guarantees that if a reconnect cancels the task between scheduling and execution, the offline event is never fired — even in tight timing scenarios.

### No behavior change when grace period is disabled

When `offlineGracePeriod == 0` (the default), the code takes the `else` branch and calls `updateDevice` immediately, identical to the previous behavior.

## Testing

Manually tested the following scenarios:
- Device disconnects and does **not** reconnect → `device_offline` event fires correctly after grace period
- Device disconnects and reconnects **within** grace period → no offline event fired
- Device disconnects multiple times rapidly → only the latest timer is active (previous ones are cancelled)
- Server shutdown with pending tasks → scheduler shuts down gracefully via `@PreDestroy`

## Notes for reviewers

- No new config keys are introduced; `Keys.EVENT_DEVICE_OFFLINE_MIN_DURATION` was already defined
- No database schema changes
- No API changes
- Fully backward compatible — existing deployments without the config key are unaffected

I would really appreciate a review from @tananaev if possible. Happy to adjust the implementation based on feedback.

### DIFF
--- a/src/main/java/org/traccar/session/ConnectionManager.java
+++ b/src/main/java/org/traccar/session/ConnectionManager.java
@@ -115,7 +115,7 @@ public class ConnectionManager implements BroadcastInterface {
 
     @PreDestroy
     public void shutdown() {
-        offlineGraceScheduler.shutdownNow();
+        offlineGraceScheduler.shutdown();
     }
 
     public DeviceSession getDeviceSession(long deviceId) {
@@ -219,11 +219,7 @@ public class ConnectionManager implements BroadcastInterface {
                             if (existing != null) {
                                 existing.cancel(false);
                             }
-                            // Fetch device now while it's still in cache, so the lambda
-                            // does not fall back to a DB query when the grace period fires.
-                            Device cachedDevice = cacheManager.getObject(Device.class, deviceId);
                             ScheduledFuture<?> task = offlineGraceScheduler.schedule(() -> {
-                                // Atomic check: only proceed if this task was not already cancelled
                                 if (pendingOfflineTasks.remove(deviceId) != null) {
                                     updateDevice(deviceId, Device.STATUS_OFFLINE, null);
                                 }


tananaev:
1. It sounds like you're misusing status notification. To me it seems like what you really want is device inactivity, which is already supported.
2. Please provide links to support this claim:
> This is a commonly requested feature in the community

eosfomogame:
Hi @tananaev, thanks for the feedback.

Regarding device inactivity: I agree it's a valid mechanism, but it doesn't address this specific case. The problem is that certain mobile operators (notably some European and Asian carriers) forcibly terminate TCP connections at regular intervals — typically every 24 hours — after which the device reconnects within a few seconds and resumes sending positions normally. There is zero actual inactivity; the device is working perfectly.

This causes a device_offline → device_online notification pair on every forced TCP reset, which reads as a false alarm to end users and erodes trust in the notification system. I've personally experienced this, and a search of the Traccar community forum shows others have too:
https://www.traccar.org/forums/topic/device-connectdisconnect-every-10-minutes/
https://www.traccar.org/forums/topic/devices-going-constantly-offline-and-online2/
https://www.traccar.org/forums/topic/disconnected-devices/
https://www.traccar.org/forums/topic/st-901-and-st-905-h02-devices-going-offline-and-online-at-regular-intervals/

I fully understand your concern that delaying the offline event changes the semantics of the TCP connection status. That's a fair architectural objection.
I was wandering if you have a better idea to solve this.

tananaev:
Why do the users need connection status notifications? That's typically for debugging only. If you want a notification when device stopped sending data, then inactivity is what you need.

eosfomogame:
thanks!

PR #5839, opened 2026-04-03
Fix: Handle missing alarmName in notification template

kundan175:
### Problem
The notification was showing `$alarmName` as a literal string instead of the actual alarm type. This happened because the template logic failed to resolve `$alarmKey` and `$alarmName` when the expected 'alarm' attribute was missing or when the translation key didn't exist.

### Solution
- Added a fallback check for the `alarms` attribute if `alarm` is not present.
- Added defensive logic to use the raw `alarmKey` if a specific translation is not found in the translation files.
- Provided a default fallback string `"alarm"` to ensure the UI never shows the literal `$alarmName` placeholder.

### Changes
- Modified `templates/notifications/en/alarm.vm` to include these safety checks and fallbacks.

### DIFF
--- a/templates/notifications/en/alarm.vm
+++ b/templates/notifications/en/alarm.vm
@@ -1,6 +1,16 @@
 #set($subject = "$esc.html($device.name): alarm!")
 #set($alarmKey = $event.getString('alarm'))
-#set($alarmName = $translations.get("alarm${alarmKey.substring(0, 1).toUpperCase()}${alarmKey.substring(1)}"))
+#if(!$alarmKey)
+  #set($alarmKey = $event.getString('alarms'))
+#end
+#if($alarmKey)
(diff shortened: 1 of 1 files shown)


tananaev:
What is the `alarms` attribute?

Unfortunately I don't think this fixes anything.

kundan175:
You are right that standard Traccar uses the **alarm** attribute. However, in some cases (potentially custom integrations or misconfigured payloads), if the alarm attribute is missing or misnamed as **alarms**, the current template fails to resolve variables, leaving the raw $alarmName string in the notification.

The main goal of this PR is two-fold:

**Robustness:** It provides fallback logic so that if the translation for a specific alarm key is missing, it displays the raw key instead of the unresolved variable string.
**Default Value:** It ensures that even if no alarm information is found, a generic "alarm" string is shown instead of **$alarmName**.

kundan175:
Screenshot 2026-04-03 at 7 06 51 PM

tananaev:
In my opinion if something is misconfigured, it's fine to keep `$alarmName`.

PR #5847, opened 2026-04-15
Added Pushover sound configuration support

AussieTraccar:
**Purpose**: All Pushover notifications default to the `pushover (default)` sound.

**Solution:** This PR implements user configurable Pushover notification sounds.

### DIFF
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -1312,6 +1312,13 @@ public final class Keys {
             List.of(KeyType.CONFIG));
 
     /**
+     * Pushover notification sound.
+     */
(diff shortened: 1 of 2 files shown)


tananaev:
To be honest I think this is too niche of a configuration. We don't even support it for Firebase.

AussieTraccar:
That is one of the reasons for this PR. Pushover is a good alternative to the Traccar Manager (Firebase) notifications.

My users have provided feedback that they at times miss notifications and want to see an events list for validation. The Pushover client also provides a list of all events and with the sound configuration option, this provides a distinctive alert to them.

I don't see it as being too niche, but rather providing better support for the Pushover protocol.

tananaev:
It is a niche because it's a non-essential configuration option for fairly low used notification channel. Maybe if we can make it more generic and used by other channels, we can include it.

PR #5849, opened 2026-04-16
fix: RFC 8414 §3 path-suffixed well-known URL for path-based issuers

Kalabint:
`WellKnownServlet` only serves `/.well-known/oauth-authorization-server`. 

RFC 8414 §3 requires the path-suffixed form (`/.well-known/oauth-authorization-server/api/oidc`)
when the issuer has a path component. 

Strict clients construct this URL and get a 404.

### DIFF
--- a/src/main/java/org/traccar/web/WellKnownServlet.java
+++ b/src/main/java/org/traccar/web/WellKnownServlet.java
@@ -49,7 +49,7 @@ public class WellKnownServlet extends HttpServlet {
             case "/openid-configuration" -> openIdConfiguration();
             case "/oauth-authorization-server" -> authorizationServerConfiguration();
             case "/oauth-protected-resource" -> protectedResourceConfiguration();
-            default -> null;
+            default -> path.startsWith("/oauth-authorization-server") ? authorizationServerConfiguration() : null;
         };
         if (payload != null) {
             resp.setContentType("application/json");


src/main/java/org/traccar/web/WellKnownServlet.java
```diff
@@ -35,21 +36,36 @@ public class WellKnownServlet extends HttpServlet {
...
         String path = req.getPathInfo();
+        // RFC 8414 §3: for an issuer with a path component (e.g. /api/oidc), the
+        // metadata URL is /.well-known/oauth-authorization-server{issuer-path}.
+        // Accept both the plain path and the issuer-path-suffixed variant.
         Map<String, Object> payload = switch (path) {
             case "/openid-configuration" -> openIdConfiguration();
             case "/oauth-authorization-server" -> authorizationServerConfiguration();
             case "/oauth-protected-resource" -> protectedResourceConfiguration();
-            default -> null;
+            default -> ("/oauth-authorization-server" + issuerPath).equals(path)
```
tananaev: Why can't we just check the prefix?

Kalabint: Yes, we can just check the prefix, the old code was a bit too exact

src/main/java/org/traccar/web/WellKnownServlet.java
```diff
@@ -49,7 +49,7 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IO
             case "/openid-configuration" -> openIdConfiguration();
             case "/oauth-authorization-server" -> authorizationServerConfiguration();
             case "/oauth-protected-resource" -> protectedResourceConfiguration();
-            default -> null;
+            default -> path.startsWith("/oauth-authorization-server") ? authorizationServerConfiguration() : null;
```
tananaev, line 52: Now we just need to combine this with the `/oauth-authorization-server` branch.

Kalabint:
Solved in #5972

PR #5850, opened 2026-04-16
fix: standard Jetty CrossOriginFilter for MCP endpoint

Kalabint:
MCP bypasses Jersey so `CorsResponseFilter` doesn't cover it. 

Adds Jetty's `CrossOriginFilter` on the MCP path when `web.origin` is configured.

`chainPreflight=false` so OPTIONS never hits `McpAuthFilter`.

### DIFF
--- a/src/main/java/org/traccar/web/WebServer.java
+++ b/src/main/java/org/traccar/web/WebServer.java
@@ -20,7 +20,6 @@ import com.google.inject.servlet.GuiceFilter;
 import jakarta.servlet.DispatcherType;
 import jakarta.servlet.SessionCookieConfig;
 import jakarta.servlet.http.HttpServletRequest;
-import org.eclipse.jetty.ee10.servlets.CrossOriginFilter;
 import org.eclipse.jetty.compression.server.CompressionHandler;
 import org.eclipse.jetty.ee10.proxy.AsyncProxyServlet;
 import org.eclipse.jetty.ee10.servlet.FilterHolder;
@@ -33,6 +32,7 @@ import org.eclipse.jetty.http.HttpCookie;
 import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.RequestLogWriter;
 import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.handler.CrossOriginHandler;
 import org.eclipse.jetty.session.DatabaseAdaptor;
 import org.eclipse.jetty.session.DefaultSessionCache;
 import org.eclipse.jetty.session.JDBCSessionDataStoreFactory;
@@ -44,7 +44,6 @@ import org.glassfish.jersey.servlet.ServletContainer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.traccar.LifecycleObject;
-import org.traccar.api.CorsResponseFilter;
 import org.traccar.api.DateParameterConverterProvider;
 import org.traccar.api.ResourceErrorHandler;
 import org.traccar.api.StreamWriter;
@@ -62,7 +61,9 @@ import java.nio.file.Files;
 import java.nio.file.LinkOption;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Duration;
 import java.util.EnumSet;
+import java.util.Set;
 
 public class WebServer implements LifecycleObject {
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/web/WebServer.java
```diff
@@ -176,9 +177,23 @@ private void initApi(ServletContextHandler servletHandler) {
...
+            String webOrigin = config.getString(Keys.WEB_ORIGIN);
+
+            if (webOrigin != null && !webOrigin.isEmpty()) {
+                FilterHolder corsHolder = new FilterHolder(CrossOriginFilter.class);
+                corsHolder.setInitParameter(CrossOriginFilter.ALLOWED_ORIGINS_PARAM, webOrigin);
+                corsHolder.setInitParameter(CrossOriginFilter.ALLOWED_METHODS_PARAM, "GET,POST,DELETE,OPTIONS");
+                corsHolder.setInitParameter(CrossOriginFilter.ALLOWED_HEADERS_PARAM,
+                        "Authorization,Content-Type,Accept");
+                corsHolder.setInitParameter(CrossOriginFilter.CHAIN_PREFLIGHT_PARAM, "false");
+                servletHandler.addFilter(corsHolder, mcpPath, EnumSet.of(DispatcherType.REQUEST));
```
tananaev, line 191: Can we use this filter for everything?

Kalabint: As far as I understand it, yes, we could use it for any requests. Should I open a PR for this, and remove the filter from this PR?

tananaev: Is there something else other than filter in this PR?

Kalabint: I was a bit too fixated on "one PR per scope", but it doesn't really make sense to rescind this and create a new PR then. So, rework the filter to work globally?

tananaev: Yep.

Kalabint: I noticed, that the suggested CrossOriginFilter is deprecated and replaced it with the current CrossOriginHandler, is this ok?

src/main/java/org/traccar/web/WebServer.java
```diff
@@ -176,9 +193,11 @@ private void initApi(ServletContextHandler servletHandler) {
             var mcpServletHolder = new ServletHolder(mcpServerHolder.getServlet());
             mcpServletHolder.setAsyncSupported(true);
             servletHandler.addServlet(mcpServletHolder, McpServerHolder.PATH);
+
+            String mcpPath = McpServerHolder.PATH + "/*";
```
tananaev, line 197: Do we still need this?

src/main/java/org/traccar/web/WebServer.java
```diff
@@ -86,6 +88,7 @@ public WebServer(Injector injector, Config config) throws IOException {
 
         ServletContextHandler servletHandler = new ServletContextHandler(ServletContextHandler.SESSIONS);
         JettyWebSocketServletContainerInitializer.configure(servletHandler, null);
+
```
tananaev, line 91: undo?

src/main/java/org/traccar/web/WebServer.java
```diff
@@ -101,7 +104,21 @@ public WebServer(Injector injector, Config config) throws IOException {
         initClientProxy(servletHandler);
         handlers.addHandler(servletHandler);
         handlers.addHandler(new CompressionHandler());
-        server.setHandler(handlers);
+
+        String webOrigin = config.getString(Keys.WEB_ORIGIN);
+        if (webOrigin != null && !webOrigin.isEmpty()) {
+            CrossOriginHandler corsHandler = new CrossOriginHandler();
```
tananaev, line 110: What about `CorsResponseFilter`? It looks like you haven't removed it?

src/main/java/org/traccar/web/WebServer.java
```diff
@@ -188,7 +207,6 @@ private void initApi(ServletContextHandler servletHandler) {
                 ObjectMapperContextResolver.class,
                 DateParameterConverterProvider.class,
                 SecurityRequestFilter.class,
-                CorsResponseFilter.class,
```
tananaev, line 191: Have you validated that it's working? Can you please provide a screenshot.

Issue #1756, opened 2016-02-27
WARNING: Unable to find permissions manager


johansid:
Hai Anton,

On my wrapper log I got repeated message:
```
INFO   | jvm 1    | 2016/02/27 08:00:53 | Feb 27, 2016 8:00:53 AM org.traccar.web.server.model.BackendRefresher refreshWithReflection
INFO   | jvm 1    | 2016/02/27 08:00:53 | WARNING: Unable to find permissions manager
INFO   | jvm 1    | 2016/02/27 08:14:27 | Feb 27, 2016 8:14:27 AM org.traccar.web.server.model.BackendRefresher refreshWithReflection
INFO   | jvm 1    | 2016/02/27 08:14:27 | WARNING: Unable to find permissions manager
```
**below another logs before above message:**
```
INFO   | jvm 1    | 2016/02/27 07:54:54 | [qtp1741615821-39] WARN / - Key[type=org.traccar.web.server.model.RESTApiServlet, annotation=[none]]: Error during method 'login' call: null
INFO   | jvm 1    | 2016/02/27 07:54:54 | java.lang.reflect.InvocationTargetException
INFO   | jvm 1    | 2016/02/27 07:54:54 | 	at sun.reflect.GeneratedMethodAccessor34.invoke(Unknown Source)
INFO   | jvm 1    | 2016/02/27 07:54:54 | 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
INFO   | jvm 1    | 2016/02/27 07:54:54 | 	at java.lang.reflect.Method.invoke(Method.java:606)
INFO   | jvm 1    | 2016/02/27 07:54:54 | 	at org.traccar.web.server.model.RESTApiServlet.makeRestCall(RESTApiServlet.java:110)
INFO   | jvm 1    | 2016/02/27 07:54:54 | 	at org.traccar.web.server.model.RESTApiServlet.doPost(RESTApiServlet.java:63)
INFO   | jvm 1    | 2016/02/27 07:54:54 | 	at javax.servlet.http.HttpServlet.service(HttpServlet.java:707)
INFO   | jvm 1    | 2016/02/27 07:54:54 | 	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
INFO   | jvm 1    | 2016/02/27 07:54:54 | 	at com.google.inject.servlet.ServletDefinition.doService(ServletDefinition.java:263)
[...]

PR #5852, opened 2026-04-16
feat: deferred OIDC auth for unauthenticated authorize requests

Kalabint:
`/api/oidc/authorize` without a session currently calls `issueCode(userId=0)`.
Instead: park PKCE params server-side, set an httpOnly cookie, redirect to `/`.

Custom `/authorize/resume` endpoint issues the real code after the user
authenticates. No spec equivalent — the standard doesn't address deferred
auth for SPAs that intercept the authorize redirect.

Pair with traccar-web PR [#1756](https://github.com/traccar/traccar-web/pull/1756)


tananaev:
What is the use case exactly?

Kalabint:
This solves the 401 error from [Suggestion 1](https://www.traccar.org/forums/topic/mcp-function-how-can-the-documentation-can-found/#post-117737)

tananaev:
How does it work in a standard?

Kalabint:
As far as I can tell, the OIDC Standard isn't defining this part, just the following:

> > **3.1.2.3. Authorization Server Authenticates End-User**
>
> If the request is valid, the Authorization Server attempts to Authenticate the End-User or determines whether the End-User is Authenticated, depending upon the request parameter values used. The methods used by the Authorization Server to Authenticate the End-User (e.g., username and password, session cookies, etc.) are beyond the scope of this specification. **_An Authentication user interface MAY be displayed by the Authorization Server_**, depending upon the request parameter values used and the authentication methods used.
>
> The Authorization Server MUST attempt to Authenticate the End-User in the following cases:
> - The End-User is not already Authenticated.
> - The Authentication Request contains the `prompt` parameter with the value `login`. In this case, the Authorization Server MUST reauthenticate the End-User even if the End-User is already authenticated.
>
> The Authorization Server MUST NOT interact with the End-User in the following case:
> - The Authentication Request contains the `prompt` parameter with the value `none`. In this case, the Authorization Server MUST return an error if an End-User is not already Authenticated or could not be silently Authenticated.
>
> When interacting with the End-User, the Authorization Server MUST employ appropriate measures against Cross-Site Request Forgery and Clickjacking as described in Sections 10.12 and 10.13 of OAuth 2.0 [RFC6749].

Source: https://openid.net/specs/openid-connect-core-1_0.html#Authenticates

I tested it with my own Mapping Application and Claude Connector.

tananaev:
Then why do we need `/authorize/resume`?

Kalabint:
No, we don't and its not a good way to do it. It seems I was coding up a custom solution, based on my own OIDC Traccar fixes, which influenced the path forward.

Reworked the whole to a stateless JWT Cookie based system (had a look how other IdP work), which can also work on Load balanced traccar system.

I hope this is better

tananaev:
This still looks overcomplicated. Why can't we just save values in the current session?

Kalabint:
Reworked to have it work without any cookies, and removed the need for server storage/logic entirely.

src/main/java/org/traccar/api/resource/OidcResource.java
```diff
@@ -79,7 +80,7 @@ public Response authorize(
             @QueryParam("response_type") String responseType,
             @QueryParam("code_challenge") String codeChallenge,
             @QueryParam("code_challenge_method") String codeChallengeMethod,
-            @QueryParam("nonce") String nonce) {
+            @QueryParam("nonce") String nonce) throws GeneralSecurityException, JOSEException, StorageException {
```
tananaev, line 83: Do we still need these?

src/main/java/org/traccar/api/resource/OidcResource.java
```diff
@@ -91,6 +92,28 @@ public Response authorize(
             throw new WebApplicationException(Response.Status.BAD_REQUEST);
         }
 
+        if (getUserId() == 0) {
```
tananaev, line 95: Can you please update description to explain how this implementation works.

PR #5858, opened 2026-04-20
Fix NPE in OpenID login when user has no groups

### LINKED ISSUE #5856: OpenID login fails with NPE when user has no groups in token
## Description

OpenID login fails with a NullPointerException when a user does not have any group memberships in their OpenID token.

### Root Cause

In `OpenIdProvider.java`, the `authenticate()` method calls `userGroups.contains()` without checking if `userGroups` is null:

```java
boolean administrator = adminGroup != null ? userGroups.contains(adminGroup) : null;
```

If a user doesn't have any groups in their OpenID token, `userGroups` will be null, causing an NPE.

### Fix

Add null checks for `userGroups` before calling `contains()`:

```java
boolean administrator = adminGroup != null && userGroups != null && userGroups.contains(adminGroup);

if (!(administrator || allowGroup == null || (userGroups != null && userGroups.contains(allowGroup)))) {
    throw new GeneralSecurityException("Your OpenID Groups do not permit access");
}
```

### Additional Improvement

Also adds error handling in `SessionResource.java` to redirect users to an error page (`?openid=error&message=...`) when exceptions occur during the OpenID callback, rather than crashing.

rossigee:
Fixes a NullPointerException when authenticating via OpenID if the user has no group memberships in their token.

### OpenIdProvider.java
- Add null check for userGroups before calling contains() in admin group check
- Add null check for userGroups before calling contains() in allow group check

### SessionResource.java
- Add error handling to redirect to error page when exceptions occur during OpenID callback
- Added getBaseUrl() method to OpenIdProvider to support error redirects

Fixes #5856


src/main/java/org/traccar/database/OpenIdProvider.java
```diff
@@ -174,9 +174,9 @@ public URI handleCallback(String queryParameters, HttpServletRequest request)
         UserInfo userInfo = getUserInfo(bearerToken);
 
         List<String> userGroups = userInfo.getStringListClaim(groupsClaimName);
-        Boolean administrator = adminGroup != null ? userGroups.contains(adminGroup) : null;
+        boolean administrator = adminGroup != null && userGroups != null && userGroups.contains(adminGroup);
```
tananaev, line 177: This should stay and nullable value. Otherwise it will reset admin flag on the acount.

src/main/java/org/traccar/api/resource/SessionResource.java
```diff
@@ -180,11 +181,17 @@ public Response openIdAuth() {
...
             throw new UnsupportedOperationException("OpenID not enabled");
         }
-        return Response.seeOther(openIdProvider.handleCallback(request.getQueryString(), request)).build();
+        try {
+            return Response.seeOther(openIdProvider.handleCallback(request.getQueryString(), request)).build();
+        } catch (GeneralSecurityException e) {
+            URI errorUri = openIdProvider.getBaseUrl().resolve("?openid=error&message=" +
+                    java.net.URLEncoder.encode(e.getMessage(), java.nio.charset.StandardCharsets.UTF_8));
+            return Response.seeOther(errorUri).build();
+        }
```
tananaev, line 194: Not sure about this part. Let's revert it for now.

src/main/java/org/traccar/database/OpenIdProvider.java
rossigee: I see. It's more complicated than it looks. Maybe best to put it in functions for better readability/testability.

tananaev: What is complicated about it?

Issue #5810, opened 2026-03-06
protocol x25


EmanuelANDR:
alguém ja conseguiu fazer este rastreador receber o ID DO MOTORISTA ?
[XT40-TM Protocol rev1.05 (1).pdf](https://github.com/user-attachments/files/25785171/XT40-TM.Protocol.rev1.05.1.pdf)

tananaev:
> Has anyone managed to get this tracker to receive the DRIVER ID?

PR #5837, opened 2026-03-30
Add support for AOVX VL100 model

jcardus:
[AOVX VL100 Protocol.pdf](https://github.com/user-attachments/files/26362944/AOVX.VL100.Protocol.pdf)

PR #5860, opened 2026-04-25
correct battery voltage for x3tech xt40 devices

ivanfmartinez:
x3tech XT40 devices uses almost same protocol of NT20, but have battery voltage that cannot be direct converted to percent using 6V as maximum level.

apparently all other data are correct using the same decoding of NT20.

This PR define KEY_BATTERY instead of KEY_BATTERY_LEVEL for devices defined with model XT40

Protocol definition already attached on [#5810](https://github.com/traccar/traccar/issues/5810) - https://github.com/user-attachments/files/25785171/XT40-TM.Protocol.rev1.05.1.pdf

```
5.5.1.20. Battery Voltage
Battery voltage, value with one house multiplied by 10
```

### DIFF
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
@@ -237,7 +237,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
             case MSG_LBS_ALARM:
                 return true;
             case MSG_GPS_LBS_2:
-                return model != null && Set.of("NT20", "VL100").contains(model.toUpperCase());
+                return model != null && Set.of("NT20", "VL100","XT40").contains(model.toUpperCase());
             case 0xA3: // MSG_FENCE_SINGLE / MSG_STATUS_3
                 return variant == Variant.SEEWORLD;
             default:
@@ -479,7 +479,8 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
         String model = deviceSession != null ? getDeviceModel(deviceSession) : null;
         boolean modelLW = model != null && model.toUpperCase().startsWith("LW");
         boolean modelSW = "SEEWORLD".equalsIgnoreCase(model);
-        boolean modelNT = model != null && Set.of("NT20", "VL100").contains(model.toUpperCase());
+        boolean modelXT = "XT40".equalsIgnoreCase(model);
+        boolean modelNT = modelXT || (model != null && Set.of("NT20", "VL100").contains(model.toUpperCase()));
         boolean modelVL = model != null && Set.of("VL103", "LL303", "VL512", "G18").contains(model.toUpperCase());
 
         if (type == MSG_LOGIN) {
@@ -876,13 +877,18 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
                         position.set(Position.KEY_POWER, buf.readUnsignedShort() * 0.01);
                     } else {
                         int battery = buf.readUnsignedByte();
-                        if (modelNT && type == MSG_GPS_LBS_2) {
-                            battery = battery / 10;
-                        }
-                        if (battery <= 6) {
-                            position.set(Position.KEY_BATTERY_LEVEL, battery * 100 / 6);
-                        } else if (battery <= 100) {
-                            position.set(Position.KEY_BATTERY_LEVEL, battery);
+                        if (modelXT && type == MSG_GPS_LBS_2) {
+                            position.set(Position.KEY_BATTERY, battery / 10.0);
+                        } else {
+                            if (modelNT && type == MSG_GPS_LBS_2) {
+                                battery = battery / 10;
+                            }
+
+                            if (battery <= 6) {
+                                position.set(Position.KEY_BATTERY_LEVEL, battery * 100 / 6);
+                            } else if (battery <= 100) {
+                                position.set(Position.KEY_BATTERY_LEVEL, battery);
+                            }
                         }
                     }
                     position.set(Position.KEY_RSSI, buf.readUnsignedByte());
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
@@ -237,7 +237,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
             case MSG_LBS_ALARM:
                 return true;
             case MSG_GPS_LBS_2:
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -237,7 +237,7 @@ private boolean hasStatus(int type, String model) {
             case MSG_LBS_ALARM:
                 return true;
             case MSG_GPS_LBS_2:
-                return model != null && Set.of("NT20", "VL100").contains(model.toUpperCase());
+                return model != null && Set.of("NT20", "VL100","XT40").contains(model.toUpperCase());
```
tananaev:
```suggestion
                return model != null && Set.of("NT20", "VL100", "XT40").contains(model.toUpperCase());
```

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -479,7 +479,8 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
         String model = deviceSession != null ? getDeviceModel(deviceSession) : null;
         boolean modelLW = model != null && model.toUpperCase().startsWith("LW");
         boolean modelSW = "SEEWORLD".equalsIgnoreCase(model);
-        boolean modelNT = model != null && Set.of("NT20", "VL100").contains(model.toUpperCase());
+        boolean modelXT = "XT40".equalsIgnoreCase(model);
+        boolean modelNT = modelXT || (model != null && Set.of("NT20", "VL100").contains(model.toUpperCase()));
```
tananaev, line 483:
```suggestion
        boolean modelNT = model != null && Set.of("NT20", "VL100", "XT40").contains(model.toUpperCase());
```

src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
```diff
@@ -876,13 +877,18 @@ private Object decodeBasic(Channel channel, SocketAddress remoteAddress, ByteBuf
...
                     } else {
                         int battery = buf.readUnsignedByte();
-                        if (modelNT && type == MSG_GPS_LBS_2) {
-                            battery = battery / 10;
-                        }
-                        if (battery <= 6) {
-                            position.set(Position.KEY_BATTERY_LEVEL, battery * 100 / 6);
-                        } else if (battery <= 100) {
-                            position.set(Position.KEY_BATTERY_LEVEL, battery);
+                        if (modelXT && type == MSG_GPS_LBS_2) {
```
tananaev, line 880: Why do we need a special case here?

ivanfmartinez: because the battery came as voltage and cannot be directly converted to percent using simple * 100 / 6 because lithium batteries have SoC based on curve and not by direct 0-100% conversion.

tananaev: Correct, but we already handle that case.

ivanfmartinez: > Correct, but we already handle that case.

where exactly ? 
This nt devices processing only set key_battery_level which is percentage

ivanfmartinez: and the nt device existing processing divide using integer truncating the value.

tananaev: Truncating value is by design. Battery level is not accurate enough to give you anything meaningful below 1%.

ivanfmartinez: You are wrong in this case at least. The documentation which I linked in description of this PR, and copied the relevant part indicates that field sends Voltage of battery with one decimal place. 

These lithium batteries normally range from 3.x to 4.x. The current NT code divides the number by 10 and truncate, which will give results of 3 and 4 for this case. After that the <= 6 if, it will transform these values to a percent considering maximum value of 6. This will give only 50% and 66% values for battery level.

The current code that does this :
https://github.com/traccar/traccar/pull/5860/changes#diff-4659e05c0ba6e4ba0939c431f33ddc4f3677a982dce0d95f68f1c7f1bbd81618R883-R891


After seeing this consistent 50-66% values I have checked the documentation and found the wrong processing of battery voltage data based on protocol documentation. I made a PR that keeps exactly the same behavior for who is using NT20 model and changes only behavior for who set the modem to XT40. I have also added a test to confirm the NT20 code keeps with same behavior before making my changes.

Errors are summed after each processing of data, your saying that original value does not give accuracy, and by removing the decimal place for the cases where value is less than 6 even for NT20 are making the precision worst.

Docs that I have read on other PRs about the NT20 protocol https://github.com/traccar/traccar/pull/5837 and https://github.com/traccar/traccar/pull/5474   also indicates the voltage and examples with 3.x and 4.x Volts, but as this code is already running for many users I have preferred to keep without changing that part of code and creating an specific model and dont change existing behavior without having other devices to test.

If you prefer, I can just use existing code block for NT and keep KEY_BATTERY_LEVEL as is and just add the KEY_BATTERY (with one decimal precision).  this can be better for existing NT20 / VL100 users which will get new voltage value as their docs also indicate. For me at least I have no problem getting an wrong value if I get the correct one also, but for now  with one 50 and 66% I cannot know how really charged is the battery.

ivanfmartinez: @tananaev what you think about this other approach ?

https://github.com/traccar/traccar/compare/master...ivanfmartinez:traccar:master_gt06_XT40_v2

will keep old behavior of percentage (that at least for XT40 does not work correctly), and add the direct data for BATTERY VOLTAGE as defined on protocol documentation.

tananaev: Looks like you're right. It was incorrectly implemented in https://github.com/traccar/traccar/pull/5474. It should have been battery voltage. It also looks like there are more device models we should add: NT11, NT20, NT26, NT40, NT46. We should just do if/else. Something like this:
```
int battery = buf.readUnsignedByte();
if (modelNT && type == MSG_GPS_LBS_2) {
    bposition.set(Position.KEY_BATTERY, battery / 10.0);
} else if (battery <= 6) {
    position.set(Position.KEY_BATTERY_LEVEL, battery * 100 / 6);
} else if (battery <= 100) {
    position.set(Position.KEY_BATTERY_LEVEL, battery);
}
```

ivanfmartinez:
Closing this after discussion in favor of new simplified PR on  https://github.com/traccar/traccar/pull/5862

PR #5862, opened 2026-04-25
correct battery voltage decoding for gt06 nt devices and add more device models

ivanfmartinez:
This PR replaces original idea of creating a separate support for XT40 https://github.com/traccar/traccar/pull/5860

this makes a breaking change for NT devices protocol removing KEY_BATTERY_LEVEL and adding KEY_BATTERY with received battery voltage.


src/test/java/org/traccar/protocol/Gt06ProtocolDecoderTest.java
```diff
@@ -585,6 +585,25 @@ public void testDecode() throws Exception {
...
+
+        verifyAttribute(decoder, binary(
+                "78783c220109999999999999991a04171433101a041714310fb00288a5a8050754500008000800000000000000000604e32933000200468b000025008d87ce0d0a"),
+                Position.KEY_POWER, 12.51);
+        verifyAttribute(decoder, binary(
+                "78783c220109999999999999991a04171433101a041714310fb00288a5a8050754500008000800000000000000000604e32933000200468b000025008d87ce0d0a"),
+                Position.KEY_BATTERY, 4.1);
+        verifyAttribute(decoder, binary(
+                "78783c220109999999999999991a04171433101a041714310fb00288a5a8050754500008000800000000000000000604e32933000200468b000025008d87ce0d0a"),
+                Position.KEY_BATTERY_LEVEL, null);
```
tananaev: I think we probably don't need all of these tests. I think just power and battery is enough. And we don't need both models as the decoding logic is exactly the same.

ivanfmartinez: I have put the X40 block with messages from my device, at this moment I just found this problem and still analyzing more data. 

I have just removed the IMEI from my packets and realized that is no CRC check on the code, as the messages came using TCP the packets are already checked, but if they are broken before entering TCP mode the decoder will not detect.

The tests are to make sure if some future commit change the behavior someone will check in detail.

But if you don't like I can remove.

tananaev: Let's just keep the two tests.

PR #5863, opened 2026-04-27
added obd ev parameters for teltonika 

ivanfmartinez:
Based on documentation :  https://wiki.teltonika-gps.com/view/OBDII_EV_Parameters

I got most of then from my Bolt EV using FMC003, some attributes are only available on specific cars https://wiki.teltonika-gps.com/view/OBD_supported_vehicle_list

Don't know if these parameters can came on other models that support bluetooth obd connection, for now it used a predicate to match the listed obd devices on teltonika website.

High Voltage battery parameters with **hv** prefix and generic ev/hybrid with **ev** prefix. 

Distance Remaining apparently can be applied also for non electrified cars

I have more parameters to add that came later on other PR, but are not part of this OBDII EV parameters.


src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -195,6 +195,7 @@ private static void register(int id, Predicate<String> predicate, BiConsumer<Pos
     static {
         Predicate<String> any = (m) -> true;
         Predicate<String> fmbXXX = (m) -> m != null && m.matches("FM[B-Z]...|MTB100|MSP500");
+        Predicate<String> obd = (m) -> m != null && m.matches("FM[BCM]...");
```
tananaev: Where did you get this expression from? I see conflicting information from the docs that you linked.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -280,8 +279,23 @@ private static void register(int id, Predicate<String> predicate, BiConsumer<Pos
         register(175, fmbXXX, (p, b) -> {
             p.addAlarm(b.readUnsignedByte() > 0 ? Position.ALARM_GEOFENCE_ENTER : Position.ALARM_GEOFENCE_EXIT);
         });
+        register(410, obd, (p, b) -> p.set("evCharging", b.readUnsignedByte()));
```
tananaev: Why is there an "ev" prefix?

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -280,8 +279,23 @@ private static void register(int id, Predicate<String> predicate, BiConsumer<Pos
...
+        register(1378, obd, (p, b) -> p.set("hvMinCellVoltage", b.readUnsignedShort() * 0.01));
+        register(1379, obd, (p, b) -> p.set("hvMaxCellVoltage", b.readUnsignedShort() * 0.01));
+        register(1380, obd, (p, b) -> p.set("hvCurrent", b.readUnsignedShort() * 0.1));
+        register(1381, obd, (p, b) -> p.set("hvVoltage", b.readUnsignedShort() * 0.1));
+        register(1382, obd, (p, b) -> p.set("hvCapacity", b.readUnsignedShort() * 0.01));
+        register(1383, obd, (p, b) -> p.set("hvEnergy", b.readUnsignedShort() * 0.01));
+        register(1386, obd, (p, b) -> p.set("hvMinCellTemp", b.readShort() * 0.1));
+        register(1387, obd, (p, b) -> p.set("hvMaxCellTemp", b.readShort() * 0.1));
+        register(1388, obd, (p, b) -> p.set("hvLifetimeEnergyUsed", b.readInt()));
+        register(1389, obd, (p, b) -> p.set("hvLifetimeEnergyCharged", b.readUnsignedInt()));
```
tananaev: Why do we need HV prefix?

ivanfmartinez: is data from high voltage battery. I used same indication that came on the AVL doc.

tananaev: HV just mean high voltage. We're missing battery. So the abbreviation in this case makes little sense. But I would also just call it battery. We almost always prefer full words.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
ivanfmartinez: not device / tracker charging but the EV/HYBRID CAR is charging. The current KEY_CHARGE is used to indicate when tracker is charging. We can change this, but I think that we cannot use the KEY_CHARGE

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
ivanfmartinez: The docs/wiki Teltonika are not exact (I already found some attributes that came and are not listed for other case).

Apparently FM.00. are all devices that connect on the OBD port and can get data. 
FMB old series FMC/FMM new devices with LTE support

Some of the FMB devices, with other numeric codes, support can or bluetooth connection to get data. 
I did not find these IDs conflicting on the lists that I have checked. Then I use not so restrictive expression. 

https://wiki.teltonika-gps.com/view/How_to_connect_OBD_II_Bluetooth_Dongle_to_FMB_device

ivanfmartinez: If you think is better to limit to exactly what is on the URL **(FMB001, FMx003, FMx00A)** I can use **(FMB001|FM.00[3A])**

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
ivanfmartinez: just battery also is not clear, as the EV/Hybrid hormally have the Low Voltage Battery (12-24V) and the High Voltage Battery (over 300V). And in this case with the tracker we have also the Tracker battery.

What is your suggestion for this case ?

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
tananaev: Let's make it more strict for now.

tananaev: Also need some better variable name because OBD can be supported by other devices on other IDs.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
tananaev: Maybe we can call everything "ev"?

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -188,6 +188,7 @@ private static void register(int id, Predicate<String> predicate, BiConsumer<Pos
     static {
         Predicate<String> any = (m) -> true;
         Predicate<String> fmbXXX = (m) -> m != null && m.matches("FM[B-Z]...|MTB100|MSP500");
+        Predicate<String> obdEV = (m) -> m != null && m.matches("(FMB001|FM.00[3A])");
```
tananaev, line 191: I think the naming is inconsistent. We need to find something better.

Also don't forget to re-request review when you think it's ready for review.

ivanfmartinez: @tananaev the parameters are defined on Teltonika wiki as OBDII EV Parameters.


In my mind using they as the manufacturer defines is best for checking with official information.

But I can use different name, what you want it to be ?

tananaev: Parameters might be called OBD EV, but the grouping is by device models. By the way, the wiki page you originally linked doesn't seem to work anymore.

ivanfmartinez: Just checked and apparently they have moved the parameters inside the devices definition

https://wiki.teltonika-gps.com/view/FMC003_Teltonika_Data_Sending_Parameters_ID

There is an OEM EV block now that I did not remember exists before (last edit was after this PR created but history changes details is not available).

In this case the group means device models that support OBD EV parameters, not just a model group that support features because some other devices that support OBD apparently does not support this parameters.

If you want I can make a comment to make it clear and in future if we discover other condition the name can be easily refactored.

tananaev: Maybe we can call it `fmbEV` or something like that?

PR #5876, opened 2026-05-11
Update Hyn600ProtocolDecoder.java

tjeerd13:
Fix incorrect device time decoding for Hyn600 status reports

Some Hyn600 RPT packets (e.g. status/event type 2) do not include a generated
time field, but may still contain trailing bytes. These bytes were previously
decoded unconditionally as deviceTime, resulting in invalid dates (e.g. year
> 16000) and MySQL datetime truncation errors.

The decoder now only reads generated time for event reports where it is
semantically defined (event type 7), and ensures deviceTime is set to fixTime
as a fallback when no generated time is present.

This prevents invalid timestamps while preserving correct handling of crash
and event reports.


tananaev:

tjeerd13:
https://www.traccar.org/forums/topic/assistant-telematics-avt20/#post-119582

tananaev:
This is the right fix - https://github.com/traccar/traccar/commit/764e65719a6cbd3c41e4c1c5ff5f08c997992b4a.

PR #5884, opened 2026-05-20
Add basic .editorconfig

timbze:
My code editor has other defaults, this would help to easily have defaults consistent with the codebase.

Is this something you want to add?


tananaev:
It makes sense to add, but you should split backend and web configs.

timbze:
Fixed now. Will make PR on web repo too.

PR #5886, opened 2026-05-23
add:omni

shm379:
This pull request introduces support for the new "Omni" protocol in the codebase, including all necessary configuration, protocol logic, and supporting constants. The most significant changes are the addition of protocol encoder/decoder logic, protocol registration, and new alarm constants.

**Omni Protocol Support**

* Added a new `OmniProtocol` class that registers the protocol, sets up supported commands, and configures the Netty pipeline for message handling. (`src/main/java/org/traccar/protocol/OmniProtocol.java`)
* Added a new `OmniProtocolEncoder` class that implements encoding logic for a wide range of commands supported by the Omni protocol, including custom command parsing, command normalization, and value extraction. (`src/main/java/org/traccar/protocol/OmniProtocolEncoder.java`)

**Configuration Updates**

* Registered the Omni protocol port (`5267`) in the `PortConfigSuffix` configuration map. (`src/main/java/org/traccar/config/PortConfigSuffix.java`)

**Model Extensions**

* Added new alarm type constants to `PositionExt` for controller failure, communication failure, and motor hall failure. (`src/main/java/org/traccar/model/PositionExt.java`)


tananaev:
What is Omni protocol? You have to provide documentation and more context about the devices.

shm379:
> What is Omni protocol? You have to provide documentation and more context about the devices.

https://www.omnismartiot.com/

shm379:
https://www.omnismartiot.com/Sharing-Bikes/bike-share-m136-iot-device.html

tananaev:
I don't see any protocol documentation there.

shm379:
[Omni  IoT device TCP interface protocol - V133-2_260505_111953.pdf](https://github.com/user-attachments/files/28184861/Omni.IoT.device.TCP.interface.protocol.-.V133-2_260505_111953.pdf)

tananaev:
This protocol is already supported, as far as I know. If you want to extend it, you should add to existing one instead of creating a completely new decoder.

shm379:
This protocol is already supported, as far as I know. If you want to extend it, you should add to existing one instead of creating a completely new decoder.

tananaev:
What?

shm379:
Sorry, I mean that the existing supported protocol you are referring to seems to be `xrb28`, but the Omni protocol is not fully compatible with it in practice.

It may look similar in some parts, but the current `xrb28` decoder does not work with the real Omni device messages I tested. The message format, commands, responses, and some fields are different enough that the device is not parsed correctly.

My intention was not to duplicate an existing protocol unnecessarily. I added a new decoder because the existing implementation does not handle these Omni devices correctly.

If you prefer, I can rework this PR to extend the existing `xrb28` decoder instead of creating a separate `OmniProtocolDecoder`, but we need to add the missing Omni-specific parsing and command handling because the current support is not working for these devices.

tananaev:
As I said, adding it to existing decoder is absolutely the right way to go. I would also strongly recommend smaller PRs. Split it into smaller pieces.

PR #5889, opened 2026-05-31
Lazy AMQP connection and forwarding error logging

mutasimissa:
AMQP is the only forwarder that connects synchronously in its constructor. If the broker is unavailable, `factory.newConnection()` throws and the `RuntimeException` propagates through Guice provisioning, causing `System.exit(1)` — Traccar fails to start entirely.

Kafka, MQTT and JSON forwarders don't have this issue because they connect lazily or asynchronously.

This change defers the AMQP connection to the first `publishMessage()` call and adds `channel.isOpen()` checks so forwarding resumes automatically if the broker restarts. Follows the same lazy-connect pattern used in `NetworkForwarder`.

Also adds `LOGGER.warn` to `EventForwarderAmqp` and `PositionForwarderAmqp` for forwarding errors, matching existing logging in `NetworkForwarder` and `PositionForwardingHandler`.


mutasimissa:
@tananaev Hi, this PR has been open for quite some time without any maintainer activity. Could you please confirm whether it is still being considered, or if any changes are required from my side? thanks :)

src/main/java/org/traccar/forward/AmqpClient.java
```diff
@@ -29,32 +32,49 @@
...
+        channel.exchangeDeclare(exchange, BuiltinExchangeType.TOPIC, true);
     }
 
-    public void publishMessage(String message) throws IOException {
-        channel.basicPublish(
-                exchange, topic, MessageProperties.PERSISTENT_TEXT_PLAIN, message.getBytes(StandardCharsets.UTF_8));
+    public synchronized void publishMessage(String message) throws IOException {
+        if (channel == null || !channel.isOpen()) {
+            try {
+                connect();
```
tananaev, line 84: I think resource management is broken here. We need to close both connection and channel. You're basically setting channel to null and don't even do anything with connection.

mutasimissa: That's right will update this!

mutasimissa:
Fixed — Connection is now stored as a field and both channel and connection are properly closed before reconnecting.

src/main/java/org/traccar/forward/AmqpClient.java
```diff
@@ -29,32 +32,70 @@
...
-            Connection connection = factory.newConnection();
-            channel = connection.createChannel();
-            channel.exchangeDeclare(exchange, BuiltinExchangeType.TOPIC, true);
-        } catch (IOException | TimeoutException e) {
-            throw new RuntimeException("Error while creating and configuring RabbitMQ channel", e);
+    private void disconnect() {
+        if (channel != null) {
+            try {
+                channel.close();
+            } catch (IOException | TimeoutException e) {
```
tananaev: Why do we need `TimeoutException` here?

src/main/java/org/traccar/forward/AmqpClient.java
```diff
@@ -29,32 +32,70 @@
...
+                disconnect();
+                LOGGER.warn("AMQP connection error", e);
+                return;
+            }
+        }
+        try {
+            channel.basicPublish(
+                    exchange, topic, MessageProperties.PERSISTENT_TEXT_PLAIN, message.getBytes(StandardCharsets.UTF_8));
+        } catch (IOException e) {
```
tananaev: Do we need separate catches?

src/main/java/org/traccar/forward/AmqpClient.java
```diff
@@ -29,32 +32,70 @@
...
-        channel.basicPublish(
-                exchange, topic, MessageProperties.PERSISTENT_TEXT_PLAIN, message.getBytes(StandardCharsets.UTF_8));
+    public synchronized void publishMessage(String message) throws IOException {
+        if (channel == null || !channel.isOpen()) {
+            try {
+                connect();
+            } catch (IOException | TimeoutException e) {
+                disconnect();
+                LOGGER.warn("AMQP connection error", e);
+                return;
```
tananaev: This doesn't look correct.

src/main/java/org/traccar/forward/AmqpClient.java
```diff
@@ -29,32 +32,70 @@
...
+                LOGGER.warn("AMQP channel close error", e);
+            }
+            channel = null;
         }
+        if (connection != null) {
+            try {
+                connection.close();
+            } catch (IOException e) {
+                LOGGER.warn("AMQP connection close error", e);
+            }
```
tananaev, line 63: Maybe we can reuse this part?

mutasimissa: Deduped — both close paths now call the shared `closeQuietly(AutoCloseable)` helper.

src/main/java/org/traccar/forward/AmqpClient.java
mutasimissa: Good catch — the silent `return` swallowed connect failures so callers never reported them. Now `publishMessage` always throws on failure, so `EventForwarderAmqp` / `PositionForwarderAmqp` propagate it via `resultHandler.onResult(false, e)`.

src/main/java/org/traccar/forward/AmqpClient.java
mutasimissa: Merged into a single `try` around connect + publish — both failure modes now share one `catch` that disconnects and rethrows.

src/main/java/org/traccar/forward/AmqpClient.java
mutasimissa: Removed — the close paths now go through a shared `closeQuietly(AutoCloseable)` helper that catches `Exception`, so `TimeoutException` no longer needs to be named.

PR #5890, opened 2026-05-31
Added C5 and C5L model specific handling

nilei:
## Summary

Extend JT808 custom command support for C5 and C5L devices.

Previously, text-based custom commands were only encoded as JT808 text messages (`0x8300`) for the BSJ model. This change adds support for C5 and C5L devices, which use the same command transport mechanism.

## Changes

* Added `C5` and `C5L` to the JT808 encoder text message model list.
* Custom commands for `BSJ`, `C5`, and `C5L` are now encoded using JT808 message type `0x8300` (Send Text Message).
* Added a unit test covering all three models to ensure they follow the same encoding path.

## Test Coverage

Added a test that verifies:

* `BSJ`, `C5`, and `C5L` use the same encoder branch.
* Custom commands are encoded as JT808 text messages (`0x8300`).
* The command payload is prefixed with the required flag byte (`0x01`).
* The command text is correctly included in the encoded message.

Example command used in the test:

`Test, Command, 123#`

## Motivation

C5 and C5L devices accept SMS-style text commands through the JT808 text message mechanism, matching the existing BSJ implementation. This change enables command delivery without requiring raw hexadecimal payloads and keeps behavior consistent across supported device models.


src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
```diff
@@ -1,12 +1,47 @@
...
 public class Jt808ProtocolEncoderTest extends ProtocolTest {
 
+    @Test
+    public void testEncodeCustomTextMessageModels() throws Exception {
+
+        var encoder = inject(new Jt808ProtocolEncoder(null));
+        var decoder = inject(new Jt808ProtocolDecoder(null));
+        var channel = new EmbeddedChannel(decoder);
+
+        for (String model : List.of("BSJ", "C5", "C5L")) {
```
tananaev: this doesn't make sense

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
```diff
@@ -1,12 +1,47 @@
...
+import java.util.List;
+
 public class Jt808ProtocolEncoderTest extends ProtocolTest {
 
+    @Test
+    public void testEncodeCustomTextMessageModels() throws Exception {
+
+        var encoder = inject(new Jt808ProtocolEncoder(null));
+        var decoder = inject(new Jt808ProtocolDecoder(null));
+        var channel = new EmbeddedChannel(decoder);
```
tananaev: This is not how we write encoder tests.

nilei:
Good morning Anton. I spent a long time searching through the xxxProtocolEncoderTest.java files to find an example of how to test the encoding for a specific model. Unfortunately, I didn't really have any luck. So this is the best I could come up with. Do you have a test I can use as an example to better align with "how we write encoder tests"?

tananaev:
Literally any other encoder test.

nilei:
Hi Anton. 

I’ve first identified the encoders that are tested for the `TYPE_CUSTOM` command type. I used the `UlbotechProtocolEncoderTest` and the `RuptelaProtocolEncoderTest` as an example. However, it doesn’t check for a specific model or anything like that.

-- Cheers, Nils

tananaev:
I'm pretty sure there are encoder tests with specific model override.

nilei:
Hi Anton.

I did a `grep -Ruin setModelOverride . |grep Encoder` and found one (1) other Encoder test that uses it: `Minifinder2ProtocolEncoderTest`.

What I still don't understand is how I can verify the result with `verifyCommand` if I don't know the `uniqueId`. Or is there a fixed `uniqueId` for the device with ID 1?

-- Cheers, Nils

tananaev:
There is a mocked id. Please take your time and carefully study how existing tests work.

nilei:
Hi Anton,

I've been trying to figure out the code for a really long time now, but I just can't figure out what I'm doing wrong. In fact, the only test that was already in the JT808 Encoder Test to test `TYPE_ENGINE_STOP` is set to "Disabled." When I set it to "Enable," I get the same error message as in my test:

```
Jt808ProtocolEncoderTest > testEncode() FAILED
    org.opentest4j.AssertionFailedError at Jt808ProtocolEncoderTest.java:20

```
```
org.opentest4j.AssertionFailedError: buffer is null ==> expected: not <null>
	at org.junit.jupiter.api.Assertions.assertNotNull(Assertions.java:345)
	at org.traccar.ProtocolTest.verifyFrame(ProtocolTest.java:352)
	at org.traccar.ProtocolTest.verifyCommand(ProtocolTest.java:348)
	at org.traccar.protocol.Jt808ProtocolEncoderTest.testEncode(Jt808ProtocolEncoderTest.java:20)
```

What else am I doing wrong?

-- Cheers, Nils

nilei:
Hi Anton.

I've made some progress. I've noticed that some encoders are written like this:

`protected Object encodeCommand(Channel channel, Command command) {`

and some like this:

`protected Object encodeCommand(Command command) {`

However, `verifyCommand` only calls the latter case:

```
    protected void verifyCommand(
            BaseProtocolEncoder encoder, Command command, ByteBuf expected) {
        verifyFrame(expected, encoder.encodeCommand(command));
    }
```

This then causes the `assertNotNull` to fail in `verifyFrame`.

```
    protected void verifyFrame(ByteBuf expected, Object object) {
        assertNotNull(object, "buffer is null");
        assertInstanceOf(ByteBuf.class, object, "not a buffer");
        assertEquals(ByteBufUtil.hexDump(expected), ByteBufUtil.hexDump((ByteBuf) object));
    }
```

So I first looked for all encoders that also expect a channel:

```
./StartekProtocolEncoder.java:40:    protected Object encodeCommand(Channel channel, Command command) {
./WatchProtocolEncoder.java:135:    protected Object encodeCommand(Channel channel, Command command) {
./Jt808ProtocolEncoder.java:48:    protected Object encodeCommand(Channel channel, Command command) {
./SuntechProtocolEncoder.java:31:    protected Object encodeCommand(Channel channel, Command command) {
./T800xProtocolEncoder.java:54:    protected Object encodeCommand(Channel channel, Command command) {
./Xrb28ProtocolEncoder.java:34:    protected Object encodeCommand(Channel channel, Command command) {
```

... and then I took a look at the test files:

For example, in `StartekProtocolEncoderTest`, instead of `verifyCommand`, it contains the following:

`        assertEquals("xxx", encoder.encodeCommand(null, command));`

`WatchProtocolEncoderTest` uses `verifyFrame`:

`        verifyFrame(buffer("xxx"), encoder.encodeCommand(null, command));`

`T800xProtocolEncoderTest` uses `verifyFrame`, but with `binary()`:

`        verifyFrame(binary("xxx"), encoder.encodeCommand(null, command));`

`Xrb28ProtocolEncoderTest` uses `assertEquals`:

`        assertEquals("xxx", encoder.encodeCommand(null, command));`

There is no encoder test available for `Suntech`.

However, when I use `encoder.encodeCommand(null, command)` like this in `Jt808ProtocolEncoderTest`:

```
    @Test
    public void testEncodeCustom() throws Exception {

        var encoder = inject(new Jt808ProtocolEncoder(null));

        Command command = new Command();
        command.setDeviceId(1);
        command.setType(Command.TYPE_CUSTOM);

        command.set(Command.KEY_DATA, "123");
        assertEquals("313233", encoder.encodeCommand(null, command));

    }
```

... I get an error:

```
java.lang.NullPointerException: Cannot invoke "io.netty.channel.Channel.pipeline()" because "channel" is null
	at org.traccar.protocol.Jt808ProtocolEncoder.encodeCommand(Jt808ProtocolEncoder.java:51)
	at org.traccar.protocol.Jt808ProtocolEncoderTest.testEncodeCustom(Jt808ProtocolEncoderTest.java:37)
```

It seems that `Channel` cannot be `null` in `Jt808ProtocolEncoder`.

Unfortunately, I have absolutely no idea what "Channel" even is. 

I would appreciate any advice on how to proceed from here so that I can have a "working test file" in order to proceed with the actual use case testing.

-- Cheers, Nils

tananaev:
This fixes test: https://github.com/traccar/traccar/commit/b4348b20137110fe990191fb34fad3ea0c3fb663.

nilei:
Good morning Anton.

Great! With that fix, I was able to integrate the tests and run them successfully. Are you satisfied with the tests now?

-- Cheers, Nils

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
```diff
@@ -20,4 +20,31 @@ public void testEncode() throws Exception {
...
+        command.set(Command.KEY_DATA, "Test, Command, 123#");
+        verifyCommand(encoder, decoder, command, binary("7e830000140b3a73ce2ff2000001546573742c20436f6d6d616e642c2031323323a57e"));
+
+        encoder.setModelOverride("C5");
+        command.set(Command.KEY_DATA, "Test, Command, 123#");
+        verifyCommand(encoder, decoder, command, binary("7e830000140b3a73ce2ff2000001546573742c20436f6d6d616e642c2031323323a57e"));
+
+        encoder.setModelOverride("C5L");
+        command.set(Command.KEY_DATA, "Test, Command, 123#");
+        verifyCommand(encoder, decoder, command, binary("7e830000140b3a73ce2ff2000001546573742c20436f6d6d616e642c2031323323a57e"));
```
tananaev: We don't need to test all 3. It's redundant because the logic is exactly the same.

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
```diff
@@ -20,4 +20,31 @@ public void testEncode() throws Exception {
 
     }
 
+    @Test
+    public void testEncodeCustom() throws Exception {
```
tananaev: I would also combine it with the main test, so we don't need to duplicate things.

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
nilei: Reduced to one test with BSJ model name.

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
nilei: Combined all into testEncode()

PR #5895, opened 2026-06-02
Decode Xchinchun 0xE1 cell tower data

nilei:
This pull request adds support for decoding Xinchun-specific 0xE1 location extension data as cell tower information.

When the 0xE1 payload matches the Xinchun base station format, the decoder now extracts MCC, MNC, LAC, Cell ID, and RSSI and adds them to the position network.

The existing fallback behavior for other 0xE1 payload formats is preserved.

Evidence can be found in the Xchinchun protocol documentation [Xchinchun JT808.docx](https://github.com/user-attachments/files/28529600/Xchinchun.JT808.docx) on page 24.

image


src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -831,7 +831,18 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
                 case 0xD4:
                 case 0xE1:
                 case 0xE9:
-                    if (length == 1) {
```
tananaev, line 834: We should keep this as the first option.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -831,7 +831,18 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
...
+                    if (subtype == 0xE1 && length > 4 && (length - 4) % 8 == 0) {
```
tananaev: Shouldn't it be at least length 12?

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
nilei: 👍

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
nilei: 🚀 very good catch - Thank you!

nilei:
@tananaev  All changes have been incorporated.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -838,6 +838,17 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
                         } else {
                             position.set(Position.KEY_BATTERY_LEVEL, batteryLevel);
                         }
+                    } else if (subtype == 0xE1 && length > 12 && (length - 4) % 8 == 0) {
```
tananaev: Probably this?
```suggestion
                    } else if (subtype == 0xE1 && length >= 12 && (length - 4) % 8 == 0) {
```

nilei: > Probably this?

I am probably stupid... sometimes 😉 

Thank you!

PR #5900, opened 2026-06-06
feat(notification): add custom base URL support for Telegram

ghzserg:
Introduce a new configuration key 'notificator.telegram.base.url' to allow overriding the default Telegram Bot API endpoint.

This is highly useful for environments or countries where the official Telegram API (api.telegram.org) is blocked or restricted, enabling the use of reverse proxies or self-hosted Telegram Bot API servers.

- Added NOTIFICATOR_TELEGRAM_BASE_URL to Keys.java
- Updated NotificatorTelegram.java to fallback to default URL if not specified


tananaev:
Why not proxy connection from the server? It doesn't seem right to add a workaround directly in the server.

ghzserg:
In Russia, Roskomnadzor has learned to block proxy servers outside of Russia.Replacing the base URL works much better and is not blocked. Nowadays, many Telegram libraries are adding the ability to change the base URL.

tananaev:
How is your custom URL working if proxy is blocked?

ghzserg:
Rkn can distinguish a proxy request from a regular request for data from a website. And a request through a proxy if the proxy outside of Russia is blocked. 

A custom URL does not look like a proxy request. It looks like a link to a legitimate site that actually works.

There's a website called test.com, and it works like a standard website. It is located outside of Russia. But the URL test.com/about/filial5 is actually a reverse proxy for Telegram and easily passes Roskomnadzor's filters.

```
location /about/filial5/bot {
        rewrite ^/about/filial5(/bot.*)$ $1 break;

        proxy_pass https://api.telegram.org;

        proxy_set_header Host api.telegram.org;
        proxy_set_header Connection "";

        proxy_connect_timeout 60s;
        proxy_send_timeout 60s;
        proxy_read_timeout 60s;

        proxy_buffering on;
        proxy_buffer_size 4k;
        proxy_buffers 8 32k;

        proxy_http_version 1.1;
    }
```

PR #5904, opened 2026-06-10
Added dedicated protocols for MT700 and MT532 with configuration encoders

dtayme:
Added dedicated protocols for the mictrack MT700 and MT532 with defined alarms, and tested with real packets. Additionally added encoders to support pushing device configuration. Also added unit tests for the protocols.


tananaev:
This is not how we do things. It doesn't make sense to duplicate the whole protocol just because of a small variation.

Also make sure you provide documentation.

PR #5905, opened 2026-06-10
Add Mictrack MT700/MT700W/MT600/MT530/MT532 protocol support

dtayme:
## Summary

Extends `MictrackProtocolDecoder` and `MictrackProtocol` to support four additional Mictrack device families (MT700, MT700W, MT600, MT530, MT532) on a single unified protocol class and configurable port.

## Design

A new `MictrackFrameDecoder` inspects the first byte of each TCP frame to select the framing strategy:

| First byte | Format | Devices |
|------------|--------|---------|
| `*` | `*HQ,ID,VN,...#` terminated | MT532 |
| `#` | `#IMEI#MODEL#...##` terminated | MT700, MT700W, MT600, MT530 |
| other | `MT;...` / `IMEI$...` newline terminated | existing MT-series |

`MictrackProtocolDecoder` routes to the appropriate decode path and writes a `mictrack.variant` channel attribute (`"hq"` or `"mt700"`) on first contact so `MictrackProtocolEncoder` can select the correct command format without a separate protocol class per variant.

## Comparison with existing multi-format protocols

This follows the same first-byte dispatch pattern established by `H02Protocol`, which uses a custom `H02FrameDecoder` to route between its text (`*`) and binary (`$`) wire formats within a single protocol class and decoder. `MictrackFrameDecoder` applies the same technique across three distinct frame formats.

The `TLT-2H` protocol is a useful reference point because its wire format is nearly identical to the MT700-style `##`-terminated frame (same GPRMC/WiFi body structure). Without the unified framing approach used here, supporting both TLT-2H-style and HQ-style Mictrack devices would require a separate protocol file per variant.

The one addition beyond H02's approach is propagating the detected variant to the encoder via a Netty `AttributeKey` on the channel. H02 has no encoder; for protocols that do, this avoids having the encoder re-inspect packet content or maintain its own state to determine which command format to send.

## Changes

- **`MictrackFrameDecoder`** (new) - multi-format byte-level framing
- **`MictrackProtocolDecoder`** - adds `decodeHQ()` (V1/V5/V6 position, V4 heartbeat + R12 ACK, 4-byte vehicle status bitmask) and `decodeMT700()` (GPRMC + WiFi location, LBS cell towers, voltage heuristic `> 100 ? /1000 : /10`)
- **`MictrackProtocolEncoder`** (new) - HQ commands (D1, S20, SF/CF) and MT700 commands (MODE, REBOOT, 804, RCONF) selected by channel attribute
- **`MictrackProtocol`** - single server entry with the unified pipeline; all 10 command types declared

## Alarm mapping

Alarm codes differ between device families; the decoder uses the `MODEL` field from the `##`-frame header to select the correct mapping:

| Event | MT700 / MT700W | MT600 / MT530 |
|-------|----------------|---------------|
| `DEF` | `ALARM_REMOVING` (light sensor / tamper) | `ALARM_POWER_CUT` (external power cut) |
| `HT` | mode indicator — not an alarm | `ALARM_TEMPERATURE` |
| `TOWED` | `ALARM_TOW` | `ALARM_TOW` |
| `SHAKE` | `ALARM_VIBRATION` | — |
| `BLP` / `CLP` | `ALARM_LOW_BATTERY` | `ALARM_LOW_BATTERY` |

HQ vehicle status bitmask decoding covers: SOS, overspeed, power-on, low battery, tow, power cut, tamper, vibration, door, geofence, ignition.

## Test plan

- [x] `MictrackProtocolDecoderTest` passes — all existing standard/low-altitude tests plus new HQ (V1, V5, V6, V4 heartbeat, vehicle status bits) and MT700 tests (GPS, WiFi, LBS, SHAKE, TOWED, MT700 DEF vs MT600 DEF)
- [x] `./gradlew test --tests "org.traccar.protocol.MictrackProtocolDecoderTest"` — BUILD SUCCESSFUL


tananaev:
And again not right. I suggest you discuss it first if you don't understand what the problem is.

dtayme:
> And again not right. I suggest you discuss it first if you don't understand what the problem is.

I am curious what I am missing? I am only created the pull request as I thought it would be helpful.

tananaev:
You're trying to squish completely different protocols into one, duplicating code in the process. Basically the same issue as your original PR.

dtayme:
I will go back to the drawing board a last time. This PR was an attempt to consolidate the functionality based on the "vendor" protocol definitions. My original PR was trying to keep the 3 different protocol definitions streamlined and compact.  While the implementations are similar to the TLT2H and H02 protocols there are deviations for these models and I want to support configuration items for these models.

PR #5906, opened 2026-06-15
gt06: decode MSG_GPS_LBS_RFID (0x17) frames with length 0x27

paulobautz:
Some gt06 trackers send the GPS+LBS position in a `0x17` (`MSG_GPS_LBS_RFID`) frame with **length `0x27`** — one byte shorter than the RFID variant's `0x28` (they omit the trailing validity byte).

`decodeVariant()` only selected `Variant.RFID` for length `0x28`, so these frames fell through to `Variant.STANDARD`, were parsed as `MSG_WIFI`, and overran the buffer:

```
WARN: [Txxxxxxxx] error - readerIndex(44) + length(1) exceeds writerIndex(44): ...
```

The channel is then closed and the device reconnects in a loop ("connects and drops, never reports a position").

This accepts length `0x27` for `MSG_GPS_LBS_RFID` so the frame is decoded as GPS+LBS through the existing RFID path. The trailing RFID validity read lands on the serial bytes (no overrun) and the position decodes correctly. The change is additive — it only affects `type 0x17 && length 0x27`, which currently always crashes.

### Sample

```
787827171a060e160e33cd022e10560453b05500592102d40b1bd7002a9f4606500e020129f800238ffa0d0a
```

Decodes to `2026-06-14 22:14:51`, lat `-20.31848`, lon `-40.32972`, course `289°`. Verified against a running `6.14.4` instance with this patch, and added as a `verifyPosition` test case.

### DIFF
--- a/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Gt06ProtocolDecoder.java
@@ -1601,7 +1601,7 @@ public class Gt06ProtocolDecoder extends BaseProtocolDecoder {
             variant = Variant.SEEWORLD;
         } else if (header == 0x7878 && type == MSG_STATUS_3 && length == 0x0c) {
             variant = Variant.SEEWORLD;
-        } else if (header == 0x7878 && type == MSG_GPS_LBS_RFID && length == 0x28) {
+        } else if (header == 0x7878 && type == MSG_GPS_LBS_RFID && (length == 0x28 || length == 0x27)) {
             variant = Variant.RFID;
         } else if (header == 0x7878 && type == MSG_GPS_LBS_STATUS_5 && length == 0x40) {
             variant = Variant.LW4G;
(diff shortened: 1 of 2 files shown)


tananaev:
Please provide the official protocol documentation for your device.

paulobautz:
This is an unbranded Chinese device — no model markings on the case or PCB, and the SIM does not accept SMS, so I can't query the firmware version. I identified it as gt06 by following the protocol-identification guide (https://www.traccar.org/identify-protocol/), but I don't have the manufacturer's protocol manual.

In case the decoded sample is useful, the device sends its position as a `0x17` frame with length `0x27` — the same GPS+LBS layout as the existing RFID variant (`0x28`), minus the trailing validity byte:

```
787827171a060e160e33cd022e10560453b05500592102d40b1bd7002a9f4606500e020129f800238ffa0d0a
```

- `7878` `27`(len) `17`(type)
- `1a060e160e33` date/time → 2026-06-14 22:14:51
- `cd` gps len/sats, `022e1056` lat, `0453b055` lon, `00` speed, `5921` course/flags → -20.31848, -40.32972, course 289°
- `02d4` MCC 724 (Brazil), `0b` MNC, `1bd7` LAC, `002a9f` CID
- `4606500e020129f8` 8 trailing bytes
- `0023` serial, `8ffa` checksum

Because the length is `0x27`, it currently falls through to `MSG_WIFI` and overruns the buffer. Completely understand if you'd prefer not to merge without official documentation — just sharing the data in case it helps.

paulobautz:
Closing this — after more digging it looks like a custom/one-off firmware variant, and we only have a single unit, so it isn't worth carrying upstream. We'll keep it as a local patch on our side. Thanks for taking a look, and for Traccar!

PR #5907, opened 2026-06-15
Decode power voltage for BSJ and JH119 trackers

cycsmail:
Two JT808 variants report the vehicle supply voltage in fields that are currently dropped or misread:

- BSJ-KG-12A: extended block `0xEB` sub-id `0x002D` carries a 4-byte value in mV. The existing 2-byte case (a backup battery) is kept; the 4-byte variant is decoded as `power`, discriminated by length.
- JH119: additional id `0xE1` with length 2 carries a 0.1 V value, decoded as `power`.

Tests included for both fields.


tananaev:

cycsmail:
There is no public specification for these — they are proprietary BSJ / JH119 sub-fields, not part of the base JT/T 808 standard. Both changes extend fields the decoder already handles:

- `0x002D` is already decoded as battery (2 bytes). The BSJ-KG-12A sends it as 4 bytes; the length check preserves the existing case and decodes the 4-byte form as power.
- `0xE1` already handles length 1 (battery level) and the cell-tower form; this adds length 2 as power.

I verified both are the main supply voltage from captured frames: over several days the value tracks a parked vehicle battery, dropping ~12.5 → 11.4 V as the units drain with no charging — it moves like a battery, not a constant or status code. The test frames in the PR are real captures with device IDs and coordinates zeroed for privacy. Happy to share raw logs if that helps.

cycsmail:
Concrete data backing this — daily average of the decoded field, both units sitting unpowered (wired to a vehicle battery, engine off, so the battery is discharging with no charging source):

| date | BSJ `0x002D` | JH119 `0xE1` |
|------|------|------|
| 06-11 | 12.65 V | 12.46 V |
| 06-12 | 12.65 V | 12.45 V |
| 06-13 | 12.63 V | 12.42 V |
| 06-14 | 12.62 V | 12.39 V |
| 06-15 | 12.02 V | 11.61 V |

Both decline monotonically over days, then fall off as they approach the lead-acid discharge knee — the signature of a battery under load with no charging. A constant or a status code would not track that. This is averaged across >1,000 frames per device over the period.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -1068,7 +1070,11 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
                                             buf.readCharSequence(6, StandardCharsets.US_ASCII).toString()));
                                     break;
                                 case 0x002D:
-                                    position.set(Position.KEY_BATTERY, buf.readUnsignedShort() / 1000.0);
+                                    if (extendedLength >= 6) {
```
tananaev: Why `>= 6`?

cycsmail: Ah yeah, `>= 6` is unclear. The length includes the 2-byte type, so it just means the value is 4 bytes (the power form) vs the existing 2-byte battery one. I'll switch it to `== 6`.

PR #5910, opened 2026-06-17
Decode JT808 0x0210 hibernation battery reports

nilei:
Add handling for JT808 0x0210 hibernation battery level updates.

Some devices use message 0x0210 with a 7-byte body to report battery level during sleep:
- 1 byte battery percentage
- 6 byte BCD timestamp

The existing decoder treated all 0x0210 messages as batch location uploads. This caused 7-byte sleep battery reports to be parsed incorrectly.

This change adds a length-based special case for 0x0210 messages with a 7-byte body, decodes the battery level and timestamp into a Position object, and keeps existing batch upload handling unchanged for other 0x0210 payloads.

A regression test is included using a real device frame: 7e0210000704546000231900884f2606180229579c7e

**Evidence** can be found in the [Xchinchun JT808.docx](https://github.com/user-attachments/files/29068497/Xchinchun.JT808.docx) on page 41, section **3.20 Hibernation Battery Level Update (0x 0210)**

### DIFF
--- a/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
@@ -342,6 +342,8 @@ public class Jt808ProtocolDecoder extends BaseProtocolDecoder {
         int type = buf.readUnsignedShort();
         int attribute = buf.readUnsignedShort();
 
+        int bodyLength = BitUtil.to(attribute, 10);
+
         protocolVersion = BitUtil.check(attribute, 14) ? (int) buf.readUnsignedByte() : null;
         ByteBuf id = buf.readSlice(protocolVersion != null ? 10 : (delimiter == 0xe7 ? 7 : 6));
 
@@ -395,8 +397,6 @@ public class Jt808ProtocolDecoder extends BaseProtocolDecoder {
 
             getLastLocation(position, null);
 
-            int bodyLength = BitUtil.to(attribute, 10);
-
             buf.readUnsignedShort(); // response serial number
 
             String result = buf.readCharSequence(bodyLength - 2, StandardCharsets.UTF_16BE).toString().trim();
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -442,6 +442,20 @@ protected Object decode(
 
             return decodeLocation2(deviceSession, buf, type);
 
+        } else if (type == MSG_LOCATION_BATCH_2 && BitUtil.to(attribute, 10) == 7) {
```
tananaev: Is `BitUtil.to(attribute, 10)` length field? We should probably store it in a variable.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -442,6 +442,20 @@ protected Object decode(
...
+
+            sendGeneralResponse(channel, remoteAddress, id, type, index);
+
+            Position position = new Position(getProtocolName());
+            position.setDeviceId(deviceSession.getDeviceId());
+
+            getLastLocation(position, null);
+
+            position.set(Position.KEY_BATTERY_LEVEL, buf.readUnsignedByte());
+            position.setTime(readDate(buf, deviceSession.get(DeviceSession.KEY_TIMEZONE)));
```
tananaev: This is incorrect. You should instead pass time to `getLastLocation`.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
nilei: Moved and reused bodyLength variable.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
nilei: Submitted time into getLastLocation() instead of position.setTime()

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -442,6 +442,22 @@ protected Object decode(
...
             return decodeLocation2(deviceSession, buf, type);
 
+        } else if (type == MSG_LOCATION_BATCH_2 && bodyLength == 7) {
+
+            sendGeneralResponse(channel, remoteAddress, id, type, index);
+
+            Position position = new Position(getProtocolName());
+            position.setDeviceId(deviceSession.getDeviceId());
+
+            int batteryLevel = buf.readUnsignedByte();
```
tananaev: Why do we need this variable?

nilei: I felt like it was easier to read that way. Should I undo it?

tananaev: It's not easier to read when you split what could be one line into two lines with some unrelated code in between.

nilei: I initially tried to avoid a temporary variable by reading the battery value from the buffer after parsing the timestamp, but that turned out to be fragile because it relies on buffer index calculations.

In this case the payload format is:

`[1 byte battery level][6 byte BCD timestamp]`

Reading the battery value first and then parsing the timestamp is the most straightforward approach:

```java
position.set(Position.KEY_BATTERY_LEVEL, buf.readUnsignedByte());
getLastLocation(position, readDate(buf, deviceSession.get(DeviceSession.KEY_TIMEZONE)));
```

This follows the payload structure directly, avoids buffer index arithmetic, and does not require any assumptions about the reader position after `readDate()`.

Is that okay?

PR #5912, opened 2026-06-17
Add JT808 0xF6 IMEI decoding

nilei:
### Summary

Adds support for decoding the Xinchun-specific JT808 `0x0200` additional information field `0xF6` as an IMEI upload field.

### Details

According to the Xinchun JT808 protocol documentation, additional field `0xF6` with length `8` contains the device IMEI encoded as BCD. The decoder previously handled `0xF6` only as air pressure or as a generic event/sensor field, so Xinchun IMEI uploads were not exposed.

This change adds explicit handling for `0xF6` when the field length is `8`:

* reads the 8-byte BCD value
* removes the leading BCD padding zero
* stores the decoded value as `imei`
* keeps the existing `0xF6` handling for other lengths unchanged

### Test

Adds a decoder test using a minimal `0x0200` position report with:

`f6 08 08 67 56 70 41 10 07 69`

Expected decoded value:

`imei = 867567041100769`

### Evidence

Evidence can be found in the [Xchinchun JT808.docx](https://github.com/user-attachments/files/29069443/Xchinchun.JT808.docx) page 27, 0xF6 IMEI Upload.


tananaev:
IMEI is usually an identifier, not something we store in position attributes.

nilei:
> IMEI is usually an identifier, not something we store in position attributes.

Would you still consider adding that field once the failed test has been fixed? We also process the ICCID. I think that's a similar type of information.

-- Cheers, Nils

tananaev:
It would be inconsistent with all other protocols.

tananaev:
Which failed test are you referring to?

nilei:
> Which failed test are you referring to?

I run `./gradlew test --tests org.traccar.protocol.Jt808ProtocolDecoderTest` without errors, but here on GitHub there was a test failing that said:

```
Error: eckstyle] [ERROR] /home/runner/work/traccar/traccar/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java:1207: Line is longer than 120 characters (found 143). [LineLength]
```

tananaev:
That's not test and that's a failure in your own new code.

nilei:
> That's not test and that's a failure in your own new code.

Not sure where it came from, but it is fixed now. All checks are green.

nilei:
> IMEI is usually an identifier, not something we store in position attributes.

Would it then instead make sense for you to include a comparison and only add the IMEI attribute if the value differs from the device ID?

-- Cheers, Nils

tananaev:
It would still be inconsistent with all other protocols. Is it different in your case? And why do you need it?

nilei:
> It would still be inconsistent with all other protocols. Is it different in your case? And why do you need it?

Yes, the two pieces of information are completely different. For example, the ID is 045460002319 and the IMEI is 860422072160927. In fact, I have devices from four different manufacturers here at work, and only one of them has an ID that matches the IMEI (Xexun protocol). 

I have all the data forwarded to our backend, where the IMEI is an important piece of information for us. One that we currently have to retrieve from the SIM card provider via cumbersome additional API calls. However, this requires us to store the device ID there, which doesn’t always work 100% error-free, since the number is simply stored in a free-text field. All in all, it would be much more deterministic if the detected and transmitted information were also simply forwarded. That, at least, is our use case.

-- Cheers, Nils

tananaev:
Let's decode it as a custom attribute for now.

nilei:
I'm not entirely sure what you mean by "custom attributes." Is it then just doing this instead?

`                        position.set("imei", ByteBufUtil.hexDump(`

-- Cheers, Nils

tananaev:
Yep

nilei:
Done

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -1215,6 +1215,12 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
                 case 0xF6:
                     if (length == 2) {
                         position.set("airPressure", buf.readUnsignedShort());
+                    } else if (length == 8) {
+                        position.set("imei", ByteBufUtil.hexDump(
+                            buf.readSlice(length))
+                            .replaceFirst("^0", "")
+                            .replaceAll("f", "")
```
tananaev: Why do we need to do this? IMEI should have fixed format.

src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java
```diff
@@ -357,6 +357,10 @@ public void testDecode() throws Exception {
         verifyPosition(decoder, binary(
                 "7e02000072045460005635007300000000000000000236d51c00823f19000000000000260602220701010400000000300111310100e4020064e50100e60100e7080000000600000000e10c010c000600bbe4001ada1500ec23f0090dba115abb5091e3194157b9ce20bd25d93fb5ce20bd85d93fb57e9cd82cadd5b1f50101717e"));
 
+         verifyAttribute(decoder, binary(
```
tananaev: Indentation seems off

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
nilei: True! Done.

src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java
nilei: You've got an eagle eye, don't you? 😉 Done!

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -1215,6 +1215,8 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
                 case 0xF6:
                     if (length == 2) {
                         position.set("airPressure", buf.readUnsignedShort());
+                    } else if (length == 8) {
```
nilei, line 1218: @tananaev Do you think that 

`    if (length == 8 && isBcd(buf, buf.readerIndex(), length)) {`

would be a valuable addition?

PR #5914, opened 2026-06-17
Add local.properties to .gitignore

harrowmykel:
Add local.properties to .gitignore

local.properties is auto-generated when server project is built. This fix makes it impossible for contributors to accidentally commit local.properties

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ out
 build
 .vscode
 bin
+local.properties


tananaev:
What tool generates it and why?

harrowmykel:
Android studio auto generates it. Used by Gradle

image

tananaev:
I'm very confused. How is Android Studio related to this project?

harrowmykel:
I use android studio to write and test java code.

tananaev:
It sounds like a problem with your specific environment setup. I don't think we should adjust configuration for that.

harrowmykel:
Alright. A lot of developers use Android Studio for java, since it is free unlike Intellij IDEA - so i thought it made sense to just add it to the gitignore to avoid accidental commits for me and other devs in the future. But it's OK if you think it is unnecessary.

PR #5915, opened 2026-06-18
#5909: Support dynamic extra keys in OsmAnd json requests

### LINKED ISSUE #5909: Extend osman json payload to be dynamic
Possible to extend decodeJson() to accept additional data?
 by adding maps with predefined value type
e.g

```
extras
    double_values: {}
    string_values: {}
    boolean_values: {}
```

in https://www.traccar.org/osmand/

file => traccar-master\src\main\java\org\traccar\protocol\OsmAndProtocolDecoder.java

```java
---


 if (location.containsKey("extras")) {
    JsonObject extras = location.getJsonObject("extras");
    if (extras.containsKey("alarm")) {
        position.set(Position.KEY_ALARM, extras.getString("alarm"));
    }
    if (extras.containsKey("double_values")) {
        JsonObject doubleValues = extras.getJsonObject("double_values");
        for (Map.Entry<String, JsonValue> entry : doubleValues.entrySet()) {
            try {
                position.set(entry.getKey(), Double.parseDouble(entry.getValue()));
            } catch (NumberFormatException e) {
            }
        }
    }
    if (extras.containsKey("string_values")) {
        JsonObject stringValues = extras.getJsonObject("string_values");
        for (Map.Entry<String, JsonValue> entry : stringValues.entrySet()) {
            position.set(entry.getKey(), entry.getValue());
        }
    }
    if (extras.containsKey("boolean_values")) {
        JsonObject stringValues = extras.getJsonObject("boolean_values");
        for (Map.Entry<String, JsonValue> entry : stringValues.entrySet()) {
            position.set(entry.getKey(), entry.getValue());
        }
    }
}
```

harrowmykel:
this fixes #5909

### DIFF
--- a/src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
@@ -316,12 +316,7 @@ public class OsmAndProtocolDecoder extends BaseHttpProtocolDecoder {
                     case TRUE -> position.set(extraEntry.getKey(), true);
                     case FALSE -> position.set(extraEntry.getKey(), false);
                     case STRING -> {
-                        String entryStrValue = ((JsonString) extraEntry.getValue()).getString();
-                        switch (entryStrValue) {
-                            case "true" -> position.set(extraEntry.getKey(), true);
-                            case "false" -> position.set(extraEntry.getKey(), false);
-                            default -> position.set(extraEntry.getKey(), entryStrValue);
-                        }
+                        position.set(extraEntry.getKey(), ((JsonString) extraEntry.getValue()).getString());
                     }
                     default -> {}
                 }
--- a/src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
@@ -309,31 +309,25 @@ public class OsmAndProtocolDecoder extends BaseHttpProtocolDecoder {
                 JsonValue.ValueType valueType = extraEntryValue.getValueType();
 
                 switch (valueType) {
-                    case NUMBER:
+                    case NUMBER -> {
                         JsonNumber jsonNumber = (JsonNumber) extraEntryValue;
                         if (jsonNumber.isIntegral()) {
                             position.set(extraEntryKey, jsonNumber.longValue());
                         } else {
                             position.set(extraEntryKey, jsonNumber.doubleValue());
                         }
-                        break;
-                    case TRUE:
-                        position.set(extraEntryKey, true);
-                        break;
-                    case FALSE:
-                        position.set(extraEntryKey, false);
-                        break;
-                    case STRING:
+                    }
+                    case TRUE -> position.set(extraEntryKey, true);
+                    case FALSE -> position.set(extraEntryKey, false);
+                    case STRING -> {
                         String entryStrValue = ((JsonString) extraEntryValue).getString();
                         switch (entryStrValue) {
                             case "true" -> position.set(extraEntryKey, true);
                             case "false" -> position.set(extraEntryKey, false);
                             default -> position.set(extraEntryKey, entryStrValue);
                         }
-                        break;
-                    default:
-                        position.set(extraEntryKey, extraEntryValue.toString());
-                        break;
+                    }
+                    default -> position.set(extraEntryKey, extraEntryValue.toString());
                 }
             }
         }
--- a/src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
@@ -304,30 +304,26 @@ public class OsmAndProtocolDecoder extends BaseHttpProtocolDecoder {
                 position.set(Position.KEY_ALARM, extras.getString("alarm"));
             }
             for (Map.Entry<String, JsonValue> extraEntry : extras.entrySet()) {
-                String extraEntryKey = extraEntry.getKey();
-                JsonValue extraEntryValue = extraEntry.getValue();
-                JsonValue.ValueType valueType = extraEntryValue.getValueType();
-
-                switch (valueType) {
+                switch (extraEntry.getValue().getValueType()) {
                     case NUMBER -> {
-                        JsonNumber jsonNumber = (JsonNumber) extraEntryValue;
+                        JsonNumber jsonNumber = (JsonNumber) extraEntry.getValue();
                         if (jsonNumber.isIntegral()) {
-                            position.set(extraEntryKey, jsonNumber.longValue());
+                            position.set(extraEntry.getKey(), jsonNumber.longValue());
                         } else {
-                            position.set(extraEntryKey, jsonNumber.doubleValue());
+                            position.set(extraEntry.getKey(), jsonNumber.doubleValue());
                         }
                     }
-                    case TRUE -> position.set(extraEntryKey, true);
-                    case FALSE -> position.set(extraEntryKey, false);
+                    case TRUE -> position.set(extraEntry.getKey(), true);
+                    case FALSE -> position.set(extraEntry.getKey(), false);
                     case STRING -> {
-                        String entryStrValue = ((JsonString) extraEntryValue).getString();
+                        String entryStrValue = ((JsonString) extraEntry.getValue()).getString();
                         switch (entryStrValue) {
-                            case "true" -> position.set(extraEntryKey, true);
-                            case "false" -> position.set(extraEntryKey, false);
-                            default -> position.set(extraEntryKey, entryStrValue);
+                            case "true" -> position.set(extraEntry.getKey(), true);
+                            case "false" -> position.set(extraEntry.getKey(), false);
+                            default -> position.set(extraEntry.getKey(), entryStrValue);
                         }
                     }
-                    default -> position.set(extraEntryKey, extraEntryValue.toString());
+                    default -> {}
                 }
             }
         }
--- a/src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
@@ -23,7 +23,10 @@ import io.netty.handler.codec.http.HttpHeaderValues;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.QueryStringDecoder;
 import jakarta.json.Json;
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
```diff
@@ -294,11 +297,45 @@ private Object decodeJson(
...
             JsonObject extras = location.getJsonObject("extras");
             if (extras.containsKey("alarm")) {
                 position.set(Position.KEY_ALARM, extras.getString("alarm"));
             }
+            for (Map.Entry<String, JsonValue> extraEntry : extras.entrySet()) {
+                String extraEntryKey = extraEntry.getKey();
+                JsonValue extraEntryValue = extraEntry.getValue();
+                JsonValue.ValueType valueType = extraEntryValue.getValueType();
+
+                switch (valueType) {
```
tananaev: Use the new switch type, like above.

src/test/java/org/traccar/protocol/OsmAndProtocolDecoderTest.java
```diff
@@ -15,6 +15,16 @@ public void testDecodeJson() throws Exception {
...
+        verifyAttribute(decoder, requestData, "device_string_test", "long_device_id");
+        verifyAttribute(decoder, requestData, "device_bool_test", true);
+        verifyAttribute(decoder, requestData, "device_bool_false_test", false);
+        verifyAttribute(decoder, requestData, "device_integer_test", 1L);
+        verifyAttribute(decoder, requestData, "device_integer_twelve_test", 12L);
+        verifyAttribute(decoder, requestData, "device_double_test", 1.345);
+        verifyAttribute(decoder, requestData, "device_very_long_double_test", 1763636.3442);
```
tananaev: Use `verifyAttribute`. You don't need to do all of them. One is enough.

harrowmykel: Testing different value types

tananaev: Yeah, I can see what you're doing, but we don't really have a policy of 100% coverage. I think one value check is enough in this case.

harrowmykel: Fixed

src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
harrowmykel: Fixed

src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
```diff
@@ -294,11 +297,39 @@ private Object decodeJson(
...
+        }
+        if (location.containsKey("extras")) {
             JsonObject extras = location.getJsonObject("extras");
             if (extras.containsKey("alarm")) {
                 position.set(Position.KEY_ALARM, extras.getString("alarm"));
             }
+            for (Map.Entry<String, JsonValue> extraEntry : extras.entrySet()) {
+                String extraEntryKey = extraEntry.getKey();
+                JsonValue extraEntryValue = extraEntry.getValue();
+                JsonValue.ValueType valueType = extraEntryValue.getValueType();
```
tananaev: Inline all of these.

src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
```diff
@@ -294,11 +297,39 @@ private Object decodeJson(
...
+                    case FALSE -> position.set(extraEntryKey, false);
+                    case STRING -> {
+                        String entryStrValue = ((JsonString) extraEntryValue).getString();
+                        switch (entryStrValue) {
+                            case "true" -> position.set(extraEntryKey, true);
+                            case "false" -> position.set(extraEntryKey, false);
+                            default -> position.set(extraEntryKey, entryStrValue);
+                        }
+                    }
+                    default -> position.set(extraEntryKey, extraEntryValue.toString());
```
tananaev: What is the other type here?

harrowmykel: Array, object/map, null

tananaev: All those are not supported and should not be mapped to a string.

harrowmykel: done

src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
harrowmykel: done

src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
```diff
@@ -294,11 +297,35 @@ private Object decodeJson(
...
+                            position.set(extraEntry.getKey(), jsonNumber.doubleValue());
+                        }
+                    }
+                    case TRUE -> position.set(extraEntry.getKey(), true);
+                    case FALSE -> position.set(extraEntry.getKey(), false);
+                    case STRING -> {
+                        String entryStrValue = ((JsonString) extraEntry.getValue()).getString();
+                        switch (entryStrValue) {
+                            case "true" -> position.set(extraEntry.getKey(), true);
+                            case "false" -> position.set(extraEntry.getKey(), false);
```
tananaev: What is this? This is wrong.

harrowmykel: Trying to keep the same behavior as decodeQuery(). 

Why is it wrong?

tananaev: JSON already has a dedicated boolean type. We don't need to decode boolean strings.

harrowmykel: Fixed. Should be ready now.

src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
```diff
@@ -294,11 +297,28 @@ private Object decodeJson(
...
         if (location.containsKey("alarm")) {
             position.set(Position.KEY_ALARM, location.getString("alarm"));
-        } else if (location.containsKey("extras")) {
+        }
+        if (location.containsKey("extras")) {
             JsonObject extras = location.getJsonObject("extras");
             if (extras.containsKey("alarm")) {
                 position.set(Position.KEY_ALARM, extras.getString("alarm"));
             }
+            for (Map.Entry<String, JsonValue> extraEntry : extras.entrySet()) {
```
tananaev, line 302: How do you dedupe alarm?

src/test/java/org/traccar/protocol/OsmAndProtocolDecoderTest.java
```diff
@@ -15,6 +15,10 @@ public void testDecodeJson() throws Exception {
...
 
+        verifyAttribute(decoder, request(HttpMethod.POST, "/", new ReadOnlyHttpHeaders(true, "Content-Type", "application/json"), buffer(
+            "{\"location\":{\"timestamp\":\"2025-06-15T13:45:12.862Z\",\"coords\":{\"latitude\":37.4219983,\"longitude\":-122.084,\"accuracy\":5,\"speed\":0,\"heading\":-1,\"altitude\":5},\"is_moving\":false,\"odometer\":0,\"event\":\"motionchange\",\"battery\":{\"level\":1,\"is_charging\":false},\"activity\":{\"type\":\"still\"},\"extras\":{\"device_string_test\":\"long_device_id\",\"device_bool_test\":true,\"device_bool_false_test\":false,\"device_integer_test\":1,\"device_integer_twelve_test\":12,\"device_double_test\":1.345,\"device_very_long_double_test\":1763636.3442},\"_\":\"&id=48241179&lat=37.4219983&lon=-122.084&timestamp=2025-06-15T13:45:12.862Z&\"},\"device_id\":\"48241179\"}")),
+            "device_string_test", "long_device_id");
```
tananaev: I would test an int instead.

src/main/java/org/traccar/protocol/OsmAndProtocolDecoder.java
harrowmykel: Why should the user send the alarm twice?

I can do this:
```java

        if (location.containsKey("extras")) {
            JsonObject extras = location.getJsonObject("extras");
            if (extras.containsKey("alarm") && !location.containsKey("alarm")) {
                position.set(Position.KEY_ALARM, extras.getString("alarm"));
            }
....

}


```


This retains the old behavior and prioritises the location.alarm

tananaev: It won't send it twice. Your code decodes it twice.

harrowmykel: Yes. because this was the  old code

```java
    if (location.containsKey("alarm")) {
            position.set(Position.KEY_ALARM, location.getString("alarm"));
        } else if (location.containsKey("extras")) {
            JsonObject extras = location.getJsonObject("extras");
            if (extras.containsKey("alarm")) {
                position.set(Position.KEY_ALARM, extras.getString("alarm"));
            }
        }

```

The old code checks twice as well

tananaev: That's not your code. I said the issue is in YOUR code.

harrowmykel: Yes. I am saying my code tries to retain old behaviour.

The old code checks
data.location.alarm and data.location.extras.alarm.

So i am doing the same
checking data.location.alarm and data.extras.location.alarm.

I think you mean dedupe as in to avoid overwriting the value of alarm, right?

If that is what you mean, 
I can use data.location.alarm and if data.location.alarm is not set, i use data.location.extras.alarm

harrowmykel: oh i understand what you mean now.

```java

            for (Map.Entry<String, JsonValue> extraEntry : extras.entrySet()) {
                if(extraEntry.getKey().equals("alarm")){
                    continue;
                }
...
}
```

tananaev: Yes, but I think we should just remove old alarm check instead of doing this.

PR #5916, opened 2026-06-18
converted old switch to new format and added test

harrowmykel:
Using the new switch syntax seems to be a preference. So I converted the rest of the file

### DIFF
--- a/src/test/java/org/traccar/protocol/OsmAndProtocolDecoderTest.java
+++ b/src/test/java/org/traccar/protocol/OsmAndProtocolDecoderTest.java
@@ -4,6 +4,7 @@ import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.ReadOnlyHttpHeaders;
 import org.junit.jupiter.api.Test;
 import org.traccar.ProtocolTest;
+import org.traccar.model.Position;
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/traccar/protocol/OsmAndProtocolDecoderTest.java
```diff
@@ -42,19 +43,23 @@ public void testDecodeQuery() throws Exception {
...
+
         verifyPosition(decoder, request(
                 "/?lat=60.0&lon=30.0&speed=0.0&heading=0.0&vacc=0&hacc=0&altitude=0&deviceid=123456"));
 
         verifyPosition(decoder, request(
                 "/?id=861001000719969&lat=41.666667&lon=-0.883333&altitude=350.059479&speed=0.000000&batt=87"));
 
+        verifyAttribute(decoder, request(
+                "/?id=861001000719969&lat=41.666667&lon=-0.883333&altitude=350.059479&speed=0.000000&batt=87"),
+            Position.KEY_BATTERY_LEVEL, 87.0);
```
tananaev: Looks like indentation is off here. Also I don't think we need a new test.

harrowmykel: done

PR #5921, opened 2026-06-21
Add hybridTime threshold config and unit tests for TimeHandler

eriansv:
When Traccar Server have to hande hybrid devices, some sending the right deviceTime and other wiht bug deviceTime (lot of days or years ago), TimeHanderl_Hybrid filter bad deviceTime using the HYBRID_TIME_DAYS config value (Default: 7 days).

### DIFF
--- a/src/main/java/org/traccar/config/Keys.java
+++ b/src/main/java/org/traccar/config/Keys.java
@@ -1749,6 +1749,14 @@ public final class Keys {
             List.of(KeyType.CONFIG, KeyType.DEVICE));
 
     /**
+     * Number of days to use for hybrid time threshold. If not defined, defaults to 7 days.
+     */
+    public static final ConfigKey<Long> HYBRID_TIME_DAYS = new LongConfigKey(
+            "time.hybridDays",
+            List.of(KeyType.CONFIG),
+            7L);
+
+    /**
      * List of protocols to enable. If not specified, Traccar enables all protocols that have port numbers listed.
      * The value is a comma-separated list of protocol names.
      * Example value: teltonika,osmand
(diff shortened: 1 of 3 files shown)


tananaev:
You have to provide a bit more details. I don't understand the context.

eriansv:
Some GPS devices have firmware bugs that cause them to report incorrect timestamps lagging several years behind.

Using the existing serverTime option to fix this is problematic because it overwrites the timestamps for all devices. This leads to two critical issues:

1. Accurate devices lose precision: Devices without bugs lose their actual GPS-acquired timestamps.
2. Loss of buffered data history: For devices that were offline and later uploaded their cached/buffered historical data, their original event times are completely overwritten by the current server time, destroying historical data accuracy.

Solution:
I introduced the hybridTime case. It checks if the device's time is lagging behind the server time by a specific threshold (hybridTimeMillis). If it is severely outdated (buggy), it falls back to the server time. Otherwise, it preserves the device’s original timestamp, successfully protecting buffered historical data from normal devices.

tananaev:
It sounds like you have a rollover problem, which should be fixed in a different way.

eriansv:
Yes, I am aware. However, replacing or updating all the faulty devices currently deployed would be too expensive and time-consuming. Since we cannot modify the hardware, handling this anomaly on the server side is our only viable choice to maintain data integrity. Implementing this hybridTime feature fixes our current issue immediately while offering a valuable fallback option for other users facing similar legacy hardware limitations.

That being said, if you believe this should be handled differently, could you please point me in the right direction or suggest what the most viable server-side solution would be for this type of problem?

tananaev:
Something like this:

https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/handler/TimeHandler.java#L58

PR #5923, opened 2026-06-22
Add support for JT808 terminal attribute report (0x0107) ICCID decoding

nilei:
## Summary

Add support for decoding ICCID information from JT808 terminal attribute report messages (0x0107).

## Details

Some devices report terminal information using message ID 0x0107 immediately after connection. The message contains device metadata including terminal model, terminal ID, ICCID, hardware version, and firmware version.

The current decoder ignores 0x0107 messages, causing ICCID information to be unavailable until a vendor-specific (e.g. 0x1107) message is received. This change adds support for extracting the ICCID from the standard JT808 terminal attribute report.

## Protocol Reference

[Xchinchun JT808.docx](https://github.com/user-attachments/files/29208352/Xchinchun.JT808.docx)

- Page 33
- Section 2.22: Query Terminal Attribute Reporting (0x0107)

The specification defines the following fields in the 0x0107 message:

- Terminal Type
- Manufacturer ID
- Terminal Model
- Terminal ID
- Terminal SIM Card ICCID (BCD[10])
- Hardware Version
- Firmware Version

This change decodes the standard ICCID field from the terminal attribute report and stores it in `Position.KEY_ICCID`.

## Testing

Added a unit test using a real C5L terminal attribute report frame containing:

- Model: C5L
- ICCID: 89883030000131882455
- Firmware Version: C5L_V2.6 2026/06/04 15:31

The test verifies that the decoder correctly extracts and stores the ICCID from the 0x0107 message.

### DIFF
--- a/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
@@ -558,7 +558,7 @@ public class Jt808ProtocolDecoder extends BaseProtocolDecoder {
 
             return position;
 
-        }        
+        }
 
         return null;
     }
--- a/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
@@ -436,12 +436,12 @@ public class Jt808ProtocolDecoder extends BaseProtocolDecoder {
 
             getLastLocation(position, null);
 
-            buf.skipBytes(2); // terminal type
+            buf.readUnsignedShort(); // terminal type
             buf.skipBytes(5); // manufacturer id
             buf.skipBytes(20); // terminal model
             buf.skipBytes(7); // terminal id
 
-            position.set(Position.KEY_ICCID, ByteBufUtil.hexDump(buf.readSlice(10)).replaceAll("f", ""));
+            position.set(Position.KEY_ICCID, ByteBufUtil.hexDump(buf.readSlice(10)).substring(0, 20));
 
             return position;
 
--- a/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
@@ -441,7 +441,7 @@ public class Jt808ProtocolDecoder extends BaseProtocolDecoder {
             buf.skipBytes(20); // terminal model
             buf.skipBytes(7); // terminal id
 
-            position.set(Position.KEY_ICCID, ByteBufUtil.hexDump(buf.readSlice(10)).substring(0, 20));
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -426,6 +427,24 @@ protected Object decode(
...
+        } else if (type == MSG_TERMINAL_ATTRIBUTES) {
+
+            sendGeneralResponse(channel, remoteAddress, id, type, index);
+
+            Position position = new Position(getProtocolName());
+            position.setDeviceId(deviceSession.getDeviceId());
+
+            getLastLocation(position, null);
+
+            buf.skipBytes(2); // terminal type
```
tananaev:
```suggestion
            buf.readUnsignedShort(); // terminal type
```

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -426,6 +427,24 @@ protected Object decode(
...
+            buf.skipBytes(5); // manufacturer id
+            buf.skipBytes(20); // terminal model
+            buf.skipBytes(7); // terminal id
+
+            position.set(Position.KEY_ICCID, ByteBufUtil.hexDump(buf.readSlice(10)).replaceAll("f", ""));
```
tananaev: The ICCID length shouldn't be changing, so I think you can use substring instead of replacement.

nilei: Good point. According to the protocol specification, the ICCID field is fixed at `BCD[10]`, so it should always contain 20 digits. Using `substring(0, 20)` is more appropriate here than removing all trailing `f` characters. I have updated it accordingly like so:

`position.set(Position.KEY_ICCID, ByteBufUtil.hexDump(buf.readSlice(10)).substring(0, 20));`

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
nilei: Done.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
tananaev: This doesn't make sense. 10 bytes will always result in 20 digits.

nilei: Yes, so - In that case neither the replacement nor the substring is necessary, right?

Would using the value directly be what you had in mind?

```java
position.set(Position.KEY_ICCID, ByteBufUtil.hexDump(buf.readSlice(10)));
```

tananaev: Sounds like it.

nilei: Done

PR #5924, opened 2026-06-22
Feature/support monthly weekly yearly summary report

harrowmykel:
Added support for monthly weekly and yearly report summary.

In the current version daily summary is supported, so i extended it by adding weekly, monthly and yearly.

NB: Currently, the code does not try to correct the user. For example: if the user sets a range of 1 month (from=01.06.2026 to=30.06.206) and chooses yearly view. The code respects the user choice and calculates the sum for that year WITHIN the date range, in this case within 1 month and not within that full year.

Screenshot_4


tananaev:
What is the use case? I haven't heard a need for this, so I think we should collect some interest before we incorporate this change.

harrowmykel:
I need it in my analytics project. I am working on a dashboard where admins can compare different cars and drivers - how good each driver drive compared to another.

My use case is particularly how many times events (like for example hardbrake, hardAcceleration) occured per driver within a particular period and then drawing a graph. 

So I need the Api route. I dont want to be too invasive though. 

Since the API is already available, it was just a minor change to the existing route.

tananaev:
It definitely doesn't look like a minor change.

harrowmykel:
I understand but I only added a string reportInterval instead of boolean daily and a switch.

tananaev:
I see 100 lines of code added. I also see web app commit update for some reason.

harrowmykel:
Its just easier to do on the server than sending full daily report of a year over the net or multiple api queries with monthly from and to.

harrowmykel:
> I see 100 lines of code added. I also see web app commit update for some reason.

The web app is not necessary. I only added it since the feature is available on the api. Only the server is useful. 

Web was just 2 files. String translation and Select field

tananaev:
If you can find a cleaner way, we might be able to incorporate it, but as it is, I think it adds to much overhead. It seems like you're also inconsistently replaced daily in some places and left it in others.

harrowmykel:
I did not replace it anywhere, except in the stat provider file, where i used enum instead of bool. In the report resource file, i just added a new query parameter.

tananaev:
And that's definitely wrong.

harrowmykel:
Adding param or using enum?

tananaev:
I am talking about adding duplicate parameter.

harrowmykel:
Just trying to retain backward compatibility, since the boolean daily datatype only supports 2 states. 

The current solution prioritizes the boolean daily else checks for reportInterval. If none is provided, it skips the restriction.

PR #5926, opened 2026-06-24
Add support for the Xinchun time zone synchronization command (0x1007) by sending a general response (0x8001)

nilei:
## Details

Some devices send message ID 0x1007 after authentication and before the first location report. The current decoder does not explicitly handle this message and does not send a response.

This change adds support for 0x1007 and acknowledges the message with a standard JT808 general response (0x8001).

No position data is contained in this message and no device attributes are currently updated.

## Protocol Reference

[Xchinchun JT808.docx](https://github.com/user-attachments/files/29297571/Xchinchun.JT808.docx)

* Page 41
* Section 3.7: Time Zone Synchronization Command (0x1007)

The protocol defines a 4-byte payload containing time zone information.

Example frame observed from a C5L device:

```text
7e10070004045460002319090640000000567e
```

Payload:

```text
40 00 00 00
```

According to the protocol specification, the first byte represents the configured time zone. In this example, the value corresponds to GMT+8.

## Future Improvements

The time zone value is currently not stored or applied by the decoder.

While the protocol provides time zone information, this change intentionally limits the implementation to acknowledging the message. The observed devices report GMT+8, which matches the existing default behavior and does not currently provide any additional information required by the decoder.

Future work could include:

* Decoding the reported time zone value into a human-readable UTC offset
* Storing the reported device time zone as a device attribute
* Applying the reported time zone to device-specific date parsing where appropriate

These enhancements can be implemented later if devices are observed reporting different time zone values or if additional use cases require access to the reported configuration.

## Testing

Added a unit test using a real device frame to verify that the decoder accepts and processes the 0x1007 message without generating a position.

The message has been observed consistently on C5L devices immediately after authentication and before the first location report following a reconnect.

### DIFF
--- a/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
@@ -424,7 +424,7 @@ public class Jt808ProtocolDecoder extends BaseProtocolDecoder {
             }
 
         } else if (type == MSG_TERMINAL_AUTH || type == MSG_HEARTBEAT_2
-                || type == MSG_PHOTO || type == MSG_TERMINAL_LOGOUT) {
+                || type == MSG_PHOTO || type == MSG_TERMINAL_LOGOUT || type == MSG_TIMEZONE_SYNC) {
 
             sendGeneralResponse(channel, remoteAddress, id, type, index);
 
@@ -494,10 +494,6 @@ public class Jt808ProtocolDecoder extends BaseProtocolDecoder {
                 channel.writeAndFlush(new NetworkMessage(
                         formatMessage(MSG_TERMINAL_REGISTER_RESPONSE, id, false, response), remoteAddress));
             }
-
-        } else if (type == MSG_TIMEZONE_SYNC) {
-
-            sendGeneralResponse(channel, remoteAddress, id, type, index);
 
         } else if (type == MSG_ACCELERATION) {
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -494,6 +495,10 @@ protected Object decode(
                         formatMessage(MSG_TERMINAL_REGISTER_RESPONSE, id, false, response), remoteAddress));
             }
 
+        } else if (type == MSG_TIMEZONE_SYNC) {
```
tananaev: Why do we need a completely new block for this?

nilei: To be honest, I hadn't thought about or looked for whether there was another part of the decoder that simply does nothing more than send a generic response. After reading your comment, I found this part:

```
        } else if (type == MSG_TERMINAL_AUTH || type == MSG_HEARTBEAT_2
                || type == MSG_PHOTO || type == MSG_TERMINAL_LOGOUT) {

            sendGeneralResponse(channel, remoteAddress, id, type, index);
```

Would it be better to just include the key there?

-- Cheers, Nils

tananaev: Yes, that's exactly what it's for there.

nilei: Done

PR #5928, opened 2026-06-25
Decode Xinchun 0xF8 Extension as Step Count for C5/C5L Devices

nilei:
### Description

This PR adds model-specific decoding for the proprietary Xinchun extension **`0xF8`**.

### Changes

* Decode extension **`0xF8`** as **step count** (`steps`) for **C5** and **C5L** devices.
* Preserve the existing behavior for all other JT808 models, where `0xF8` continues to be decoded as the second external temperature sensor.
* Keep the implementation fully backward compatible by applying the new interpretation only to the affected device models.

### Motivation

The Xinchun protocol assigns a different meaning to extension `0xF8` than the current JT808 implementation. According to the vendor documentation, `0xF8` contains the device's **step count** for C5/C5L pet trackers, while the current decoder interprets it as an external temperature value.

Restricting the change to the relevant models preserves compatibility with existing devices while allowing C5/C5L trackers to expose the correct data.

### Evidence

See the [Xchinchun JT808.docx](https://github.com/user-attachments/files/29336626/Xchinchun.JT808.docx):

* **Section:** Location Report – Extended Information Items
* **Extension ID:** `0xF8`
* **Definition:** Step Count (WORD, 2 bytes)
* **Page:** 10 (Extended Information Items section)


src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -1272,7 +1272,11 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
                     }
                     break;
                 case 0xF8:
-                    position.set(Position.PREFIX_TEMP + 2, buf.readUnsignedShort() / 10.0 - 50);
+                    if (length == 2 && Set.of("C5", "C5L").contains(model)) {
```
tananaev:
```suggestion
                    if (Set.of("C5", "C5L").contains(model)) {
```

src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java
```diff
@@ -406,6 +406,11 @@ public void testDecode() throws Exception {
                 "7E020000834EB6FB4AF2C1010F00000800000C000B01AAB48204EA7E57000801AD00B626032715524501040000045130011F310112EB095554432D30343A3030140400000001150400000004642F0000000B00040100000000002A000801AAB33804EA7E542603271552450400303037343234322603271552450B0500E81120050E200102008C0006010400070200069B7E"),
                 "alarmLabel", "303037343234322603271552450b0500");
 
+        decoder.setModelOverride("C5L");
+
+        verifyAttribute(decoder, binary(
+                "7e02000020000000000000000100000000000000000000000000000000000000000000260625120000f8023039c77e"),
+                "steps", 12345);
```
tananaev:
```suggestion
                "steps", 12345);

```

src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java
```diff
@@ -406,6 +406,12 @@ public void testDecode() throws Exception {
                 "7E020000834EB6FB4AF2C1010F00000800000C000B01AAB48204EA7E57000801AD00B626032715524501040000045130011F310112EB095554432D30343A3030140400000001150400000004642F0000000B00040100000000002A000801AAB33804EA7E542603271552450400303037343234322603271552450B0500E81120050E200102008C0006010400070200069B7E"),
                 "alarmLabel", "303037343234322603271552450b0500");
 
+        decoder.setModelOverride("C5L");
+
+        verifyAttribute(decoder, binary(
+                "7e02000020000000000000000100000000000000000000000000000000000000000000260625120000f8023039c77e"),
```
tananaev: Wait is this real data you got from device? Or made up fake message?

nilei: Sorry. I of course have logs with real data that I am normally use for this. It is added now. Additinally, there was an error on line 107 of the decoder test due to the removed length check when model was null. Fixed this with a different check for the model.

PR #5930, opened 2026-06-26
Add configurable IOTM output control handling

nickiv69:
This PR adds server-side support for IOTM `outputControl` commands.

Changes included:

- Add IOTM MQTT output control command encoding.
- Map Traccar `outputControl` commands for OUT1 and OUT2 to the IOTM command ids.
- Add `iotm.permanentOutputControl` server config option.
- When `iotm.permanentOutputControl=true`, output commands use static/latching output ids.
- When the option is missing or `false`, the existing timed output behavior is preserved.
- Add/update encoder tests for default timed output, configured static output, and output id changes.


tananaev:
Do you have the protocol documentation?

nickiv69:
Yes, I have.
Just for info, we are running your service on
https://service.geotrack24.com/
And all commits are tested live with LX41.

tananaev:
Great, but I still don't see it attached unfortunately.

nickiv69:
Let's try with a link this time.
https://drive.google.com/file/d/16m3eme7O0Q-MIY7IwMypRgX7XEpcB19g/view?usp=drive_link

src/main/java/org/traccar/protocol/IotmProtocolDecoder.java
```diff
@@ -139,9 +143,20 @@ private void decodeSensor(Position position, ByteBuf record, int sensorType, int
     protected Object decode(
             DeviceSession deviceSession, MqttPublishMessage message) throws Exception {
 
-        List<Position> positions = new LinkedList<>();
+        try {
+            return decodePayload(deviceSession, message.payload());
+        } catch (IndexOutOfBoundsException | IllegalArgumentException error) {
```
tananaev: Why is this needed?

src/main/java/org/traccar/protocol/IotmProtocolEncoder.java
```diff
@@ -21,13 +21,24 @@
...
+import org.traccar.config.Keys;
 import org.traccar.helper.Checksum;
 import org.traccar.model.Command;
 
+import java.util.concurrent.atomic.AtomicInteger;
+
 public class IotmProtocolEncoder extends BaseProtocolEncoder {
 
-    public IotmProtocolEncoder(Protocol protocol) {
+    private static final int OUTPUT_ID_STATIC_SIGNAL = 0x08;
```
tananaev, line 33: Where in the documentation can find this?

src/main/java/org/traccar/protocol/IotmProtocolEncoder.java
```diff
@@ -48,22 +59,39 @@ protected Object encodeCommand(Command command) {
...
+        }
 
         buf.writeByte(Checksum.sum(buf.nioBuffer()));
 
         return MqttMessageBuilders.publish()
                 .topicName(uniqueId + "/OUTC")
                 .qos(MqttQoS.AT_LEAST_ONCE)
                 .payload(buf)
-                .messageId(0)
+                .messageId(messageId.updateAndGet(value -> value == 0xffff ? 1 : value + 1))
```
tananaev: Please explain this.

src/test/java/org/traccar/protocol/IotmProtocolDecoderTest.java
```diff
@@ -1,12 +1,74 @@
...
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.clearInvocations;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
 
 public class IotmProtocolDecoderTest extends ProtocolTest {
 
+    private static class TestIotmProtocolDecoder extends IotmProtocolDecoder {
```
tananaev: All your unit tests would need to be completely re-done. They're very inconsistent with the current way we do tests.

tananaev:
Please reply properly to GitHub commits.

src/main/java/org/traccar/protocol/IotmProtocolEncoder.java
nickiv69: The OUTC command is sent as an MQTT PUBLISH with QoS 1
(MqttQoS.AT_LEAST_ONCE). For QoS 1, MQTT requires a non-zero packet
identifier so the receiver can acknowledge the exact message with PUBACK.

Using messageId(0) is invalid for QoS 1. The new code generates
a changing 16-bit MQTT packet identifier in the valid range 1..0xffff and
wraps back to 1 after 0xffff, avoiding 0.

tananaev:
If you continue replying via email, we won't be able to merge this. GitHub email responses are an unusable disaster. Your comments are not formatted properly and they don't land in the right threads.

src/main/java/org/traccar/protocol/IotmProtocolDecoder.java
nickiv69: It is not required for the outputControl / keep-alive change itself.

I added it only as defensive handling for malformed or incomplete IOTM
payloads,
so one bad packet would be ignored instead of propagating an
IndexOutOfBoundsException from ByteBuf parsing.

But if this is not desired in Traccar style, I can remove the try/catch and
keep
the original behavior, where decode errors are handled by the common
pipeline.

tananaev: Have you ever seen malformed packets? And they were still validated by TCP checksum?

src/main/java/org/traccar/protocol/IotmProtocolEncoder.java
tananaev: Can't we just use 1?

src/main/java/org/traccar/protocol/IotmProtocolEncoder.java
nickiv69: The value 0x08 comes from XG_IOTM protocol V1.7.pdf, section 8
"Output data structures", where output id 0x08 is defined as
STATIC SIGNAL1 CONTROL. Static signals are sequentially assigned:
0x08 = S1, 0x09 = S2, 0x0A = S3, etc. Therefore 0x08 is used as the
base id for static signal output control.

src/test/java/org/traccar/protocol/IotmProtocolDecoderTest.java
nickiv69: I don't have experience with other devices in Traccar.
If you give me some basic instructions, I can rewrite it.

tananaev: Just look at existing tests.

nickiv69:
> If you continue replying via email, we won't be able to merge this. GitHub email responses are an unusable disaster. Your comments are not formatted properly and they don't land in the right threads.

Sorry, I sow your messages too late.

src/main/java/org/traccar/protocol/IotmProtocolDecoder.java
nickiv69: A packet can still be application-level invalid while TCP checksum is valid, for example because of a device firmware/protocol issue, but I do not have such a real case here. So I agree it should not be part of this change. The change is made during search of other session problems with LX41.

src/main/java/org/traccar/protocol/IotmProtocolEncoder.java
nickiv69: Yes, we can use 1 if the goal is only to avoid the invalid zero packet id.

I used an incrementing value because this field is the MQTT QoS1 packet
identifier, so normally it identifies each in-flight PUBLISH/PUBACK exchange.
Using a changing id avoids ambiguity if multiple commands are sent on the same
connection before the previous PUBACK is processed.

But for this implementation we don't track outbound MQTT state, and the device
ACKs the command immediately, so a constant non-zero value should be enough.

src/test/java/org/traccar/protocol/IotmProtocolDecoderTest.java
nickiv69: Understood. I will remove the custom TestIotmProtocolDecoder subclass and rewrite the tests following the existing ProtocolTest pattern used by other protocol decoder tests.

nickiv69:
I pushed cleanup changes based on the review:
- removed unrelated malformed-payload handling
- reworked/cleaned up IOTM tests to match existing Traccar test style
- regarding simplifying MQTT messageId handling -
A constant messageId(1) would probably work with LX41 in practice because commands are ACKed immediately, but the current implementation follows the MQTT model more closely. I can simplify it if you prefer.

The PR now focuses only on IOTM outputControl support and MQTT ping handling.

src/main/java/org/traccar/protocol/IotmProtocolEncoder.java
tananaev: It looks like this is not resolved.

src/test/java/org/traccar/protocol/IotmProtocolDecoderTest.java
```diff
@@ -1,38 +1,88 @@
...
 import org.traccar.ProtocolTest;
 
+import java.net.SocketAddress;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.mock;
+
 public class IotmProtocolDecoderTest extends ProtocolTest {
```
tananaev, line 8: What happened to all the test cases? You have to follow the standard style.

nickiv69: Review last commit.
Is there anything else you would like to be changed?

tananaev: Still doesn't look good. You're still doing too much in your tests. Please I'm asking again to look at all other tests as examples.

nickiv69: Simplified the IOTM tests further to match the existing protocol test style. Removed custom MQTT/ByteBuf test logic and kept only minimal decode/encode coverage.

src/test/java/org/traccar/protocol/IotmProtocolDecoderTest.java
```diff
@@ -33,6 +33,32 @@ public void testDecode() throws Exception {
...
+                "02020800faafb360e4160300015900c34b306a0302b0030b30050030eb2e050430800c03034003014003"
+                        + "0240031d30030c300301a0031ba00300200e00d0e0402d42ca56dd4100000b0e0000820101940000"
+                        + "95000401200e040520350013000002000308200307201c"))
+                .qos(MqttQoS.AT_LEAST_ONCE).messageId(2).build(), "out1", true);
+
+        verifyAttribute(decoder, MqttMessageBuilders.publish().payload(binary(
+                "02020800faafb360e4160300015900c34b306a0302b0030b30050030eb2e050430800c03034003014003"
+                        + "0240031d30030c300301a0031ba00300200e00d0e0402d42ca56dd4100000b0e0000820101940000"
+                        + "95000401200e040520350013000002000308200307201c"))
+                .qos(MqttQoS.AT_LEAST_ONCE).messageId(2).build(), "out2", false);
```
tananaev: HEX should be single line. Also why do we need 4 tests? Are these real samples from a real device?

src/main/java/org/traccar/protocol/IotmProtocolEncoder.java
```diff
@@ -48,22 +58,39 @@ protected Object encodeCommand(Command command) {
...
-        buf.writeByte(0); // output command index
-        buf.writeByte(3); // length
-        buf.writeByte(command.getInteger(Command.KEY_DATA));
-        buf.writeByte(0xB0);
-        buf.writeByte(0xB1);
+        int index = command.getInteger(Command.KEY_INDEX);
+        if (permanentOutputControl) {
+            if (index < 1 || index > 2) {
+                throw new IllegalArgumentException("Unsupported permanent output index");
+            }
```
tananaev: These checks are unnecessary.

src/test/java/org/traccar/protocol/IotmProtocolDecoderTest.java
nickiv69: The samples are copied from Traccar tracker-server.log while the device was connected to my development server. They are not synthetic packets.

The four cases cover the full mapping introduced by this change:

index=1,data=1 -> S1 ON
index=1,data=0 -> S1 OFF
index=2,data=1 -> S2 ON
index=2,data=0 -> S2 OFF

The mapping is not only ON/OFF, but also output index to static signal id:
index 1 -> static signal 0x08
index 2 -> static signal 0x09

src/main/java/org/traccar/protocol/IotmProtocolEncoder.java
nickiv69: the lines are removed.

src/test/java/org/traccar/protocol/IotmProtocolDecoderTest.java
nickiv69: HEX is now single line.

PR #5931, opened 2026-06-27
Make memory storage concurrent-safe

BastyJuice:
Update `MemoryStorage` to use concurrent collections for shared in-memory object and permission maps.

The previous implementation used regular `HashMap` instances and exposed lazy streams over `values()`. With `database.memory=true` and `database.registerUnknown=true`, concurrent device logins can read from `getObjectsStream()` while another thread registers or updates an object, which can trigger `ConcurrentModificationException`.

This change replaces the shared maps with `ConcurrentHashMap`, centralizes object-map creation through `getObjectMap()`, and uses concurrent permission sets. This keeps the in-memory storage behavior unchanged while making concurrent access from decoder/session threads safer.

### DIFF
--- a/src/main/java/org/traccar/storage/MemoryStorage.java
+++ b/src/main/java/org/traccar/storage/MemoryStorage.java
@@ -49,7 +49,9 @@ public class MemoryStorage extends Storage {
         Server server = new Server();
         server.setId(1);
         server.setRegistration(true);
-        objects.computeIfAbsent(Server.class, key -> new ConcurrentHashMap<>()).put(server.getId(), server);
+        Map<Long, Object> servers = new ConcurrentHashMap<>();
+        servers.put(server.getId(), server);
+        objects.put(Server.class, servers);
     }
 
     @Override
--- a/src/main/java/org/traccar/storage/MemoryStorage.java
+++ b/src/main/java/org/traccar/storage/MemoryStorage.java
@@ -29,19 +29,19 @@ import org.traccar.storage.query.Request;
 
 import java.lang.invoke.MethodHandle;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Stream;
 
 public class MemoryStorage extends Storage {
 
-    private final Map<Class<?>, Map<Long, Object>> objects = new HashMap<>();
-    private final Map<Pair<Class<?>, Class<?>>, Set<Pair<Long, Long>>> permissions = new HashMap<>();
+    private final Map<Class<?>, Map<Long, Object>> objects = new ConcurrentHashMap<>();
+    private final Map<Pair<Class<?>, Class<?>>, Set<Pair<Long, Long>>> permissions = new ConcurrentHashMap<>();
 
     private final AtomicLong increment = new AtomicLong();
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/storage/MemoryStorage.java
```diff
@@ -29,27 +29,31 @@
...
 
     public MemoryStorage() {
         Server server = new Server();
         server.setId(1);
         server.setRegistration(true);
-        objects.put(Server.class, Map.of(server.getId(), server));
+        getObjectMap(Server.class).put(server.getId(), server);
+    }
+
+    private Map<Long, Object> getObjectMap(Class<?> clazz) {
```
tananaev: Why do we need this?

BastyJuice:
The goal is to avoid `ConcurrentModificationException` when `MemoryStorage` is accessed concurrently.

In our forwarder-only setup we use:

- database.memory=true
- database.registerUnknown=true

With many concurrent Teltonika logins, one thread can read from `getObjectsStream()` while another thread auto-registers a device through `addObject()`. The current implementation uses `HashMap` and returns lazy streams over `values()`, so concurrent writes can fail with `ConcurrentModificationException`.

The helper is only to make sure all object maps are created consistently as `ConcurrentHashMap`.

tananaev:
I think we can inline the helper.

src/main/java/org/traccar/storage/MemoryStorage.java
```diff
@@ -29,27 +29,27 @@
...
+    private final Map<Pair<Class<?>, Class<?>>, Set<Pair<Long, Long>>> permissions = new ConcurrentHashMap<>();
 
     private final AtomicLong increment = new AtomicLong();
 
     public MemoryStorage() {
         Server server = new Server();
         server.setId(1);
         server.setRegistration(true);
-        objects.put(Server.class, Map.of(server.getId(), server));
+        objects.computeIfAbsent(Server.class, key -> new ConcurrentHashMap<>()).put(server.getId(), server);
```
tananaev: I don't think this requires `computeIfAbsent`.

src/main/java/org/traccar/storage/MemoryStorage.java
BastyJuice: We don't strictly need it. I added it to avoid repeating the `computeIfAbsent(... new ConcurrentHashMap<>())` expression, but I agree if done it now by inline to keep the change smaller. I updated the patch accordingly.

src/main/java/org/traccar/storage/MemoryStorage.java
BastyJuice: I can changed it to initialize the server map directly instead of using `computeIfAbsent`.

src/main/java/org/traccar/storage/MemoryStorage.java
```diff
@@ -29,27 +29,29 @@
...
+    private final Map<Pair<Class<?>, Class<?>>, Set<Pair<Long, Long>>> permissions = new ConcurrentHashMap<>();
 
     private final AtomicLong increment = new AtomicLong();
 
     public MemoryStorage() {
         Server server = new Server();
         server.setId(1);
         server.setRegistration(true);
-        objects.put(Server.class, Map.of(server.getId(), server));
+        Map<Long, Object> servers = new ConcurrentHashMap<>();
```
tananaev: You can still keep it as one-liner. Do we actually need `ConcurrentHashMap` here?

BastyJuice: Good point. The initial Server map does not need to be mutable. I will keep that as Map.of(...) and use ConcurrentHashMap only for maps that are modified at runtime.

PR #5934, opened 2026-06-28
fixed #5925: added dynamic periodinterval

### LINKED ISSUE #5925: Support for periodInterval in summary report
Adding periodInterval would make requests for time periods faster - > weekly, monthly, yearly.

Use case:
I am working on a dashboard where admins can compare different cars and drivers - how good each driver drive compared to another.

My use case is particularly how many times events (like for example hardbrake, hardAcceleration) occured per driver within a particular period and then drawing a graph.
The Api for this already exists, however it only supports "daily report" or "total" between two timeframes: from and to. If a user requires "monthly reports" between a period of one year for multiple devices n , this would require 12 months * n API Requests.

Adding this feature would reduce it to n requests.

----

Important: This issue was created to track commits. Please do NOT work on it.

harrowmykel:
This is a fix for #5925 <-- read this to see what it does

and cleaner overwrite for #5924 

NB: Currently, the code does not try to correct the user. For example: if the user sets a range of 1 month (from=01.06.2026 to=30.06.206) and chooses yearly view. The code respects the user choice and calculates the sum for that year WITHIN the date range, in this case within 1 month and not within that full year.

### DIFF
--- a/src/main/java/org/traccar/helper/DateUtil.java
+++ b/src/main/java/org/traccar/helper/DateUtil.java
@@ -15,9 +15,12 @@
  */
 package org.traccar.helper;
 
+import java.time.DayOfWeek;
 import java.time.Instant;
 import java.time.ZoneId;
+import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
 import java.util.Calendar;
 import java.util.Date;
 
@@ -78,6 +81,54 @@ public final class DateUtil {
         } else {
             return LOCAL_DATE_TIME.format(date.toInstant());
         }
+    }
+
+    public enum SummaryReportInterval {
+        NONE,
+        DAILY,
+        WEEKLY,
+        MONTHLY,
+        YEARLY
+    }
+
+    public static SummaryReportInterval nextSummaryReportInterval(String reportInterval, boolean daily) {
+        if (daily) {
+            return SummaryReportInterval.DAILY;
+        }else if (reportInterval == null) {
+            return SummaryReportInterval.NONE;
+        }
+        switch (reportInterval) {
+            case "daily" -> {
+                return SummaryReportInterval.DAILY;
+            }
+            case "weekly" -> {
+                return SummaryReportInterval.WEEKLY;
+            }
+            case "monthly" -> {
+                return SummaryReportInterval.MONTHLY;
+            }
+            case "yearly" -> {
+                return SummaryReportInterval.YEARLY;
+            }
+            default -> {}
+        }
+        return SummaryReportInterval.NONE;
+    }
+
+    public static ZonedDateTime startOfZDTDay(ZonedDateTime date) {
+        return date.truncatedTo(ChronoUnit.DAYS);
+    }
+
+    public static ZonedDateTime startOfZDTWeek(ZonedDateTime date) {
+        return date.with(DayOfWeek.MONDAY).truncatedTo(ChronoUnit.DAYS);
+    }
+
+    public static ZonedDateTime startOfZDTMonth(ZonedDateTime date) {
+        return date.withDayOfMonth(1).truncatedTo(ChronoUnit.DAYS);
+    }
+
+    public static ZonedDateTime startOfZDTYear(ZonedDateTime date) {
+        return date.withDayOfYear(1).truncatedTo(ChronoUnit.DAYS);
     }
 
 }
--- a/src/main/java/org/traccar/helper/DateUtil.java
+++ b/src/main/java/org/traccar/helper/DateUtil.java
@@ -94,7 +94,7 @@ public final class DateUtil {
     public static SummaryReportInterval nextSummaryReportInterval(String reportInterval, boolean daily) {
         if (daily) {
             return SummaryReportInterval.DAILY;
-        }else if (reportInterval == null) {
+        } else if (reportInterval == null) {
             return SummaryReportInterval.NONE;
         }
         switch (reportInterval) {
--- a/src/main/java/org/traccar/helper/DateUtil.java
+++ b/src/main/java/org/traccar/helper/DateUtil.java
@@ -83,38 +83,6 @@ public final class DateUtil {
         }
     }
 
-    public enum SummaryReportInterval {
-        NONE,
-        DAILY,
-        WEEKLY,
-        MONTHLY,
-        YEARLY
-    }
-
-    public static SummaryReportInterval nextSummaryReportInterval(String reportInterval, boolean daily) {
-        if (daily) {
-            return SummaryReportInterval.DAILY;
-        } else if (reportInterval == null) {
-            return SummaryReportInterval.NONE;
-        }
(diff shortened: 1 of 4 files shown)


src/main/java/org/traccar/helper/DateUtil.java
```diff
@@ -80,4 +83,52 @@ public static String formatDate(Date date, boolean zoned) {
...
+    public enum SummaryReportInterval {
+        NONE,
+        DAILY,
+        WEEKLY,
+        MONTHLY,
+        YEARLY
+    }
+
+    public static SummaryReportInterval nextSummaryReportInterval(String reportInterval, boolean daily) {
+        if (daily) {
```
tananaev: Why does this have some daily parameter?

src/main/java/org/traccar/helper/DateUtil.java
```diff
@@ -80,4 +83,52 @@ public static String formatDate(Date date, boolean zoned) {
...
 
+    public enum SummaryReportInterval {
+        NONE,
+        DAILY,
+        WEEKLY,
+        MONTHLY,
+        YEARLY
+    }
+
+    public static SummaryReportInterval nextSummaryReportInterval(String reportInterval, boolean daily) {
```
tananaev: Why generic date util know about summary report? And why is interval have to be specific to summary?

tananaev:
Based on my first brief look it doesn't seem like the issue is fixed. Added some comments, but we would need much cleaner solution before we can merge.

src/main/java/org/traccar/helper/DateUtil.java
harrowmykel: The old api support boolean "daily" as query parameter. So that we convert this boolean into enumeration value in a single place, it is passed.

src/main/java/org/traccar/helper/DateUtil.java
harrowmykel: Indeed, Interval can be generic, however since we are fixing summary. My first instinct was to make it summary specific. 

The enum is placed in DateUtil since it is an helper that relates to a date/time period conversion

src/main/java/org/traccar/helper/DateUtil.java
tananaev: It definitely doesn't belong here. Didn't we already discuss that we shouldn't be support backward compatibility? Why did you create a completely new set of PRs? I feel like we lost some discussions there.

harrowmykel: Yeah. We had a previous discussion in #5924 but there was a weird traccar-web file in the changed files, that i thought you were complaining about. 

Will put it in mind not to retain backward compatibility.

harrowmykel: Where would you put the enum and the string to enum conversion function?

tananaev: Can we use something standard, like `ChronoUnit`?

harrowmykel: I can do that. Chronounit was my first solution as well until i changed it because i felt chronounit provided too many values than i needed.

Should the helper function still remain in dateutil or moved to summaryreportprovider class as a private function?

tananaev: Please think what makes sense from good engineering point of view.

harrowmykel: I will leave it in dateutil but make it generic so it can be used anywhere else by changing it to 

```java
public static ChronoUnit intervalToChronoUnit(String interval)
....
```

tananaev: There's no standard conversion?

harrowmykel: There is no standard conversion of string period intervals in this codebase. 

I think i just will rename the function 
```java 
public static ChronoUnit parseDateInterval(String interval)

```
And add it to helper/Parser.java file since most generic datetime parsers are there

tananaev: OK, let's see what it looks like.

harrowmykel: added the changes

src/main/java/org/traccar/reports/SummaryReportProvider.java
```diff
@@ -123,17 +124,49 @@ private Collection<SummaryReportItem> calculateDeviceResult(
...
+                case ChronoUnit.DAYS -> {
+                    while (DateUtil.startOfZDTDay(from).isBefore(DateUtil.startOfZDTDay(to))) {
+                        ZonedDateTime fromDay = DateUtil.startOfZDTDay(from);
+                        ZonedDateTime nextDay = fromDay.plusDays(1);
+                        results.addAll(calculateDeviceResult(device, Date.from(from.toInstant()),
+                            Date.from(nextDay.toInstant()), fast));
+                        from = nextDay;
+                    }
+                }
+                case ChronoUnit.WEEKS -> {
```
tananaev, line 142: There has to be a better way. You're using `ChronoUnit`, but then you're still hardcoding every case.

src/main/java/org/traccar/helper/Parser.java
```diff
@@ -362,4 +363,26 @@ public Date nextDateTime() {
         return nextDateTime(DateTimeFormat.YMD_HMS, null);
     }
 
+    public static ChronoUnit parseDateInterval(String reportInterval) {
```
tananaev, line 366: Why is this part of parser?

src/main/java/org/traccar/helper/Parser.java
```diff
@@ -362,4 +363,26 @@ public Date nextDateTime() {
...
+        if (reportInterval == null) {
+            return null;
+        }
+        switch (reportInterval) {
```
tananaev: Use return switch.

src/main/java/org/traccar/helper/DateUtil.java
```diff
@@ -80,4 +83,20 @@ public static String formatDate(Date date, boolean zoned) {
...
+        return date.with(DayOfWeek.MONDAY).truncatedTo(ChronoUnit.DAYS);
+    }
+
+    public static ZonedDateTime startOfZDTMonth(ZonedDateTime date) {
+        return date.withDayOfMonth(1).truncatedTo(ChronoUnit.DAYS);
+    }
+
+    public static ZonedDateTime startOfZDTYear(ZonedDateTime date) {
+        return date.withDayOfYear(1).truncatedTo(ChronoUnit.DAYS);
+    }
```
tananaev, line 100: Why do we need this?

harrowmykel: when we get weekly, monthly intervals - we need to find the first of each month, week, year between from and to as applicable.
These are helper functions

tananaev: Yeah, but they're all used in a single place. I think the breakdown of what should be helper functions is not right.

PR #5936, opened 2026-06-30
Limit MemoryStorage to latest position per device

BastyJuice:
**The implementation is intentionally minimal and only affects `MemoryStorage`.**

Currently, when `database.memory=true` is used, every stored position is kept in memory. For high-traffic deployments this can cause the in-memory storage to grow over time.

Following the suggested approach, this PR updates `MemoryStorage` to keep only the latest position per device.

### What Changed

This change updates `MemoryStorage` position handling:

1. When a new `Position` is added, existing positions for the same device are checked.
2. If there is no newer position for the device, older positions for that device are removed and the new position is stored.
3. If a newer position already exists, the new older position is not stored.
4. SQL-backed storage remains unchanged.

### Benefits

1. Prevents unbounded in-memory growth from stored positions when `database.memory=true`.
2. Preserves the latest known position for each device.
3. Keeps the existing persistence flow intact.
4. Does not add any new configuration options.
5. Does not affect SQL-backed storage.

### Notes

The comparison only uses `fixTime`, keeping the implementation small and local to `MemoryStorage`.

### DIFF
--- a/src/main/java/org/traccar/storage/MemoryStorage.java
+++ b/src/main/java/org/traccar/storage/MemoryStorage.java
@@ -205,7 +205,7 @@ public class MemoryStorage extends Storage {
         var items = objects.computeIfAbsent(entity.getClass(), key -> new ConcurrentHashMap<>());
 
         if (entity instanceof Position position) {
-            Iterator<Object> iterator = items.values().iterator();
+            var iterator = items.values().iterator();
             while (iterator.hasNext()) {
                 Position item = (Position) iterator.next();
                 if (item.getDeviceId() == position.getDeviceId()) {
--- a/src/main/java/org/traccar/storage/MemoryStorage.java
+++ b/src/main/java/org/traccar/storage/MemoryStorage.java
@@ -43,7 +43,6 @@ public class MemoryStorage extends Storage {
 
     private final Map<Class<?>, Map<Long, Object>> objects = new ConcurrentHashMap<>();
     private final Map<Pair<Class<?>, Class<?>>, Set<Pair<Long, Long>>> permissions = new ConcurrentHashMap<>();
-    private final Map<Long, Long> latestPositionIds = new ConcurrentHashMap<>();
 
     private final AtomicLong increment = new AtomicLong();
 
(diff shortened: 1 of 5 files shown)


tananaev:
Is your main issue with memory enabled? Then I think the solution should be different. I think we should just limit how many data points is stored.

BastyJuice:
Yes, the main issue is with database.memory=true.

My reasoning was that in a forwarding-only setup I don’t really need Traccar to keep the forwarded positions locally, because the receiving backend stores and processes them.

Since Traccar already has several forwarding-related options, I thought an explicit option to skip local persistence could fit this use case as well.

But I understand your point. A limit for data points stored in memory would be more generic and would preserve the normal Traccar behavior better than disabling persistence completely.

If that is the preferred approach, I can rework the PR to limit the number of positions stored when database.memory=true.

tananaev:
I would basically store only the latest one. It also probably makes sense to just enable it by default and not make it configurable.

BastyJuice:
Thanks Anton, that makes sense.

I reworked the PR accordingly. The new configuration options have been removed, and the change now only applies to MemoryStorage.

MemoryStorage now keeps only the latest position per device by default when database.memory=true is used. SQL-backed storage remains unchanged.

It also avoids replacing the stored latest position with older out-of-order data by comparing fixTime and serverTime.

PS: I also kept the implementation inline and minimal.

tananaev:
I think what you have is very much over-engineered. The only time you need to check is fixed time. I also don't think we need `latestPositionIds`.

BastyJuice:
I removed latestPositionIds and only compare wuith fixTime Now. The change is limited to Memorysorage and does not add any configuration options.

src/main/java/org/traccar/storage/MemoryStorage.java
```diff
@@ -200,7 +201,26 @@ private Object retrieveValue(Object object, String key) {
     @Override
     public <T> long addObject(T entity, Request request) {
         long id = increment.incrementAndGet();
-        objects.computeIfAbsent(entity.getClass(), key -> new ConcurrentHashMap<>()).put(id, entity);
+        var items = objects.computeIfAbsent(entity.getClass(), key -> new ConcurrentHashMap<>());
+
+        if (entity instanceof Position position) {
+            boolean latest = true;
+            for (Object item : items.values()) {
```
tananaev: Can we use an iterator, so that we can remove old position immediately without doing another lookup?

tananaev: That way we can probably also skip `latest` variable.

BastyJuice: Done. I simplified the implementation as suggested. It now uses an iterator, removes older positions immediately during iteration, and skips the latest variable.

src/main/java/org/traccar/storage/MemoryStorage.java
```diff
@@ -200,7 +202,24 @@ private Object retrieveValue(Object object, String key) {
...
+                if (item.getDeviceId() == position.getDeviceId()) {
+                    if (item.getFixTime().after(position.getFixTime())) {
+                        return id;
+                    }
+                    iterator.remove();
+                }
+            }
+            items.put(id, entity);
+        } else {
+            items.put(id, entity);
```
tananaev: Why do we need to duplicate this?

src/main/java/org/traccar/storage/MemoryStorage.java
```diff
@@ -200,7 +202,24 @@ private Object retrieveValue(Object object, String key) {
     @Override
     public <T> long addObject(T entity, Request request) {
         long id = increment.incrementAndGet();
-        objects.computeIfAbsent(entity.getClass(), key -> new ConcurrentHashMap<>()).put(id, entity);
+        var items = objects.computeIfAbsent(entity.getClass(), key -> new ConcurrentHashMap<>());
+
+        if (entity instanceof Position position) {
+            Iterator<Object> iterator = items.values().iterator();
```
tananaev:
```suggestion
            var iterator = items.values().iterator();
```

src/main/java/org/traccar/storage/MemoryStorage.java
```diff
@@ -200,7 +202,24 @@ private Object retrieveValue(Object object, String key) {
...
+            while (iterator.hasNext()) {
+                Position item = (Position) iterator.next();
+                if (item.getDeviceId() == position.getDeviceId()) {
+                    if (item.getFixTime().after(position.getFixTime())) {
+                        return id;
+                    }
+                    iterator.remove();
```
tananaev, line 214: Probably need to add break?

src/main/java/org/traccar/storage/MemoryStorage.java
BastyJuice: That was my copy-paste enthusiasm showing. Fixed now, the put call is no longer duplicated.

src/main/java/org/traccar/storage/MemoryStorage.java
BastyJuice: Have a break, have a KitKat :-) Done.

PR #5937, opened 2026-07-01
Allow generating session tokens for other users

Eruqb13:
## Summary
Add optional `userId` parameter to `POST /api/session/token` so
admin or manager accounts can request an access token on behalf of
another user without logging in as them.

## Changes
- `SessionResource.java`: new optional `userId` form param; permission
  check delegated to `PermissionsService.checkUser` (admin / manager
  with managed-user permission).
- `openapi.yaml`: document the new `userId` parameter.

## Backwards compatibility
The `userId` parameter is optional. Existing clients that omit it
keep working exactly as before (token is generated for the
authenticated user).

## Testing
- Admin can request tokens for any user.
- Manager can request tokens for their managed users (validated by
  existing `checkUser` logic).
- Non-admin / non-manager receives 403 when targeting another user.

### DIFF
--- a/src/main/java/org/traccar/api/resource/SessionResource.java
+++ b/src/main/java/org/traccar/api/resource/SessionResource.java
@@ -145,14 +145,18 @@ public class SessionResource extends BaseResource {
     @Path("token")
     @POST
     public String requestToken(
-            @FormParam("expiration") Date expiration) throws StorageException, GeneralSecurityException, IOException {
+            @FormParam("expiration") Date expiration,
+            @FormParam("userId") Long userId) throws StorageException, GeneralSecurityException, IOException {
+        long currentUserId = getUserId();
+        long targetUserId = userId != null ? userId : currentUserId;
+        permissionsService.checkUser(currentUserId, targetUserId);
         Date currentExpiration = (Date) request.getSession().getAttribute(SessionHelper.EXPIRATION_KEY);
-        if (currentExpiration != null && currentExpiration.before(expiration)) {
+        if (currentExpiration != null && (expiration == null || currentExpiration.before(expiration))) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/api/resource/SessionResource.java
```diff
@@ -145,14 +145,18 @@ public Response remove() {
     @Path("token")
     @POST
     public String requestToken(
-            @FormParam("expiration") Date expiration) throws StorageException, GeneralSecurityException, IOException {
+            @FormParam("expiration") Date expiration,
+            @FormParam("userId") Long userId) throws StorageException, GeneralSecurityException, IOException {
+        long currentUserId = getUserId();
```
tananaev, line 150: I would inline this.

src/main/java/org/traccar/api/resource/SessionResource.java
```diff
@@ -145,14 +145,18 @@ public Response remove() {
...
+        long targetUserId = userId != null ? userId : currentUserId;
```
tananaev, line 151: And reuse userId variable here.

src/main/java/org/traccar/api/resource/SessionResource.java
```diff
@@ -145,14 +145,18 @@ public Response remove() {
...
+        permissionsService.checkUser(currentUserId, targetUserId);
         Date currentExpiration = (Date) request.getSession().getAttribute(SessionHelper.EXPIRATION_KEY);
-        if (currentExpiration != null && currentExpiration.before(expiration)) {
+        if (currentExpiration != null && (expiration == null || currentExpiration.before(expiration))) {
```
tananaev, line 154: Why do we need this check?

PR #5940, opened 2026-07-03
Forgotten BCE decoder issue where some BCE packets could fail completely

BastyJuice:
**The implementation is intentionally minimal and only affects `BceProtocolDecoder`.**

Currently, BCE coordinates are written directly to `Position`. If a BCE record contains an invalid coordinate value, `Position.setLatitude` or `Position.setLongitude` throws an exception.

In the affected sample, one record contains a longitude outside the valid `-180..180` range. This causes decoding to fail with `Longitude out of range`.

This case was already present in the codebase as a disabled test with the comment `Needs to be fixed`.

This PR fixes that case.

### What Changed

This change updates BCE coordinate handling:

1. Longitude and latitude are read into local variables first.
2. The values are checked before they are set on `Position`.
3. The position is only marked as valid if both coordinates are in range.
4. Invalid BCE coordinate values no longer throw from the decoder.
5. The existing disabled BCE test case has been enabled again.

### Benefits

1. Fixes the known BCE `Longitude out of range` failure.
2. Keeps the existing strict validation in `Position`.
3. Keeps the change local to `BceProtocolDecoder`.
4. Keeps the implementation small.
5. Adds coverage for the known failing BCE sample.

### Verification

`./gradlew test --tests org.traccar.protocol.BceProtocolDecoderTest --tests org.traccar.protocol.BceFrameDecoderTest`

### Question

While checking BCE, I noticed that the digital input value is currently stored only as the raw `input` bitmask.

Would it make sense to also expose this 16-bit bitmask as `in1` through `in16`?

Suggested implementation:

`int input = buf.readUnsignedShortLE();`
`position.set(Position.KEY_INPUT, input);`
`for (int i = 0; i < 16; i++) {`
`    position.set(Position.PREFIX_IN + (i + 1), BitUtil.check(input, i));`
`}`

This would keep the existing `input` value unchanged and additionally expose the individual input states.


tananaev:
This is not the right "fix".

BastyJuice:
Thanks for checking. 

I checked the BCE protocol document again: https://www.traccar.org/protocol/5080-bce/SimpleBCEdeviceprotocol.pdf

The current coordinate group decoding seems to match the specification. The document describes `Coord. group1` as 17 bytes, which matches the current decoder structure:

- longitude float LE
- latitude float LE
- speed
- status
- course
- altitude
- odometer

Do you mean that the decoder should not silently ignore the invalid coordinate value here?

The failing sample contains one record where the longitude field decodes to a value outside the valid range, while the rest of the BCE structure still appears to parse correctly. My intention was to avoid dropping the whole message because of that one invalid coordinate.

Would you prefer skipping the whole record instead, or do you think the BCE payload is being decoded incorrectly before this point?

tananaev:
We need to figure out why coordinates are invalid there. We should not just silence the error.

BastyJuice:
I have a lot of BCE Devices in my Fleet so i will try to collect raw BCE packets from the affected device/logs and compare the decoded structure, masks, and coordinate fields.

BastyJuice:
I collected a raw BCE packet from production and decoded it against the current structure.

The parser does not seem to be generally misaligned. In the same BCE message, the first full GPS record decodes to an invalid longitude:

`lon = 1295.2332763671875`
`lat = 49.40214920043945`

The following records in the same message decode correctly, for example:

`lon = 7.55549955368042`
`lat = 49.40235900878906`

and several more valid records after that.

So I agree that silently ignoring invalid coordinates on the same `Position` is probably not the right fix. A better approach would be to skip only the affected BCE record when `Coord. group1` contains out-of-range coordinates, while continuing to decode the remaining records in the same BCE message.

Would that approach be acceptable?

tananaev:
So you're saying device is actually reporting invalid coordinates? Or we're not decoding something correctly?

BastyJuice:
Based on the raw frames I collected, it looks like the device is reporting invalid coordinates in one record, while the rest of the same BCE message decodes correctly.

I decoded two production frames manually. In both cases the structure stays aligned and the following GPS records are valid.

Example 1:

- IMEI: `867459045613205`
- GPS records: `10`
- valid GPS records: `9`
- invalid GPS records: `1`
- invalid record longitude: `1303.0`
- invalid record latitude: `49.342384338378906`

Example 2:

- IMEI: `867459045614344`
- GPS records: `10`
- valid GPS records: `9`
- invalid GPS records: `1`
- invalid record longitude: `1295.2332763671875`
- invalid record latitude: `49.40214920043945`

In both frames the invalid and valid GPS records use the same masks:

- mask1: `0xc003`
- mask2: `0x8000`
- mask3: `0x0040`

The following records in the same frames decode to plausible coordinates around `7.55, 49.40`, so I do not see a general offset issue in the current decoder.

So my current assumption is that the device reports invalid coordinates for one record, but I cannot fully rule out an undocumented BCE variant or some special case in the first GPS record.

tananaev:
Can you provide the sample?

BastyJuice:
Sure, here is one raw BCE frame that reproduces the issue:

```text
0863450cf31403003002a5820a367d64166f07434352302e377d641603c00080400077e7a144cd9b4542131cad560100000000f880060102ad0279a2003b00000000000000000a567d64166f07434352302e577d641603c000804000a7c6f140049c4542141caa590100000000f980060102ad0279a2003b00000000000000000a667d641600a0434352300a967d64166f07434352302e977d641603c00080400083c4f140819c4542181caa5d0100000000f980060102ad0279a2003b00000000000000000aa67d64166f07434352302ea77d641603c00080400005c4f140a09c4542171caa5d0100000000f980060102ad0279a2003b00000000000000000aa67d641600a0434352300ac67d641600a0434352302e477e641603c000804000d6bff140c09d4542131c05610100000000f980060102ad0279a2004700000000000000002e577e641603c00080400069c0f140d99d4542141c11610100000000f980060102ad0279a2004700000000000000002e777

File: src/main/java/org/traccar/protocol/BceProtocolDecoder.java [hunk 1]
```diff
@@ -52,9 +52,13 @@ public BceProtocolDecoder(Protocol protocol) {
     private void decodeMask1(ByteBuf buf, int mask, Position position) {
 
         if (BitUtil.check(mask, 0)) {
-            position.setValid(true);
-            position.setLongitude(buf.readFloatLE());
-            position.setLatitude(buf.readFloatLE());
+            double longitude = buf.readFloatLE();
+            double latitude = buf.readFloatLE();
+            if (longitude >= -180 && longitude <= 180 && latitude >= -90 && latitude <= 90) {
```
tananaev: Instead of doing this, let's add a catch for `IllegalArgumentException`.

src/test/java/org/traccar/protocol/BceProtocolDecoderTest.java
```diff
@@ -1,19 +1,16 @@
...
 import org.junit.jupiter.api.Test;
 import org.traccar.ProtocolTest;
 import org.traccar.model.Position;
 
 public class BceProtocolDecoderTest extends ProtocolTest {
 
-    @Disabled
     @Test
-    public void testDecodeFail() throws Exception {
+    public void testDecodeInvalidLocation() throws Exception {
```
tananaev: Move the test to normal `testDecode`.

tananaev: Can you also add your new test sample.

BastyJuice: sure done it :-)

BastyJuice:
should be done now

src/main/java/org/traccar/protocol/BceProtocolDecoder.java
```diff
@@ -302,30 +302,34 @@ protected Object decode(
...
+                            decodeMask3(buf, mask, position);
+                        }
+
+                        if (masks.size() >= 4) {
+                            mask = masks.get(3);
+                            decodeMask4(buf, mask, position);
+                        }
+                    } catch (IllegalArgumentException e) {
+                        position = null;
                     }
```
tananaev: This doesn't make sense. Why do we to wrap all this in try? Why do we set position to null?

BastyJuice: The reason I originally wrapped the full mask decoding was to skip the whole affected BCE record when one coordinate value was invalid.

From the production samples, one GPS record in the BCE message contained an invalid longitude, but the following records in the same message decoded correctly. My goal was to avoid dropping the entire BCE message because of one bad record.

Setting `position = null` was meant as a simple way to prevent that affected record from being added to the result list after the reader index was moved to `structEnd`.

But I agree this is too broad. It catches more than just invalid coordinates and makes the control flow less clear.

I updated it to catch `IllegalArgumentException` only around the coordinate decoding in `decodeMask1`, which is the actual failing part.

src/main/java/org/traccar/protocol/BceProtocolDecoder.java
```diff
@@ -52,9 +52,14 @@ public BceProtocolDecoder(Protocol protocol) {
     private void decodeMask1(ByteBuf buf, int mask, Position position) {
 
         if (BitUtil.check(mask, 0)) {
-            position.setValid(true);
-            position.setLongitude(buf.readFloatLE());
-            position.setLatitude(buf.readFloatLE());
+            int index = buf.readerIndex();
```
tananaev: Index usually means something else, so I would rename this to maybe `startIndex` or something like that. Otherwise looks good.

tananaev:
Thanks.

PR #5942, opened 2026-07-07
Fix NullPointerException in PathAwayProtocolDecoder

### LINKED ISSUE #5933: PathAway Protocol NullPointerException
Error:
```
WARN: Cannot invoke "java.util.List.get(int)" because the return value of 
"java.util.Map.get(Object)" is null - NullPointerException
(PathAwayProtocolDecoder:67 < ExtendedObjectDecoder:72)
```
What happened:

A scanner sent GET / HTTP/1.0 to the PathAway protocol port. The PathAwayProtocolDecoder at line 67 attempts to call .get(int) on the result of a Map lookup without first checking whether the map returned null for that key.

Root Cause:

Missing null guard in PathAwayProtocolDecoder:67. The decoder assumes a required map entry always exists, but the HTTP request payload produced a parsing state where the key was absent.

Impact:

The server handles it gracefully — the session disconnects cleanly with no crash
No real PathAway device is affected since this only triggers on malformed HTTP input
However, if a real PathAway device sent an unexpected or truncated first packet, the same NPE could occur, causing that device's session to drop silently

bhavanagangavaram:
### Problem
PathAwayProtocolDecoder throws a NullPointerException at line 67 when receiving malformed HTTP requests (e.g., GET / HTTP/1.0) that lack the required UserName and LOC query parameters.

The root cause is that decoder.parameters().get(UserName) returns 
ull when the parameter is absent, and calling .get(0) on 
ull causes the NPE.

### Changes
**PathAwayProtocolDecoder.java**:
- Extract UserName and LOC parameter lists into local variables
- Add null/empty checks before accessing .get(0) — return 
ull gracefully if either parameter is missing
- Added import java.util.List

**PathAwayProtocolDecoderTest.java**:
- Added test for bare GET / request (no query params) — verifies null return, no NPE
- Added test for request with UserName only (no LOC) — verifies null return, no NPE

### Impact
- Malformed/scanner HTTP requests are now handled gracefully (return null, session disconnects cleanly)
- No behavior change for valid PathAway device requests
- Prevents silent session drops if a real PathAway device sends a truncated first packet

Fixes #5933

### DIFF
--- a/src/main/java/org/traccar/protocol/PathAwayProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/PathAwayProtocolDecoder.java
@@ -32,6 +32,7 @@ import org.traccar.helper.PatternBuilder;
 import org.traccar.model.Position;
 
 import java.net.SocketAddress;
+import java.util.List;
 import java.util.regex.Pattern;
 
 public class PathAwayProtocolDecoder extends BaseProtocolDecoder {
@@ -63,13 +64,23 @@ public class PathAwayProtocolDecoder extends BaseProtocolDecoder {
         FullHttpRequest request = (FullHttpRequest) msg;
         QueryStringDecoder decoder = new QueryStringDecoder(request.uri());
 
+        List<String> userNames = decoder.parameters().get("UserName");
+        if (userNames == null || userNames.isEmpty()) {
+            return null;
+        }
+
         DeviceSession deviceSession = getDeviceSession(
-                channel, remoteAddress, decoder.parameters().get("UserName").get(0));
+                channel, remoteAddress, userNames.get(0));
         if (deviceSession == null) {
             return null;
         }
 
-        Parser parser = new Parser(PATTERN, decoder.parameters().get("LOC").get(0));
+        List<String> locations = decoder.parameters().get("LOC");
+        if (locations == null || locations.isEmpty()) {
+            return null;
+        }
+
+        Parser parser = new Parser(PATTERN, locations.get(0));
         if (!parser.matches()) {
             return null;
         }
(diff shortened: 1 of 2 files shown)


src/test/java/org/traccar/protocol/PathAwayProtocolDecoderTest.java
```diff
@@ -9,7 +9,11 @@ public class PathAwayProtocolDecoderTest extends ProtocolTest {
     public void testDecode() throws Exception {
 
         var decoder = inject(new PathAwayProtocolDecoder(null));
-        
+
+        verifyNull(decoder, request("/"));
+
+        verifyNull(decoder, request("/?UserName=name"));
```
tananaev, line 15: Is this real data?

src/main/java/org/traccar/protocol/PathAwayProtocolDecoder.java
```diff
@@ -63,13 +64,23 @@ protected Object decode(
         FullHttpRequest request = (FullHttpRequest) msg;
         QueryStringDecoder decoder = new QueryStringDecoder(request.uri());
 
+        List<String> userNames = decoder.parameters().get("UserName");
+        if (userNames == null || userNames.isEmpty()) {
```
tananaev, line 68: Why do we need to check for both?

src/main/java/org/traccar/protocol/PathAwayProtocolDecoder.java
```diff
@@ -63,13 +64,23 @@ protected Object decode(
...
         DeviceSession deviceSession = getDeviceSession(
-                channel, remoteAddress, decoder.parameters().get("UserName").get(0));
+                channel, remoteAddress, userNames.get(0));
         if (deviceSession == null) {
             return null;
         }
 
-        Parser parser = new Parser(PATTERN, decoder.parameters().get("LOC").get(0));
+        List<String> locations = decoder.parameters().get("LOC");
+        if (locations == null || locations.isEmpty()) {
```
tananaev, line 79: Same here.

PR #5945, opened 2026-07-10
Add Matrix notificator

MaBurn:
# Add Matrix notificator

## Background

We are a non-profit association (Verein) that has been using Traccar to track our vehicles and equipment, and we're very happy with it. For event notifications, however, none of the existing channels fit our needs:

- **WhatsApp and Telegram** are not acceptable to us for data-protection reasons — we don't want notification data (device names, positions, events) flowing through those services.
- **SMS** would add recurring per-message costs we'd rather avoid.

Our association already runs [Matrix](https://matrix.org) as its internal messenger, and we prefer it in general as an open, self-hostable and end-to-end-capable protocol. So we built a Matrix notificator for Traccar (with the help of Claude) and would like to contribute it back to the project.

## What this adds

A new `matrix` notificator that delivers notifications to a Matrix homeserver (Synapse, Dendrite, Conduit, …) via the Client-Server API v3. Messages are sent by a bot account identified by an access token.

- **Message body** uses the standard `NotificationMessage.digest()`, so notifications go through Traccar's existing notification templates — no custom formatting, consistent with the other notificators.
- **Default room:** if a user has no personal Matrix address configured, notifications go to `notificator.matrix.defaultRoomId`. This accepts either a room id (`!abc:example.org`) or a room alias (`#alerts:example.org`, resolved via the directory API).
- **Per-user delivery:** if a Traccar user has the `matrixUser` attribute set (e.g. `@jane:example.org`), the notificator looks for an existing room shared between the bot and that user. If none exists and `notificator.matrix.createUserRooms` is enabled, it creates a private direct room and invites the user. Resolved rooms are cached in memory.

## Configuration

```xml
<entry key='notificator.types'>web,mail,matrix</entry>
<entry key='notificator.matrix.homeserverUrl'>https://matrix.example.org</entry>
<entry key='notificator.matrix.accessToken'>syt_...</entry>
<entry key='notificator.matrix.defaultRoomId'>#alerts:example.org</entry>
<entry key='notificator.matrix.createUserRooms'>true</entry>
```

## Companion UI change

The web UI counterpart (adding the `matrixUser` user attribute and the notificator label) is in a separate pull request against `traccar/traccar-web`: .

## Testing

This has been running in our own deployment against a Synapse homeserver, covering both delivery paths: the shared default room and per-user direct rooms, including automatic room creation and reuse of existing/cached rooms.

We're of course happy to adjust anything to match the project's conventions.

### DIFF
--- a/src/main/java/org/traccar/notificators/NotificatorMatrix.java
+++ b/src/main/java/org/traccar/notificators/NotificatorMatrix.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright 2026 Anton Tananaev (anton@traccar.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.notificators;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+import jakarta.ws.rs.client.Client;
+import jakarta.ws.rs.client.Entity;
+import jakarta.ws.rs.client.WebTarget;
+import jakarta.ws.rs.core.Response;
+import org.traccar.config.Config;
+import org.traccar.config.Keys;
+import org.traccar.model.Event;
+import org.traccar.model.Position;
+import org.traccar.model.User;
+import org.traccar.notification.MessageException;
+import org.traccar.notification.NotificationFormatter;
+import org.traccar.notification.NotificationMessage;
+
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+
+@Singleton
+public class NotificatorMatrix extends Notificator {
+
+    private static final String USER_ATTRIBUTE = "matrixUser";
+
+    private final Client client;
+    private final String homeserverUrl;
+    private final String accessToken;
+    private final Map<String, String> userRooms = new ConcurrentHashMap<>();
+
+    public static class JoinedRoomsResponse {
+        @JsonProperty("joined_rooms")
+        private List<String> joinedRooms = List.of();
+    }
+
+    public static class JoinedMembersResponse {
+        @JsonProperty("joined")
+        private Map<String, Object> joined = Map.of();
+    }
+
+    public static class CreateRoomRequest {
+        @JsonProperty("preset")
+        private String preset = "trusted_private_chat";
+        @JsonProperty("is_direct")
+        private boolean direct = true;
+        @JsonProperty("invite")
+        private List<String> invite;
+    }
+
+    public static class CreateRoomResponse {
+        @JsonProperty("room_id")
+        private String roomId;
+    }
+
+    public static class Message {
+        @JsonProperty("msgtype")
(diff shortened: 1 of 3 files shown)


tananaev:
Can you provide the API documentation.

MaBurn:
The extension uses the Matrix Client-Server API (v3). 

The full specification is available at https://spec.matrix.org/latest/client-server-api/. 

The specific endpoints used are 
`PUT /rooms/{roomId}/send/{eventType}/{txnId}` (send message), 
`GET /joined_rooms` and `GET /rooms/{roomId}/joined_members` (find a user's room), 
`POST /createRoom` (create a direct room),
`GET /directory/room/{roomAlias}` (resolve a room alias).

src/main/java/org/traccar/notificators/NotificatorMatrix.java
```diff
@@ -0,0 +1,248 @@
...
+
+    private static String trimTrailingSlash(String value) {
+        value = normalize(value);
+        if (value != null && value.endsWith("/")) {
+            return value.substring(0, value.length() - 1);
+        }
+        return value;
+    }
+
+    private static String normalize(String value) {
```
tananaev, line 187: Why is this needed?

src/main/java/org/traccar/notificators/NotificatorMatrix.java
```diff
@@ -0,0 +1,248 @@
...
+        }
+    }
+
+    private void validateConfiguration() throws MessageException {
+        if (homeserverUrl == null || accessToken == null) {
+            throw new MessageException("Missing Matrix configuration");
+        }
+    }
+
+    private String resolveRoomId(String matrixUser) throws MessageException {
```
tananaev, line 115: Why do we need such complicated logic. That doesn't align with what we do for other notificators, especially creating a new room.

src/main/java/org/traccar/notificators/NotificatorMatrix.java
MaBurn: Good catch — that was unnecessary defensive code. Removed it; the homeserver URL is now used as configured, consistent with the other notificators.

src/main/java/org/traccar/notificators/NotificatorMatrix.java
MaBurn: I've trimmed this down. There's no longer a default room and no alias resolution — every notification now goes to a per-user room, keyed by the user's matrixUser attribute.

The per-user room resolution is, unfortunately, unavoidable for Matrix specifically. Unlike Telegram's chat id, Matrix has no "send to a user" primitive — you can only send an event to a room. So to notify a user, the bot needs a room it shares with them. The remaining logic is just that:

reuse the cached room id, otherwise
find an existing room the bot already shares with the user (joined_rooms + joined_members), otherwise
create a private room and invite the user.
Step 2 is the important one: without it we'd have to create a brand-new room for every notification (Matrix would happily do that, leaving the user buried in duplicate rooms) or persist room ids ourselves. Discovery keeps a single stable room per user and survives restarts, since the in-memory cache is only an optimization.

It's also considerably more user-friendly: the admin only needs to set each user's Matrix id, and the bot resolves or provisions the room automatically — no manual room creation, no copying room ids into user profiles. Step 3 (auto-creation) is what makes that hands-off setup possible. Happy to adjust the trade-off if you'd prefer something else.

tananaev: Telegram is also only to a room.

MaBurn: I can change it to use Matrix room ID directly.

My intention behind using only the user's Matrix ID and resolving the room automatically was to make the setup easier for users who may not be familiar with Matrix room IDs.

Just to make sure I understand your preference correctly: would you prefer the implementation to stay as simple and consistent as possible by using the room ID directly, rather than resolving the room automatically from the user's Matrix ID?

tananaev: Correct. Let's start with dead simple and consistent with existing implementation.

Issue #5892, opened 2026-06-01
Edit accumulator total distance not working when fix time is the same due to CacheManager changes?


timbze:
I tried updating total distance, and it seemed to work, but soon the change was reverted.

With some digging I think it may be because of CacheManager changes between 6.13.3 and 6.14.0. When fix time remains the same it takes the old position values instead of the newer value where I updated the total distance (I think).

tananaev:
This should be fixed in 6.14.2.

Issue #5882, opened 2026-05-20
Odometer/total distance resets after manual accumulator update


kundan175:
When I manually update a device’s odometer or total distance using the accumulator API, the value appears to be updated successfully at first, but after some time it gets overwritten or appears to revert back.

This seems to happen because the device processing flow recalculates distance from incoming positions later, so the manual update does not stay as the long-term source of truth.

**Steps to Reproduce**
Update a device’s accumulators using the API.
Set a new value for total distance and engine hours.
Wait for the device to send new position data.
Check the device again after some time.
The odometer/total distance may no longer match the manually updated value.

**Expected Behavior**
After updating the device accumulators, the updated odometer/total distance should remain consistent and should not be unexpectedly overwritten by later processing.

**Actual Behavior**
The value looks correct immediately after update, but later it changes or appears to go back.

**Possible Cause**
The accumulator update only changes the latest position record, while later incoming positions are processed by distance calculation logic and may overwrite the displayed total distance.

**Relevant Areas**
Device accumulator update endpoint
Distance calculation during position processing
Latest position cache/state handling

**Impact**
This causes inconsistent odometer readings and makes manual distance corrections unreliable.

**Suggested Fix**
The update flow should ensure that manual accumulator updates are preserved, or the distance calculation logic should respect the manually updated baseline.

tananaev:
Race condition is possible in some cases and it is expected currently due to the fact that the value is stored in positions.

PR #5948, opened 2026-07-15
Fix accumulators getting overwritten bug due to caching

timbze:
This fix makes sure that cache always picks the updated position as latest position.

Addresses the following 2 issues (already closed).

#5892
#5882

Side note: I wonder if it would be a better idea to store accumulator values in device table or another new table.

Note: Tests are mostly written by AI (Grok 4.5), but I have confirmed they both fail without fix, and pass with fix.

### DIFF
--- a/src/main/java/org/traccar/api/resource/DeviceResource.java
+++ b/src/main/java/org/traccar/api/resource/DeviceResource.java
@@ -49,6 +49,7 @@ import jakarta.ws.rs.core.Response;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.util.Date;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


src/test/java/org/traccar/handler/AccumulatorsDistanceTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.ws.rs.core.SecurityContext;
+
+/**
+ * Covers totalDistance reset after manual accumulator update when a status
+ * reply (same fixTime) is processed. Exercises DeviceResource.updateAccumulators
+ * against a live CacheManager so the serverTime refresh fix is actually required.
+ */
+public class AccumulatorsDistanceTest {
```
tananaev: this is overkill. drop

timbze: Removed test file now

timbze: Or did you not mean entire test file?

PR #5954, opened 2026-07-22
Include conflicting port in log

### LINKED ISSUE #5552: Log the conflicting port when BindException occurs
Currently, when a BindException is thrown while starting a TrackerConnector, the log message does not indicate which port caused the conflict. This makes it difficult to debug when multiple connectors are involved.

Proposed Improvement:
Enhance the error logging to include the port number in the message. For example:

mdiniz97:
Fixes #5552

Include the connector port in the warning logged when a BindException occurs, while preserving the original exception.

Expose the configured port through TrackerConnector so ServerManager can include it in the warning.

Testing:
- ./gradlew check

### DIFF
--- a/src/main/java/org/traccar/ServerManager.java
+++ b/src/main/java/org/traccar/ServerManager.java
@@ -73,7 +73,7 @@ public class ServerManager implements LifecycleObject {
             try {
                 connector.start();
             } catch (BindException e) {
-                LOGGER.warn("Port disabled due to conflict", e);
+                LOGGER.warn("Port {} disabled due to conflict", connector.getPort(), e);
             } catch (ConnectException e) {
                 LOGGER.warn("Connection failed", e);
             }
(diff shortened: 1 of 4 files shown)


src/test/java/org/traccar/ServerManagerTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class ServerManagerTest {
```
tananaev: I don't think we need a test for it.

mdiniz97: Removed the test.

PR #5965, opened 2026-07-31
Support {uniqueId} placeholder in custom commands

### LINKED ISSUE #5957: Placeholders for custom commands?
Hello,
I have two times the same GPS tracker (PAJ Pet Finder 4G) which I configured to talk to my traccar instance. Everything is working fine so far. But I want to save some custom commands to use for both devices, but I can't find anything regarding some placeholders.

Somthing like
`*HQ,{uniqueId},X2,062108,1#` to turn on the LED on the tracker. 

Is there really no placeholder and do I need to create every command twice for my devices?

Thanks in advance,
Dominic

ctucker1984:
## Summary
- Expand `{uniqueId}` in custom command `data` at send time from the target device.
- Central hook in `CommandsManager` so live, SMS, queue, group, and notificator sends all get the same behavior.
- Fixes https://github.com/traccar/traccar/issues/5957

## Test plan
- [x] Unit tests (`CommandUtilTest`)
- [x] Lab on 6.14.5: outbound log showed real device id substituted for `{uniqueId}`
- [x] Developer's Production Instance: one saved custom command works across multiple devices, each getting its own id
- [ ] Reviewer: one saved custom command, two devices, each gets its own id

Made with [Cursor](https://cursor.com)


src/main/java/org/traccar/helper/model/CommandUtil.java
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.traccar.helper.model;
+
+import org.traccar.model.Command;
+import org.traccar.model.Device;
+
+public final class CommandUtil {
```
tananaev: We don't need a separate class.

src/main/java/org/traccar/helper/model/CommandUtil.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private static final String PLACEHOLDER_UNIQUE_ID = "{uniqueId}";
```
tananaev: Inline.

src/main/java/org/traccar/helper/model/CommandUtil.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private CommandUtil() {
+    }
+
+    public static void expandPlaceholders(Command command, Device device) {
+        if (!Command.TYPE_CUSTOM.equals(command.getType())) {
```
tananaev: I think just gating by `KEY_DATA` is enough.

src/test/java/org/traccar/helper/model/CommandUtilTest.java
```diff
@@ -0,0 +1,67 @@
+package org.traccar.helper.model;
+
+import org.junit.jupiter.api.Test;
+import org.traccar.model.Command;
+import org.traccar.model.Device;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+public class CommandUtilTest {
```
tananaev: Not needed. It's a trivial implementation.

src/main/java/org/traccar/database/CommandsManager.java
```diff
@@ -79,6 +80,7 @@ public QueuedCommand sendCommand(Command command) throws Exception {
         long deviceId = command.getDeviceId();
         Device device = storage.getObject(Device.class, new Request(
                 new Columns.All(), new Condition.Equals("id", deviceId)));
+        CommandUtil.expandPlaceholders(command, device);
```
tananaev: Are there cases when we reuse command object? Like sending it to a group of devices? In that case it will break.

ctucker1984:
Thanks — addressed in the latest commit:

- Removed \`CommandUtil\` and the unit test; expansion is inlined in \`CommandsManager\`.
- Gated only on \`KEY_DATA\` (no type check).
- Copy via \`QueuedCommand.fromCommand(...).toCommand()\` before replacing \`{uniqueId}\`, so a shared/group template is not mutated across devices. \`CommandResource\` group send already copied per device; this makes \`sendCommand\` safe either way.

src/main/java/org/traccar/database/CommandsManager.java
ctucker1984: I think the intent here is to fire the command once per selected device, not across a whole group all at once.

tananaev: This method can be used for a group as well. That's why I'm asking. Have you traced it? I see you changed this, but is it actually needed?

ctucker1984: Sorry, I wasn't totally sure what you were getting at first — took me a sec to realize you meant whether the same `Command` object gets reused across devices, and that mutating `data` in place would leak one device's `{uniqueId}` into the next.

So I traced the callers: `CommandResource` (group + single) and `NotificatorCommand`. The group path already builds a fresh command per device with `QueuedCommand.fromCommand(entity).toCommand()` before calling `sendCommand`, and the other two only send once. So nothing actually reuses the object across devices.

That means the extra copy I'd added in `CommandsManager` wasn't needed — removed it and kept just the placeholder expansion.

meydominic:
Hey, any updates on this pr? :)

ctucker1984:
In case field testing is required, I was able to push this to my production instance at the time of authoring this PR. It's been running successfully ever since across multiple devices and users.

PR #5968, opened 2026-08-05
Add JT808 storage failure alarm

emmanuelh-dev:
## Problem

JT/T 1078 location extension `0x14` includes video-related alarm flags. Traccar currently stores the raw `videoAlarm` value, but a storage unit failure (bit 2) is not exposed as an alarm event. Extension `0x17`, which identifies the affected storage unit, is also skipped.

As a result, the failure is present in the device packet but cannot be shown or used as a distinct alarm in Traccar.

## Change

- add the `storageFailure` alarm value
- map bit 2 of JT808 extension `0x14` to the new alarm
- expose extension `0x17` as `storageAlarm`
- add regression coverage using a captured JT808 packet with `videoAlarm = 4` and `storageAlarm = 1`

## Related pull request

- Traccar Web translations: https://github.com/traccar/traccar-web/pull/1923

The backend pull request defines and emits the alarm value. The linked Traccar Web pull request only adds the corresponding display translations.

## Validation

- `./gradlew test --tests org.traccar.protocol.Jt808ProtocolDecoderTest`
- `./gradlew checkstyleMain`
- `git diff --check`


src/main/java/org/traccar/model/Position.java
```diff
@@ -123,6 +123,7 @@ public class Position extends Message {
     public static final String ALARM_LOW_POWER = "lowPower";
     public static final String ALARM_LOW_BATTERY = "lowBattery";
     public static final String ALARM_FAULT = "fault";
+    public static final String ALARM_STORAGE_FAILURE = "storageFailure";
```
tananaev, line 126: We don't add one off alarms like this. It has to be something used by at least two protocols.

emmanuelh-dev: Can I replace the new alarm with the generic `fault` alarm, keep `storageAlarm` as the decoded attribute?

tananaev: Generic alarm is probably fine, but it should be unconditional then.

PR #5969, opened 2026-08-05
MCP server: fix OAuth discovery/config bugs, harden startup, expand and optimize toolset

shanelord01:
## Heads up - this was AI-assisted ("vibe coded")

I built this with Claude Code, prompting and reviewing as I went rather than
writing every line myself. I've tested it against a real production instance
and it's been running stable, but I have not personally audited every line
against Traccar's conventions the way a from-scratch human PR would be. Please
review this thoroughly - I'd rather over-flag that up front than have it
treated as more vetted than it is. Happy to address anything in review.

## Motivation

This started from friction documented across a multi-page forum thread trying
to get the MCP server working:
https://www.traccar.org/forums/topic/mcp-function-how-can-the-documentation-can-found/

Several people (including me) hit the same OAuth discovery/config issues with
no clear fix, and separately asked for more MCP tools than just
`device-position`. This PR addresses both.

## What's in here

**Fix RFC 8414/9728 discovery + unclear config error**
- `WellKnownServlet` only served the bare `/.well-known/<name>` path. RFC
  8414/9728-compliant clients request the path-suffixed form instead
  (`/.well-known/oauth-authorization-server/api/oidc`) and got a 404 - this is
  the root cause behind the "blank page when authorizing" reports in the
  thread. Now serves both forms.
- A malformed `openid.clients` entry (e.g. a stale 2-field config) threw a raw
  `ArrayIndexOutOfBoundsException` with no indication of what was wrong. Now
  throws a clear error naming the bad entry and the expected format.

**Expand and fix the MCP toolset**
- Added `device-list`, `device-route`, `device-trips`, `device-summary` (the
  most-requested features in the thread).
- The new tools' report providers pull in `ReportUtils`, which takes the
  request-scoped `PermissionsService` directly. `McpServerHolder` is a
  singleton built once at server startup, outside any HTTP request - injecting
  those providers directly crashed the server on boot with a Guice
  `OutOfScopeException`. Fixed by injecting `Provider<X>` and resolving lazily
  inside each tool handler, matching the existing pattern already used for
  `PermissionsService` in this same file.

**Token efficiency**
- `device-route` over a full day from a 5-second-interval tracker was
  returning 17k+ raw positions (3.2MB) - unusable for an LLM context. Now:
  consecutive stationary positions collapse to first/last of each run (opt
  out via `dedup: false`), results downsample to a configurable `limit`
  (default 200, always keeping first/last), and per-point output is curated
  instead of dumping the full raw `Position` object.
- Added `device-summary` (via the existing `SummaryReportProvider`) for
  whole-range totals - distance, fuel, engine hours - without needing a full
  trip-by-trip list for long ranges.
- Added tool descriptions, per-parameter schema descriptions, `readOnlyHint`/
  non-destructive tool annotations, server-level `instructions`, and a
  `trip-report` prompt tying the tools together (the `prompts` capability was
  already declared but nothing was registered).

## Testing

- `./gradlew build` passes (compile, checkstyle, full existing test suite).
- Added unit tests for the discovery path-normalization logic and the
  `openid.clients` validation logic.
- Manually verified end-to-end against a live production instance: discovery
  endpoints (both old and new paths), a deliberately malformed `openid.clients`
  entry, and every MCP tool exercised with real device data.
- No test coverage added for the MCP tool handlers themselves (no existing
  test infrastructure for `McpServerHolder` to build on - flagged as a gap).

### DIFF
--- a/src/main/java/org/traccar/api/resource/OidcResource.java
+++ b/src/main/java/org/traccar/api/resource/OidcResource.java
@@ -173,12 +173,22 @@ public class OidcResource extends BaseResource {
         }
         Map<String, ClientConfig> clients = new LinkedHashMap<>();
         for (String entry : value.split(",")) {
-            String[] values = entry.split(":", 3);
+            String[] values = parseClientEntry(entry);
             clients.put(values[0], new ClientConfig(values[1], Arrays.stream(values[2].split("\\|"))
                     .map(URI::create)
                     .collect(Collectors.toSet())));
         }
         return clients;
+    }
+
+    static String[] parseClientEntry(String entry) {
+        String[] values = entry.split(":", 3);
+        if (values.length != 3 || values[0].isBlank() || values[1].isBlank() || values[2].isBlank()) {
+            throw new IllegalArgumentException(
+                    "Invalid openid.clients entry '" + entry
+                            + "': expected format 'clientId:clientSecret:redirectUri'");
+        }
+        return values;
     }
 
 }
--- a/src/main/java/org/traccar/web/McpServerHolder.java
+++ b/src/main/java/org/traccar/web/McpServerHolder.java
@@ -34,8 +34,15 @@ import org.traccar.api.security.PermissionsService;
 import org.traccar.config.Config;
 import org.traccar.config.Keys;
 import org.traccar.geocoder.Geocoder;
+import org.traccar.helper.DateUtil;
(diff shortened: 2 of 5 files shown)


shanelord01:
Closing to replace with a cleaner PR - caught a few issues in my own review pass after opening this one and wanted the history to reflect the fixes rather than layering follow-up commits on top.

PR #5970, opened 2026-08-05
MCP server: fix OAuth discovery/config bugs, harden startup, expand and optimize toolset

shanelord01:
## Heads up - this was AI-assisted ("vibe coded")

I built this with Claude Code, prompting and reviewing as I went rather than
writing every line myself. I've tested it against a real production instance
and it's been running stable. It also went through an automated code-review
pass that caught 3 real issues (details below) - all fixed in this commit, not
left outstanding. Still flagging for thorough human review rather than treating
it as fully vetted. Happy to address anything else in review.

## Motivation

This started from friction documented across a multi-page forum thread trying
to get the MCP server working:
https://www.traccar.org/forums/topic/mcp-function-how-can-the-documentation-can-found/

Several people (including me) hit the same OAuth discovery/config issues with
no clear fix, and separately asked for more MCP tools than just
`device-position`. This PR addresses both.

## What's in here

**Fix RFC 8414/9728 discovery + unclear config error**
- `WellKnownServlet` only served the bare `/.well-known/<name>` path. RFC
  8414/9728-compliant clients request the path-suffixed form instead
  (`/.well-known/oauth-authorization-server/api/oidc`) and got a 404 - this is
  the root cause behind the "blank page when authorizing" reports in the
  thread. Now serves both forms.
- A malformed `openid.clients` entry (e.g. a stale 2-field config) threw a raw
  `ArrayIndexOutOfBoundsException` with no indication of what was wrong. Now
  throws a clear error naming the bad entry and the expected format.

**Expand and fix the MCP toolset**
- Added `device-list`, `device-route`, `device-trips`, `device-summary` (the
  most-requested features in the thread).
- The new tools' report providers pull in `ReportUtils`, which takes the
  request-scoped `PermissionsService` directly. `McpServerHolder` is a
  singleton built once at server startup, outside any HTTP request - injecting
  those providers directly crashed the server on boot with a Guice
  `OutOfScopeException`. Fixed by injecting `Provider<X>` and resolving lazily
  inside each tool handler, matching the existing pattern already used for
  `PermissionsService` in this same file.

**Token efficiency**
- `device-route` over a full day from a 5-second-interval tracker was
  returning 17k+ raw positions (3.2MB) - unusable for an LLM context. Now:
  consecutive stationary positions collapse to first/last of each run (opt
  out via `dedup: false`), results downsample to a configurable `limit`
  (default 200, always keeping first/last), and per-point output is curated
  instead of dumping the full raw `Position` object.
- Added `device-summary` (via the existing `SummaryReportProvider`) for
  whole-range totals - distance, fuel, engine hours - without needing a full
  trip-by-trip list for long ranges.
- Added tool descriptions, per-parameter schema descriptions, `readOnlyHint`/
  non-destructive tool annotations, server-level `instructions`, and a
  `trip-report` prompt tying the tools together (the `prompts` capability was
  already declared but nothing was registered).

## Fixed during review (before merge, not left outstanding)

An automated code-review pass on the first draft of this PR caught 3 issues,
all fixed in the commit as it stands now:
- `device-summary` had no result-count bound, unlike `device-route`/
  `device-trips` - a `daily: true` call over an unbounded range could return
  an unbounded array. Added a `limit` param (default 60, clamped 1-1000).
- The stationary-position dedup in `device-route` treated any position
  *missing* the `motion` attribute entirely as stationary (since the getter
  defaults to `false` when the key is absent), not just positions explicitly
  known to be stationary. That could silently collapse a real route recorded
  before the motion handler existed, or written via a path that bypasses it,
  down to just its first and last point. Now only positions where `motion` is
  explicitly present and `false` are treated as stationary; positions missing
  the attribute are kept as-is.
- `device-route`'s description claimed `rawCount` reflects "positions before
  filtering," but the server's own `report.maxPositions` setting can already
  truncate the data before this tool ever sees it. Description updated to be
  accurate about that.

## Testing

- `./gradlew build` passes (compile, checkstyle, full existing test suite).
- Added unit tests for the discovery path-normalization logic and the
  `openid.clients` validation logic.
- Manually verified end-to-end against a live production instance: discovery
  endpoints (both old and new paths), a deliberately malformed `openid.clients`
  entry, and every MCP tool exercised with real device data.
- No test coverage added for the MCP tool handlers themselves (no existing
  test infrastructure for `McpServerHolder` to build on).

### DIFF
--- a/src/main/java/org/traccar/api/resource/OidcResource.java
+++ b/src/main/java/org/traccar/api/resource/OidcResource.java
@@ -173,12 +173,22 @@ public class OidcResource extends BaseResource {
         }
         Map<String, ClientConfig> clients = new LinkedHashMap<>();
         for (String entry : value.split(",")) {
-            String[] values = entry.split(":", 3);
+            String[] values = parseClientEntry(entry);
             clients.put(values[0], new ClientConfig(values[1], Arrays.stream(values[2].split("\\|"))
                     .map(URI::create)
                     .collect(Collectors.toSet())));
         }
         return clients;
+    }
+
+    static String[] parseClientEntry(String entry) {
+        String[] values = entry.split(":", 3);
+        if (values.length != 3 || values[0].isBlank() || values[1].isBlank() || values[2].isBlank()) {
+            throw new IllegalArgumentException(
+                    "Invalid openid.clients entry '" + entry
+                            + "': expected format 'clientId:clientSecret:redirectUri'");
+        }
+        return values;
     }
 
 }
--- a/src/main/java/org/traccar/web/McpServerHolder.java
+++ b/src/main/java/org/traccar/web/McpServerHolder.java
@@ -34,8 +34,15 @@ import org.traccar.api.security.PermissionsService;
 import org.traccar.config.Config;
 import org.traccar.config.Keys;
 import org.traccar.geocoder.Geocoder;
+import org.traccar.helper.DateUtil;
+import org.traccar.helper.model.DeviceUtil;
 import org.traccar.model.Device;
 import org.traccar.model.Position;
+import org.traccar.reports.RouteReportProvider;
+import org.traccar.reports.SummaryReportProvider;
+import org.traccar.reports.TripsReportProvider;
+import org.traccar.reports.model.SummaryReportItem;
+import org.traccar.reports.model.TripReportItem;
 import org.traccar.storage.Storage;
 import org.traccar.storage.StorageException;
 import org.traccar.storage.query.Columns;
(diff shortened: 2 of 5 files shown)


tananaev:
We would need to break this down into smaller parts.

shanelord01:
> We would need to break this down into smaller parts.

I will close this and submit as smaller part PR's. Thanks.

shanelord01:
Thanks - broken up and reopened as nine separate PRs. Closing this one.

Two independent bug fixes, reviewable in any order:

- #5972 - serve path-suffixed well-known discovery endpoints (RFC 8414/9728)
- #5973 - validate `openid.clients` configuration entries

Then the MCP work, as a series that applies in order:

1. #5974 - descriptions and read-only annotations on the existing tools
2. #5975 - `device-list`
3. #5976 - `device-summary` (carries the lazy `Provider` injection fix and the shared range helpers)
4. #5977 - `device-trips`
5. #5978 - `device-route`
6. #5979 - server instructions and `trip-report` prompt
7. #5980 - curated `device-position` output

Each has exactly one commit of its own, so while the series is stacked the GitHub diff of a later PR includes its parents until they merge - each description names the single commit to review. I'll rebase the rest as they land, and I'm happy to drop or rework any of them individually; 6 and 7 in particular are optional relative to the tools themselves.

Every branch builds and passes `./gradlew build` (compile, checkstyle, tests) on its own, not just at the tip.

While splitting this up I also reduced some duplication that was in the original: argument parsing and limit clamping for the range-based tools are now shared helpers introduced in #5976 rather than repeated in each handler.

PR #5972, opened 2026-08-06
Serve path-suffixed well-known discovery endpoints

shanelord01:
`WellKnownServlet` matches only the bare metadata paths, for example `/.well-known/oauth-authorization-server`.

RFC 8414 (section 3.1) and RFC 9728 (section 3.1) specify that the resource path component is inserted *between* the well-known URI and the rest of the path, so a compliant client requests `/.well-known/oauth-authorization-server/api/oidc` instead. That returns 404 today, which leaves MCP clients unable to complete authorization discovery and is the cause of the blank authorization page reported in the [forum thread](https://www.traccar.org/forums/topic/mcp-function-how-can-the-documentation-can-found/).

This normalizes the request path so both the bare and path-suffixed forms resolve to the same metadata. The suffix has to match the endpoint it belongs to, so `/oauth-protected-resource/api/oidc` and `/openid-configuration/api/mcp` still fall through to 404 rather than silently returning the wrong document.

Testing:
- `./gradlew build`
- Added `WellKnownServletTest` covering bare paths, correctly suffixed paths, mismatched suffixes, unknown paths and a null path.

This is the first of a series splitting up #5970, as requested. The parts are independent of each other except where noted.

---

### DIFF
--- a/src/main/java/org/traccar/web/WellKnownServlet.java
+++ b/src/main/java/org/traccar/web/WellKnownServlet.java
@@ -46,9 +46,11 @@ public class WellKnownServlet extends HttpServlet {
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
         String path = req.getPathInfo();
         Map<String, Object> payload = switch (path) {
-            case "/openid-configuration" -> openIdConfiguration();
-            case "/oauth-authorization-server" -> authorizationServerConfiguration();
-            case "/oauth-protected-resource" -> protectedResourceConfiguration();
+            case "/openid-configuration", "/openid-configuration/api/oidc" -> openIdConfiguration();
+            case "/oauth-authorization-server", "/oauth-authorization-server/api/oidc" ->
+                    authorizationServerConfiguration();
+            case "/oauth-protected-resource", "/oauth-protected-resource/api/mcp" ->
+                    protectedResourceConfiguration();
             default -> null;
         };
         if (payload != null) {


src/test/java/org/traccar/web/WellKnownServletTest.java
```diff
@@ -0,0 +1,43 @@
+package org.traccar.web;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Test;
+import org.traccar.config.Config;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class WellKnownServletTest {
```
tananaev: Please drop all non essential tests from your PRs. This is overkill and not really needed for a trivial change like this.

src/main/java/org/traccar/web/WellKnownServlet.java
```diff
@@ -59,6 +62,22 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IO
...
+        }
+        if (path.equals("/openid-configuration" + OIDC_PATH_SUFFIX)) {
+            return "/openid-configuration";
+        }
+        if (path.equals("/oauth-authorization-server" + OIDC_PATH_SUFFIX)) {
+            return "/oauth-authorization-server";
+        }
+        if (path.equals("/oauth-protected-resource" + MCP_PATH_SUFFIX)) {
+            return "/oauth-protected-resource";
+        }
```
tananaev: Why can't we explicitly list these options under the `switch`?

src/test/java/org/traccar/web/WellKnownServletTest.java
shanelord01: Dropped the test file.

src/main/java/org/traccar/web/WellKnownServlet.java
shanelord01: Good point - refactored to list both forms explicitly as switch cases, which also removed the helper method and constants. Diff is now 5 lines.

PR #5973, opened 2026-08-06
Validate openid.clients configuration entries

shanelord01:
`getClients()` splits each `openid.clients` entry into three fields and indexes them directly. An entry that does not have three fields, for example one left in an older two-field format, fails with a bare `ArrayIndexOutOfBoundsException` that names neither the offending entry nor the expected format. Working out what was wrong from that took a while, and others hit the same thing in the [forum thread](https://www.traccar.org/forums/topic/mcp-function-how-can-the-documentation-can-found/).

This validates that each entry has three non-blank fields and otherwise throws an `IllegalArgumentException` naming the entry and the expected `clientId:clientSecret:redirectUri` format.

Testing:
- `./gradlew build`
- Added `OidcResourceTest` covering a valid entry, an entry with multiple pipe-separated redirect URIs, a missing field and a blank field.

Part of the split of #5970. Independent of the other parts.

---

### DIFF
--- a/src/main/java/org/traccar/api/resource/OidcResource.java
+++ b/src/main/java/org/traccar/api/resource/OidcResource.java
@@ -173,12 +173,22 @@ public class OidcResource extends BaseResource {
         }
         Map<String, ClientConfig> clients = new LinkedHashMap<>();
         for (String entry : value.split(",")) {
-            String[] values = entry.split(":", 3);
+            String[] values = parseClientEntry(entry);
             clients.put(values[0], new ClientConfig(values[1], Arrays.stream(values[2].split("\\|"))
                     .map(URI::create)
                     .collect(Collectors.toSet())));
         }
         return clients;
+    }
+
+    static String[] parseClientEntry(String entry) {
+        String[] values = entry.split(":", 3);
+        if (values.length != 3 || values[0].isBlank() || values[1].isBlank() || values[2].isBlank()) {
+            throw new IllegalArgumentException(
+                    "Invalid openid.clients entry '" + entry
+                            + "': expected format 'clientId:clientSecret:redirectUri'");
+        }
+        return values;
     }
 
 }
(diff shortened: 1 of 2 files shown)


src/test/java/org/traccar/api/resource/OidcResourceTest.java
```diff
@@ -0,0 +1,39 @@
+package org.traccar.api.resource;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class OidcResourceTest {
```
tananaev, line 9: please drop

src/main/java/org/traccar/api/resource/OidcResource.java
```diff
@@ -173,12 +173,22 @@ private Map<String, ClientConfig> getClients() {
...
                     .map(URI::create)
                     .collect(Collectors.toSet())));
         }
         return clients;
     }
 
+    static String[] parseClientEntry(String entry) {
+        String[] values = entry.split(":", 3);
+        if (values.length != 3 || values[0].isBlank() || values[1].isBlank() || values[2].isBlank()) {
+            throw new IllegalArgumentException(
```
tananaev, line 187: Thinking about it again, I think this PR we would have to reject. The philosophy of our project is to keep implementation as minimal and maintainable is possible. Adding validations like this is unnecessary in my opinion. The reason people hit it early on because we had to change the format, but I don't expect this to be an ongoing issue.

shanelord01:
That's fair - with the format now stable it's a one-time migration issue rather than an ongoing one, and the minimal-implementation philosophy makes sense. Closing.

PR #5974, opened 2026-08-06
Add read-only annotations to MCP tools

shanelord01:
The MCP tools carry no annotations, so a client cannot tell that they only read data and may prompt for confirmation before calling them.

This marks both tools read-only and non-destructive. No behaviour change.

Testing: `./gradlew build`

Part of the split of #5970. This is the base of the MCP series; the remaining MCP parts build on it in order.

---

### DIFF
--- a/src/main/java/org/traccar/web/McpServerHolder.java
+++ b/src/main/java/org/traccar/web/McpServerHolder.java
@@ -52,6 +52,9 @@ public class McpServerHolder implements AutoCloseable {
 
     public static final String PATH = "/api/mcp";
 
+    private static final McpSchema.ToolAnnotations READ_ONLY_ANNOTATIONS = new McpSchema.ToolAnnotations(
+            null, true, false, true, false, null);
+
     private final Storage storage;
     private final Provider<PermissionsService> permissionsService;
     private final Geocoder geocoder;
@@ -110,6 +113,7 @@ public class McpServerHolder implements AutoCloseable {
                 .name("traccar-version")
                 .title("Returns server version name")
                 .inputSchema(inputSchema)
+                .annotations(READ_ONLY_ANNOTATIONS)
                 .build();
 
         return McpServerFeatures.AsyncToolSpecification.builder()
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/web/McpServerHolder.java
```diff
@@ -109,7 +116,9 @@ private McpServerFeatures.AsyncToolSpecification createVersionTool() {
         var toolSchema = McpSchema.Tool.builder()
                 .name("traccar-version")
                 .title("Returns server version name")
+                .description("Returns the Traccar server version string.")
```
tananaev: This seems duplicative with the title. What makes you think title is not enough?

shanelord01: In MCP, title is a human-facing display label, while description is what LLM clients actually read when deciding whether and how to call a tool. It can carry information the title cannot, like units (speed in knots), what the id means, and when to prefer another tool. The readOnlyHint annotation also lets clients skip write-confirmation prompts. You were right on this specific one though - the traccar-version description just restated the title, so I dropped it. The remaining description on device-position carries information the title does not.

tananaev: Are you saying LLM doesn't read the title? I think one field should be enough. Whether it's title or description.

shanelord01: Sorry I'm sure it will - I was just following the latest schema as a guide.

https://modelcontextprotocol.io/specification/2026-07-28/server/tools#data-types

tananaev: If it can read both, then why are we changing anything? Is something not working?

shanelord01: Agreed on one field. I went with description, since that is the field clients pass to the model, and updated the PR so every tool sets only description and none set both.

tananaev: If model can read both, I think title is better because it works for both humans and models. Also I think you stuff that doesn't belong there in the description. You're describing output format, which if really needed, belongs in the output schema.

shanelord01: Fair enough. Switched everything back to title and dropped the descriptions across the series, so each tool sets title only and none of them describe the output. That leaves this PR as just the read-only annotations, so I have retitled it to match.

PR #5975, opened 2026-08-06
Add MCP device-list tool

shanelord01:
There is no way to discover devices over MCP, so a caller has to already know a numeric device id before `device-position` is usable. This was one of the most frequently asked-for additions in the [forum thread](https://www.traccar.org/forums/topic/mcp-function-how-can-the-documentation-can-found/).

This adds a `device-list` tool returning the devices accessible to the requesting user, via the existing `DeviceUtil.getAccessibleDevices`, in the same format as `/api/devices`. It takes an optional case-insensitive substring filter on the name and a result limit (default 200, clamped to 1-1000).

Testing: `./gradlew build`

---

Part of the split of #5970. Builds on #5974 - until that merges the diff here includes it. The commit to review is `3c1be4e1e`, the single top commit.

### DIFF
--- a/src/main/java/org/traccar/web/McpServerHolder.java
+++ b/src/main/java/org/traccar/web/McpServerHolder.java
@@ -36,14 +36,17 @@ import org.traccar.config.Keys;
 import org.traccar.geocoder.Geocoder;
 import org.traccar.model.Device;
 import org.traccar.model.Position;
+import org.traccar.model.User;
 import org.traccar.storage.Storage;
 import org.traccar.storage.StorageException;
 import org.traccar.storage.query.Columns;
 import org.traccar.storage.query.Condition;
+import org.traccar.storage.query.Order;
 import org.traccar.storage.query.Request;
 import reactor.core.publisher.Mono;
 
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


shanelord01:
Updated along with the feedback on #5974 and #5978: the tool now returns Device objects in the same format as /api/devices instead of a reshaped subset, and sets a single description field.

shanelord01:

src/main/java/org/traccar/web/McpServerHolder.java
```diff
@@ -191,6 +222,38 @@ private Mono<McpSchema.CallToolResult> getDevicePosition(
...
+            return Mono.just(errorResult("User context is missing"));
+        }
+
+        Object nameValue = request.arguments().get("name");
+        String nameFilter = nameValue instanceof String s && !s.isBlank()
+                ? s.toLowerCase(Locale.ROOT) : null;
+
+        Object limitValue = request.arguments().get("limit");
+        int limit = limitValue instanceof Number number ? number.intValue() : 200;
+        limit = Math.max(1, Math.min(limit, 1000));
```
tananaev: Why do we need hardcoded limits?

shanelord01: No good reason, and they are gone.

The tool now runs the same query DeviceResource does: Condition.Permission for access, Condition.Contains over name, uniqueId, phone, model and contact when a keyword is given, and limit and offset passed to Order instead of applied after the fact. No default and no clamp, so omitting limit returns everything, the same as /api/devices.

The old version filtered in memory on name alone. That was a second implementation of something the storage layer already does, and a worse one: searching for a VIN held in the phone field now works, because it is the same five columns REST searches.

One thing I would like your view on rather than just adding it. Device attributes come back in full here, the same as /api/devices, and on my own server that includes a device's push notification token. Over REST the caller is a browser session, but over MCP the caller is a language model and the response may leave the machine. DeviceResource already has excludeAttributes mapping to Columns.Exclude("attributes"), so exposing the same argument is a couple of lines. Worth adding, and if so should it match the REST default of including attributes, or default the other way for MCP?

PR #5976, opened 2026-08-06
Add MCP device-summary tool

shanelord01:
This adds a `device-summary` tool exposing the existing `SummaryReportProvider`, so a caller can get totals for distance, average and max speed, fuel, odometer and engine hours over a time range, optionally broken down per day. Results are returned in the same format as `/api/reports/summary`. The result count is bounded by a `limit` argument (default 60, clamped to 1-1000), which matters when `daily` is true over a long range.

One thing worth review attention. `SummaryReportProvider` takes the request-scoped `PermissionsService` directly, but `McpServerHolder` is a `@Singleton` constructed at server startup, outside any HTTP request. Injecting the report provider eagerly fails startup with a Guice `OutOfScopeException`. It is therefore injected as a `Provider` and resolved inside the tool handler, which is the same approach already used for `PermissionsService` in this class. The two follow-up tools do the same.

This part also adds the helpers shared by the range-based tools (the common `deviceId`/`from`/`to` input schema, argument parsing, and limit clamping) so that `device-trips` and `device-route` stay small. `device-list` is switched to the limit helper rather than repeating the clamp.

Testing: `./gradlew build`

---

Part of the split of #5970. Builds on #5975 - until that merges the diff here includes it. The commit to review is `d967ddf06`, the single top commit.

### DIFF
--- a/src/main/java/org/traccar/web/McpServerHolder.java
+++ b/src/main/java/org/traccar/web/McpServerHolder.java
@@ -36,14 +36,17 @@ import org.traccar.config.Keys;
 import org.traccar.geocoder.Geocoder;
 import org.traccar.model.Device;
 import org.traccar.model.Position;
+import org.traccar.model.User;
 import org.traccar.storage.Storage;
 import org.traccar.storage.StorageException;
 import org.traccar.storage.query.Columns;
 import org.traccar.storage.query.Condition;
+import org.traccar.storage.query.Order;
 import org.traccar.storage.query.Request;
 import reactor.core.publisher.Mono;
 
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
@@ -88,7 +91,7 @@ public class McpServerHolder implements AutoCloseable {
         server = McpServer.async(transport)
                 .serverInfo("traccar-mcp", "1.0.0")
                 .capabilities(capabilities)
-                .tools(createVersionTool(), createDevicePositionTool())
+                .tools(createVersionTool(), createDevicePositionTool(), createDeviceListTool())
                 .build();
     }
 
@@ -102,6 +105,10 @@ public class McpServerHolder implements AutoCloseable {
             return McpTransportContext.EMPTY;
         }
         return McpTransportContext.create(contextData);
+    }
+
+    private Map<String, Object> schemaProperty(String type, String description) {
+        return Map.of("type", type, "description", description);
     }
 
     private McpServerFeatures.AsyncToolSpecification createVersionTool() {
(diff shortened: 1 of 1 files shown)


shanelord01:
Updated along with the feedback on #5974 and #5978: the tool now returns SummaryReportItem objects in the same format as /api/reports/summary instead of reshaping fields, and sets a single description field.

shanelord01:

shanelord01:
Reworked around your point on #5978 about not duplicating the report API.

There is now a single report tool factory. It builds the deviceId/from/to input schema, runs the permission checks, calls the provider and returns the result unchanged. Registering a report is one entry naming the provider call, so there is no handler per report any more. device-summary is the first one through it.

Two things the hand-written handlers were missing that the REST path does. ReportResource calls checkRestriction(getDisableReports) on every report endpoint and the MCP tools called it on none, so a user with reports disabled could still pull report data over MCP. The shared handler does it now, and I checked against a running server that a non-admin with disableReports set is refused on both paths. Date parsing also goes through DateUtil.parseDate, the same converter the REST query parameters use.

SummaryReportProvider still needs to be injected as a Provider and resolved inside the handler. McpServerHolder is a singleton built at startup, outside any request scope, so injecting it eagerly fails with a Guice OutOfScopeException.

tananaev:
I don't see any factories. I also don't see proper commits for your changes. I think we need to discuss the approach first before doing anything. Also make sure you keep proper git history, so I can see what the changes are.

shanelord01:
The force-pushes are on me. Every update to these branches replaced the whole branch, so your review comments went outdated and there is nothing to diff against your last look. From here on changes go on top as new commits.

The four PRs all target master, so #5976, #5977 and #5978 each carry the earlier branches' commits and the commit list does not match the title. GitHub will not let me base a fork PR on another fork branch, so the choice is one PR for the report tools, or opening them one at a time after the previous one merges. I lean to one PR, with #5975 staying separate since it does not touch reports.

For the approach itself, the factory I mentioned is createReportTool in db7d132, but it only removes duplication between the MCP tools. It still parses arguments and runs the permission checks itself, next to ReportResource, so it does not do what you asked for on #5978. Two ways I can see to reuse the existing API: a single definition per report (name, provider call, parameters) that both ReportResource and the MCP server register from, or the MCP handler calling into ReportResource directly. Which of those, or something else?

I will not push anything until you've shared how you would prefer to proceed.

tananaev:
So what's your proposal to unify API and MCP?

shanelord01:
Proposal: MCP tool calls go through the existing REST layer in process. Jersey's ServletContainer exposes its ApplicationHandler, which takes a ContainerRequest without HTTP. The MCP handler builds a GET for the resource path from the tool arguments, passes the Authorization header through, and returns the response entity. SecurityRequestFilter, PermissionsService, the date converter, ResourceErrorHandler and LogAction all run as they do for REST, so MCP gets the same behaviour, including the audit rows it does not write today.

Input schemas come from the resource model (Resource.from(ReportResource.class) gives parameter names and types), so exposing an endpoint is one entry with a name, a path and a description. device-position stays as it is because of the geocoding.

I checked this on a local branch, not pushed: a reports/summary tool gives the same result, the same 404 on a bad date and the same empty list for a device the user cannot see as the REST call, and writes the same tc_actions row. It needs two Jersey types from internal packages for the ContainerRequest properties and the HttpServletRequest injection. If the direction is right I will open one PR replacing these four, in small commits.

For transparency, I used Claude to help audit the API and MCP code while working this out.

tananaev:
Maybe you can create a draft prototype to see what it would look like?

shanelord01:
Let me get started on that... :)

tananaev:
Please keep it as minimal as possible just to see what the wiring looks like.

PR #5977, opened 2026-08-06
Add MCP device-trips tool

shanelord01:
This adds a `device-trips` tool exposing the existing `TripsReportProvider`, giving a leg-by-leg breakdown of motion-based segments over a time range in the same format as `/api/reports/trips`. The result count is bounded by a `limit` argument (default 100, clamped to 1-1000).

`TripsReportProvider` is injected as a `Provider` and resolved inside the handler for the same request-scope reason described in #5976.

Testing: `./gradlew build`

---

Part of the split of #5970. Builds on #5976 - until that merges the diff here includes it. The commit to review is `bd7f8f61a`, the single top commit.

### DIFF
--- a/src/main/java/org/traccar/web/McpServerHolder.java
+++ b/src/main/java/org/traccar/web/McpServerHolder.java
@@ -40,6 +40,7 @@ import org.traccar.model.Position;
 import org.traccar.model.User;
 import org.traccar.model.UserRestrictions;
 import org.traccar.reports.SummaryReportProvider;
+import org.traccar.reports.TripsReportProvider;
 import org.traccar.storage.Storage;
 import org.traccar.storage.StorageException;
 import org.traccar.storage.query.Columns;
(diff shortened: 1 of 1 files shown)


shanelord01:
Updated along with the feedback on #5974 and #5978: the tool now returns TripReportItem objects in the same format as /api/reports/trips instead of reshaping fields, and sets a single description field.

shanelord01:

src/main/java/org/traccar/web/McpServerHolder.java
```diff
@@ -191,6 +318,98 @@ private Mono<McpSchema.CallToolResult> getDevicePosition(
...
+                            "trips", result,
+                            "returnedCount", result.size(),
+                            "rawCount", trips.size()))
+                    .build());
+        } catch (StorageException | SecurityException | IllegalArgumentException e) {
+            return Mono.just(errorResult(e.getMessage()));
+        }
+    }
+
+    private Mono<McpSchema.CallToolResult> getDeviceSummary(
```
tananaev: This is still hardcoded and a lot of duplication. I think we need something better.

shanelord01: Agreed, and the shape has changed. Each of the three report tools had its own 30-line handler differing only in the provider call, which is what you were looking at.

The factory in #5976 holds all of that now. This PR is what adding a report costs afterwards: 12 lines, one entry naming TripsReportProvider.getObjects, and no handler of its own. Stops, events and combined would each be the same.

The limits are gone as well. Nothing here defaults or clamps a result count.

PR #5978, opened 2026-08-06
Add MCP device-route tool

shanelord01:
This adds a `device-route` tool exposing the existing `RouteReportProvider`, for callers that need the recorded path rather than trip or summary aggregates.

Raw route data cannot be returned verbatim over MCP. A full day from a five-second-interval tracker is 17k+ positions, around 3.2 MB, which does not fit in a model context window. The tool downsamples evenly to at most `limit` positions (default 200, clamped to 1-2000), always keeping the first and last point. Positions themselves are returned unchanged, in the same format as `/api/reports/route`, and `rawCount` is reported alongside `returnedCount` so the caller can see when downsampling happened.

Testing: `./gradlew build`

---

Part of the split of #5970. Builds on #5977 - until that merges the diff here includes it. The commit to review is `ce941f3c3`, the single top commit.

### DIFF
--- a/src/main/java/org/traccar/web/McpServerHolder.java
+++ b/src/main/java/org/traccar/web/McpServerHolder.java
@@ -40,6 +40,7 @@ import org.traccar.model.Position;
 import org.traccar.model.User;
 import org.traccar.model.UserRestrictions;
 import org.traccar.reports.SummaryReportProvider;
+import org.traccar.reports.TripsReportProvider;
 import org.traccar.storage.Storage;
 import org.traccar.storage.StorageException;
 import org.traccar.storage.query.Columns;
@@ -69,6 +70,7 @@ public class McpServerHolder implements AutoCloseable {
     private final Geocoder geocoder;
     private final boolean geocodeOnRequest;
     private final Provider<SummaryReportProvider> summaryReportProvider;
+    private final Provider<TripsReportProvider> tripsReportProvider;
 
     private final HttpServletStreamableServerTransportProvider transport;
     private final McpAsyncServer server;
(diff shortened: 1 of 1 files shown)


tananaev:
I think we need a better way of organizing and adding functionality to MCP. Just hardcoding everything doesn't seem clean and sustainable.

shanelord01:
Agreed. My suggestion: a small McpTool interface (name, tool schema, call handler) with one class per tool in an org.traccar.mcp package, injected as a set into McpServerHolder so the holder only wires transport and registration. That keeps each tool self-contained and adding a tool becomes adding one class. Would you prefer that refactor as a separate precursor PR that the tool PRs rebase onto, or folded into this one?

tananaev:
That just solves the problem of large single class. But doesn't solve the duplication of code. We already have API that returns all the reports. Ideally we reuse that somehow or unify the implementation.

shanelord01:
Fair point. I went back over the series with that in mind. The tools were already fetching data through the same report providers the REST API uses, but they then reshaped the output into their own format, which duplicated what Jackson already produces for /api/reports/*. I have removed all of that reshaping across the series: device-summary, device-trips and device-route now return the provider objects unchanged, so the output format is identical to the corresponding REST endpoint and there is one implementation to maintain. device-list returns plain Device objects for the same reason. What remains per tool is the input schema, the permission check and the provider call.

The one piece I kept is the position limit with downsampling in this tool, because a long range can hold tens of thousands of positions and the response has to fit in a model context window.

If you would prefer a more general mechanism for exposing existing endpoints as MCP tools, I am happy to work on that instead, but I did not want to build a framework before agreeing on the shape.

shanelord01:

shanelord01:
Rebased onto the report tool factory from #5976, so this is one entry calling RouteReportProvider plus the sampling helper.

On the limit: it is optional with no default and no clamp, so leaving it out returns the full route exactly as /api/reports/route does. When it is set the tool samples evenly across the range and keeps the first and last position.

I estimated the size in the original description and the estimate was low. Measuring a Teltonika unit on a real touring day, position ids run contiguously while the device is logging, and one 24 hour window spans 22,670 of them. A single 2h54m driving leg covered 4,805 positions, one every 2.2 seconds. At the size these serialise to, that day is roughly 8 MB rather than the 3.2 MB I wrote.

Asking for 8 positions across that day returned a route whose first and last totalDistance differ by 713,888 m, against 713,888 m reported by device-summary for the same window, so the sampled result still reconstructs the day.

If you would rather the server did no sampling at all I will drop it and let callers narrow the range instead. It is the only thing here that is not a straight pass-through, so I would rather ask than assume.

PR #5979, opened 2026-08-06
Add MCP server instructions and trip-report prompt

shanelord01:
The server declares the `prompts` capability but registers no prompts, and offers no server-level guidance on how the tools relate to each other.

This adds server `instructions` pointing clients at `device-list` for discovery and steering them towards the aggregate tools rather than `device-route` for wide ranges, and registers a `trip-report` prompt for the common case of summarizing a device over a date range.

Testing: `./gradlew build`

---

Part of the split of #5970. Rebased on master, so the diff is only this commit (`0c304d746`). The instructions and prompt text reference the tools added in #5975 to #5978, so this one should land after those. It can be dropped without affecting them.

### DIFF
--- a/src/main/java/org/traccar/web/McpServerHolder.java
+++ b/src/main/java/org/traccar/web/McpServerHolder.java
@@ -84,8 +84,14 @@ public class McpServerHolder implements AutoCloseable {
 
         server = McpServer.async(transport)
                 .serverInfo("traccar-mcp", "1.0.0")
+                .instructions(
+                        "Call device-list first to find valid device ids. Prefer device-summary for totals over "
+                                + "long ranges (weeks or more), device-trips for a leg-by-leg breakdown, and "
+                                + "device-route only when the exact path is needed, ideally over a narrow "
+                                + "sub-range - it returns far more data than the other tools.")
                 .capabilities(capabilities)
                 .tools(createVersionTool(), createDevicePositionTool())
+                .prompts(createTripReportPrompt())
                 .build();
     }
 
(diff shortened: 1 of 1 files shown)


tananaev:
It looks like this also has unrelated changes.

shanelord01:
You're right, the branch had the whole tool stack in it. Rebuilt it against master so the diff is now only the server instructions and the trip-report prompt (38 lines, one file). Note the instructions and prompt text reference the tools added in the other PRs, so this one should land after those.

shanelord01:
Rebased on current master (MCP SDK 2.0.1). Diff is still the single commit.

src/main/java/org/traccar/web/McpServerHolder.java
```diff
@@ -84,8 +84,14 @@ public McpServerHolder(
 
         server = McpServer.async(transport)
                 .serverInfo("traccar-mcp", "1.0.0")
+                .instructions(
+                        "Call device-list first to find valid device ids. Prefer device-summary for totals over "
+                                + "long ranges (weeks or more), device-trips for a leg-by-leg breakdown, and "
+                                + "device-route only when the exact path is needed, ideally over a narrow "
+                                + "sub-range - it returns far more data than the other tools.")
```
tananaev, line 91: This doesn't look right. Specific descriptions belong to specific functions.

src/main/java/org/traccar/web/McpServerHolder.java
```diff
@@ -146,6 +152,38 @@ private McpServerFeatures.AsyncToolSpecification createDevicePositionTool() {
...
+                List.of(
+                        new McpSchema.PromptArgument(
+                                "deviceId", "Device id, see device-list for available ids", true),
+                        new McpSchema.PromptArgument("from", "Start of the range, ISO-8601", true),
+                        new McpSchema.PromptArgument("to", "End of the range, ISO-8601", true)));
+
+        return new McpServerFeatures.AsyncPromptSpecification(prompt, this::getTripReportPrompt);
+    }
+
+    private Mono<McpSchema.GetPromptResult> getTripReportPrompt(
```
tananaev, line 168: Why do we need this?

shanelord01:
Both points are fair. Guidance about a tool belongs on that tool rather than in a server-level string that names them, and the prompt was a workflow written into the server that a client can hold just as easily. Closing this one. The tool PRs do not depend on it.

PR #5980, opened 2026-08-06
Return curated fields from MCP device-position

shanelord01:
`device-position` passes the whole `Position` object as structured content, so every response carries internal and network columns that are of no use to a client and inflate the result.

This returns an explicit set of fields instead — device id, coordinates, altitude, speed, course, validity, address, fix time, and the protocol attributes map — with geofence ids included only when the position has any.

Note this is a change to an existing tool's output shape, unlike the other parts of the series which only add things. It is last in the series for that reason and can be dropped independently.

Testing: `./gradlew build`

---

Part of the split of #5970. **Builds on #5979** — until that merges the diff here includes it. The commit to review is `5743cff31`, the single top commit.

### DIFF
--- a/src/main/java/org/traccar/web/McpServerHolder.java
+++ b/src/main/java/org/traccar/web/McpServerHolder.java
@@ -44,6 +44,7 @@ import org.traccar.storage.query.Request;
 import reactor.core.publisher.Mono;
 
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/web/McpServerHolder.java
```diff
@@ -181,16 +402,279 @@ private Mono<McpSchema.CallToolResult> getDevicePosition(
...
+                            "trips", curated,
+                            "returnedCount", curated.size(),
+                            "rawCount", trips.size()))
+                    .build());
+        } catch (StorageException | SecurityException | IllegalArgumentException e) {
+            return Mono.just(errorResult(e.getMessage()));
+        }
+    }
+
+    private Mono<McpSchema.CallToolResult> getDeviceSummary(
```
tananaev: This seems unrelated.

shanelord01: You're right, the branch had unrelated changes from the other tool branches in it. Rebuilt against master so the diff is now only the device-position output change (18 lines, one file).

tananaev:
Which columns are of no use? Network is valuable information.

shanelord01:
Fair point, network is useful and for a single position the savings from trimming fields are small. The size concern only really matters for bulk output, which #5978 now handles with a position limit instead of field trimming. Closing this one, the full Position output stays.

PR #5984, opened 2026-08-19
Support Akroz reports in TAIP protocol

MateusAzevedo-AK:
Adds support for the reports used by Akroz Telematics trackers, as part of the existing TAIP protocol rather than as a new decoder.

You were right that it is a TAIP variation. The devices reuse the TAIP envelope — the `>` … `;ID=…;#…;*XX<` framing and the same encoding for coordinates, speed, course and status flags — with their own report formats. The manufacturer identifies the protocol as `NT003`, which is the fixed second field of every message, and the reports are prefixed `RUV` followed by the format number.

Protocol documentation: the manufacturer publishes it at https://wiki.newtectelemetria.com.br/pt-br/vl08-guia-homologacao — that page is access-restricted, so I've attached the relevant sections as a PDF to this PR.
Sample messages: included in the unit tests, captured from a device in the field.

### Reports

- `RUV00` — presentation report sent on connection: serial number, firmware and hardware versions, supply voltages and ICCID. It carries no position, so it reports the last known location.
- `RUV01` — position and status: battery and power voltage, engine hours, odometer, RPM, coolant temperature, oil pressure, fuel level, driver unique id
- `RUV02` — end of trip: trip duration, trip distance, fuel used
- `RUV03` — CAN/engine report: throttle, engine hours, odometer, RPM, coolant temperature, oil pressure, fuel level, fuel used, OBD speed, engine torque, engine brake, cruise control, clutch, parking brake, service brake

`RUV01` to `RUV03` share a status block with the date, position, fix quality, ignition, main power, digital inputs and HDOP. It is decoded by a separate pattern because the header carries the report format and the protocol identifier ahead of the position.

### Acknowledgement

These devices retransmit indefinitely until a report is acknowledged, so the acknowledgement is sent before decoding, independently of whether the device is known or the payload parses. They also compute the XOR up to the last `;`, excluding the `*`, which is the same convention already handled for Lantrix devices — with the default calculation the acknowledgement is rejected.

### Events

Event codes are reported as `KEY_EVENT`. The subset with a standard equivalent is additionally reported as an alarm: overspeed, high RPM, idle, temperature, hard acceleration, hard braking, hard cornering, power cut, power restored and low battery.

### Tests

Seven cases added to `TaipProtocolDecoderTest`, covering the four report formats with both real device captures and samples from the manufacturer documentation. Verified against a live device as well.

[Protocolo_XVM_NT003_RUV.pdf](https://github.com/user-attachments/files/31203646/Protocolo_XVM_NT003_RUV.pdf)


tananaev:
It looks like a variation of TAIP-based protocol, so it shouldn't be a completely new decoder.

MateusAzevedo-AK:
Ok. What would be the next step?

tananaev:
What do you mean? Next step is to implement it as a part of the existing protocol, not as a new one.

MateusAzevedo-AK:
> What do you mean? Next step is to implement it as a part of the existing protocol, not as a new one.

Reworked as part of the TAIP protocol, as suggested. Also verified against a live device.

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -237,21 +285,182 @@ protected Object decode(
...
+            position.set("oilPressure", Integer.parseInt(values[5]));
+            position.set(Position.KEY_FUEL_LEVEL, Integer.parseInt(values[6]));
+            position.set(Position.KEY_FUEL_USED, Long.parseLong(values[7]) / 10.0);
+            position.set(Position.KEY_OBD_SPEED, Integer.parseInt(values[9]));
+            position.set("engineTorque", Integer.parseInt(values[10]));
+            position.set("engineBrake", Integer.parseInt(values[12]));
+            position.set("cruiseControl", Integer.parseInt(values[16]) == 1);
+            position.set("clutchState", Integer.parseInt(values[17]) == 64);
+            position.set("parkingBrake", Integer.parseInt(values[18]) == 4);
+            position.set("serviceBrake", Integer.parseInt(values[19]) == 8);
```
tananaev: This looks odd. Maybe we should check bits instead?

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -151,6 +191,14 @@ protected Object decode(
             sentence = sentence.substring(beginIndex + 1);
         }
 
+        if (sentence.startsWith("RUV")) {
+            sendRuvResponse(channel, remoteAddress, sentence);
```
tananaev: Why are we sending response here and ignore response in the other place?

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -151,6 +191,14 @@ protected Object decode(
...
+            if (sentence.startsWith("RUV00")) {
+                return decodeRuvStatus(channel, remoteAddress, sentence);
+            }
+            return decodeRuv(channel, remoteAddress, sentence);
```
tananaev: There is some duplication between the two message types, so I don't think it make sense to just split them.

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -140,6 +141,45 @@ private String decodeAlarm2(int value) {
...
+            .expression(",([^,]*)")              // firmware version
+            .expression(",([^,]*)")              // hardware version
+            .expression(",[^,]*")                // script version
+            .expression(",[^,]*")                // peripheral
+            .expression(",[^,]*,[^,]*,[^,]*")    // settings
+            .expression(",([^;,]*)")             // iccid
+            .any()
+            .compile();
+
+    private static final Pattern PATTERN_RUV = new PatternBuilder()
```
tananaev: I also see duplication with existing formats. It seems like you just mechanically copied separate implementation instead of actually thinking what can be reused and unified.

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -237,21 +285,182 @@ protected Object decode(
...
+        position.set(Position.KEY_HDOP, parser.nextInt());
+
+        position.set(Position.KEY_EVENT, event);
+        position.addAlarm(decodeAlarmRuv(event));
+
+        String[] values = parser.next().split(",");
+        switch (format) {
+            case 1 -> decodeRuvBasic(position, values);
+            case 2 -> decodeRuvTrip(position, values);
+            case 3 -> decodeRuvCan(position, values);
```
tananaev: It doesn't make sense to split there into separate sub-functions. It makes it harder to see the whole picture.

src/test/java/org/traccar/protocol/TaipProtocolDecoderTest.java
```diff
@@ -117,6 +117,33 @@ public void testDecode() throws Exception {
         verifyPosition(decoder, text(
                 ">RPV19105+4538405-0739518900000012;ID=9999;*7A<\r\n"));
 
+        verifyAttribute(decoder, text(
```
tananaev, line 120: Please make sure only use real samples from real device, not fake ones.

MateusAzevedo-AK:
Reworked everything based on your comments — thanks for the push, the result is
a lot smaller.

**Reused the existing pattern.** RUV is just another type in the alternation
now, and the whole shared body (coordinates, speed, course, fix status, input
mask) is reused as is. The coordinates turned out to be the same encoding
already. Only a few things actually differ and they're handled inline: speed is
km/h, fix status goes 0-9 instead of 0/2/3, bit 6 of the input mask is main
power, and the two digits before HDOP are unused rather than an event code.
That last one was a real bug I only hit once the code was merged — the existing
branch reads them as the event, which was wiping the actual event.

**Single flow.** The presentation report has no position, which is why it was
split before. Made the position block optional in the pattern instead, so all
four report types go through the same path. Sub-functions are gone, the payload
is just split by comma from a fixed index.

**Single response path.** Dropped the separate method — the ack goes through
decodeAttributes like everything else. Only difference is the checksum, these
devices leave the '*' out, so it reuses the Lantrix flag.

**Bit checks** for cruise control, clutch and brakes, as suggested.

**Real samples only.** All test captures are from a device in the field now.

Diff went from +427 (original separate protocol) to +127.

Also ran it against a live device: 636 positions, 22 distinct event codes across
all four report formats, and the alarm mappings confirmed in the field —
overspeed, highRpm, idle, hardAcceleration and hardBraking all fired correctly,
with the matching "end of violation" events correctly producing no alarm.

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -49,12 +49,15 @@ public TaipProtocolDecoder(Protocol protocol) {
             .groupEnd("?")
             .number("(d{5})")                    // seconds
             .or()
-            .expression("(?:RGP|RCQ|RCV|RBR|RUS00|RPI),?") // type
-            .number("(dd)?")                     // event
-            .number("(dd)(dd)(dd)")              // date (mmddyy)
+            .expression("(?:RGP|RCQ|RCV|RBR|RUS00|RPI|RUV\\d\\d),?") // type
```
tananaev: I think this is easier to read:
```suggestion
            .expression("(?:RGP|RCQ|RCV|RBR|RUS00|RPI|RUV[0-9]{2}),?") // type
```

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -156,6 +160,8 @@ protected Object decode(
             return null;
         }
 
+        boolean akroz = sentence.startsWith("RUV");
```
tananaev:
```suggestion
        boolean isAkroz = sentence.startsWith("RUV");
```

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -176,22 +182,29 @@ protected Object decode(
...
         }
 
-        position.setSpeed(convertSpeed(parser.nextDouble(0), "mph"));
+        position.setSpeed(convertSpeed(parser.nextDouble(0), akroz ? "kmh" : "mph"));
         position.setCourse(parser.nextDouble(0));
 
         if (parser.hasNext(2)) {
-            valid = parser.nextInt() > 0;
+            int fixMode = parser.nextInt();
+            valid = akroz ? fixMode < 8 : fixMode > 0;
```
tananaev: What is this value for akroz?

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -228,30 +244,104 @@ protected Object decode(
 
         position.setValid(valid == null || valid);
 
+        if (akroz) {
+            String[] values = sentence.substring(0, sentence.indexOf(';')).split(",", -1);
```
tananaev, line 248: Typically the way we iterate through these is an `index` variable. I recommend using the same pattern here.

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -228,30 +244,104 @@ protected Object decode(
...
+                    position.set("serviceBrake", BitUtil.check(Integer.parseInt(values[22]), 3));
+                }
+            }
+        }
+
         if (event != null) {
             position.set(Position.KEY_EVENT, event);
-            if (sentence.charAt(5) == ',') {
+            if (akroz) {
+                position.addAlarm(decodeAlarmRuv(event));
```
tananaev: It looks like we have two terms "akroz" and "ruv". What's the difference between them? Should we unify?

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -228,30 +244,104 @@ protected Object decode(
...
 
-        String[] attributes = null;
-        beginIndex = sentence.indexOf(';');
-        if (beginIndex != -1) {
-            int endIndex = sentence.indexOf('<', beginIndex);
-            if (endIndex == -1) {
-                endIndex = sentence.length();
-            }
-            attributes = sentence.substring(beginIndex, endIndex).split(";");
+        Position result = decodeAttributes(channel, remoteAddress, position, splitAttributes(sentence), akroz);
```
tananaev: Why are we splitting this?

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
MateusAzevedo-AK: It's the GPS fix status. The manufacturer defines 0-9: 0-3 is a fix with 1 to 4+
satellites, 4-5 are least-squares 2D/3D, 6 is dead reckoning, 7 means no
movement detected since the last valid fix, 8 is a shorted GPS antenna and 9 is
no position at all. So anything below 8 carries a usable position.

MateusAzevedo-AK:
I followed some of your suggestions

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
tananaev: You can't have a position with 1 or 2 satellites.

MateusAzevedo-AK: Right  0 and 1 claim a fix with 1 or 2 satellites, which isn't possible. Changed
to require at least 3 (value 2 and up), so 0, 1, 8 (antenna fault) and 9 (no
position) are all invalid now.

One more in the same vein: 6 is dead reckoning, calculated without satellites
from speed and heading. I left it as valid since the coordinates are still
usable, but happy to exclude it too if you'd rather only accept real fixes.

tananaev: I also don't think we need fixMode < 8 because no position at all is gated in a different way.

tananaev:
Merged, thank you.

src/main/java/org/traccar/protocol/TaipProtocolDecoder.java
```diff
@@ -228,30 +244,112 @@ protected Object decode(
...
+            case 128 -> Position.ALARM_POWER_RESTORED;
+            case 129 -> Position.ALARM_LOW_BATTERY;
+            default -> null;
+        };
     }
 
     private Position decodeAttributes(
-            Channel channel, SocketAddress remoteAddress, Position position, String[] attributes) {
+            Channel channel, SocketAddress remoteAddress, Position position, String[] attributes,
+            boolean plainChecksum) {
```
tananaev, line 352: Just noticed this one is unused parameter. Is that expected?

PR #5987, opened 2026-08-23
fix: GL320M parsing using doubles

MrCyjaneK:
the device reports battery level using doubles so my log was flooded with:

```
WARN: [xxxxxxxxx] error - For input string: "31.3" - NumberFormatException (... < Gl200TextProtocolDecoder:1000 < *:1747 < Gl200ProtocolDecoder:54 < ExtendedObjectDecoder:72 < ... < WrapperContext:102 < ...)
```

I've ran the build on the server that the device is connected to and it started reporting routes without any issues and warnings.


tananaev:
Please add the test. Make sure it's real sample from logs.

MrCyjaneK:
Hey! I missed this comment - will do that as soon as I'm back

PR #5992, opened 2026-08-24
Decode JT808 location query response (0x0201)

onethings:
Split out the decoder part from the JT808 command types PR.

### Changes
- Add `MSG_LOCATION_QUERY_RESPONSE` (0x0201) message constant
- Decode the 0x0201 location query response: send general response and parse the embedded location
- Add `Jt808ProtocolDecoderTest.testDecodeLocationQueryResponse` test

This PR only touches the decoder; the encoder side (new downlink command types) is kept in #5985.

All tests pass (`Jt808ProtocolDecoderTest`).

PR #5997, opened 2026-08-26
Decode JT808 location query and parameter query responses

onethings:
### Summary
Split out the **decoder** part of the JT808 work. The encoder side (new command types) is kept in a separate PR.

### Decoder changes (`Jt808ProtocolDecoder.java`)
- Decode `0x0201` (location query response): reply with a general response and parse the embedded location.
- Decode `0x0104` (parameter query response) and map the parameters real devices actually report:
  - `0x0001` → heartbeat interval
  - `0x0010` → APN
  - `0x0013` → server address
  - `0x0018` → server TCP port
  - `0x0027` → ignition-off interval
  - `0x0028` → report interval
  - `0x0029` → ignition-on interval
  - `0x0055` → overspeed speed threshold
  - `0x0056` → overspeed duration
  - `0x0080` → odometer

### Tests
- `testDecodeLocationQueryResponse` uses a **real device capture**.
- `testDecodeParameterQueryResponse` uses a **real device capture**.
- All `Jt808ProtocolDecoderTest` tests pass.


tananaev:
Missing documentation and the source of the sample.

onethings:
0x0201 (014538236391)
```
2026-08-25 10:32:36  INFO: [105670f2: huabao > 136.228.172.33] HEX: 7e8201000001453823639100002e7e
2026-08-25 10:32:37  INFO: [105670f2: huabao < 136.228.172.33] HEX: 7e0201003c0145382363912db0000000000000000000020101313605bb3b58000c000001292608251203040104000b620beb16000c00b28995018202107439115800060089fffffffece7e
```

(014141610847)
```
2026-08-25 11:13:17  INFO: [1c175960: huabao > 136.228.172.48] HEX: 7e810400000141416108470000aa7e
2026-08-25 11:13:17  INFO: [1c175960: huabao < 136.228.172.48] HEX: 7e600202f5014141610847045a0000000f2600000000000000c201027d018005b9fad000080000000226082510153001040007984a030200003200000000000000c201027d018005b9fad000080000000226082510184901040007984a03020000eb0a000800233039382e37343200000000000000c201027d018005b9fad000080000000226082510204801040007984a03020000eb0a000800233039382e37343200000000000000c201027d018005b9fad000080000000226082510233301040007984a03020000eb0a000800233039382e37363200000000000000c201027d018005b9fad000080000000226082510252001040007984a03020000eb0a000800233039382e37383200000000000000c201027d018005b9fad000080000000226082510281301040007984a03020000eb0a000800233039382e37383200000000000000c201027d018005b9fad000080000000226082510310801040007984a03020000eb0a000800233039382e37393200000000000000c201027d018005b9fad000080000000226082510325301040007984a03020000eb0a000800233039382e37393200000000000000c201027d018005b9fad000080000000226082510361401040007984a03020000eb0a000800233039382e37393200000000000000c201027d018005b9fad000080000000226082510384601040007984a03020000eb0a000800233039382e37393200000000000000c201027d018005b9fad000080000000226082510404701040007984a03020000eb0a000800233039382e38303200000000000000c201027d018005b9fad000080000000226082510434601040007984a03020000eb0a000800233039382e38303200000000000000c201027d018005b9fad000080000000226082510452501040007984a03020000eb0a000800233039382e38303200000000000000c201027d018005b9fad000080000000226082510485301040007984a03020000eb0a000800233039382e38303200000000000000c201027d018005b9fad000080000000226082510514301040007984a03020000eb0a000800233039382e3831f27e
```

onethings:
Document found in chinese version from gov.cn, not found english version.  https://github.com/traccar/traccar/issues/5993

tananaev:
Link please.

onethings:
Speed limit are set by me in other chinese platform, limit 100km, duration 35 sec(through their platform tcp command).  Upload interval are when Acc on 10 second, at acc off are 120 second(through sms).

onethings:
Device model are BSJ K9A, 2013 jt808 version. 098.742→098.810 are fuel data.

onethings:
https://jtt.shandong.gov.cn/module/download/downfile.jsp?classid=0&filename=da0134589d3d407081f943b7802e4286.pdf

onethings:
JT/T 808-2019
Screenshot 2026-08-26 at 12 13 06
https://jtst.mot.gov.cn/hb/search/stdHBView?id=2eabe27a635213c9af12f3aeadd2c906




JT/T 809-2019
Screenshot 2026-08-26 at 12 11 08
https://jtst.mot.gov.cn/hb/search/stdHBView?id=a71bf1e0e866c7207daf0bc436642be8





and i upload here 
[JTT808-2019.pdf](https://github.com/user-attachments/files/31449627/JTT808-2019.pdf)
[JTT809-2019.pdf](https://github.com/user-attachments/files/31449628/JTT809-2019.pdf)

onethings:
```
2026-08-25 12:20:24  INFO: [1c175960: huabao > 136.228.172.48] HEX: 7e8103001e014141610847000102000000130f3136382e3134342e3130342e323234000000180400001397157e

2026-08-25 12:20:25  INFO: [1c175960: huabao < 136.228.172.48] HEX: 7e600202f501414161084708b10000000f2600000000000000c201027d018005b9fad000080000000226082510153001040007984a030200003200000000000000c201027d018005b9fad000080000000226082510184901040007984a03020000eb0a000800233039382e37343200000000000000c201027d018005b9fad000080000000226082510204801040007984a03020000eb0a000800233039382e37343200000000000000c201027d018005b9fad000080000000226082510233301040007984a03020000eb0a000800233039382e37363200000000000000c201027d018005b9fad000080000000226082510252001040007984a03020000eb0a000800233039382e37383200000000000000c201027d018005b9fad000080000000226082510281301040007984a03020000eb0a000800233039382e37383200000000000000c201027d018005b9fad000080000000226082510310801040007984a03020000eb0a000800233039382e37393200000000000000c201027d018005b9fad000080000000226082510325301040007984a03020000eb0a000800233039382e37393200000000000000c201027d018005b9fad000080000000226082510361401040007984a03020000eb0a000800233039382e37393200000000000000c201027d018005b9fad000080000000226082510384601040007984a03020000eb0a000800233039382e37393200000000000000c201027d018005b9fad000080000000226082510404701040007984a03020000eb0a000800233039382e38303200000000000000c201027d018005b9fad000080000000226082510434601040007984a03020000eb0a000800233039382e38303200000000000000c201027d018005b9fad000080000000226082510452501040007984a03020000eb0a000800233039382e38303200000000000000c201027d018005b9fad000080000000226082510485301040007984a03020000eb0a000800233039382e38303200000000000000c201027d018005b9fad000080000000226082510514301040007984a03020000eb0a000800233039382e3831157e
```

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -565,6 +575,70 @@ protected Object decode(
...
+                        break;
+                    case 0x0029:
+                        position.set("ignitionOnInterval", buf.readUnsignedInt());
+                        break;
+                    case 0x0046:
+                        position.set(Position.KEY_SPEED_LIMIT, buf.readUnsignedShort());
+                        break;
+                    case 0x55:
+                        position.set(Position.KEY_SPEED_LIMIT, (int) buf.readUnsignedInt());
+                        break;
```
tananaev: This is incorrect mapping. Speed limit here is actual street speed limit, not device configuration.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -565,6 +575,70 @@ protected Object decode(
...
+            int endIndex = buf.readerIndex() + bodyLength - 3;
+            while (buf.readerIndex() < endIndex) {
+                int parameterId = buf.readInt();
+                int length = buf.readUnsignedByte();
+                int valueEndIndex = buf.readerIndex() + length;
+                switch (parameterId) {
+                    case 0x0001:
+                        position.set("heartbeatInterval", buf.readUnsignedInt());
+                        break;
+                    case 0x10:
```
tananaev: Why are we using 4 digits for some and 2 for other?

src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java
```diff
@@ -432,4 +432,48 @@ public void testDecode() throws Exception {
...
+        var decoder = inject(new Jt808ProtocolDecoder(null));
+
+        verifyPosition(decoder, binary(
+                "7e0201003c0145382363912db0000000000000000000020101313605bb3b58000c000001292608251203040104000b620beb16000c00b28995018202107439115800060089fffffffece7e"),
+                position("2026-08-25 04:03:04.000", true, 16.85535, 96.156504));
+
+    }
+
+    @Test
+    public void testDecodeParameterQueryResponse() throws Exception {
```
tananaev: We don't need to test every single parameter. One test is fine. Make sure you keep the test formatting as other tests. And also collapse all tests into existing single function.

tananaev:
Checked the `0x0104` parameter map against the JT/T 808 standard (verified on both the 2013-based spec and the official JT/T 808-2019 table), and against the real capture used in `testDecodeParameterQueryResponse`.

**Wrong parameter mappings:**

- `0x0002` → set as `"port"`. Per spec this is the TCP message response timeout (seconds), unrelated to port.
- `0x0011` → set as `"plateNumber"`. Per spec this is the main server dial-up **username** (STRING). The actual plate-number parameter is `0x0083`, which *is* present in the test capture (`d4c142383838383820202020`) but currently falls through to `default` and is never decoded.
- `0x0021` → set as `"heartbeatInterval"`. Per spec this is the location-report *scheme* (a 0/1 enum), not an interval — and it duplicates the key already used for `0x0001`.
- `0x0028` → set as `"reportInterval"`. Per spec this is specifically the emergency-alarm report interval, not the general/default interval.
- `0x0046` → set as `Position.KEY_SPEED_LIMIT` via `buf.readUnsignedShort()`. Per spec this is "max single call duration" (DWORD, seconds), unrelated to speed. It's also read as 2 bytes when the field is a 4-byte DWORD, which will desync the TLV loop and corrupt parsing of any parameters that follow it in the same response.

None of `0x0002`, `0x0011`, `0x0021`, or `0x0046` appear in the real capture in the test, so these are unverified against sample data and contradict the spec.

**Mislabeled (byte-correct, wrong name):**

- `0x0027` → `"ignitionOffInterval"`. Spec defines this as the sleep-mode report interval — no ignition semantics.
- `0x0029` → `"ignitionOnInterval"`. Spec defines this as the default/normal report interval, not tied to ignition state.

**`0x0201` handler:** it calls `sendGeneralResponse(...)` before parsing the location. Looking at the existing decoder, messages that are themselves the terminal's direct answer to a platform query (e.g. `MSG_TEXT_MESSAGE_RESPONSE`, and this PR's own `0x0104` handler) don't send a general response — only spontaneous reports (`0x0200`, `0x0102`, heartbeat, etc.) do. `0x0201` is the direct reply to `0x8201`, so this extra ack looks unintended/copied from the `0x0200` handler.

`0x0080` (odometer, `raw * 100L` for the 1/10 km unit) and `0x0055`/`0x0056` (speed limit / duration) all check out against the spec.

onethings:
Testing Device are BSJ KM02, Jt808-2013 version

Testing Power Off
```
2026-08-27 00:31:26  INFO: [51a8b591: huabao > 37.111.13.243] HEX: 7e81050001024530313349000002bb7e
2026-08-27 00:31:28  INFO: [51a8b591: huabao < 37.111.13.243] HEX: 7e0001000502453031334900500000810500ec7e
```
Testing Factory Reset
```
2026-08-27 00:52:53  INFO: [1f932ca2: huabao > 37.111.15.28] HEX: 7e81050001024530313349000003ba7e
2026-08-27 00:52:53  INFO: [1f932ca2: huabao < 37.111.15.28] HEX: 7e0001000502453031334900eb0000810500577e
2026-08-27 00:53:01  INFO: [1f932ca2: huabao < 37.111.15.28] HEX: 7e0002000002453031334900ecd27e
```
Check Factory Reset Working or not
```
2026-08-27 01:04:33  INFO: [dabd95b2: huabao > 37.111.13.48] HEX: 7e810400000245303133490000b97e
2026-08-27 01:04:34  INFO: [dabd95b2: huabao < 37.111.13.48] HEX: 7e01040184024530313349013d00002c0000000104000000b400000002040000000000000003040000000000000004040000000000000005040000000000000006040000000000000007040000000000000010066d70746e657400000011000000001200000000130e36382e3138332e3232342e31333500000014000000001500000000160000000017000000001804000013970000001904000000000000002004000000000000002104000000000000002204000000000000002704000001f400000028040000000000000029040000000a0000002c04000000000000003004000000140000005004000000000000005204000000000000005304000000000000005504000000000000005604000000050000005704000000000000005804000000000000005904000000000000005a04000000000000007004000000000000007104000000000000007204000000000000007304000000000000007404000000000000008004000000000000008102002c0000008202012f000000830cd4c142383838383820202020000000840101a27e
```
Both Result are not working with Power Off and Factory Reset Command in BSJ KM02.

onethings:
Testing Device are BSJ KM02, Jt808-2013 version

Switch ip and port, success.
```
2026-08-28 17:07:15  INFO: [6b938156: huabao > 37.111.12.165] HEX: 7e8103001d024530313349000102000000130e3132302e37372e3134342e3132390000001804000003289c7e
2026-08-28 17:07:16  INFO: [6b938156: huabao < 37.111.12.165] HEX: 7e0001000502453031334900eb0001810300507e
2026-08-28 17:07:17  INFO: [6b938156: huabao < 37.111.12.165] HEX: 7e0002000002453031334900ecd27e
```

onethings:
Testing Device are BSJ KM02, Jt808-2013 version

Engine Stop
```
2026-08-28 19:34:36  INFO: [866da9af: huabao > 37.111.14.239] HEX: 7e81050001024530313349000164dc7e
2026-08-28 19:34:37  INFO: [866da9af: huabao < 37.111.14.239] HEX: 7e0001000502453031334900160001810500ab7e
2026-08-28 19:34:37  INFO: [866da9af: huabao < 37.111.14.239] HEX: 7e600600150245303133490017002a2acce1cabe3ac7d0b6cfd3cdb5e7b3c9b9a6215d7e
```


Engine Resume
```
2026-08-28 19:39:02  INFO: [1d4c0627: huabao > 37.111.14.222] HEX: 7e81050001024530313349000165dd7e
2026-08-28 19:39:03  INFO: [1d4c0627: huabao < 37.111.14.222] HEX: 7e00010005024530313349003c0001810500817e
2026-08-28 19:39:03  INFO: [1d4c0627: huabao < 37.111.14.222] HEX: 7e60060015024530313349003d002a2acce1cabe3abbd6b8b4d3cdb5e7b3c9b9a621787e
```

Engine Stop and Engine Resume success on BSJ KM02.


0x8105, 0x64 for engine stop, 0x65 for engine resume.
[jt-808通讯协议(1).doc](https://github.com/user-attachments/files/31555534/jt-808.1.doc)
[接线-808通讯协议25.01.01.doc](https://github.com/user-attachments/files/31555538/-808.25.01.01.doc)

onethings:
BSJ Km02 - jt808-2013 version retest again.

1. 復位 rebootDevice → 0x8105 0x04
```
2026-08-29 08:47:13  INFO: [dd4925b1: huabao > 37.111.12.104] HEX: 7e81050001024530313349000004bd7e
2026-08-29 08:47:14  INFO: [dd4925b1: huabao < 37.111.12.104] HEX: 7e00010005024530313349003200008105008e7e
2026-08-29 08:47:14  INFO: [dd4925b1: huabao < 37.111.12.104] HEX: 7e00010005024530313349003300008105008f7e
2026-08-29 08:47:35  INFO: [ae577232: huabao < 37.111.13.79] HEX: 7e0102000e024530313349000161757468656e7469636174696f6e307e
2026-08-29 08:47:35  INFO: [ae577232: huabao > 37.111.13.79] HEX: 7e8001000502453031334900010001010200bb7e
```


2. 位置查詢 positionSingle → 0x8201
```
2026-08-29 08:49:44  INFO: [4d170fd7: huabao > 37.111.14.229] HEX: 7e820100000245303133490000bf7e
2026-08-29 08:49:45  INFO: [4d170fd7: huabao < 37.111.14.229] HEX: 7e0201003c02453031334900140000000000000000000201000f9005bbd430000000000000070113102041010400000000eb16000c00b28995062412035796032900060089ffffffffb87e
2026-08-29 08:49:52  INFO: [4d170fd7: huabao < 37.111.14.229] HEX: 7e0002000002453031334900152b7e
```


3. 上報間隔 positionPeriodic 10s → 0x8103 0x29
```
2026-08-29 08:51:35  INFO: [8b2c1e05: huabao > 37.111.12.33] HEX: 7e8103000a02453031334900000100000029040000000a927e
2026-08-29 08:51:35  INFO: [8b2c1e05: huabao < 37.111.12.33] HEX: 7e00010005024530313349002000008103009a7e
2026-08-29 08:51:36  INFO: [8b2c1e05: huabao < 37.111.12.33] HEX: 7e0200003a0245303133490021000000000000000201000f9005bbd430000000000002070113102232010400000000eb16000c00b28995062412035796032900060089fffffffff97e
```


4. 限速 100 setSpeedLimit → 0x8103 0x55+0x56
```
2026-08-29 08:53:35  INFO: [ff1b2f19: huabao > 37.111.13.31] HEX: 7e81030013024530313349000002000000550400000064000000560400000023eb7e
2026-08-29 08:53:35  INFO: [ff1b2f19: huabao < 37.111.13.31] HEX: 7e00010005024530313349002f0000810300957e
2026-08-29 08:53:36  INFO: [ff1b2f19: huabao < 37.111.13.31] HEX: 7e0200003a0245303133490030000000000000000201000f9005bbd430000000000000070113102431010400000000eb16000c00b28995062412035796032900060089ffffffffef7e
```


5. 文字訊息 "Hello" → 0x8300 flag=0x04
```
2026-08-29 08:55:34  INFO: [7e5c970d: huabao > 37.111.13.230] HEX: 7e8300000602453031334900000448656c6c6fff7e
2026-08-29 08:55:34  INFO: [7e5c970d: huabao < 37.111.13.230] HEX: 7e0704016b024530313349003f000601003a000000000000000301000f9005bbd430000000000009070113102542010400000000eb16000c00b28995062412035796032900060089ffffffff003a000000000000000301000f9005bbd430000000000009070113102543010400000000eb16000c00b28995062412035796032900060089ffffffff003a000000000000000301000f9005bbd430000000000009070113102552010400000000eb16000c00b28995062412035796032900060089ffffffff003a000000000000000301000f9005bbd430000000000009070113102602010400000000eb16000c00b28995062412035796032900060089ffffffff003a000000000000000301000f9005bbd430000000000000070113102612010400000000eb16000c00b28995062412035796032900060089ffffffff003a000000000000000301000f9005bbd430000000000000070113102622010400000000eb16000c00b28995062412035796032900060089ffffffff0f7e7e0001000502453031334900400000830000fb7e
2026-08-29 08:55:34  INFO: [7e5c970d: huabao < 37.111.13.230] HEX: 7e0200003a0245303133490041000000000000000301000f9005bbd430000000000000070113102630010400000000eb16000c00b28995062412035796032900060089ffffffff9c7e
```


6. 語音 "Hello" → 0x8300 flag=0x08
```
2026-08-29 08:57:39  INFO: [6eb3f450: huabao > 37.111.15.21] HEX: 7e8300000602453031334900000848656c6c6ff37e
2026-08-29 08:57:40  INFO: [6eb3f450: huabao < 37.111.15.21] HEX: 7e0001000502453031334900530000830000e87e
```


7. 停止跟蹤 positionStop → 0x8202
```
2026-08-29 08:58:51  INFO: [c70a042d: huabao > 37.111.14.85] HEX: 7e820200060245303133490000000000000000ba7e
2026-08-29 08:58:52  INFO: [c70a042d: huabao < 37.111.14.85] HEX: 7e00010005024530313349005d0000820200e57e
2026-08-29 08:58:53  INFO: [c70a042d: huabao < 37.111.14.85] HEX: 7e00020000024530313349005e607e
```


8. 輸出控制 outputControl index=1 → 0x8500
```
2026-08-29 09:00:16  INFO: [ecfd12b7: huabao > 37.111.12.36] HEX: 7e85000001024530313349000001b97e
2026-08-29 09:00:37  INFO: [e3ab7657: huabao < 37.111.15.89] HEX: 7e0102000e024530313349006961757468656e7469636174696f6e587e
2026-08-29 09:00:37  INFO: [e3ab7657: huabao > 37.111.15.89] HEX: 7e8001000502453031334900010069010200d37e
2026-08-29 09:00:37  INFO: [e3ab7657: huabao < 37.111.15.89] HEX: 7e0704007b024530313349006a000201003a000000000000000201000f9005bbd430000000000001070113103041010400000000eb16000c00b28995062412035796032900060089ffffffff003a000000000000000301000f9005bbd430000000000001070113103040010400000000eb16000c00b28995062412035796032900060089ffffffff2d7e
2026-08-29 09:00:38  INFO: [e3ab7657: huabao < 37.111.15.89] HEX: 7e0200003a024530313349006b000000000000000201000f9005bbd430000000000001070113103134010400000000eb16000c00b28995062412035796032900060089ffffffffa57e
```


9. 設防 alarmArm → 0x8103 0x24
```
2026-08-29 09:03:15  INFO: [63b30aed: huabao > 37.111.12.26] HEX: 7e8103000b02453031334900000100000024050175736572857e
2026-08-29 09:03:15  INFO: [63b30aed: huabao < 37.111.12.26] HEX: 7e00010005024530313349007f0000810300c57e
```


10. 撤防 alarmDisarm → 0x8103 0x24
```
2026-08-29 09:04:38  INFO: [e8a7d0f7: huabao > 37.111.15.46] HEX: 7e8103000b02453031334900000100000024050075736572847e
2026-08-29 09:04:39  INFO: [e8a7d0f7: huabao < 37.111.15.46] HEX: 7e00010005024530313349008a0000810300307e
```


11. 報警確認 alarmDismiss → 0x8203
```
2026-08-29 09:06:35  INFO: [201b5988: huabao > 37.111.12.156] HEX: 7e820300060245303133490000000000000000bb7e
2026-08-29 09:06:35  INFO: [201b5988: huabao < 37.111.12.156] HEX: 7e00010005024530313349009700008203002e7e
2026-08-29 09:06:35  INFO: [201b5988: huabao < 37.111.12.156] HEX: 7e0200003a0245303133490098000000000000000201000f9005bbd430000000000000070113103731010400000000eb16000c00b28995062412035796032900060089ffffffff547e
2026-08-29 09:06:37  INFO: [201b5988: huabao < 37.111.12.156] HEX: 7e000200000245303133490099a77e
```


12. 拍照 requestPhoto → 0x8801
```
2026-08-29 09:08:13  INFO: [3097801a: huabao > 37.111.15.34] HEX: 7e8801000c0245303133490000000001000000000000000000b87e
2026-08-29 09:08:14  INFO: [3097801a: huabao < 37.111.15.34] HEX: 7e0001000502453031334900a50000880100147e
```


中文訊息 GBK "測試"
```
2026-08-29 09:09:47  INFO: [d399bf64: huabao > 37.111.14.146] HEX: 7e830000050245303133490000049c79d487087e
2026-08-29 09:09:47  INFO: [d399bf64: huabao < 37.111.14.146] HEX: 7e0001000502453031334900b100008300000a7e
2026-08-29 09:09:54  INFO: [d399bf64: huabao < 37.111.14.146] HEX: 7e0002000002453031334900b28c7e
```


-------
No.8 outputControl retest
```
2026-08-29 09:19:03  INFO: [4e80f445: huabao > 37.111.14.6] HEX: 7e85000001024530313349000001b97e
2026-08-29 09:19:04  INFO: [4e80f445: huabao < 37.111.14.6] HEX: 7e0002000002453031334900f2cc7e
2026-08-29 09:19:04  INFO: [4e80f445: huabao < 37.111.14.6] HEX: 7e0001000502453031334900f300008500004e7e
2026-08-29 09:19:12  INFO: [4e80f445: huabao < 37.111.14.6] HEX: 7e0002000002453031334900f4ca7e
```

No.12. Km02 no camera.

.gitignore
```diff
@@ -12,3 +12,4 @@ out
 build
 .vscode
 bin
+.gradle-home/
```
tananaev: This is unrelated. Please revert.

tananaev:
It seems like your comments are unrelated to this PR. I don't see commands. Have you mixed up pull requests?

onethings:
Thanks for the review.

1. The unrelated `.gitignore` change has been reverted and pushed.

2. Sorry for the confusion — those device test captures (engine stop/resume, output control, alarm arm/disarm, photo, etc.) are command tests that belong to the **encoder PR #5998**, not this decoder PR. They were posted here by mistake. The relevant command work is tracked separately in #5998.

onethings:
Hi @tananaev, thanks for the review.

I've cleaned up both PRs — the unrelated `.gitignore` change has been reverted (in both #5997 and #5998), and those command-test comments I posted earlier were meant for the encoder PR #5998.

Could you let me know what you'd like me to do next? In particular, is splitting this into a decoder PR (#5997) and an encoder PR (#5998) the right approach, or would you prefer any changes before a re-review?

src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java
```diff
@@ -433,6 +433,20 @@ public void testDecode() throws Exception {
                 "7E09000049487060054219000E54100000000000000000000000000000000000240112034640001F2401120345210C4100F4011C00F3FFE80008004003FE001F2401120345510C4100F4011C00F3FFF0000800A003FEB67E"),
                 Position.KEY_DEVICE_TEMP, 24.4);
 
+        verifyPosition(decoder, binary(
```
tananaev, line 436: Why do we need this test?

src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java
```diff
@@ -433,6 +433,20 @@ public void testDecode() throws Exception {
...
+                "7e0201003c0145382363912db0000000000000000000020101313605bb3b58000c000001292608251203040104000b620beb16000c00b28995018202107439115800060089fffffffece7e"),
+                position("2026-08-25 04:03:04.000", true, 16.85535, 96.156504));
+
+        verifyAttribute(decoder, binary(
```
tananaev, line 440: Please reformat for this to have HEX on a single line, like we do for all other tests.

src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java
onethings: Test for request location update from real device(to replace generated samples for tests). https://github.com/traccar/traccar/pull/5992/changes/BASE..083ed8edc3820be42bc98f8f5ea23f277002b297#diff-495de3442cfaec31fc7c21302db0dd86ac3efe447efbf9f77f3a1e9c0894d75c

src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java
onethings: Fixed — the 0x0104 test HEX is now on a single line, matching the rest of the file. I also switched it to a real device capture (BSJ KM02, terminal 024530313349).

src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java
tananaev: I think you misunderstood my question. Why do you need two test cases for this PR?

tananaev: Oh I see now. You added two message types.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
tananaev: It looks like this is not resolved.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -565,6 +573,64 @@ protected Object decode(
...
+                        position.set("defaultReportInterval", buf.readUnsignedInt());
+                        break;
+                    case 0x0055:
+                        position.set(Position.KEY_SPEED_LIMIT, (int) buf.readUnsignedInt());
+                        break;
+                    case 0x0056:
+                        position.set("speedLimitDuration", (int) buf.readUnsignedInt());
+                        break;
+                    case 0x0080:
+                        position.set(Position.KEY_ODOMETER, buf.readUnsignedInt() * 100L);
```
tananaev: Why 100L instead of just 100?

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -565,6 +573,64 @@ protected Object decode(
...
+                        position.set("emergencyReportInterval", buf.readUnsignedInt());
+                        break;
+                    case 0x0029:
+                        position.set("defaultReportInterval", buf.readUnsignedInt());
+                        break;
+                    case 0x0055:
+                        position.set(Position.KEY_SPEED_LIMIT, (int) buf.readUnsignedInt());
+                        break;
+                    case 0x0056:
+                        position.set("speedLimitDuration", (int) buf.readUnsignedInt());
```
tananaev: Why `(int)`?

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -565,6 +573,64 @@ protected Object decode(
...
+
+            Position position = new Position(getProtocolName());
+            position.setDeviceId(deviceSession.getDeviceId());
+
+            getLastLocation(position, null);
+
+            buf.readUnsignedShort(); // response serial number
+            buf.readUnsignedByte(); // parameter count
+
+            int endIndex = buf.readerIndex() + bodyLength - 3;
```
tananaev: I believe you can move this a big higher and avoid this `- 3` stuff.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -565,6 +573,64 @@ protected Object decode(
...
+            while (buf.readerIndex() < endIndex) {
+                int parameterId = buf.readInt();
+                int length = buf.readUnsignedByte();
+                int valueEndIndex = buf.readerIndex() + length;
```
tananaev: The naming seems inconsistent with the rest of the decoder. I think instead of parameter id we should use subtype or something. And end index should be consistent.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: I have ask Deepseek v4 flash before why add L behind the number, ai asnwer this are long number or something.

onethings: Screenshot 2026-08-30 at 09 43 12
And i complaint with deepseek v4 flash about L charater behind the number, and give me other sample that code have before. I search 1000L in code and found a lot in code, may be can have a traccar java format standard can let ai do more efficient work?

tananaev: It's not about formatting. It's about type conversion. Do you understand why some places have it and some don't?

onethings: Now understand a little, u said is a type coversion, for unit coversion.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: Fixed. The 0x0046 case has been removed entirely. Per the spec it is not a speed parameter. 0x0055 (overspeed alarm threshold) now maps to KEY_SPEED_LIMIT, which is the device configured speed limit and is confirmed in the real device capture.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: Fixed. All parameter IDs now use a consistent 4-digit hex format (0x0001, 0x0010, ..., 0x0083).

src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java
onethings: Fixed. The parameter query tests are consolidated into the single testDecode() method, and the 0x0104 test now uses one real device capture (BSJ KM02, terminal 024530313349) with HEX on a single line.

.gitignore
onethings: Fixed. The unrelated .gitignore change has been reverted and pushed.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: Understood. readUnsignedInt() returns a long because a 32-bit unsigned value can exceed Integer.MAX_VALUE, so the arithmetic must stay in long to avoid overflow. Where the code uses * 100L it marks the literal as long explicitly; where it uses * 100, the long operand from readUnsignedInt() already forces the promotion. Either way the result stays in long.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: Fixed. Removed the (int) casts. readUnsignedInt() already returns the value directly.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: Fixed. The loop now uses buf.readableBytes() > 2 directly instead of bodyLength - 3.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: Fixed. Renamed to subtype and endIndex to match the rest of the decoder.

onethings:
Finally , found have to lookup this five file, after split. 
PR #5997 → 看 jt808-decoder 這個 branch
順序	檔案
1️⃣ 主邏輯	Jt808ProtocolDecoder.java
2️⃣ 測試	Jt808ProtocolDecoderTest.java
/Users/kevinyang/Documents/traccar/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
/Users/kevinyang/Documents/traccar/src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java

https://github.com/onethings/traccar/blob/jt808-decoder/src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
https://github.com/onethings/traccar/blob/jt808-decoder/src/test/java/org/traccar/protocol/Jt808ProtocolDecoderTest.java



PR #5998 → 看 jt808-encoder 這個 branch
順序	檔案
1️⃣ 主邏輯	Jt808ProtocolEncoder.java
2️⃣ 註冊	Jt808Protocol.java
3️⃣ 測試	Jt808ProtocolEncoderTest.java
/Users/kevinyang/Documents/traccar/src/main/java/org/traccar/protocol/Jt808ProtocolEncoder.java
/Users/kevinyang/Documents/traccar/src/main/java/org/traccar/protocol/Jt808Protocol.java
/Users/kevinyang/Documents/traccar/src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java

https://github.com/onethings/traccar/blob/jt808-encoder/src/main/java/org/traccar/protocol/Jt808ProtocolEncoder.java

https://github.com/onethings/traccar/blob/jt808-encoder/src/main/java/org/traccar/protocol/Jt808ProtocolEncoder.java
https://github.com/onethings/traccar/blob/jt808-encoder/src/main/java/org/traccar/protocol/Jt808Protocol.java
https://github.com/onethings/traccar/blob/jt808-encoder/src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java

onethings:
Should i close this or not?

tananaev:
It depends on you. There are still issues. One is merge conflict and another is `KEY_SPEED_LIMIT`.

onethings:
Hi @tananaev, both issues are now addressed — thanks for flagging them.

**Merge conflict:** the branch was stale relative to `master` (the encoder constants from #5998 landed upstream in the same block). I merged the latest `master` in and combined the constant declarations — there's no value overlap, and the PR now reports as mergeable.

**`KEY_SPEED_LIMIT`:** you're right, that mapping was wrong. `Position.KEY_SPEED_LIMIT` is set by `SpeedLimitHandler` from map data for the actual street speed limit, and the overspeed event handlers rely on that semantic. The JT808 `0x0055` value in a parameter query response is the device-configured overspeed threshold, so it should not reuse that key. I now store it as a plain attribute (`overspeedThreshold`) like the other reported device parameters.

Also removed a redundant comment in the parameter query loop.

`Jt808ProtocolDecoderTest` and `Jt808ProtocolEncoderTest` both pass locally. Head is now `b3b844f26`. Could you take another look?

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -575,6 +583,67 @@ protected Object decode(
...
+
+            Position position = new Position(getProtocolName());
+            position.setDeviceId(deviceSession.getDeviceId());
+
+            getLastLocation(position, null);
+
+            buf.readUnsignedShort(); // response serial number
+            buf.readUnsignedByte(); // parameter count
+
+            while (buf.readableBytes() > 2) {
```
tananaev: This check looks incorrect. You're reading 5 bytes immediately. Why are we only checking for 2?

onethings: You're right, thanks. The > 2 guard was copied from the network sub-block loop in decodeLocation, where each entry header is only 2 bytes (1-byte subtype + 1-byte length). Here each parameter entry is a 4-byte ID plus a 1-byte length, so a single iteration reads 5 bytes up front, and readInt() would underflow if only 3–4 bytes were left. I'll change the condition to buf.readableBytes() >= 5.

tananaev: This is still incorrect because you'r not accounting for the checksum and delimiter.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -575,6 +583,67 @@ protected Object decode(
...
+                }
+                switch (subtype) {
+                    case 0x0001:
+                        position.set("heartbeatInterval", buf.readUnsignedInt());
+                        break;
+                    case 0x0010:
+                        position.set("apn", buf.readCharSequence(length, StandardCharsets.US_ASCII).toString());
+                        break;
+                    case 0x0013:
+                        position.set("server", buf.readCharSequence(length, StandardCharsets.US_ASCII).toString());
```
tananaev: Don't we need to use `GBK` here was well?

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: You are right — the frame decoder keeps the trailing 1-byte checksum and the closing 0x7e in the buffer, so readableBytes() includes them. Each parameter entry needs a 5-byte header (4-byte id + 1-byte length), so the loop now only continues while at least 5 real parameter bytes remain: buf.readableBytes() - 2 >= 5. Fixed in b85537e43.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: Good catch. APN and server addresses are ASCII in practice, but GBK is backward-compatible with ASCII and 0x0083 (plateNumber) already used it, so I now resolve the charset once before the loop and use it for all string parameters (apn, server, plateNumber). Fixed in 2408b78a2.

onethings: Good catch. APN and server addresses are ASCII in practice, but GBK is backward-compatible with ASCII and 0x0083 (plateNumber) already used it, so I now resolve the charset once before the loop and use it for all string parameters (apn, server, plateNumber). Fixed in 2408b78a2.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -575,6 +583,83 @@ protected Object decode(
...
+                            position.set("overspeedThreshold", value.readUnsignedInt());
+                        }
+                        break;
+                    case 0x0056:
+                        if (length == 4) {
+                            position.set("speedLimitDuration", value.readUnsignedInt());
+                        }
+                        break;
+                    case 0x0080:
+                        if (length == 4) {
```
tananaev: Why are we checking length in every single case?

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -575,6 +583,83 @@ protected Object decode(
...
+                            position.set(Position.KEY_ODOMETER, value.readUnsignedInt() * 100); // 0.1 km units
```
tananaev: Unnecessary comment.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -575,6 +583,83 @@ protected Object decode(
...
+            Charset charset = Charset.isSupported("GBK") ? Charset.forName("GBK") : StandardCharsets.US_ASCII;
+
+            while (buf.readableBytes() - 2 >= 5) {
+
+                int subtype = buf.readInt();
+                int length = buf.readUnsignedByte();
+                if (length > buf.readableBytes() - 2) {
+                    break;
+                }
+                ByteBuf value = buf.readSlice(length);
```
tananaev: I don't think makes sense. Why did you have to rewrite the whole decoding? This triggers more unnecessary back and forth.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -575,6 +583,83 @@ protected Object decode(
...
+                    case 0x0001:
+                        if (length == 4) {
+                            position.set("heartbeatInterval", value.readUnsignedInt());
+                        }
+                        break;
+                    case 0x0010:
+                        position.set("apn", value.toString(charset));
+                        break;
+                    case 0x0013:
+                        position.set("server", value.toString(charset));
```
tananaev: Do we need trim here?

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -575,6 +583,83 @@ protected Object decode(
...
+
+            Position position = new Position(getProtocolName());
+            position.setDeviceId(deviceSession.getDeviceId());
+
+            getLastLocation(position, null);
+
+            buf.readUnsignedShort(); // response serial number
+            buf.readUnsignedByte(); // parameter count
+
+            Charset charset = Charset.isSupported("GBK") ? Charset.forName("GBK") : StandardCharsets.US_ASCII;
```
tananaev: It looks like we do something similar in other places. Maybe we can define once. Either decoder or encoder.

onethings: Done — I defined it once as a CHARSET_GBK static constant in the decoder and reuse it here and in the text-message handler. The encoder copies came from the already-merged encoder PR, so I left that file untouched in this decoder PR.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: You're right, sorry. I restructured this loop into a bounded-slice version with per-case length checks, which wasn't needed and diverged from the rest of the decoder. I've reverted it to the endIndex-style parsing already used in decodeLocation, keeping only the trailing checksum/delimiter guard.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: Yes — the real device pads the plate number with trailing spaces (the capture value ends in ...38 38 38 38 38 20 20 20), so without trim() the stored value would keep the trailing spaces. The test expects the clean 粤B88888.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: Removed — the per-case length == 4 guards are gone now that the loop is back to the endIndex style.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: Removed.

Issue #1957, opened 2016-05-18
API for querying positions by GroupID


m-menon:
hi, just wanted to confirm - is there a way to query for positions using the API - but I can see that the API doesn't have the groupid parameter or filter. Thanks

tananaev:
Filtering is done on the client side.

m-menon:
alright thanks :+1:

PR #5998, opened 2026-08-26
Add JT808 downlink command types

onethings:
### Summary
Split out the **encoder** part of the JT808 work. The decoder changes are kept in a separate PR (#5997).

### Encoder changes (`Jt808ProtocolEncoder.java`)
Encode the standard JT808 downlink commands and register them in `Jt808Protocol.java`:
- `TYPE_SET_CONNECTION` → `0x8103` (server address `0x13` + TCP port `0x18`)
- `TYPE_POSITION_SINGLE` / `TYPE_POSITION_STOP` → `0x8201` / `0x8202`
- `TYPE_ALARM_DISMISS` → `0x8203`
- `TYPE_POWER_OFF` / `TYPE_FACTORY_RESET` → `0x8105` terminal control
- `TYPE_MESSAGE` / `TYPE_VOICE_MESSAGE` → `0x8300` (GBK)
- `TYPE_REQUEST_PHOTO` → `0x8801`
- `TYPE_SET_SPEED_LIMIT` → `0x8103` (overspeed threshold `0x55`; duration `0x56` written only when provided)
- `TYPE_OUTPUT_CONTROL` → `0x8500`
- `TYPE_CONFIGURATION` → `0x8104`

Existing command behavior (`TYPE_REBOOT_DEVICE`, `TYPE_POSITION_PERIODIC`, `TYPE_ALARM_ARM`/`TYPE_ALARM_DISARM`, and others) is left untouched in this PR. Any change to existing behavior will be proposed separately for review first.

### Tests
All `Jt808ProtocolEncoderTest` tests pass. The new downlink command tests are consolidated into the existing `testEncode()` method.


tananaev:
Have you reviewed all the other docs for conflicts?

tananaev:
It looks like you haven't responded to my question here. Also I see you again have some changes to decoder here mixed with commands.

onethings:
@tananaev — sorry, let me address both points.

**On reviewing the docs:** Yes. I reviewed the JT/T 808-2013 and JT/T 808-2019 specs (including the linked standards) for conflicts. The downlink command types and parameter IDs are aligned with the standard tables in both versions: `POSITION_PERIODIC` → `0x0028` (report interval), `ALARM_ARM`/`ALARM_DISARM` → `0x0024` (arm/disarm), overspeed threshold/duration → `0x0055`/`0x0056`.

**On the decoder changes mixed in:** You're right, those don't belong in this PR. I'll move the decoder-side changes (the downlink message constants, `decodeLocation` boundary check, cell-tower fix, and the decoder test) out of this PR — they belong in the decoder PR #5997. This PR will be kept to just the encoder/command changes.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -108,6 +113,10 @@ public Integer getProtocolVersion() {
         return protocolVersion;
     }
 
+    void setProtocolVersion(Integer protocolVersion) {
```
tananaev: We don't usually use default visibility. This should probably be public. Also why not `int` argument?

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -701,9 +710,6 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
             int subtype = buf.readUnsignedByte();
             int length = buf.readUnsignedByte();
             int endIndex = buf.readerIndex() + length;
-            if (endIndex > buf.writerIndex()) {
```
tananaev, line 704: Why are we removing this?

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
```diff
@@ -909,12 +915,8 @@ private Position decodeLocation(DeviceSession deviceSession, ByteBuf buf) {
                         int mcc = buf.readUnsignedShort();
                         int mnc = buf.readUnsignedShort();
                         while (buf.readerIndex() < endIndex) {
-                            int lac = buf.readUnsignedMedium();
-                            long cid = buf.readUnsignedInt();
-                            int rssi = buf.readUnsignedByte();
-                            if (lac > 0 && lac <= 0xffff && cid <= 0xfffffff) {
```
tananaev, line 915: And this?

tananaev:
Are you reviewing your own code? Please make sure you do before pushing.

onethings:
> Are you reviewing your own code? Please make sure you do before pushing.

Thank for review, all write by Deepseek v4 flash. I don't understand java. All suggest are appericate and i will let gemini for double check to follow traccar java writing style compare with other protocol.

tananaev:
AI should be able to explain you the changes, so if you can't review it yourself, you should review it with the help of AI. When you contribute, the expectation is that you understand the code and what it's doing.

onethings:
Yeah, that should be a standard procedure. Thanks.

src/main/java/org/traccar/protocol/Jt808ProtocolDecoder.java
onethings: Change to public anud use int.

onethings:
Hi @tananaev,

I have addressed the previous feedback on this PR:

- Decoder-side changes are now moved out — this PR is kept to just the encoder / command-registration changes. The decoder work lives in #5997.
- Applied the feedback on `setProtocolVersion` (public visibility, `int` argument).
- As stated in the PR description, all existing command behavior is left untouched; any behavioral change would be proposed separately for review first.

I also went through the changes with AI assistance to make sure I understand what each command encoding does, as you suggested.

Could you please take another look when you have a chance? Thanks!

tananaev:
You need to cut down on documentation in the encoder. Only keep absolutely essential short comments.

And we also need to cleanup test cases. We don't need a 100% coverage. We just need to test what makes sense the most. Maybe 3-4 tests max.

tananaev:
To request review - just click the button.

onethings:
> To request review - just click the button.

Ok.

src/main/java/org/traccar/protocol/Jt808ProtocolEncoder.java
```diff
@@ -127,6 +140,67 @@ protected Object encodeCommand(Channel channel, Command command) {
...
+                    return decoder.formatMessage(
+                            Jt808ProtocolDecoder.MSG_TAKE_PHOTO, id, false, data);
+                case Command.TYPE_SET_SPEED_LIMIT:
+                    data.writeByte(2); // number of parameters
+                    data.writeInt(0x0055); // parameter id: overspeed threshold
+                    data.writeByte(4); // parameter value length
+                    data.writeInt(command.getInteger(Command.KEY_DATA));
+                    data.writeInt(0x0056); // parameter id: overspeed duration
+                    data.writeByte(4); // parameter value length
+                    data.writeInt(35); // duration in seconds
```
tananaev: Where did 35 seconds come from?

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
```diff
@@ -55,4 +55,79 @@ public void testEncodeJimiCustom() throws Exception {
 
     }
 
+    @Test
+    public void testEncodeCommands() throws Exception {
```
tananaev: Why is this separate from other tests? I still see more tests than needed. I though you fixed it?

src/main/java/org/traccar/protocol/Jt808ProtocolEncoder.java
onethings: I set 100km/h , duration for 35 second, for speeding alarm. So 100 and 35 are user have to input to define speed rate and time duration for define user's speeding. For BSJ jt808 device's speeding alarm needed.

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
onethings: All move to testEncoder().

One question  for Video comand, if the device have no camera but user still saw the Video command. I though full test for each model will force us to define more subdivision like by model, manufacture, port, every model have it niche market and device separate funtion come to market have it own reason. Each device model life just 4-5 year only.I though with ai we can fully support device features by model or by manufacture, or port( like 6015 fot jt808 with camera or 5215, port are the best way to split and most clean way)?, cause want to give user a clear device feature command that he can use, otherwise extra explain service have to answer to user)
. Or more protocol mutation can support in config(traccar.xml)? In reality market some device from china like bsj km02 supoort multi protocol by flash different firmware, common are jt808, gt06, h02 that i have found(some are limitation by regions).

src/main/java/org/traccar/protocol/Jt808ProtocolEncoder.java
tananaev: What? This doesn't answer my question. Why is the value hardcore?

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
tananaev: That discussion is out of scope for this PR. We already have too much back and forth here. Please don't expand scope.

src/main/java/org/traccar/protocol/Jt808ProtocolEncoder.java
onethings: Thanks for the feedback, and you're right — the hardcoded value didn't make sense. The 35-second duration is no longer hardcoded; it's now a user-input parameter.

Changes:

Jt808ProtocolEncoder.java: in TYPE_SET_SPEED_LIMIT, the overspeed duration (parameter 0x56) now reads command.getInteger(Command.KEY_DURATION) instead of the fixed 35.
Command.java: added a new command key KEY_DURATION ("duration").
Frontend: added the setSpeedLimit command with two fields — speed limit (data) and duration (duration, in seconds) — so the user enters both the threshold (e.g. 100 km/h) and the continuous-duration (e.g. 35 s). See https://github.com/traccar/traccar-web/pull/1957
So both values are now fully user-configurable.

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
onethings: Understood — keeping the scope tight and tests minimal. Thanks.

onethings: > Why is this separate from other tests? I still see more tests than needed. I though you fixed it?

@tananaev — fixed properly now. The hardcoded `35` was a leftover placeholder
from testing; it's removed.

Final behavior in e69d48027:
- `0x0056` is no longer written with a fixed value.
- The encoder writes `0x0056` only when `duration` is actually provided on the
  command (`KEY_DURATION`); otherwise it sends only the threshold (`0x0055`)
  and the terminal keeps its current duration setting.
- Web UI (traccar-web #1957): the `duration` field was dropped to keep the scope
  small, so the UI only sets the speed limit.

A test was added for the no-duration case.

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
```diff
@@ -35,6 +35,67 @@ public void testEncode() throws Exception {
...
+            encodeCommand(encoder, decoder, command));
+
+        command.setType(Command.TYPE_SET_SPEED_LIMIT);
+        command.set(Command.KEY_DATA, 100);
+        command.set(Command.KEY_DURATION, 35);
+        verifyFrame(
+            binary("7e810300130b3a73ce2ff2000002000000550400000064000000560400000023867e"),
+            encodeCommand(encoder, decoder, command));
+
+        Command speedLimitCommand = new Command();
```
tananaev: Why do we need special variable here?

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
```diff
@@ -35,6 +35,67 @@ public void testEncode() throws Exception {
...
+        verifyFrame(
+            binary("7e8103000a0b3a73ce2ff2000001000000550400000064ed7e"),
+            encodeCommand(encoder, decoder, speedLimitCommand));
+
+        command.setType(Command.TYPE_CONFIGURATION);
+        verifyFrame(
+            binary("7e810400000b3a73ce2ff20000d47e"),
+            encodeCommand(encoder, decoder, command));
+
+        // JT/T 808-2019 protocol (10-byte id)
```
tananaev: Remove comment

src/main/java/org/traccar/protocol/Jt808ProtocolEncoder.java
```diff
@@ -127,6 +140,70 @@ protected Object encodeCommand(Channel channel, Command command) {
...
+                    data.writeByte(command.getInteger(Command.KEY_INDEX, 0)); // channel
+                    data.writeShort(1); // take one photo
+                    data.writeShort(0);
+                    data.writeByte(0);
+                    data.writeByte(0);
+                    data.writeByte(0);
+                    data.writeByte(0);
+                    data.writeByte(0);
+                    data.writeByte(0);
+                    data.writeByte(0);
```
tananaev: There has to be a better way.

src/main/java/org/traccar/protocol/Jt808ProtocolEncoder.java
```diff
@@ -127,6 +140,70 @@ protected Object encodeCommand(Channel channel, Command command) {
...
+                    return decoder.formatMessage(
+                            Jt808ProtocolDecoder.MSG_TAKE_PHOTO, id, false, data);
+                case Command.TYPE_SET_SPEED_LIMIT:
+                    boolean hasDuration = command.hasAttribute(Command.KEY_DURATION);
+                    data.writeByte(hasDuration ? 2 : 1); // number of parameters
+                    data.writeInt(0x0055); // parameter id: overspeed threshold
+                    data.writeByte(4); // parameter value length
```
tananaev: Can we cleanup comments more.

src/main/java/org/traccar/protocol/Jt808ProtocolEncoder.java
onethings: Thanks. Replaced the seven writeByte(0) calls with a single data.writeZero(9), which writes the photo interval and the default image parameters in one call.

src/main/java/org/traccar/protocol/Jt808ProtocolEncoder.java
onethings: Done. Removed the redundant "parameter value length" comments and shortened the parameter-id comments in TYPE_SET_SPEED_LIMIT.

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
onethings: Good point. Removed the separate speedLimitCommand and now reuse the shared command: first the case without duration, then set KEY_DURATION for the two-parameter case.

src/test/java/org/traccar/protocol/Jt808ProtocolEncoderTest.java
onethings: Removed.

PR #6003, opened 2026-08-28
Fix Bug in flespi decoder on unknow attribute

### LINKED ISSUE #6002: Bug in flespi decoder on unknow attribute
There is a bug    in  [org.traccar.protocol.FlespiProtocolDecoder.java](https://github.com/traccar/traccar/blob/master/src/main/java/org/traccar/protocol/FlespiProtocolDecoder.java)

in this method

```
private void decodeUnknownParam(String name, JsonValue value, Position position) {
        switch (value) {
            case JsonNumber jsonNumber -> {
                if (jsonNumber.isIntegral()) {
                    position.set(name, jsonNumber.longValue());
                } else {
                    position.set(name, jsonNumber.doubleValue());
                }
                position.set(name, jsonNumber.doubleValue());  // <<--  forcing the value to a double, even when it is an integer.
            }
            case JsonString jsonString -> position.set(name, jsonString.getString());
            case JsonValue v when v == JsonValue.TRUE || v == JsonValue.FALSE ->
                    position.set(name, v == JsonValue.TRUE);
            default -> {}
        }
    }
```
the row

`position.set(name, jsonNumber.doubleValue());`

after if block

```
 if (jsonNumber.isIntegral()) {
        position.set(name, jsonNumber.longValue());
    } else {
        position.set(name, jsonNumber.doubleValue());
    }
```

overrides the correct behavior by forcing the value to a double, even when it is an integer.

jackom83:
fix  #6002

### DIFF
--- a/src/main/java/org/traccar/protocol/FlespiProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/FlespiProtocolDecoder.java
@@ -295,7 +295,6 @@ public class FlespiProtocolDecoder extends BaseHttpProtocolDecoder {
                 } else {
                     position.set(name, jsonNumber.doubleValue());
                 }
-                
             }
             case JsonString jsonString -> position.set(name, jsonString.getString());
             case JsonValue v when v == JsonValue.TRUE || v == JsonValue.FALSE ->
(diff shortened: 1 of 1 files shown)


src/main/java/org/traccar/protocol/FlespiProtocolDecoder.java
```diff
@@ -295,7 +295,7 @@ private void decodeUnknownParam(String name, JsonValue value, Position position)
                 } else {
                     position.set(name, jsonNumber.doubleValue());
                 }
-                position.set(name, jsonNumber.doubleValue());
+
```
tananaev: Why do we need a space here?

jackom83: No, it was just an oversight; I removed it.

PR #6008, opened 2026-09-02
Withdrawn

pbm9z95m6z-hue:
Withdrawn by the author.


tananaev:
1. Do you have the protocol documentation for your device?
2. Is the sample from real device?
3. The right fix is probably check the length upfront.

PR #6009, opened 2026-09-02
Dispatch MCP tool calls through the REST API

shanelord01:
Draft prototype for the MCP and REST unification discussed in #5976, kept to the minimum needed to show the wiring. One tool, reports-summary, builds a ContainerRequest from the tool arguments and runs it through the API application handler in process, so authentication, permission checks, date parsing, error mapping and action logging all stay in the REST layer.

Tested locally against H2 with a device feeding OsmAnd positions. The summary result matches the REST endpoint, a bad date returns the same 404, and a non-admin user with disableReports is refused the same way.


tananaev:
This doesn't look very clean to me.

shanelord01:
I opened #6010 as an alternative.

PR #6010, opened 2026-09-02
Call REST API from MCP tools

shanelord01:
Alternative to #6009 for the MCP and REST unification discussed in #5976. The MCP tool reuses the REST endpoint, but the call goes over HTTP to the server's own API using the JAX-RS Client that is already injected. The tool passes the caller's bearer token through, so authentication, permission checks, date parsing, error mapping and action logging are the normal REST path.

I've tried to simplify it and clean it up - this is my last go - if not right I'm not sure if I'm up to this.


tananaev:
Decided to implement something myself: https://github.com/traccar/traccar/commit/bca9628dba83f9beb67788ac832623fefd867221.

PR #6012, opened 2026-09-04
Fix LTE cell decoding for Topin protocol

BSyndrom:
The `MSG_LBS_WIFI_2` message uses 32-bit TAC and ECI fields for LTE cells, while the current decoder reads both values as 16-bit fields.

This causes the cell data to become misaligned and produces incorrect network information. Remaining bytes can also be interpreted as an alarm.

This change decodes LTE cells using a 32-bit TAC, unsigned 32-bit ECI and signal byte, while keeping the existing format for the other LBS message types.

Added a regression test using a packet captured from a Topin ZX909 device.

### DIFF
--- a/src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/TopinProtocolDecoder.java
@@ -252,8 +252,15 @@ public class TopinProtocolDecoder extends BaseProtocolDecoder {
             int mcc = buf.readUnsignedShort();
             int mnc = buf.readUnsignedByte();
             for (int i = 0; i < cellCount; i++) {
-                network.addCellTower(CellTower.from(
-                        mcc, mnc, buf.readUnsignedShort(), buf.readUnsignedShort(), buf.readUnsignedByte()));
+                if (type == MSG_LBS_WIFI_2) {
+                    int tac = buf.readInt();
+                    long eci = buf.readUnsignedInt();
+                    int signal = buf.readUnsignedByte();
+                    network.addCellTower(CellTower.from(mcc, mnc, tac, eci, signal));
+                } else {
+                    network.addCellTower(CellTower.from(
+                            mcc, mnc, buf.readUnsignedShort(), buf.readUnsignedShort(), buf.readUnsignedByte()));
+                }
             }
 
             if (buf.readableBytes() > 2) {
--- a/src/test/java/org/traccar/protocol/TopinProtocolDecoderTest.java
+++ b/src/test/java/org/traccar/protocol/TopinProtocolDecoderTest.java
@@ -2,6 +2,8 @@ package org.traccar.protocol;
 
 import org.junit.jupiter.api.Test;
 import org.traccar.ProtocolTest;
+import org.traccar.model.CellTower;
+import org.traccar.model.Network;
 import org.traccar.model.Position;
 
 public class TopinProtocolDecoderTest extends ProtocolTest {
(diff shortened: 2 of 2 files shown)


BSyndrom:
For additional context: the same device also uses message type 0x1B as an offline variant of 0x1A, with the same 32-bit TAC/ECI cell format. I have a separate tested change for 0x1B ready and can submit it as a follow-up PR.

tananaev:
Please provide the protocol documentation and logs from real device.

BSyndrom:
Thanks. I found an older Topin protocol document hosted on the Traccar website:

https://www.traccar.org/protocol/5023-gt06/ZhongXun%20Topin%20Locator%20Communication%20Protocol.pdf

It does not document the newer 0x1A message used by this ZX909, but it documents the same Topin cell record structure with 2-byte MCC, 1-byte MNC, 4-byte LAC, 4-byte cell ID and 1-byte signal value.

For the newer 0x1A message I verified the layout against a real Topin ZX909 running firmware:

ZX909_EU_V1.3.7_2025-11-17_09-43-24

Example packet received by Traccar:

7878001a260904114653030106010000119f01e91105210000119f01e91105210000119f01e9110521000d0a

The cell section decodes as:

03          cell count
0106        MCC = 262
01          MNC = 1

0000119f    TAC = 0x119f = 4511
01e91105    ECI = 0x01e91105 = 32051461
21          signal = 33

The other two cell records contain the same values.

I also checked the LTE modem in the same physical device independently via its AT interface. It reported:

+CEREG: ... "119f","01e91105",7
+COPS: ... "26201",7

So the 32-bit values in the 0x1A packet match the TAC (0x119f) and ECI (0x01e91105) reported independently by the modem.

I've attached a sanitized extract from the production Traccar log containing multiple real 0x1A messages from two physical ZX909 devices. I only removed IP addresses, device IDs and GPS coordinates; the raw packet data itself is unchanged.

I also have the complete production logs from longer test sessions, including many additional 0x1A messages, normal GPS/status traffic and complete device sessions. I can provide those as well if useful.

For transparency: I used ChatGPT (GPT-5.6) to assist with analyzing the captured packets, reviewing the Traccar source code and preparing the patch. All packet samples and modem responses above are from the physical devices and were verified against the actual logs.
[zx909-0x1a-real-device.log](https://github.com/user-attachments/files/31838730/zx909-0x1a-real-device.log)

tananaev:
We would need the real documentation. How did you determine the format without a document?

BSyndrom:
I don't have an official protocol document for this newer ZX909 0x1A message. I determined the format empirically from the real device data and cross-checked it against the LTE modem in the same device.

The key observation was this packet:

7878001a260904114653030106010000119f01e91105210000119f01e91105210000119f01e9110521000d0a

After the timestamp and cell count/MCC/MNC:

03       cell count
0106     MCC 262
01       MNC 1

the remaining cell data repeats in 9-byte records:

0000119f 01e91105 21
0000119f 01e91105 21
0000119f 01e91105 21

I then queried the cellular modem in the same physical tracker through its AT interface. Independently of Traccar, it reported:

+CEREG: ... "119f","01e91105",7
+COPS: ... "26201",7

So the two 32-bit values from the packet exactly match the LTE TAC and ECI reported by the modem, while MCC/MNC also match the registered network.

I repeated this with additional real packets. The cell values change when the device changes cells; for example another captured 0x1A packet contains:

0000119f 023c0410 23

This is why I concluded that the 0x1A cell record is:

4 bytes TAC
4 bytes ECI
1 byte signal

So to be clear: this part was reverse-engineered from the device traffic and independently verified against the modem, not taken from an official 0x1A protocol document.

If an official document for this ZX909 protocol variant is required for the change, I currently don't have one.

For transparency: I used ChatGPT (GPT-5.6) to assist with analyzing the packet captures and Traccar decoder. The raw packets and AT modem responses used for the analysis are from the physical ZX909 devices.

tananaev:
Unfortunately we can't accept AI guessed protocols. You have to get proper documentation from the vendor.

BSyndrom:
Understood. Thanks for clarifying.

I'll try to obtain the protocol documentation for this firmware/protocol variant from the vendor. I'll leave the PR as-is for now and update it if I can get the documentation.

PR #6013, opened 2026-09-04
Add frame decoding for Topin protocol

BSyndrom:
The Topin protocol can send multiple CRLF-terminated messages in a single TCP read.

Without a frame decoder, `TopinProtocolDecoder` processes only the first message from the received buffer. Any additional complete messages in the same buffer are discarded.

This change adds a `CharacterDelimiterFrameDecoder` using the protocol's CRLF delimiter while preserving the delimiter expected by the existing decoder.

Added a regression test with three concatenated Topin messages captured from a real device.

### DIFF
--- a/src/main/java/org/traccar/protocol/TopinProtocol.java
+++ b/src/main/java/org/traccar/protocol/TopinProtocol.java
@@ -16,6 +16,7 @@
 package org.traccar.protocol;
 
 import org.traccar.BaseProtocol;
+import org.traccar.CharacterDelimiterFrameDecoder;
 import org.traccar.PipelineBuilder;
 import org.traccar.TrackerServer;
 import org.traccar.config.Config;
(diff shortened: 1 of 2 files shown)


src/main/java/org/traccar/protocol/TopinProtocol.java
```diff
@@ -35,6 +36,7 @@ public TopinProtocol(Config config) {
         addServer(new TrackerServer(config, getName(), false) {
             @Override
             protected void addProtocolHandlers(PipelineBuilder pipeline, Config config) {
+                pipeline.addLast(new CharacterDelimiterFrameDecoder(MAX_FRAME_LENGTH, false, "\r\n"));
```
tananaev, line 39: This is almost certainly wrong because this is binary protocol.

BSyndrom:
You're right. I used CRLF because all captured device messages are terminated by `0D0A`, but using it directly as a delimiter is not safe for a binary protocol.

The underlying issue is still reproducible: the ZX909 sends multiple complete Topin messages in a single TCP read, and without framing only the first one is decoded.

I also noticed that a generic length-field decoder won't work here because some Topin message types use the third byte for a different purpose. For example, the Topin documentation explicitly defines it as the WiFi count for `0x17` and `0x69`, and the ZX909 uses the same special framing for `0x1A`.

I'll replace this with protocol-aware binary framing and update the tests.

For transparency: I used ChatGPT (GPT-5.6) to assist with analyzing the captured packets and investigating the framing issue. The packet captures and observed device behavior are from the physical ZX909 devices.

PR #6014, opened 2026-09-05
feat: handshake decoding and test for tk103 protocol

NicholasRasi:
### Description

I have a TK103 device (which may be a clone) that sends the battery level as a hexadecimal value after `HSOP`, for example:

`072105039432BP00352672105039432HSOP4B`

Traccar currently does not decode this value.

This PR detects this message pattern and decodes the hexadecimal value after `HSOP` to determine the battery level.

### Testing

Tested with my device and verified that the battery level is correctly decoded.

### DIFF
--- a/src/main/java/org/traccar/protocol/Tk103ProtocolDecoder.java
+++ b/src/main/java/org/traccar/protocol/Tk103ProtocolDecoder.java
@@ -156,6 +156,16 @@ public class Tk103ProtocolDecoder extends BaseProtocolDecoder {
             .text(")")
             .compile();
 
+    private static final Pattern PATTERN_HANDSHAKE = new PatternBuilder()
+            .text("(")
+            .expression("(.{12})")               // device id
+            .text("BP00")
+            .number("d*")                        // imei
+            .text("HSOP")
+            .number("(xx)")                      // battery level
+            .text(")")
+            .compile();
+
     private String decodeAlarm(int value) {
         return switch (value) {
             case 1 -> Position.ALARM_ACCIDENT;
@@ -380,6 +390,27 @@ public class Tk103ProtocolDecoder extends BaseProtocolDecoder {
         return position;
     }
 
+    private Position decodeHandshake(Channel channel, SocketAddress remoteAddress, String sentence) {
+        Parser parser = new Parser(PATTERN_HANDSHAKE, sentence);
+        if (!parser.matches()) {
+            return null;
+        }
+
+        DeviceSession deviceSession = getDeviceSession(channel, remoteAddress, parser.next());
+        if (deviceSession == null) {
+            return null;
+        }
+
+        Position position = new Position(getProtocolName());
+        position.setDeviceId(deviceSession.getDeviceId());
+
+        getLastLocation(position, null);
+
+        position.set(Position.KEY_BATTERY_LEVEL, parser.nextHexInt());
+
+        return position;
+    }
+
     private Position decodeBms(Channel channel, SocketAddress remoteAddress, String sentence) {
         String id = sentence.substring(1, 13);
         DeviceSession deviceSession = getDeviceSession(channel, remoteAddress, id);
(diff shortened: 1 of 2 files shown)


tananaev:

NicholasRasi:
> Please provide the documentation.

The documentation for what?

tananaev:
The protocol documentation, so we can validate this is correct.

NicholasRasi:
Yes, I understand your requirements.

I found this documentation, which states in Chapter 3.2.1 (“Handshake signal Message + battery voltage”) that `BP00` is the command used for the handshake and battery voltage.

[gps-tracker-communication-protocol.pdf](https://github.com/user-attachments/files/31965948/gps-tracker-communication-protocol.pdf)

But, in the doc [https://www.traccar.org/protocol/5002-tk103/KX402_GPRS_V15_EN.pdf](https://www.traccar.org/protocol/5002-tk103/KX402_GPRS_V15_EN.pdf) the battery is not stated.

However, I’m not sure how reliable or applicable this documentation is.

It’s also possible that my device uses a non-strict or slightly modified version of the TK103 protocol, but I don’t know for certain.

The key point, in my opinion, is: do you think this PR could be useful to other users with the same or a similar device?

It could also be interesting to have a way to support community-provided protocol definitions that could be plugged into Traccar. But that’s probably a separate discussion/digression.

tananaev:
You should get documentation for your specific device from your vendor. There are million variations of the protocol.

NicholasRasi:
@tananaev, did you read what I wrote?

> I have a TK103 device (which may be a clone).

My device is most likely a clone. Given that, the only documentation I have is the basic instructions I received for configuring and using the device.

I opened this PR because the patch works with my device, and it may be useful to others using the same or a similar device. I cannot provide official protocol documentation because none is available to me.

If having official protocol documentation is a strict requirement for approving this PR, then there is simply no way for me to provide it. In that case, this PR cannot move forward, thanks.

tananaev:
I read what you said. I don't really understand how the fact that it's a clone is related to the documentation. Even no name devices from AliExpress usually have documentation. You just need to ask for it.

NicholasRasi:
Because, it's a clone device with no known brand, I thought it would be much more difficult and complicated to obtain the protocol documentation compared to a branded, well-known device.

However, after few requests, they sent me the documentation:

[HY200 tk202 TK600 TK200 Communication Protocol V2.pdf](https://github.com/user-attachments/files/32026998/HY200.tk202.TK600.TK200.Communication.Protocol.V2.pdf)

As you can see, the battery percentage is included in the handshake signal message (3.4).

src/main/java/org/traccar/protocol/Tk103ProtocolDecoder.java
```diff
@@ -156,6 +156,16 @@ protected void init() {
             .text(")")
             .compile();
 
+    private static final Pattern PATTERN_HANDSHAKE = new PatternBuilder()
+            .text("(")
+            .expression("(.{12})")               // device id
+            .text("BP00")
+            .number("d*")                        // imei
```
tananaev: Where is this coming from? I don't see it in the doc. Also your sample doesn't match the doc.

NicholasRasi: The documentation does not match the actual data.

The sample was taken directly from the log.

tananaev: If it is IMEI, I think the expression should be something like `d{15}?`.

NicholasRasi: Yes it is. But, I think it depends, if we also want to accept the example from the documentation, maybe `(?:d{15})?` would be a better choice.

tananaev: That's exactly what I meant.

NicholasRasi: https://github.com/traccar/traccar/pull/6014/commits/7d6cedffb59abd771325a2f255875e4b6ed1d3cd

NicholasRasi:
@tananaev, thank you for your support and for all your effort in building Traccar. It's appreciated

tananaev:
Thank you for the contribution!

PR #6016, opened 2026-09-05
Support Teltonika FT platform parameters

Kresnik003:
The FT platform (FTC880, FTM880, FTC921, ...) shares most AVL IDs with FMB, but the decoder gated them behind the `fmbXXX` predicate, so FT devices lost external voltage, battery level and several alarms, keeping them only as raw `ioNNN` attributes.

Per Teltonika's [AVL ID differences between FMB and FT platforms](https://wiki.teltonika-gps.com/view/AVL_ID_differences_between_FMB_and_FT_platforms), only IDs 70, 155, 156, 157, 358, 359, 360, 641 and 800 change meaning between the two platforms; every other shared ID keeps its FMB semantics. This change adds an `ftXXX` predicate (`FT[A-Z]\d{3}`) and only extends IDs that are **not** in that list, so their meaning is identical on FT:

- `68` battery current, `113` battery level, `200` sleep mode
- `246` towing, `249` jamming, `251` excessive idling, `252` unplug (power cut)
- `175` auto geofence

Each of those was checked against the [FTC880 parameter list](https://wiki.teltonika-gps.com/view/FTC880_Teltonika_Data_Sending_Parameters_ID) for id, size and unit, so nothing shifts inside the record: 68 is 2 bytes mA, 113 is 1 byte %, and the rest are 1-byte flags with the same polarity as on FMB.

External voltage is the one changed ID handled here: FT devices report a wider range than the two-byte AVL 66 allows, so Teltonika moved it to **AVL 800**, four bytes in mV.

It is registered for `ftXXX` only, not for every model. On [TFT100 and TST100](https://wiki.teltonika-gps.com/view/TFT100_AVL_ID_List) the same id is *Extended External Voltage*, four bytes already **in volts**, used above 65 V — dividing that by 1000 would be wrong. Those models are matched by the existing `tatXXX` predicate and are untouched here.

Accident is deliberately left untouched: FMB AVL 247 does not exist on FT, where a crash is reported through AVL 1429–1432 instead.

Verified with two assertions at the end of `testDecode`, on a real Codec 8 Extended record from an FTC880 running 6.15.3 (external voltage 800 and battery level 113). The network identifiers, odometer and timestamp in that frame were zeroed and the CRC recomputed, because they identify the vehicle; the two asserted values are exactly as the device sent them.


src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -190,6 +190,7 @@ private static void register(int id, Predicate<String> predicate, BiConsumer<Pos
         Predicate<String> fmbXXX = (m) -> m != null && m.matches("FM[B-Z]...|MTB100|MSP500");
         Predicate<String> fmb6XX = (m) -> m != null && m.matches("FM.6..");
         Predicate<String> tatXXX = (m) -> m != null && m.matches("T.T...");
+        Predicate<String> ftXXX = (m) -> m != null && m.matches("FT[A-Z]\\d{3}|TFT\\d{3}");
```
tananaev: I think we can simplify expression to avoid `\\d{3}` duplication.

src/test/java/org/traccar/protocol/TeltonikaProtocolDecoderTest.java
```diff
@@ -7,6 +7,49 @@
 
 public class TeltonikaProtocolDecoderTest extends ProtocolTest {
 
+    @Test
+    public void testDecodeFt() throws Exception {
```
tananaev: 1. Keep tests in the existing function.
2. Adding so many tests is completely unnecessary. I think one or two tests is more than enough.
3. Keep styling and formatting consistent with existing code.
4. Make sure you real data from device.

Kresnik003: All four addressed. The separate test function is gone and the assertions now live at the end of `testDecode`, next to the existing `setModelOverride` block and formatted like the surrounding calls. Cut down to two assertions on a single frame.

On real data: that frame is an actual record from my FTC880 on 6.15.3, not a synthetic one. I only zeroed the network identifiers (205, 206, 241), the odometer and the timestamp, and recomputed the CRC, since they identify the vehicle. Nothing else was touched, and the two values asserted (800 external voltage and 113 battery level) are exactly as the device sent them.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
Kresnik003: Done: `T?FT[A-Z]?\d{3}`. Covers both shapes with a single \d{3} (FTC880, FTM880 and TFT100).

src/test/java/org/traccar/protocol/TeltonikaProtocolDecoderTest.java
```diff
@@ -153,6 +153,16 @@ public void testDecode() throws Exception {
                 "00000000000000628e0100000193011f1ec8003fa9def2fc401ea4000d00000400000000000f000700ef0000f00000150500c8000045012a34492a3800000600b5001300b60011004230aa00430fb7004400002a30fb5d000200f10000c74200100000014900000000010000fbff"),
                 "eyeTemp1", -11.87);
 
+        decoder.setModelOverride("FTC880");
```
tananaev, line 165: Put new tests at the bottom, so it doesn't affect any other cases.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -190,6 +190,7 @@ private static void register(int id, Predicate<String> predicate, BiConsumer<Pos
         Predicate<String> fmbXXX = (m) -> m != null && m.matches("FM[B-Z]...|MTB100|MSP500");
         Predicate<String> fmb6XX = (m) -> m != null && m.matches("FM.6..");
         Predicate<String> tatXXX = (m) -> m != null && m.matches("T.T...");
+        Predicate<String> ftXXX = (m) -> m != null && m.matches("T?FT[A-Z]?\\d{3}");
```
tananaev: I think it should be something like this

```suggestion
        Predicate<String> ftXXX = (m) -> m != null && m.matches("(TFT|FT[A-Z])\\d{3}");
```

Kresnik003: Applied verbatim.

src/test/java/org/traccar/protocol/TeltonikaProtocolDecoderTest.java
Kresnik003: Moved to the end of `testDecode`, after the last `verifyPositions`.

One detail worth flagging: the block needs `decoder.setExtended(false)` in front of it. The frame is a real Codec 8 Extended record from the device, and with `extended` still true from the previous case the decoder expects an extra 16-byte group that the record does not carry, so it throws. `test` and `checkstyleMain` pass.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -251,30 +252,32 @@ private static void register(int id, Predicate<String> predicate, BiConsumer<Pos
...
                 case 3 -> p.addAlarm(Position.ALARM_CORNERING);
             }
         });
-        register(175, fmbXXX, (p, b) -> {
+        register(175, fmbXXX.or(ftXXX), (p, b) -> {
             p.addAlarm(b.readUnsignedByte() > 0 ? Position.ALARM_GEOFENCE_ENTER : Position.ALARM_GEOFENCE_EXIT);
         });
         register(636, fmbXXX.or(tatXXX), (p, b) -> p.set("cid4g", b.readUnsignedInt()));
         register(662, fmbXXX, (p, b) -> p.set(Position.KEY_DOOR, b.readUnsignedByte() > 0));
+        register(800, any, (p, b) -> p.set(Position.KEY_POWER, b.readUnsignedInt() / 1000.0));
```
tananaev: 1. Why is this `any`? Have you verified that it applies to all models, including legacy ones?
2. For TFT100 it looks like the value is V and doesn't need to be divided by 1000.

tananaev:
Why are you force-pushing changes as a single commit? That defeats the whole purpose of having version control.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
Kresnik003: Both points are fair, and the second one is my fault.

**On TFT100.** I brought it into this myself in an earlier reply, and I was wrong. I have no TFT100 and never checked it; I assumed the name shape meant the same platform. It does not. There the ID is a different element with a different unit:

| Model | AVL 800 | Size | Multiplier | Unit |
| --- | --- | --- | --- | --- |
| FTC880 | External Voltage | 4 | - | mV |
| TFT100 / TST100 | Extended External Voltage | 4 | - | V |

- FTC880: https://wiki.teltonika-gps.com/view/FTC880_Teltonika_Data_Sending_Parameters_ID
- TFT100: https://wiki.teltonika-gps.com/view/TFT100_AVL_ID_List

On TFT100 it is the high-range element, used when the external voltage is above 65 V, and it is already in volts, so dividing by 1000 there would be wrong. TFT100 is also matched by the existing `tatXXX` predicate already, so it was never mine to claim. Removed: the predicate is now `FT[A-Z]\d{3}`, which matches FTC880, FTM880 and FTC921 only.

**On `any`.** No, I had not verified it, and TFT100 is precisely why it was wrong: `any` would apply the mV conversion to a model that reports volts. I had only checked that 800 is absent from the FMB list and treated that as good enough, which it is not. Changed to `ftXXX`.

Sorry for the noise. Both branches are updated, `test` and `checkstyleMain` pass.

Kresnik003:
Two notes so you do not have to dig through the thread.

**Your suggested regex is not what is in the branch now.** You suggested `(TFT|FT[A-Z])\d{3}` and I replied "applied verbatim", but that suggestion only included `TFT` because I had claimed TFT100 was in scope, and that claim was wrong. The branch now has `FT[A-Z]\d{3}`. I did not want you to find your own suggestion quietly reverted.

**Everything the diff touches, checked against the official FT table** (https://wiki.teltonika-gps.com/view/FTC880_Teltonika_Data_Sending_Parameters_ID), since "have you verified" was a fair question to ask about more than one line:

| AVL | Name on FT | Size | Unit | Handler reads | Value meaning |
| --- | --- | --- | --- | --- | --- |
| 68 | Battery Current | 2 | mA | 2 bytes, /1000 | — |
| 113 | Battery level | 1 | % | 1 byte | — |
| 175 | Auto geofence | 1 | - | 1 byte | 0 left zone, 1 entered zone |
| 200 | Sleep Mode | 1 | - | 1 byte | stored raw |
| 246 | Towing detection | 1 | - | 1 byte | 0 steady, 1 towing |
| 249 | Network jamming detection | 1 | - | 1 byte | 0/1 |
| 251 | Excessive idling | 1 | - | 1 byte | 0 moving, 1 idling |
| 252 | Unplug detection | 1 | - | 1 byte | 0 present, 1 unplugged |
| 800 | External Voltage | 4 | mV | 4 bytes, /1000 | — |

Sizes and units match what the existing handler already reads for FMB, so nothing shifts inside the record.

Scope of the change, to be explicit about regression risk: 800 is a new registration, restricted to FT. Every other line only adds FT to a predicate that was `fmbXXX`, so no FMB, TAT/TST/TFT or legacy model changes behaviour.

Kresnik003:
Fair, and I will stop.

I was amending to keep the branch at one commit per logical change, but that is my habit getting in the way of your review: it throws away the diff between rounds, which is exactly what you need to see what I changed since you last looked. Same for #6017.

From here on each round of feedback goes in as a new commit on top, and you squash on merge if you want it as one.

The current tip already contains everything from this round, so I have nothing pending to push. The intermediate commits still exist on my side (`4c9fbc3` as opened, `f7cb631` after your first round, `85139eb` now). If you would rather have them in the branch so the history is there, say so and I will restore them, but that is one more rewrite, so I am not doing it unless you want it.

Kresnik003:
Sorry. You were right and I should not have made you ask.

I have restored the history instead of leaving it as a promise. #6016 is now four commits, one per state you actually reviewed, with the tree unchanged from what you last saw:

- `a788d89` the PR as you first reviewed it
- `85bd145` simplify the predicate, cut the tests down to one real frame
- `835e4ab` move the FT assertions to the end of `testDecode`
- `22cc62a` restrict AVL 800 to FT models

#6017 is rebased onto that and left as its single commit: its one change since opening (signed axis components) landed before you looked at it, so there is no review round to preserve there and I did not want to invent one.

That is the last force-push. From now on every round goes on top as a new commit.

tananaev:
The code looks good, but I think you forgot to update description.

Kresnik003:
Updated, and it needed more than a touch-up: it still described the version you rejected.

#6016 — corrected: TFT100 removed from the model list, the `any` reasoning replaced with why 800 is FT-only (Extended External Voltage in volts on TFT100/TST100), the predicate spelled out, the per-id size and unit check added, and the reference to a `testDecodeFt` function that no longer exists dropped.

#6017 — corrected the stacking sentence, and one thing I should have written there from the start: **the crash test frame is constructed, not captured.** The device has Crash Detection enabled and I verified the settings against it, but it has never had a crash, so there is no recorded crash record. The frame is built to the documented layout with a valid CRC. It was wrong to leave that implicit in a description that reads as if the data came off the device, and I have said so explicitly there now. If you would rather hold that PR until I can supply a real crash record, say so and I will.

PR #6017, opened 2026-09-06
Decode Teltonika FT crash detection data

Kresnik003:
### What is missing today

On the FT platform the Crash Detection scenario reports through dedicated AVL IDs that the decoder does not read, so the crash data arrives as opaque `ioNNN` attributes and no accident alarm is raised.

### Documentation

- [Crash trace](https://wiki.teltonika-gps.com/view/Crash_trace) — describes the Crash Detection scenario and how the crash data is reported.
- - [FTC880 Data Sending Parameters](https://wiki.teltonika-gps.com/view/FTC880_Teltonika_Data_Sending_Parameters_ID):
-   - **1429 Crash average vector**, 8 bytes, mG
-   - **1432 Crash max vector**, 8 bytes, mG
The wiki gives the layout of both as *"2B - X axis, 2B - Y axis, 2B - Z axis, 2B - magnitude in mG"*. That order does not match what the device sends — see Verification.

### What this does

```
crashAvg / crashAvgMagnitude   (AVL 1429)
crashMax / crashMaxMagnitude   (AVL 1432)
alarm = accident
```

The axes go into `crashAvg` and `crashMax` as `[x,y,z]`, the shape `KEY_G_SENSOR` already uses in the T800x, Its, Jt808, Castel and Suntech decoders. Both elements arrive in the variable length group, so they are handled there. The alarm is raised on the average vector only, not on both, to avoid a duplicated alarm value when a frame carries the full crash package.

The axis components are read with `readShort()` and the magnitude stays unsigned: an impact opposite to a sensor axis reports a negative acceleration, and the standalone axis IDs 17, 18 and 19 are signed over -8000..8000 mg.

`1430` (crash direction) and `1431` (crash duration) are left out: the wiki does not document their byte width.

### Verification

The test frame is a capture from an FTC880 during a crash test on 6 September 2026. It holds two records a second apart, one per impact. The device had no GNSS fix, so the frame carries no position. The serving cell id, area code and operator code (AVL 205, 206 and 241) are zeroed and the packet checksum recomputed, so it is still a valid packet with no location in it; nothing else is altered and none of those fields are read by this code path.

Two things came out of testing against real data rather than a constructed frame:

- **The elements arrive in the variable length group**, not the eight byte group. That group does not go through the parameter table, so handlers registered there never ran on a real device.
- - **The magnitude is the first field, not the last.** Read magnitude first and the declared value equals the modulus of the other three components on every captured average vector: 1960 against 1961, 3434 against 3435, 3263 against 3264, 3397 against 3398. Read as documented, that same field is 62429 to 65130 across those four records, which no crash magnitude in mG can be.
The captured records also carry element 247 set to 1, and the record event id is 247. The decoder maps element 247 to the accident alarm for FMB models only, and 247 is not listed in the FTC880 AVL table, so the alarm is raised from 1429, which is listed. Widening the existing 247 registration to the FT platform would also work if you prefer that.

`./gradlew checkstyleMain checkstyleTest test --tests '*TeltonikaProtocolDecoderTest'` passes on Temurin JDK 25.


tananaev:
Did you mean to make stacked PRs?

Kresnik003:
Yes, intentionally stacked: this one needs the `ftXXX` predicate introduced in #6016, so it is branched off it and only the second commit is new here. Happy to squash both into a single PR if you would rather review it that way, or to rebase this one once #6016 is in.

I also applied the signed-axis fix from the review comment below.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -251,30 +252,45 @@ private static void register(int id, Predicate<String> predicate, BiConsumer<Pos
...
+        register(175, fmbXXX.or(ftXXX), (p, b) -> {
             p.addAlarm(b.readUnsignedByte() > 0 ? Position.ALARM_GEOFENCE_ENTER : Position.ALARM_GEOFENCE_EXIT);
         });
         register(636, fmbXXX.or(tatXXX), (p, b) -> p.set("cid4g", b.readUnsignedInt()));
         register(662, fmbXXX, (p, b) -> p.set(Position.KEY_DOOR, b.readUnsignedByte() > 0));
+        register(800, any, (p, b) -> p.set(Position.KEY_POWER, b.readUnsignedInt() / 1000.0));
+        register(1429, ftXXX, (p, b) -> {
+            p.set("crashAverageX", b.readUnsignedShort());
+            p.set("crashAverageY", b.readUnsignedShort());
+            p.set("crashAverageZ", b.readUnsignedShort());
```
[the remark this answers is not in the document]
Kresnik003: Agreed, fixed. The three axis components now use `readShort()` and the magnitude stays unsigned, and there is a test with negative axes (X=-1200, Z=-2600 on 1429; X=-1500, Z=-3200 on 1432).

Worth noting the wiki lists the type of the eight-byte container rather than of each component, and the standalone axis IDs 17/18/19 are signed with a -8000..8000 mg range. Reading them signed is also the safe option under either reading: an unsigned value is only misread above 32767 mG, far beyond any crash threshold, while a signed value read as unsigned turns -1200 into 64336.

tananaev:
The problem is that you didn't actually stack it properly.

Kresnik003:
You are right, and I should have checked how it renders here rather than only in git.

The branches are stacked locally (the crash commit sits on top of the external-voltage one), but this PR targets `traccar:master`, and GitHub only lets a pull request from a fork target a branch of the base repository. It cannot point at `Kresnik003:teltonika-ft-external-voltage`, so it keeps showing both commits no matter how the branches are arranged on my side.

Two ways out, your call:

1. Leave this one as it is and ignore the first commit while reviewing. Once #6016 is merged I rebase and it drops to a single commit on its own.
2. I fold the crash change into #6016 and close this one, so you review a single PR.

I would rather not guess, so tell me which you prefer and I will do it. Both branches are pushed with your review comments applied and `test` plus `checkstyleMain` green.

Kresnik003:
Rebased onto master now that #6016 is in. This is down to one commit and 21 added lines, so the stacking problem is gone.

That was the rebase I said I would do once the first one merged, not another history rewrite: the only commits dropped are the four that are now upstream in `e760853`. Since you squashed, git could not recognise them as already applied, so I rebased the range above the old base rather than letting it replay them.

Verified against the merged master: `test`, `checkstyleMain` and `checkstyleTest` pass on Temurin JDK 25.

On the test data. As written in the description, the frame in the test is constructed to the documented layout, not captured, because the device has never had a crash. Real crash tests are being done on the device today: the firmware gets updated and then the crash is triggered on purpose. As soon as there is a captured record, I will push a commit replacing the constructed frame with the real one, and post the raw frame here.

So if you would rather not merge this on constructed data, hold it and I will come back today with the real trace. If you are fine with it as it is, the replacement commit will follow anyway.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -278,6 +278,19 @@ private static void register(int id, Predicate<String> predicate, BiConsumer<Pos
...
+            p.set("crashAverageY", b.readShort());
+            p.set("crashAverageZ", b.readShort());
+            p.set("crashAverageMagnitude", b.readUnsignedShort());
+            p.addAlarm(Position.ALARM_ACCIDENT);
+        });
+        register(1432, ftXXX, (p, b) -> {
+            p.set("crashMaximumX", b.readShort());
+            p.set("crashMaximumY", b.readShort());
+            p.set("crashMaximumZ", b.readShort());
+            p.set("crashMaximumMagnitude", b.readUnsignedShort());
```
tananaev: I think these attributes are too verbose. I think we should store X,Y,Z in the same format we store accelerometer info in other places. Also I would probably shorten Maximum and Average words to Max and Avg.

tananaev:
For tests we definitely need ONLY real data. Otherwise tests are useless.

Kresnik003:
Replaced the constructed frame with a capture from a real crash test, as you asked. Testing against it turned up two problems with what I had.

**The handlers never ran on a real device.** The FTC880 sends 1429 and 1432 in the variable length group, and that group does not go through the parameter table, so `register(1429, ...)` was dead code. The previous test only passed because the frame I built put those elements in the eight byte group, which is not where the device puts them. They are now handled in the variable length group, next to the other elements that live there.

**The field order in the wiki is wrong.** It documents the eight bytes as `2B X, 2B Y, 2B Z, 2B magnitude`, but with that order the magnitude is negative in every captured record. The magnitude is the first field. In the average vector the declared value equals the modulus of the other three components: `07a801bcf8b6fe6a` declares 1960 mG against a modulus of 1961, and `0d6afeecf359fbab` declares 3434 against 3435. Read the documented way, those same fields are -406 and -1109.

Attribute names shortened as you suggested: `crashAvgAxisX/Y/Z` and `crashAvgMagnitude`, `crashMaxAxisX/Y/Z` and `crashMaxMagnitude`.

About the frame itself: one packet from an FTC880 during a crash test, two records a second apart, one per impact. The device had no GNSS fix, so it carries no position. I zeroed the serving cell id, area code and operator code (205, 206 and 241) and recomputed the packet checksum, so it is still a valid packet with no location in it. Nothing else is altered, and none of the zeroed fields are read by this code path.

One thing worth flagging from the capture: the records also carry element 247 set to 1, and the record event id is 247. The decoder maps element 247 to the accident alarm for FMB models only. 247 is not listed in the FTC880 AVL table, so I left the alarm on 1429, which is listed. If you would rather widen the existing 247 registration to the FT platform instead, the captured data supports that too.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
Kresnik003: Done. They are now `crashAvgAxisX`, `crashAvgAxisY`, `crashAvgAxisZ` and `crashAvgMagnitude`, with `crashMax` for 1432, so the components match the `axisX/Y/Z` naming the decoder already uses for the accelerometer.

tananaev: Why did you add axis? Have you checked how we store accelerometer values in other places? Please provide decoders you checked.

Kresnik003: I checked all 271 protocol decoders. Per-axis accelerometer values are stored in exactly one place in the tree, and it is this same file:

```
TeltonikaProtocolDecoder.java:205  register(17, any, (p, b) -> p.set("axisX", b.readShort()));
TeltonikaProtocolDecoder.java:206  register(18, any, (p, b) -> p.set("axisY", b.readShort()));
TeltonikaProtocolDecoder.java:207  register(19, any, (p, b) -> p.set("axisZ", b.readShort()));
```

`grep -rE '"[a-zA-Z]*(X|Y|Z)"\s*,\s*b\.read' src/main/java/org/traccar/protocol/` matches only this decoder. Everything else that touches acceleration stores a scalar: `maxAcceleration`, `lateralAcceleration`, `rapidAcceleration`, `hardAcceleration`, `acceleratorPosition`, `acceleratorMax`, `acceleratorAvg`. `Position` has no per-axis constant either, only `KEY_ACCELERATION = "acceleration"` and `KEY_G_SENSOR = "gSensor"`.

So `axisX/Y/Z` is the only precedent, and it needs to stay distinct from it, because a crash record carries both at the same time:

```
2B (11): 1431=0043  17=1cd0  18=f120  19=0c60  67=0ff6 ...
VAR id=1429 len=8: 07a801bcf8b6fe6a
VAR id=1432 len=8: 1d4d1ca9f125f75e
```

17, 18 and 19 are the instantaneous reading at record time; 1429 and 1432 are the average and maximum during the crash. Reusing `axisX` would overwrite the live value with the crash average in the same position.

If you prefer `crashAvgX` / `crashMaxX` without the word, say so and I will push it. The only constraint is that the crash names differ from `axisX/Y/Z`; whether they carry "Axis" is your call.

tananaev: The research is incorrect. Have you checked how we store `gSensor`?

Kresnik003: You were pointing at something bigger than the naming, and you were right. I went through every decoder that touches accelerometer data:

```
AstraProtocolDecoder:113        buf.skipBytes(6);    // accelerometer
NavisetProtocolDecoder:118      buf.skipBytes(6);    // accelerometer
SigfoxProtocolDecoder:228       buf.skipBytes(6);    // accelerometer
Jt808ProtocolDecoder:1321       buf.skipBytes(6);    // accelerometer
DmtProtocolDecoder:118-120      buf.readShortLE();   // accelerometer x, y, z  (dropped)
ThinkPowerProtocolDecoder:75    buf.skipBytes(2);    // g-sensor x/y/z
AquilaProtocolDecoder:93-95     pattern, not stored  // accelerometer x, y, z
JidoProtocolDecoder:59-61       pattern, not stored  // accelerometer x, y, z
```

The convention is not a naming one. Six decoders skip exactly six bytes with an `accelerometer` comment, and the rest read the axes and throw them away. The only per-axis attributes in the whole tree are `axisX/Y/Z` in this file, and `tiltX/tiltY` in `ItsProtocolDecoder`, which is an inclinometer.

So I dropped the axes. Pushed as a separate commit: the two magnitudes stay, the six bytes are skipped, and the alarm is still raised from 1429. That is what the alarm and the severity are built on, and it takes the change from eight attributes to two.

```java
} else if ((id == 1429 || id == 1432) && length == 8) {
    String prefix = id == 1429 ? "crashAvg" : "crashMax";
    position.set(prefix + "Magnitude", buf.readUnsignedShort());
    buf.skipBytes(6); // axes
    if (id == 1429) {
        position.addAlarm(Position.ALARM_ACCIDENT);
    }
}
```

`test`, `checkstyleMain` and `checkstyleTest` pass on Temurin JDK 25 against the captured frame.

Worth noting for anyone who needs the direction of an impact: it does not disappear with this. The device also sends 1430 (crash direction) and, when crash trace is enabled, 257 with the full per-axis recording around the event, and both still arrive as raw `ioNNN`.

tananaev: I think your AI got wires crossed.

Kresnik003: You are right and my previous comment was wrong. I grepped for attribute names ending in X, Y or Z, found nothing outside this file, and concluded the axes are discarded. They are not. They go in `Position.KEY_G_SENSOR` and I missed it. Sorry for the noise.

All seven uses in the tree:

| decoder | shape |
|---|---|
| `T800xProtocolDecoder:469` | `[x,y,z]` |
| `ItsProtocolDecoder:254` | `[x,y,z]` |
| `Jt808ProtocolDecoder:1405` | `[x,y,z]` |
| `Jt808ProtocolDecoder:547` | `[[x,y,z],...]` |
| `CastelProtocolDecoder:442` | `[[x,y,z],...]` |
| `SuntechProtocolDecoder:926` | `[[x,y,z],...]`, 100 samples, buffer named `crash` |
| `ArmoliProtocolDecoder:127` | raw string from the pattern |

Pushed. The axes are back in that format:

```java
} else if ((id == 1429 || id == 1432) && length == 8) {
    String prefix = id == 1429 ? "crashAvg" : "crashMax";
    position.set(prefix + "Magnitude", buf.readUnsignedShort());
    position.set(prefix + "GSensor", String.format(
            "[%d,%d,%d]", buf.readShort(), buf.readShort(), buf.readShort()));
    if (id == 1429) {
        position.addAlarm(Position.ALARM_ACCIDENT);
    }
}
```

The magnitude stays a separate value because it is not derivable from the axes. In the maximum vector the device reports the peak of each axis independently, so the declared magnitude does not equal their modulus: `1d4d1ca9f125f75e` declares 7501 mG while the three components give 8554. In the average vector they do agree, which is what identified the field order in the first place.

Two things I would rather ask than assume:

1. **Names.** `KEY_G_SENSOR` is one key and there are two vectors in a crash record, so I prefixed rather than picking one. If you want them named differently, tell me and I will push it.

2. **Scope.** AVL 257 is the FT crash trace, and it is the same thing Suntech, Castel and Jt808 store as a nested `gSensor`. On my device a crash produces 5758 samples at about 384 Hz spread over 43 records, and it currently arrives as an opaque `io257` hex string. Do you want that decoded here too, or as a separate change?

Kresnik003: Fair. That was an AI-assisted search and it got it wrong: it grepped for attribute names ending in X, Y or Z, found none outside this file, and reported that the axes are discarded. `gSensor` holds the three values in a single attribute, so that pattern could not have matched it by construction.

It took your pointer to `gSensor` to redo the search properly. I would not have got there on my own, because I was looking for the wrong shape and the first search came back clean, which is the worst way to be wrong.

The mistake was mine for posting the conclusion before checking it, and then contradicting it four minutes later, which is worse than not having posted the first one.

The correction stands and is what is on the branch: `crashAvgGSensor` and `crashMaxGSensor` in `[x,y,z]`, magnitudes kept separate, `test` and `checkstyle` green against the captured frame.

I will stop pushing and wait for your call on the naming, and on whether 257 belongs in this PR or in a separate one.

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -598,6 +598,14 @@ private void decodeLocation(Position position, ByteBuf buf, int codec, String mo
                             }
                         }
                     }
+                } else if ((id == 1429 || id == 1432) && length == 8) {
+                    String prefix = id == 1429 ? "crashAvg" : "crashMax";
+                    position.set(prefix + "Magnitude", buf.readUnsignedShort());
+                    position.set(prefix + "GSensor", String.format(
```
tananaev:
```suggestion
                    position.set(prefix, String.format(
```

src/test/java/org/traccar/protocol/TeltonikaProtocolDecoderTest.java
```diff
@@ -172,6 +172,22 @@ public void testDecode() throws Exception {
                 "00000000000000968e010000019b76daa800000000000000000000000000000000000000001b000b00150500450000716400c80300ed0200ef0000f000012f00017f0333d20833d30a000a0011ffd30012ffe70013002300430fd400440000004600ee00b5000000b6000000ce000001820000000600100000000000c70000000000cd0000000000f100000000032000003249047c00000000000000000100003e0b"),
                 Position.KEY_BATTERY_LEVEL, 100);
 
+        verifyAttribute(decoder, binary(
```
tananaev, line 175: We don't need so many tests. One is enough.

Is this now real device data?

src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java
```diff
@@ -598,6 +598,14 @@ private void decodeLocation(Position position, ByteBuf buf, int codec, String mo
                             }
                         }
                     }
+                } else if ((id == 1429 || id == 1432) && length == 8) {
+                    String prefix = id == 1429 ? "crashAvg" : "crashMax";
+                    position.set(prefix + "Magnitude", buf.readUnsignedShort());
```
tananaev, line 603: btw what happened with the order? Why magnitude is first now?

Kresnik003:
Real device data corrected two things I had wrong. Neither was visible in the constructed frame.

**The handlers never ran.** The FT sends 1429/1432 in the variable length group, which does not consult the parameter table. The constructed frame passed only because I had put the elements in the eight byte group.

**The magnitude is the first field, not the last as documented.** On the average vector the declared value equals the modulus of the other three, on all four captured records:

| record | declared | modulus |
| --- | --- | --- |
| `07a801bcf8b6fe6a` | 1960 | 1961 |
| `0d6afeecf359fbab` | 3434 | 3435 |
| `0cbff9d3062ef6b7` | 3263 | 3264 |
| `0d4505590086f3dd` | 3397 | 3398 |

Read as documented, that same field is 62429 to 65130 unsigned across those four. I wrote "negative" in the description and in an earlier comment; the field is read unsigned, so that was the wrong way to put it. Same conclusion, description corrected.

Tests down to one. And yes, real device data: an FTC880 firing its own crash scenario under controlled bench impacts. The unit is in service on a motorcycle. Naming applied as suggested.

We run this on our own instance, where AVL 257 is decoded client side into the impact view for that same crash:

crash impact view

Still open: whether 257 belongs in this PR or a separate one.

tananaev:
What are we going to do about incorrect documentation? Have you checked with Teltonika?

Kresnik003:
Not yet. We are opening a conversation with Teltonika about the 1429/1432 layout to find out which one is correct, and I will post their answer here.

tananaev:
Let's wait.

Issue #5555, opened 2025-05-12
Speed issue in Topfly TLD2-D (Protocol 0x2626) - Incorrect speed parsing


miguelGerez:
Currently, the speed decoded for Topfly TLD2-D devices always appears as 0. After reviewing the manufacturer’s protocol documentation, it’s clear that the actual speed is located in a different position within the binary message.

Instead of being in the position expected by the current Traccar decoder, the speed is located specifically at bytes 65 and 66 of the message. This offset is not currently handled in the T880xProtocolDecoder, resulting in incorrect speed readings.

As a temporary workaround, a calculated attribute is being used with the following formula:

`(obdSpeed == 166.65) ? 0 : ((obdSpeed * 10) / 1.852)`

Example Message 
This is a real sample frame received from the device:

`262602005356f408696160608183510032012c140064001f51c000010100004004000000b187660025050814571700000000571b74c2b0fa0bc2000000bc130701500000000700000000035aff00855e0f0e20`

This helps estimate the speed using the obdSpeed attribute, but it is not reliable or accurate in all cases. The proper fix would be to update the decoder logic to read bytes 65–66 as the speed value when handling 0x2626 messages.

Port: 5094
Traccar Version: 6.6
Devices: TopflyTech TLD2-D
Message Header: 0x2626
Message Type: Position Message (0x260x260x02)

Related links and references
[Error report in TLD2-D on Traccar Forum](https://www.traccar.org/forums/topic/tld2-d-other-traccar-error/)

[Functionality issues with the Topflytech T880x protocol](https://www.traccar.org/forums/topic/functionality-issues-with-the-topflytech-t880x-protocol-on-various-devices-eg-tlw2-12bl/)

[TopflyTech protocol documentation (Google Sheet)](https://docs.google.com/spreadsheets/d/1QYuaMPlDL4rYXFmtWX-o6WALRN1KdtZm/edit#gid=1794005853)

tananaev:
Seems like there's a conflict with some other similar protocols. Speed value should be coming right after the coordinates.

miguelGerez:
This works for me, although I'm not sure if it's correct, and there are still some parameters to include.

```
        if (buf.readableBytes() >= 2) {
                position.set(Position.KEY_POWER, BcdUtil.readInteger(buf, 4) * 0.01);
            }

            if (buf.readableBytes() >= 19) {
                double obdSpeed = BcdUtil.readInteger(buf, 4) * 0.1;
[...]

PR #6024, opened 2026-09-11
Fix T800x vehicle bus data alignment

miguelGerez:
The vehicle bus block at the end of the TopFlyTech `0x2626` position message (TLD2-D, TLD1-DA/DE) is read two bytes late, and only its first single-byte value is ever read.

[3dea5bfd2](https://github.com/traccar/traccar/commit/3dea5bfd2) (#5555) started reading speed from bytes 65–66 but kept the `KEY_OBD_SPEED` read that was already consuming those two bytes, so everything after it shifted by two. Separately, since the block was added in [0caa8a3f3](https://github.com/traccar/traccar/commit/0caa8a3f3), the seven single-byte values are all evaluated from one `readUnsignedByte()`.

The attached [TOPFLYtech 0x26 0x26 Protocol 2023-08-02](https://github.com/user-attachments/files/32117883/TOPFLYtech.0x260x26.Protocol.2023-08-02.TorchX.100.TLD2-D.TLD1-DA.or.DE.xlsx), sheet `0x260x260x02`, lists the tail of the 83-byte message as:

| Bytes | Field |
|---|---|
| 65–66 | speed (BCD, 0.1 km/h) |
| 67–70 | accumulated fuel consumption (ml) |
| 71–74 | instant fuel consumption (ml/h) |
| 75–76 | RPM |
| 77–83 | air input, air pressure, coolant temperature, air inflow temperature, engine load, throttle position, remaining fuel |

It also says `0xFF` means the value can't be read from the vehicle, and that remaining fuel is in litres when its most significant bit is set and a percentage otherwise. There is no OBD speed field.

This change:
- drops the `KEY_OBD_SPEED` read and reads each single-byte value separately;
- skips unavailable values per field, including the fuel counters and RPM;
- stores the percentage case in `KEY_FUEL_LEVEL` — `KEY_FUEL` means litres since #5670/#5674, and the rename was applied mechanically here.

The fixtures are the two existing `0x2626` frames, now asserting values instead of only the position:

- The TLD2-D frame from #5555. On master: `rpm=65280`, `airTemp=93`, `engineLoad=133`, `throttle=133`, `fuel=133`. With this change: `rpm=858`, `airTemp=54`, `engineLoad=15`, `throttle=14`, `fuelLevel=32`.
- A frame where only the remaining fuel byte is available. On master: `fuelUsed=165871.615`, `fuelConsumption=4294967.295`, `rpm=65535`, no fuel level. With this change: `fuelUsed=2.53`, `fuelLevel=9`.

This also affects fuel events. `FuelEventHandler` raises fuel increase and drop events from changes in `fuel`, and on master that attribute receives the raw coolant byte — `fuel=133` in the first fixture, next to `coolantTemp=93`. With a threshold configured, a coolant temperature change of that size is reported as a fuel event. With this change `fuel` is only set when the device reports litres.

Also checked on three TLD2-D units on three vehicles. Availability differs per vehicle, even between adjacent bytes: a 2017 Ford reports air pressure but not air input, and a 2026 Toyota the opposite. The third vehicle reports nothing at all; on master every one of its positions carries `fuelUsed=65.535`, `fuelConsumption=4294967.295` and `rpm=65535`.

This is the first of a few small `0x2626` fixes; the others will follow one at a time.


src/main/java/org/traccar/protocol/T800xProtocolDecoder.java
```diff
@@ -498,32 +498,50 @@ private Position decodePosition(
...
-                position.set(Position.KEY_RPM, buf.readUnsignedShort());
+                long fuelUsed = buf.readUnsignedInt();
+                if (fuelUsed != 0xffffffffL) {
+                    position.set(Position.KEY_FUEL_USED, fuelUsed / 1000.0);
+                }
+                long fuelConsumption = buf.readUnsignedInt();
+                if (fuelConsumption != 0xffffffffL) {
+                    position.set(Position.KEY_FUEL_CONSUMPTION, fuelConsumption / 1000.0);
+                }
+                int rpm = buf.readUnsignedShort();
```
tananaev: use value here?

miguelGerez: Done. The fuel counters keep their own variables since they're unsigned 32-bit values.

PR #6025, opened 2026-09-11
Skip empty T800x 0x2626 cell towers

miguelGerez:
Follow-up to #6024.

When a TopFlyTech `0x2626` position has no GNSS fix, bytes 47–62 carry MCC, MNC and three LAC/CI pairs. If MCC and MNC are `0xFFFF` the decoder doesn't build the network, but it also doesn't consume the three pairs, so the rest of the message is read 12 bytes early.

In the [TOPFLYtech 0x26 0x26 Protocol 2023-08-02](https://github.com/user-attachments/files/32117883/TOPFLYtech.0x260x26.Protocol.2023-08-02.TorchX.100.TLD2-D.TLD1-DA.or.DE.xlsx) document (the same one attached to #6024), sheet `0x260x260x02`, the block is a fixed 16 bytes, filled with `0xFF` when there is no cell information.

A TLD2-D sends these frames when it has neither a fix nor cell data — 4 of the 429 position and alarm messages from one unit over a day. The fixture is one of them, an ignition-off alarm with the external supply at 12.12 V. On master it decodes as `power=166.65`, a speed of 1666.5 km/h, `rpm=4626` and `coolantTemp=-40`; with this change `power=12.12` and speed 0.

The skip is limited to `0x2626`, the only variant I could verify. The T880X (`0x2323`) specification has the same fixed block, but nothing follows it there, so it makes no difference. If `0x2525` and `0x2727` also always include the block, the condition can go — you'd know better than me.

### DIFF
--- a/src/test/java/org/traccar/protocol/T800xProtocolDecoderTest.java
+++ b/src/test/java/org/traccar/protocol/T800xProtocolDecoderTest.java
@@ -11,6 +11,10 @@ public class T800xProtocolDecoderTest extends ProtocolTest {
 
         var decoder = inject(new T800xProtocolDecoder(null));
 
+        verifyAttribute(decoder, binary(
+                "262604005330c40863257069370825001902581e0000003f80c0200505000000041600001b5b2800260908230233ffffffffffffffffffffffffffffffff1212000000000000ffffffffffffffffffffffffff"),
+                Position.KEY_POWER, 12.12);
+
         verifyPositions(decoder, false, binary(
                 "26260500400406086961606257315205250513182132ffffffffffffffffffffffff0000ffff0000250513182137ffffffffffffffffffffffff0770ffff0aec"));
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/traccar/protocol/T800xProtocolDecoderTest.java
```diff
@@ -11,6 +11,10 @@ public void testDecode() throws Exception {
 
         var decoder = inject(new T800xProtocolDecoder(null));
 
+        verifyAttribute(decoder, binary(
```
tananaev, line 14: Is this a real sample from a device?

miguelGerez: Yes, it's a verbatim capture, not a constructed sample. A TLD2-D logged it off the wire on 2026-09-10 at 11:38:34 — device time 2026-09-08 23:02:33, ignition-off alarm (`0x16`), external supply 12.12 V.

The long `0xFF` runs are genuine rather than blanked-out bytes: 47–62 is the empty cell block, and the tail is the vehicle bus block, which stays all `0xFF` on this unit because the car isn't ISO 15765-4/CAN.

It isn't a one-off either. That unit sent 4 such frames out of 429 position and alarm messages in a day, and since I opened this PR a second TLD2-D has produced 8 more on a different day.

Happy to attach the raw log if that's useful.

Issue #5835, opened 2026-03-25
Xexun2 Protocol (v03) - Device sends data but Server does not respond with ACK/Correct Reply


njfajardo:
**Device Info:**

- Model: Xexun X05 (Binary Protocol v03)
- Port configured: 5233 (xexun2)
- IMEI: 869412076187753

**Configuration (traccar.xml):**
XML
```
<entry key='xexun2.port'>5233</entry>
<entry key='xexun2.extended'>true</entry>
<entry key='logger.level'>all</entry>
```

The Problem: The device (and manual PowerShell simulations) successfully completes the TCP handshake and sends a valid binary packet (Version 03, Message Type 0x20). However, Traccar identifies the packet but does not send the "Correct Reply" (ACK) required by the Xexun API to maintain the connection. As a result, the device closes the connection immediately.

Logs & Network Capture: I have verified with tcpdump that the server receives the 17-byte login/data packet, but only responds with a TCP ACK (length 0) instead of the protocol-level ACK.

**Incoming Packet (HEX):** 

`fc 00 10 03 20 01 08 69 41 20 76 18 77 53 20 49 cf`

- Header: fc
- Length: 00 10
- Protocol: 03
- Type: 20 (Data upload)
- Serial: 01
- IMEI (BCD): 
- 
**Observed Behavior:**

- TCP Handshake: OK
- Device sends 0x20 Data: OK
- Server responds with TCP ACK (Length 0): OK
- Server sends Protocol ACK (fc...cf): MISSING
- Device sends FIN to close connection: FAIL

I have already tried registering the device with both the full 15-digit IMEI and the 12-digit suffix (412076187753) in the web interface, but the behavior remains the same.

Expected Behavior: The server should respond with the "Correct Reply" as per Xexun API: `FC [Length] 03 20 [Serial] 00 [Checksum] CF`

[Xexun API(2026-03-08 22_37_20).pdf](https://github.com/user-attachments/files/26250568/Xexun.API.2026-03-08.22_37_20.pdf)

tananaev:
Logs please showing the problem.

sandesh-java-21:
I’d like to add one more observation:

I’m able to successfully connect the device to the Traccar server, and the device is sending packets. However, location data is not being received. It seems either the device is not sending location packets, or Traccar is not parsing them correctly.

Device model: X05

tananaev:
Not sent and not received are two very different things. You should check logs to determine which one it is.

[...]

Issue #19, opened 2012-07-02
Installation instructions for MF2200 GPS Device


damcemomirce:
Dear Tananaev,

We have successfully install the tracking server on windows host. We open it on http://localhost:8082

The next is to add the MF2200 GPS device in application. I have see that there is windows.cfg file for configuration, is there any manual instructions for adding devices in the application and test it. We are working with MF2200 and bye it from China. We want to implement the software for tracking vehicles, but we cant find installation instructions for adding devices and testing it for sure. 

Please if there any instructions or manuals for adopting the software for the customers will be very helpful from you.

Best Regards.

We are using this software for tracking:  http://gurtam.com/en/gps_tracking/wialon_hosting.html

wialon_hosting

Issue #18, opened 2012-07-02
Generating Reports


damcemomirce:
Mr. Tananaev.

Can we generate reports for tracking devices. Our customers ask for generating daily reports for the trucks. Is there any possibility for this requirements.

best regards,

Damjan Momirovski 
Macedonia 
Skopje

Issue #17, opened 2012-06-22
Netty channel exceptions


tananaev:
Capture exceptions in channels and write something to log.

PR #6032, opened 2026-09-15
Fix Xexun3 GPS Body heading/course decoding (0x64)

nilei:
**Summary:**

Add heading field decoding to the Xexun3 GPS Body (subType 0x64). Per [Xexun API 260614 en.docx](https://github.com/user-attachments/files/32248351/Xexun.API.260614.en.docx) field 11, Heading is a U16 (big-endian) whose raw value must be divided by 10 to obtain degrees.

Previously the decoder stopped after the Speed field (field 10) and skipped the heading bytes via buf.readerIndex(subEnd), so course was always 0.0.

**Test changes:**
- Existing frame (fc0040...) changed from verifyPosition to verifyNotNull because its heading raw value (a4 00 = 41984 -> 4198.4 deg) is invalid and would fail the course <= 360 check in verifyPosition.
- New test entry with real device frame (fc00490320e8...) from IMEI 862596080926201, heading raw 00 9f = 159 -> 15.9 deg, verified via verifyAttribute course = 15.9.

**Protocol documentation references:**
- Field 10 Speed (U16): divide by 10 for km/h (already handled, but devision by 10 is missing, follow-up PR)
- Field 11 Heading (U16): divide by 10 for degrees (this PR)
- Field 12 Accuracy factor (U16): skipped by subEnd (follow-up PR)
- Field 13 Tracking duration (U8): skipped by subEnd (follow-up PR)

**Log evidence:**
```
2026-09-15 11:43:31  INFO: [T11d03210: xexun3 < 185.57.216.148] fc00490320e8086259608092620164226aa92f624049b01fff79c842401c97788f16414443084ccd400e19001c009f000e006a18630000173a2f0001a3ffffffffffffff6aa92f622101000030f4cf
2026-09-15 11:43:31  INFO: [T11d03210: xexun3 > 185.57.216.148] fc000c0320e8086259608092620100b4cbcf
2026-09-15 11:43:32  INFO: Event id: 862596080926201, time: 2026-09-15 11:43:28, type: alarm, notifications: 0
2026-09-15 11:43:32  INFO: [T11d03210] id: 862596080926201, time: 2026-09-15 11:43:09, lat: 51.37589, lon: 7.14771, course: 0.0
2026-09-15 11:43:34  INFO: [T11d03210] id: 862596080926201, time: 2026-09-15 11:43:30, lat: 51.37598, lon: 7.14792, speed: 15.1, course: 0.0
```

-- Cheers, Nils


src/test/java/org/traccar/protocol/Xexun3ProtocolDecoderTest.java
```diff
@@ -17,9 +17,13 @@ public void testDecode() throws Exception {
         verifyNull(decoder, binary(
                 "fc000b03200108610450803870158318cf"));
 
-        verifyPosition(decoder, binary(
+        verifyNotNull(decoder, binary(
```
tananaev: Why are we breaking this?

nilei: I mentioned it in the description:

Existing frame test had to be changed from verifyPosition to verifyNotNull because its course raw value was invalid and would fail the course <= 360 check in verifyPosition.

-- Cheers, Nils

tananaev: It looks like the old sample might be wrong. See doc from https://github.com/traccar/traccar/issues/5835.

nilei: What is your recommendation to fix this?

tananaev: We need to figure out of the original sample was wrong.

nilei: I don't know... The protocol docs are very different. According to the file name, mine is newer, but I cannot prove this on another way. I got the protocol documentation from Shenzhen Xexun Technology Co., Ltd. directly, as we are using their P02 devices (pet tracker). 

Here are the differences - I am happy to hear your recommendation how to handle this. 

---

## Protocol Differences: OLD (2026-03-08) vs NEW (2026-06-14)

### GPS Body (0x64) — Critical

| Field | OLD | NEW |
|---|---|---|
| #10 Speed | U16, raw = km/h (`0020` → 32 km/h) | U16, **divide by 10** (`0000` → 0 km/h) |
| #11 Heading | U16, raw = degrees (`0030` → 48°) | U16, **divide by 10** (`0000` → 0°) |
| #11 Name | "angle" | "Heading" |
| #12 Name | "Precision factor / decimeter" | "Accuracy factor (decimeters)" |
| Msg Length | 0x20 = 32 | 0x22 = 34 |

**Impact:** The NEW spec specifies `/10` for both speed and heading. The current decoder divides heading by 10 but not speed — inconsistent.

### Alarm Body (0x69) — Completely different alarm IDs

| ID | OLD | NEW |
|---|---|---|
| 2 | Low battery | Power-bank low-battery |
| 3 | Removing | Worn alarm |
| 4 | *(unclear)* | Not-worn alarm |
| 6 | — | Stationary alarm (U16) **NEW** |
| 8 | Accident | Impact alarm (U16) |
| 9 | — | Beacon-detected **NEW** |
| 10 | — | Beacon-lost **NEW** |
| 11 | Power restored | Vehicle power connected |
| 12 | Power cut | Vehicle power disconnected |
| 13 | Power on | ACC on |
| 14 | Power off | ACC off |
| 18 | — | GPS fault alarm **NEW** |
| 20 | — | Temperature alarm (2 bytes) **NEW** |

### Status Body (0x6A) — 8 new bytes in NEW

| Field | OLD | NEW |
|---|---|---|
| #17 | — | Status timestamp (U32, 4 bytes) **NEW** |
| #18 | — | Device status bits (U8, 1 byte) **NEW** |
| #19-21 | — | Device status bits ignored (3× U8) **NEW** |
| Msg Length | ~22 | 24 |

Device status bits (#18): bit0=strap, bit1=motor, bit2=wire1, bit3=wire2, bit4=wire3, bit5=motion.

### Beacon Body (0x68)

| Field | OLD | NEW |
|---|---|---|
| WiFi/BT flag | 1=WiFi, 2=Bluetooth | 1=WiFi, 2=Bluetooth, **3=Beacon movement alarm NEW** |
| #8 | "beacon power" (30=30%) | "beacon battery" (58=88%) |
| Multi-beacon | — | **Up to 10 beacons retained NEW** |
| Msg Length | 0x0D = 13 | 0x0E = 14 |

### LBS Cell Tower Body (0x66)

| Field | OLD | NEW |
|---|---|---|
| Msg Length | 0x19 = 25 | 0x11 = 17 |

Same fields (MCC/MNC/LAC/CID/RSSI), but OLD has 8 extra bytes of unspecified content.

### Other

- **Heartbeat Body (0x6B)** — present in NEW, not in OLD.
- **Invalid marking:** OLD says "all FF = invalid"; NEW says "0xFFFFFFFF = invalid" (more precise).

nilei: I might have found something:

Although issue #5835 initially refers to Binary Protocol v03, and the frame mentioned at the beginning does indeed start with FC as expected for Xexun3 (which is also confirmed in both documents), the [subsequent log example frames](https://github.com/traccar/traccar/issues/5835#issuecomment-4348414332) that are shown all begin with FAAF, which is the Xexun2 protocol, but not Xexun3. 

Not sure if this is of any help ... 

-- Cheers, Nils

nilei: I have contacted the vendor. Let's see what they say about the two very different protocol documentations. 

-- Cheers, Nils

tananaev: We don't need to compare everything between protocols. We need to figure out the course issue. I suspect old sample was incorrect because the documentation says that the course is the same. Let's wait for the vendor and see what they say.

nilei: I have now corresponded back and forth with the manufacturer and they have given me an updated protocol documentation: [Xexun API 260916 EN.docx](https://github.com/user-attachments/files/32300769/Xexun.API.260916.EN.docx).

They said that it definitely needs to be divided by 10, both the heading/course and the speed. I have asked them to check the frame that we already had in the test file to see if it is valid from their perspective. I haven't received a response yet.

The question is, once I have all of this, how should I make it available to you? I don't really want to paste the entire chat with the manufacturer here. Would it be enough if I create a summary instead?

-- Cheers, Nils

nilei: If you'd like, you could also send me your WhatsApp number in a private message, and then I can invite you to the chat so you can see for yourself. It might be a bit extreme, but I'll leave it up to you on how you'd like to be convinced.

tananaev: You can email support.

nilei: The test frame in the decoder test is exactly the same frame that was used as an example in the API specification at that time. This frame is not just compatible with the old specification, it is literally the example frame included by Xexun in the old specification. 

The manufacturer has confirmed to me that this frame does not meet the latest specifications. The frame has a message length of 33, but according to the latest specification, the message length is 34. If we want to support both variants (I'm deliberately avoiding the word "versions" because both report protocol version "3" in their data), then the lenght would probably be the best indicator for the handling switch. 

Here is a screenshot from the PDF of the **old** specification:

Image

The current or previous test frame:

`fc0040032006086104508038701564216913f223403693012f635344405c829142b302f7427f33331a2e000000a40011046a1055ffff1f0000000000ffffff04ff09ff1a30cf`

The manufacturer has confirmed that an earlier firmware version sent messages in this old format. Unfortunately, the manufacturer also states:

> These two values have been modified many times before, so we cannot confirm whether division by 10 is required for the old firmware versions. It is recommended to upgrade all devices to the latest version and adopt the new protocol.

So while we can't say for certain that our test frame comes from an actual device, at least it comes directly from the manufacturer. The latest specification no longer includes a sample frame directly in the document.

The test frame I added for the decoder test has been verified as coming from a genuine P02 device. 

`fc004903203e086259608092620164226aa936754049aff0ff6a6735401c97df94ded82041ab3333400d1c00000000000d006a18630000147e0e0007fdffffffffffffff6aa9367501010000a066cf`

The logs have already been provided.

So the question that remains is really whether we want to continue supporting the old version as best we can in future versions of the decoder to ensure backward compatibility, or whether it’s sufficient to state that the old specification is supported in an earlier version of the decoder but no longer in newer versions (Traccar versions), and that we’ll use the latest specification to test and expand the decoder.

So I see two options:
- Maintain backward compatibility for both versions of the decoder to the best of our ability.

OR

- Adapt and further develop the decoder according to the new specification.

@tananaev I will send you the full message to the support email address. I would also ask you to decide how we should proceed from here.

-- Cheers, Nils

tananaev: I think most likely the old sample is wrong. Let's remove it.

nilei: I'd be happy to do that. Should we then calculate the direction based on the new specifications? If so, I'd also take care of the speed calculation.

tananaev: Which one is correct? Is there conflict between documents?

nilei: Both are correct in that regard. However, the manufacturer naturally recommends updating the devices to the latest firmware version. The old specification is no longer supported by the new firmware. 

image


There are a few changes. I'll list the changes that affect the current decoder below. However, as I have very new devices received from the vendor (P02 devices) I would update the decoder issue by issue. 

Area | Old | New
-- | -- | --
ACK message length | 12 bytes | 16 bytes
ACK body | 0x00 only | 0x00 + U32 UTC timestamp
ACK CRC range | CRC calculated over the old 12-byte message range | CRC range must include the additional timestamp bytes
GPS message length (0x64) | Legacy GPS payload handled as 0x21 / 33 bytes | Support 0x22 / 34-byte GPS payload
GPS speed | Raw U16 interpreted directly as km/h | Raw U16 / 10 km/h
GPS angle/course | Angle not decoded / raw value not applied | Decode U16 / 10 degrees and store as course
GPS precision factor | Old interpretation / not explicitly handled | Interpret as centimeters; convert to meters if stored as accuracy
GPS tracking duration | Remaining GPS bytes skipped via subEnd | Read final U8 tracking duration in seconds
GPS invalid altitude | Float value read directly | Treat 0xFFFFFFFF as invalid and do not store altitude
LBS RSSI (0x66) | 0xFF can be read as unsigned 255 | Treat 0xFF as invalid / unavailable RSSI
Status message length (0x6A) | Legacy 0x10 / 16-byte payload | Support 0x18 / 24-byte payload
Status timestamp | No state timestamp in parsed status data | Read additional U32 State Timestamp
Device time from Status | Device time mainly depends on positioning messages | State Timestamp can be used as device time
Equipment Status 1 | Not present / skipped | Decode strap, lock, wire and motion status bits
Equipment Status 2 | Not present / skipped | Decode new-position / previous-position / never-located flags
Equipment Status 3 | Not present / skipped | Read reserved status byte
Equipment Status 4 | Not present / skipped | Read reserved status byte
Status tail handling | Remaining bytes skipped with buf.readerIndex(subEnd) | Parse the additional 8 bytes before moving to subEnd

-- Cheers, Nils

tananaev: I think let's fix the course first. Then if there's a clear conflict between two protocols, then it needs to be gated by model or some heuristic. That would be a separate PR.

nilei: Should we delete the test with the old frame, or keep it with `VerifyNotNull` as I suggested? If we go with the latter, then there would be nothing else to change in the PR.

tananaev: Delete because it seems to be invalid.

src/test/java/org/traccar/protocol/Xexun3ProtocolDecoderTest.java
```diff
@@ -17,8 +17,9 @@ public void testDecode() throws Exception {
         verifyNull(decoder, binary(
                 "fc000b03200108610450803870158318cf"));
 
-        verifyPosition(decoder, binary(
-                "fc0040032006086104508038701564216913f223403693012f635344405c829142b302f7427f33331a2e000000a40011046a1055ffff1f0000000000ffffff04ff09ff1a30cf"));
+        verifyAttribute(decoder, binary(
+                "fc00490320e8086259608092620164226aa92f624049b01fff79c842401c97788f16414443084ccd400e19001c009f000e006a18630000173a2f0001a3ffffffffffffff6aa92f622101000030f4cf"),
+                "course", 15.9);
```
tananaev: Why not verify position?

nilei: `verifyPosition` only checks that `course` is within the valid range (0–360°), not its exact value. The `position()` builder also has no way to set an expected course. To verify the exact decoded course value (`159 / 10 = 15.9°`), we must use `verifyAttribute`, which asserts `assertEquals(expected, position.getCourse())`.

-- Cheers, Nils

tananaev: That's fine to verify just range and the rest of the position.

tananaev:
What is going on?

nilei:
> What is going on?

Should be okay now.

src/test/java/org/traccar/protocol/Xexun3ProtocolDecoderTest.java
nilei: Okay, so you want the test for the correct course value deleted? Did I understand that correctly?

tananaev: Yes

nilei:
@tananaev Would you mind uploading the protocol documentation to your repository at https://www.traccar.org/protocols/ ? This would allow to reference the document in stead of uploading it multiple times with upcoming pull requests. 

[Xexun API 260916 EN.docx](https://github.com/user-attachments/files/32381780/Xexun.API.260916.EN.docx)


-- Cheers, Nils

nilei:
Further evidence from our communication with the manufacturer Xexun that the old version of the protocol is definitely outdated:

image

src/test/java/org/traccar/protocol/Xexun3ProtocolDecoderTest.java
nilei: Done

Issue #5741, opened 2026-01-12
help with the new Xexun DDX14 protocol


chesarito:
I need help developing the protocol (Decoder) for the xexun DDX14 device, which is a people anchor.

[Xexun API.pdf](https://github.com/user-attachments/files/24573938/Xexun.API.pdf)

tananaev:
Are you interested in sponsoring this work?

chesarito:
Yes, I'm interested because I have a project with these devices, and I'm having problems with LBS/WiFi and improving indoor positioning.

tananaev:
You can email support to discuss it.

chesarito:
yes [redacted]

tananaev:
Implemented. Please provide logs for testing.

PR #6035, opened 2026-09-20
Fix Xexun3 decoder speed scaling

nilei:
The [Xexun API 260916 EN.docx](https://github.com/user-attachments/files/32435141/Xexun.API.260916.EN.docx) reports speed in 0.1 km/h units. Divide the raw value by 10.0 before converting to knots, as specified in the API documentation.

Just like with the course value, I haven't added an additional test for speed. Let me know if you'd like an additional test for this value.

-- Cheers, Nils


tananaev:
There is a conflict with the older Xexun API documents:

- [Xexun API.pdf attached to #5741 in January 2026](https://github.com/user-attachments/files/24573938/Xexun.API.pdf)
- [Xexun API PDF attached to #5835 in March 2026](https://github.com/user-attachments/files/26250568/Xexun.API.2026-03-08.22_37_20.pdf)

On page 3, both documents define GPS field 10 as `Speed (km/h)`, U16, and explicitly show `00 20` decoding to `32`. That agrees with the current decoder. The September document instead says to divide the value by 10.

Can you confirm with Xexun whether this is a correction to the older documentation or a change in firmware behavior? If devices follow the older specification, this change would make their reported speed 10 times too low. Both older documents also specify protocol version `03`, so that byte alone does not distinguish the interpretations. A device capture with independently verified speed would also help resolve this before applying the scaling unconditionally.

nilei:
I'll confirm with the vendor. Stay tuned ...

nilei:
> I'll confirm with the vendor. Stay tuned ...

@tananaev 

Evidence from communication with Xexun tech service:

image

---

Also double confirmation check to be sure there is no misunderstanding:

image

---

-- Cheers, Nils

tananaev:
OK, let's merge for now.

PR #6036, opened 2026-09-20
Add GPS accuracy and tracking duration fields to Xexun3 decoder

nilei:
Parse the precision factor (field 12) and tracking duration (field 13) from the GPS message body (0x64) in Xexun3ProtocolDecoder.

- Precision is converted from centimeters to meters and set via position.setAccuracy(), consistent with other decoders (e.g. Gt06).
- Tracking duration is stored as a custom attribute since no standard Position key exists for it in Traccar.
- Both fields are guarded by buffer-length checks and sentinel-value (0xFFFF / 0xFF) validation to handle variable-length sub-messages.

Evidence via [Xexun API 260916 EN.docx](https://github.com/user-attachments/files/32436204/Xexun.API.260916.EN.docx)

-- Cheers, Nils


src/main/java/org/traccar/protocol/Xexun3ProtocolDecoder.java
```diff
@@ -132,6 +132,18 @@ protected Object decode(
                     buf.readUnsignedByte(); // signal
                     position.setSpeed(UnitsConverter.knotsFromKph(buf.readUnsignedShort()));
                     position.setCourse(buf.readUnsignedShort() / 10.0);
+                    if (subEnd - buf.readerIndex() >= 2) {
+                        int precision = buf.readUnsignedShort();
+                        if (precision != 0xFFFF) {
```
tananaev, line 137: I couldn't find this in the documentation. Where did you get this special value from?

tananaev:
The precision unit conflicts with the older Xexun API documents:

- [Xexun API.pdf attached to #5741 in January 2026](https://github.com/user-attachments/files/24573938/Xexun.API.pdf)
- [Xexun API PDF attached to #5835 in March 2026](https://github.com/user-attachments/files/26250568/Xexun.API.2026-03-08.22_37_20.pdf)

Both already include GPS field 12 (precision, U16) and field 13 (tracking duration, U8). Across pages 3–4, field 12 is labeled `Precision factor / decimeter` (the unit continues onto page 4), while tracking duration is in seconds. The September document changes precision to centimeters; tracking duration remains in seconds.

For the older documented unit, conversion to meters would be `/ 10.0`, whereas this PR uses `/ 100.0`. Devices following the older specification would therefore have accuracy distances reported 10 times too small.

Can you confirm with Xexun whether centimeters corrects an error in the older documents or reflects a firmware change? Both older documents specify protocol version `03`, so that byte alone cannot distinguish the units. The tracking-duration decoding agrees with both versions of the documentation.

nilei:
> Can you confirm with Xexun ...

Good catch! I verify this with the vendor ASAP. 

-- Cheers, Nils

nilei:
> > Can you confirm with Xexun ...
> I verify this with the vendor ASAP.

You were absolutely right. Thank you very much for your tireless attention to such important details. The manufacturer has confirmed to me that this is an error in the newer specification. The value is in decimeters and must therefore be divided by 10 to convert it to meters for calculations.

**New Specification:** 
[Xexun API 260921 EN.docx](https://github.com/user-attachments/files/32454597/Xexun.API.260921.EN.docx)


**Some more evidence:**
image

src/main/java/org/traccar/protocol/Xexun3ProtocolDecoder.java
tananaev: Looks like this is not addressed yet.

.gitignore
```diff
@@ -12,3 +12,5 @@ out
 build
 .vscode
 bin
+docker-compose.base44.yml
+.base44
```
tananaev: No.

src/main/java/org/traccar/protocol/Xexun3ProtocolDecoder.java
nilei: That is correct. The specification generally states, "0xFFFFFFFF is considered invalid", which means "all bits set" for the respective field width:

- **Precision Factor** is read using `readUnsignedShort()` (2 bytes) → all bits set = `0xFFFF`
- **Tracking Duration** is read using `readUnsignedByte()` (1 byte) → all bits set = `0xFF`

This pattern is consistent throughout the rest of the decoder: `steps != 0xFFFF` (2 bytes), `heartRate != 0xFF` (1 byte), `fuel != 0xFF` (1 byte). The checks thus correspond to the field width.

-- Cheers, Nils

tananaev: Where does it say that? Page number?

.gitignore
nilei: No worries. Thing is, that my builder platform sometimes adds unwanted files to the feature branch, and if I created the PR already from this feature branch, follow-up commits end up into the PR, too. Until now, I have no better workflow to avoid this. Sorry.

-- Cheers, Nils

src/main/java/org/traccar/protocol/Xexun3ProtocolDecoder.java
nilei: On page number 1:

image

---

In addition, the decoder already had similar checks in place. Therefore, this new section was meant to be aligned to the existing checks.

For example:

```
if (heartRate != 0xFF) {
    position.set(Position.KEY_HEART_RATE, heartRate);
}
```

tananaev: Do you have a real sample with those values?

nilei: I think this one has a few of these values for heart rate and other fields:

`fc00c703201508625960809262016a185e0000110000000000ffffffffffffff6ab081be010000006ea03041414132363036333030377c41414132363034313730327c503032000000000089883030000136467435744141412c4137363730452d4c4153435f4131323442303241373637304d362c415436353538522d354e2d33322d31432c747261636361722e63727970746f6e6f64652e64653a3532363440322c46494e444d595f4e4f545f5345542c3239353035303930383031313338392c2c302c32363230330678cf
`

The frame is a status and version frame `0x6A` + `0x6E` from `September 20, 2026 at 18:59:42 CEST`.

The `0x6A` submessage contains the current device status.

| Field                  | Value                 | Description              |
| ---------------------- | --------------------- | ------------------------ |
| Timestamp (UTC)        | `2026-09-20 16:59:42` | UTC timestamp            |
| Local Timestamp        | `2026-09-20 18:59:42` | CEST (UTC+2)             |
| Battery Level          | `94 %`                | Remaining battery        |
| Connection Duration    | `1`                   | Raw protocol value       |
| Cellular Signal (CSQ)  | `17 / 31`             | Signal quality           |
| Tracking Serial Number | `0`                   | No active tracking point |
| Movement               | `0`                   | No movement detected     |
| Charging Status        | `OFF`                 | Device is not charging   |
| Step Counter           | `0`                   | No steps recorded        |
| Temperature            | `0xFF`                | Invalid / not measured   |
| Respiratory Rate       | `0xFF`                | Not available            |
| Heart Rate             | `0xFF`                | Not available            |
| Blood Pressure SYS     | `0xFF`                | Not available            |
| Blood Pressure DIA     | `0xFF`                | Not available            |
| SpO₂                   | `0xFF`                | Not available            |
| Oil Level              | `0xFF`                | Not available            |

### Sentinel Values

```text
0xFF = unavailable / invalid / not measured
```

For this frame, `0xFF` is used for unavailable sensor measurements.


-- Cheers, Nils

nilei: Steps, for example, is currently checked against `0xFFFF`:

```
                    if (steps != 0xFFFF) {
                        position.set(Position.KEY_STEPS, steps);
                    }
```

